PR #982, opened 2017-12-21
EPUB 3.0.2

dauwhe:
I've created a rough and incomplete version of EPUB 3.0.2, based on the [Google Doc](https://docs.google.com/document/d/1r2RbLipc5VY3vUp_iuPak3oaNxI5BF9gJ5s-98qsmEY/edit) and using EPUB 3.1 as a starting point. My goal was to have something to look at, discuss, and file issues against.

I think there's a meta-issue on how to communicate to authors, reading systems, and validators the status of features restored only for backward compatibility. I want to avoid new content being authored with 'bindings' or 'switch' or 'trigger'. I've put warnings at the beginning of these sections, and even applied a background color to the entire sections. 

I also wonder if we need a more granular approach than the current categories of deprecated, superseded, and superseded and marked for removal. I know I found these confusing enough that I had to create a table :)

I hope we can also have a broad conversation within our entire community about the changes proposed. 

Note I haven't yet created a more formal "changes from 3.1" document, but that's a top priority.


murata2makoto:
Thank you,  It appears that OCF has not been updated yet.

mattgarrish:
A document or wiki that explains why this revision is happening would also be useful. The technical details are important when it comes to patching the specs, but the first thing you want to know when you hear of a change like this is why it's necessary to do and what impact it has (e.g., what does "withdrawn" mean wrt 3.1).

murata2makoto:
@mattgarrish  Is [EPUB 3.0.2 Proposal ](https://docs.google.com/document/d/1r2RbLipc5VY3vUp_iuPak3oaNxI5BF9gJ5s-98qsmEY/edit) what you want? Dave already referenced it.

Since 3.1 is merely a W3C member submission, it has no official status and thus cannot be withdrawn.  But the EPUB CG and the Publishing Business Group might want to say that 3.1 should not be used.  In ISO, I hope to see EPUB 3.0.2 as a collection of International Standards.

mattgarrish:
I don't see it serving the purpose I mentioned, no, or it doesn't provide any specific rationale for why this is being done that I spotted; only technical details of what 3.0.2 might look like.

When someone asks why 3.1 is being abandoned, for example, what are the answers? Is there a business case that spells out the advantages? I'd like to avoid the appearance that this is a capricious change, as I can see it frustrating people who are already anticipating or working toward 3.1 support (and providing ammo to the detractors of the format).

I also think we need to be clear about 3.1 support (or lack thereof), too. It's going to be very confusing that a higher numbered standard is not the most recent or supported version if it lingers on. It may look like it lives on in parallel, like 2.0 and 3.0 already do.

murata2makoto:
I believe that  no EPUB 3.1 publications have been sold, since epubcheck does not support it.  I think that this is enough to make people doubt Publishing@W3C.

The biggest adopter of EPUB 3 is Japanese publishers.  They think that business advantages of EPUB 3.1 are negligible while the cost of using EPUB 3.1 together with EPUB 3 is significant.  Thus, migration to 3.1 will not happen in Japan.  Who will then use EPUB 3.1?

As I wrote before, the version number EPUB 3.1.X will invite skepticism from Japan.  EPUB 3.0.2 has already received a number of hoorays from Japanese users.

mattgarrish:
And epubcheck is not likely to support 3.0.2 any time soon, either, which has nothing to do with the format and everything to do with the lack of available resources.

Revoking a specification is more than just an incremental change, so we should provide clarity and transparency on the issue. If there's a strong and clear case, it should be easy to create a document that details the rationale for everyone. That's what I'm asking for, not a discussion here.

murata2makoto:
OK.  I added a section "Rationale" to the EPUB 3.0.2 Proposal.

laudrain:
I completely agree and support. 
@Makato: shouldn't we add something on the difficulty to bring EPUB3.1 to ISO ?

dauwhe:
> Thank you, It appears that OCF has not been updated yet.

I've updated version numbers in OPF, and changed a reference to a package version attribute from 3.1 to 3.0. But my understanding is that no substantive changes to OCF were proposed.

therealglazou:
Are we seriously going to revive @refine and all the incredible crap that were turning implementors' lives into hell ? Seriously ?

murata2makoto:
@therealglazou
Yes, seriously.

matt-curtis:
@murata2makoto Wow, that sounds terrible (even though I've already implemented it, it was no joy at all.) What's the rationale for changing this?

Also, where was the intent to rollback to a "3.0.2" announced? What's the plan for handling 3.1, then? (I'm aware those are versions of the questions announced above.) Lastly, if I want to stay updated on decisions like these, is this the best place to do so?

murata2makoto:
I will provde more details tomorrow, but there are too many EPUB3 publications
 that heavily use ＠refines and this is extremely unlikely to change.

mattgarrish:
@matt-curtis What I've come to understand is that this is still only a proposal; no actual revision has begun. I believe a more formal announcement/process for carrying this out will be coming in the new year. If you haven't joined the community group, I'd suggest doing that.

murata2makoto:
On 2016-01-15, Digital Publishing Initiatives Japan Co., Ltd.  had 294,275 commercial EPUB3 publications.  All of them heavily use @refine.  More about this, see the EBPAJ guideline and examples, which are in a [ZIP file](http://ebpaj.jp/cmspage/wp/wp-content/uploads/2015/04/ebpaj_epub3guide_ver1.1.3-en-1.zip).  

New versions incompatible with these EPUB3 publications will be rejected by the Japanese market. 

(Markus wrote "in education US, use of refines heavily in their production".  See https://docs.google.com/document/d/1OoepgLPldn0V1dehA4ldzoJLeppQtfYkUQA3vUdOD8Y/edit?ts=56f15ecc)

mattgarrish:
If you're going to try and add the 3.1 attributes into 3.0, we need to be serious about support for new features before a specification can become a recommendation (something we should do regardless). If we can't guarantee support, we're just taking a bad situation and making it worse.

We also shouldn't release standards until we can validate them.

We may not have to follow strict W3C procedures, but if we want to improve people's faith in our standards there are processes we should start following.

murata2makoto:
One solution is to drop 3.1 attributes from 3.0.2.

But I think that IDPF was not serious about support when it shipped EPUB 3.1.

mattgarrish:
Speaking of no support, we should also remove specific mention of the satellite specs from epub, with the possible exception of CFI. And take alt style tags out of the core, where it never belonged.

They're zombies that linger on without prospects for support. They also aren't likely to be workable with EPUB 4.

GarthConboy:
To be clear, EPUB 3.0.2 is a proposal, as discussed in the initial document [here](https://docs.google.com/document/d/1r2RbLipc5VY3vUp_iuPak3oaNxI5BF9gJ5s-98qsmEY/edit?ts=5a2aaeea#).  A number of us (myself, Makoto, DaveC) hope to have the effort become an official effort of the EPUB CG; a discussion to be taken up early next year.  But, at this stage, it's simply a proposal along with some early prototyping effort.

therealglazou:
Let me summarize my point of view: you're proposing to dive into a deep mess standards-wise only because EPUB does not have a clear versioning system reflected in an attribute immediately reachable somewhere in container.xml or even better in the mimetype so you can't make a clear difference between 3.0, 3.0.1 and 3.1, something I told the IDPF EPUB WG multiple times quite long ago? I think you're using a steamroller to kill an ant here ; we could add such an attribute to the almost unused EPUB 3.1 and then have that trivial discrimination between 3.0.x and 3.1 for a much cheaper price.

therealglazou:
It also clearly indicates that the old IDPF process releasing specs without interoperable implementations was not and is not workable any more. 3.1 came too early after 3.0.1 (I said it too), tried to fix things that should *never* have made it to 3.0 that was itself released without implementations. AFAIC, BlueGriffon has three different processors for EPUB (2.x, 3.0.x and 3.1) and this is not understandable for users. It's also hell from an implementor's point of view. EPUB is, in my opinion, a dead end because it's vastly suboptimal and expensive to implement in production chains. We need to rethink publications from the ground up, *strictly* based on Web standards and getting rid of extra xml dialects, implementing extensibility mechanisms but starting with a restricted target for a Level 1. The current situation is clearly not workable long-term.

therealglazou:
FWIW, metadata refinement through the `refines` attribute has always been a very bad mechanism and remains a very bad mechanism. Getting rid of it was A Good Thing™️.

therealglazou:
FWIW the [E0 proposal](http://www.glazman.org/weblog/dotclear/index.php?post/2017/01/17/E0)

murata2makoto:
@therealglazou To me, the most important thing is to have a maintained specification for commercial EPUB3 publications in the world.  Even if a future attribute or a version attribute value different from "3.0" announces a new version of EPUB (or EPUB 3.1), such a version has no impacts on existing EPUIB3 publications.

murata2makoto:
@therealglazou Re: The current situation is clearly not workable long-term.

I agree that publication@W3C still has a lot to learn before it becomes a reliable standardization committee.   But my goal now is to ensure the longevity of existing EPUB3 publications. 

Getting rid of bad mechanisms would be nice if it is possible.  I am sure that you have a lot of experiences here.

therealglazou:
> Getting rid of bad mechanisms would be nice if it is possible. I am sure that you have a lot of experiences here.

Anything based on ID/IDREF inside a single instance should be presumed harmful. There are cases where it's not, but they're quite rare. In the vast majority of cases, the fact the ID has to be chosen by the user or machine-generated triggers an editability issue. Here, in the case of refinements, we even have nested refinements: one can refine a refinement. That's overkill.

murata2makoto:
@therealglazou  I am not saying that refines is a good thing.  I know that it is a last minute hack.  But longevity of hundreds of thousands EPUB 3 publications containing refines is more important than anything else.

therealglazou:
>  But longevity of hundreds of thousands EPUB 3 publications containing refines is more important than anything else.

Don't misunderstand me, Murata-san, I agree with you. But an ecosystem like EPUB leading to a large publishers' group (the JP publishing industry) saying the recent 3.1 effort was worthless and we should move from 3.0.1 and not 3.1 is a major failure signal for the EPUB world and its parent Standards body.

matt-curtis:
@murata2makoto This is what I find confusing, and please excuse my ignorance/naivety: Why is the fact that ePub 3.1 drops refines a problem? Yes, it makes things more work for reading system implementors and parsers, especially the more breaking changes there are (like refines), but isn't that the nature of the beast - didn't the specification always recommend as much backwards compatibility as possible? ePub 3.1 doesn't automatically mean that all ePub 3.0.1 publications have to be re-released as 3.1, and should work just fine in readers that already have 3.0.1 support, no?

@mattgarrish Thanks for the information!

murata2makoto:
@matt-curtis EPUB 3.1 provides negligible business advantages over 3.0.1 but requires non-negligible adoption costs.  Japanese publishers, the biggest adopter of EPUB 3 in the world, say so and reject it.  As of now, nobody uses EPUB 3.1.    EPUB 3.1 is very unlikely to be adopted by the market.  Slightly-different-yet-incompatible EPUB 3.1 should not have been created.

You appear to think that the EPUB 3.0.1 ecosystem is not hampered by 3.1.  Yes, it is hampered.  First, ISO/IEC standardization of 3.1 would make every EPUB 3 publication non-conformant to ISO/IEC EPUB, since ISO/IEC recognizes the latest version only.  Second, bugs and editorial mistakes of 3.0.1 are not addressed.  Third, it is not clear if miscellaneous vocabularies, which were made clear in EPUB 3.1, are applicable to EPUB 3.0.1.

therealglazou:
> EPUB 3.1 provides negligible business advantages over 3.0.1 but requires non-negligible adoption costs. Japanese publishers, the biggest adopter of EPUB 3 in the world, say so and reject it. As of now, nobody uses EPUB 3.1. EPUB 3.1 is very unlikely to be adopted by the market. Slightly-different-yet-incompatible EPUB 3.1 should not have been created.
>
>You appear to think that the EPUB 3.0.1 ecosystem is not hampered by 3.1. Yes, it is hampered. First, ISO/IEC standardization of 3.1 would make every EPUB 3 publication non-conformant to ISO/IEC EPUB, since ISO/IEC recognizes the latest version only. Second, bugs and editorial mistakes of 3.0.1 are not addressed. Third, it is not clear if miscellaneous vocabularies, which were made clear in EPUB 3.1, are applicable to EPUB 3.0.1.

Excellent arguments, I must say. I understand. One naive question though: why didn't the Japanese publishing industry veto against 3.1 during its adoption process at IDPF with these precise arguments?

therealglazou:
FWIW, i wrote the following a year ago:

> metadata fun, as always with every EPUB spec. Implementing refines in 3.0 was a bit of a hell (despite warnings to the EPUB WG...), and it's gone from 3.1, replaced by new attributes. So no forwards compatbility, no backwards compatibility. Yet another parser and serializer for EPUB-compliant user agents.

murata2makoto:
We did send a [letter](https://docs.google.com/document/d/1LDmzu7HkFLqXqj-pO5Yx_CVVfEXGz8u81cf1pyCS7gk/edit) about our concerns.  In the IDPF EPUB WG, I strongly opposed to the decision of version="3.1" but I was the only.

therealglazou:
> We did send a letter about our concerns. In the IDPF EPUB WG, I strongly opposed to the decision of version="3.1" but I was the only.

No you were not. I said the same, and thought I was the only one too.

TzviyaSiegman:
@therealglazou and @murata2makoto This was not ignored. It was discussed in the EPUB WG and on GitHub. Please refer to https://github.com/w3c/publ-epub-revision/issues/695 and https://docs.google.com/document/d/152AO1a4FzVZwZs2wmJvbVxK5w9cXAJpCvk5NQOHAlFc/edit?usp=sharing. You may not be happy with the decisions, but they were made. @GarthConboy and @murata2makoto are attempting to create something of a patch to help JEPA and other communities. 

I share the concerns listed here (and over email) about backtracking on completed work. The EPUB WG voted on this issue and resolved to move forward despite JEPA's concerns about the version attribute. I do not think this is the best place to debate policy and bylaws of standards revision. I recommend calling a meeting of the EPUB 3 CG in the new year cc @RachelComerford and @dauwhe.

therealglazou:
> I recommend calling a meeting of the EPUB 3 CG in the new yea

As far as I can tell, EPUB 3.0.2 is not in-scope for the CG, charter-wise.

murata2makoto:
> > I recommend calling a meeting of the EPUB 3 CG in the new year
>
> As far as I can tell, EPUB 3.0.2 is not in-scope for the CG, charter-wise.

I have tried to make clear that maintenance of EPUB 2 and EPUB 3(.0.1) are in the scope of the CG.  Commitment to the maintenance of old standards is a must for a good standardization committee.

RachelComerford:
@dauwhe and I are happy to discuss this in the CG (which is charge with maintaining the existing standards, as @murata2makoto states above) in the New Year. As a representative of a publisher that has in fact been working on implementing 3.1, this conversation is of particular interest to me. 

One request going into this next meeting:
Please remember that a number of the CG members are either new to the W3C or have less technical experience than @therealglazou and @murata2makoto. As we go into this next meeting, I ask that you raise issues using this framework, to help keep all members on the same page:

1. What is the problem you are trying to solve?
2. What solution are you proposing?
3. What alternatives have you considered?
4. Should this not work for others, what is your backup plan?

Many thanks - looking forward to a spirited but respectful conversation in the new year.

therealglazou:
> the CG (which is charge with maintaining the existing standards

Per Charter, the CG is in charge of
1. follow-on work from EPUB 3.1 and its constituent elements
2. EPUB 3 extension modules previously developed by IDPF
3. ancillary deliverables around EPUB 3

Can you please explain where follow-on work from EPUB 3.0.1, i.e. 3.0.2, sits in this list?

RachelComerford:
@therealglazou I'm puzzled over what you are arguing for or against here. It almost sounds as though you would prefer that there was no governing group for the maintenance of the 3.0.1 standard?

therealglazou:
> I'm puzzled over what you are arguing for or against here. It almost sounds as though you would prefer that there was no governing group for the maintenance of the 3.0.1 standard?

Rachel, did I say something like that? Certainly not. The Charter is the governing law of the Group. If the Group is not Chartered for 3.0.2 and works on it, there could be consequences, in terms of IPR for instance. This is **not** a futile question nor nitpicking.

RachelComerford:
@therealglazou Assuming this is a problem, which I think would be a valuable conversation to set some amount of time to discuss in the CG, what is the solution that you are proposing?

therealglazou:
> Assuming this is a problem, which I think would be a valuable conversation to set some amount of time to discuss in the CG, what is the solution that you are proposing?

Since the [Community and Business Group Process](https://www.w3.org/community/about/agreements/#community-process) does not define, AFAICT, how (in general) a CG Charter should be amended and this CG's Charter does not define how it should be amended, I think we should ask  Staff Contacts into the Group: @iherman , Liam. The Charter having defined a scope, all not in-scope contributions could be outside of the CG patent policy and CG CLA, this is serious matter.

mattgarrish:
> If the Group is not Chartered for 3.0.2 and works on it, there could be consequences, in terms of IPR for instance.

I asked this earlier in a private conversation, as I didn't want to head down this path and only later discover it would be rejected. What I was told is that all IP was transferred to W3C, so if 3.0.2 is the preferred course forward from 3.1 then the group can pursue that option. The charter does not require that the version number follow from 3.1, only that it is the starting point. But I expect more clarity about this will come when a revision is formalized.

RachelComerford:
Wonderful! Sounds as though we have identified one of the problems and a potential solution.  I have raised the question of revision with Ivan, who is currently traveling and so will likely not respond before the new year.

If there are any alternatives that should be considered in addition to charter revision, please provide them here so that we can discuss all potential solutions at the CG meeting. If we cannot add this to the charter, what is the proposed backup plan for addressing the problem?

dauwhe:
Do people agree that a 3.1.1 release would be within the scope of the existing CG charter?

therealglazou:
@dauwhe that's already explicitly in-scope if it derives from 3.1. What's not in-scope is something deriving from 3.0.1, even if it aggregates stuff coming from 3.1, AFAICT.

llemeurfr:
Re. the CG charter: it starts with "The EPUB 3 CG shall develop future revisions to, and maintenance of, EPUB 3 and extension modules. In scope are follow-on work from EPUB 3.1 and its constituent elements  ...". And I didn't spot other constraint that would apply to the evolution of EPUB 3. 

So, IMO the first sentence allows for a work on an EPUB 3.02, and the second sentence does not explicitly exclude such work. As @therealglazou thinks that EPUB 3.02 is NOT in the scope of the charter, I think that this important discussion requires an issue of its own.

More fundamentally, we are now in a strange situation, with a semi-explicit proposal to declare EPUB 3.1 a "dead end", in favor of an EPUB 3.02 which would be "backward compatible" with  EPUB 3.01. We've seen something like this before with XML 1.1, but it was a recognition of XML 1.0 as "good enough" and was replaced by "editions" (we're at the 5th I think) with erratas and *slight modifications*. 

My issue with it is that the objectives of the EPUB 3.02 are not clearly stated, in reference to EPUB 3.01. They are only stated in reference to EPUB 3.1 (the future dead duck). 

Before we go down the technical path of defining what would go in such "minor" evolution, I think that the reason why EPUB 3.01 is not "good enough" must be clearly stated. Maybe it has been before the work on EPUB 3.1 started, in which case I'd like to have a link to this document.

Also I think that two definitions a strongly required:

- define exactly what **backward/forward compatibility** means for such an evolution. Stating that "a recent RS read an old ebook", as @therealglazou did in a mail today, is actually true but won't help defining what can be added, relaxed etc. in this evolution. In the scope of the XML structures defined by EPUB 3, we can state that optional elements and attributes may be added, contraints on their values may be relaxed, no elements can be suppressed... but the processing model of EPUB is complex and I imagine that many other rules should be expressed to guarantee backward compatibility between EPUB 3.02 and EPUB 3.01. 

- define a logical **semantic versioning** for EPUB. Usually in software development, a major version implies that backward compatibility is broken (EPUB 4 will be a good example), a minor version (digit 1) implies that backward compatibility is maintained but features are added, and a second digit is about error corrections (e.g. relaxing a type) with no additional feature (e.g. NO addition of elements). This as clearly not the case in the history of EPUB: is there a formal IDPF doc that defines this semantic versioning? If not, I suggest we create one and follow it in the future. 

Each of these issues should get their own issue for discussion IMO.

RachelComerford:
@llemeurfr thank you! This is a helpful breakdown of the issues that are being tackled. I agree we should begin dividing up the discussions into silo'd issues so that we can dig more deeply into the problems and potential solutions without confusing them.

I highly encourage that these breakouts are structured to answer, or elicit answers to, these questions:

1. What is the problem you are trying to solve?
2. What solution are you proposing?
3. What alternatives have you considered?
4. Should this not work for others, what is your backup plan?

mattgarrish:
One thing to keep in mind is that there is no requirement that the package version match the specification number. Following on from what @llemeurfr wrote, a 3.1.1 that undoes the version change and restores/removes the features to make it compatible again with 3.0.1 would avoid the problem of 3.1 living on in its current state. It would be superseded by 3.1.1, and given that no one is using 3.1 now it would do so without version="3.1" getting any traction or 3.1 lingering on as another zombie.

The really major change anticipated for 3.1 never happened: allowing html content documents. Aside from that, a driving force of 3.1 was to undo legacy messes we brought forward from 2.0 and features of 3.0 that were going nowhere in anticipation of better compatibility with 4.0 and to try and reduce the confusion around 3.0 and what it actually supports. That's where the break necessitating a new version number came from, as we're bound to no errors/warnings because of distribution requirements.

This proposal is going to undo all of that, but there are still improvements worth keeping (getting the HTML/SVG/Unicode off specific references, improving CSS, etc.). They just aren't very major anymore, which is why it can be formulated as a 3.0.2. Keeping 3.1 as the basis for 3.0.2 still makes a lot of sense, as we put significant effort into cleaning up the documents outside of these few contentious points (not just the restructure, but clarifying a lot of ambiguities). Going back to square one would be a painful and unnecessary thing to do, whether we want to expand on this revision or not.

matt-curtis:
> @matt-curtis EPUB 3.1 provides negligible business advantages over 3.0.1 but requires non-negligible adoption costs. Japanese publishers, the biggest adopter of EPUB 3 in the world, say so and reject it. As of now, nobody uses EPUB 3.1. EPUB 3.1 is very unlikely to be adopted by the market. Slightly-different-yet-incompatible EPUB 3.1 should not have been created.

> You appear to think that the EPUB 3.0.1 ecosystem is not hampered by 3.1. Yes, it is hampered. First, ISO/IEC standardization of 3.1 would make every EPUB 3 publication non-conformant to ISO/IEC EPUB, since ISO/IEC recognizes the latest version only. Second, bugs and editorial mistakes of 3.0.1 are not addressed. Third, it is not clear if miscellaneous vocabularies, which were made clear in EPUB 3.1, are applicable to EPUB 3.0.1.

@murata2makoto Hmm. I see. Thanks for sharing, though I have some reservations about whether the only solution is rolling back to some of the deficiencies (or less than ideal parts) of 3.0.1. 

> Before we go down the technical path of defining what would go in such "minor" evolution, I think that the reason why EPUB 3.01 is not "good enough" must be clearly stated. Maybe it has been before the work on EPUB 3.1 started, in which case I'd like to have a link to this document.

@llemeurfr I agree, excellent summation. Another question I have is what this means for 3.1, and what is expected of RS implementers in terms of what should be supported. This also has repercussions for what 3.1 publications already exist.

murata2makoto:
@matt-curtis  The EPUB 3.0.2 draft created by Dave is derived from 3.1 but intended to be backward compatible with 3.0.1.  Please read the draft and also the [original proposal](https://docs.google.com/document/d/1r2RbLipc5VY3vUp_iuPak3oaNxI5BF9gJ5s-98qsmEY/edit).

murata2makoto:
I think that we should admit our serious mistake.  I think that the version number "3.1.1" hides the mistake and that "3.0.2" makes the mistake clear.

murata2makoto:
@llemeurfr 

W3C has a great document [[Editorial Draft] Extending and Versioning Languages: Terminology](https://www.w3.org/2001/tag/doc/versioning) about compatibilities and evolution.  It is still a draft but is very useful, IMHO.

Backwards compatibility and forward compatibility are defined in 1.1.1 of this document.  I tried to apply that definition to EPUB 3.1.

EPUB 3.1 is not backwards compatible since EPUB 3.1 RSs cannot process EPUB 3.0.1 publications.  EPUB 3.1 RSs (detached from EPUB 3.0 RSs) cannot be rolled out in a way that does not break existing EPUB 3.0.1 editors.

EPUB 3.1 is not forwards compatible since EPUB 3.0.1 RSs cannot correctly process EPUB 3.1 publications.  EPUB 3.1 editors cannot be deployed without breaking existing EPUB 3.0.1 RSs.

murata2makoto:
@llemeurfr 

The clause "EPUB 3.1 changes retained in EPUB 3.0.2" in [the original proposal](https://docs.google.com/document/d/1r2RbLipc5VY3vUp_iuPak3oaNxI5BF9gJ5s-98qsmEY/edit) is not a list of 3.0.1 problems, but is a list of retained solutions to them.  From this clause, we can easily construct a list of 3.0.1 problems to be addressed.  Here I provide three of them.

- The specification organization of 3.0.1 is problematic, as revealed by [my earlier experiment](https://groups.google.com/d/msg/epub-working-group/bXSHRfmUjZ4/3m9UF4UQBQAJ).
- Since the definition of vocabularies are embedded within 3.0.1, addition of new entries is difficult.
- EPUB Accessibility is detached from 3.0.1 conformance.

mattgarrish:
> I think that we should admit our serious mistake.

I don't disagree, and I don't see how any revision can avoid explaining the details of what it does. But I think we're compounding the mistake by trying to go back in time to patch into the 3.0.x line. As you say, 3.1 exists. The epub community all know about it. So let's fix it instead of trying to hide it away as a "do not use" specification. Trying to splice ourselves back into 3.0.x and hope people forget about 3.1 feels like no less of a dodge of responsibility.

If we could scrub 3.1 from existence, that would make 3.0.2 less problematic, but it seems unlikely given that it was a formally adopted specification of the IDPF.

And as @dauwhe has pointed out, what happens if we do need a major revision to add features, do we jump from 3.0.x to 3.2 because 3.1 is left in an unimplemented state? How awful would that be?

But that's my only bone of contention with this proposal, and I know you and Garth have a different opinion.

murata2makoto:
@mattgarrish I think that abandoning XHTML2 and XML1.1 was a good decision.  Sticking to them would have caused more problems.  We should do the same thing.  EPUB 3.1 is known and recognized as a failure.  Sticking to it makes people doubt Publishing@W3C seriously. 

I do not believe that we will create a major revision of 3.0.2 since the Publishing WG is doing something entirely different.  I think your point here is purely theoretical.

mattgarrish:
> I think that abandoning XHTML2 and XML1.1 was a good decision.

Neither is a fair comparison, but there are lessons to learn. XHTML2 was abandoned before it was finished, so no one questions its status. XML 1.1 is a recommendation, and a source of confusion to this day. EPUB 3.1 is going to be much more like XML 1.1 than XHTML2.

What I'd rather hear than conjecture on who might think what about what, is what we can do to mitigate the problems of 3.1's existence if it's kept around in parallel. Hoping people forget about it does everyone a disservice.

murata2makoto:
Dave's initial draft of 3.0.2 has been [translated](https://advanced-publishing-laboratory.github.io/publ-epub-revision/) to Japanese already.

therealglazou:
> Neither is a fair comparison, but there are lessons to learn. XHTML2 was abandoned before it was finished, so no one questions its status. XML 1.1 is a recommendation, and a source of confusion to this day. EPUB 3.1 is going to be much more like XML 1.1 than XHTML2.

It took 5 years to kill XHTML2. I started highlighting the intrinsic blockers of XHTML2 when I was still working at Netscape, helping Netscape's representative into the XHTML2 WG, Beth Epperson. XHTML2 completely lost traction the day, LONG before its eventual dismissal, browser vendors left the WG. The WG itself lost all common sense when it refused to handle HTML4 errata and broke backward-compatibility. In that sense, the comparison is worth something : if we design things that the rendering engines' implementors won't implement and ship, these things are probably worthless and pointless. Remember the stubbornness of the XHTML2 WG led to the creation of the WHATWG, a lesson we should carefully study.

XML 1.1 is very interesting for us: a Standard can fail and remain unimplemented and/or unused. The only problems it causes are the following: a) worthless investment into the development of that Standard equals money lost and workforce diverted from other activities that could have been more strategic at that time. b) image loss.

therealglazou:
> EPUB 3.1 is not backwards compatible since EPUB 3.1 RSs cannot process EPUB 3.0.1 publications. EPUB 3.1 RSs (detached from EPUB 3.0 RSs) cannot be rolled out in a way that does not break existing EPUB 3.0.1 editors.
>
> EPUB 3.1 is not forwards compatible since EPUB 3.0.1 RSs cannot correctly process EPUB 3.1 publications. EPUB 3.1 editors cannot be deployed without breaking existing EPUB 3.0.1 RSs.

EPUB is, whatever the version, not backwards- nor forwards-compatible (I'm of course excluding minor editions here). Having implemented all flavors of EPUB 2 and 3, I know a thing or two about that...

About the numbering, my 0.02€: our main problem is ISO/IEC. I don't think we can safely release a 3.0.2 knowing its version number is lower than a 3.1 they could consider. Since we can't rescind 3.1, our best option is probably a 3.2. Since that new document will contain things that were explicitly removed from 3.0.1 for 3.1, we'll be a bit ridiculous but that's life. Whatever the version number chosen, one thing is urgent: tell the EPUB world they should not use/implement/deploy 3.1 if that's the chosen path.

The @refines history is worth introspection: ID/IDREF mechanisms are well known, since the SGML era, by all markup experts as a potential trap for editability and maintainability. I was already discussing this problem with Jean Paoli back in 1992, when we were working on "paired elements" and "paired processing instructions" for the internal controls of GRIF's rendering engine. ID/IDREF is worth it only if there is never any human intervention in the ID selection/modification. This is not the case with the `refines` attribute because most EPUB production chains in the world still contain human manipulations. I have to add the fact that EPUB 3 metadata allow nested refinements adds a unbelievable level of complexity, and turns any potential UI for metadata editing into UX hell. All in all, I am extremely disappointed and sad to see we could not kill this ugly design, bad choice and financial pit. If there is one single reason why EPUB is sick, it's that one.

murata2makoto:
@therealglazou I am quite sure that ISO/IEC does not care the version number.  This is because ISO/IEC often has versions slightly older than the latest.   I also think that we should tell the EPUB world not to use EPUB 3.1.

mattgarrish:
> if we design things that the rendering engines' implementors won't implement and ship, these things are probably worthless and pointless

True, I agree with you on this, and 3.1 was far more of a subtractive revision than an additive one. It was trying to impose some measure of reality back on EPUB for past mistakes.

3.1 has problematic aspects, and is divisive. Let's leave it at that and look at how we can do better. We're underestimating the response to 3.0.2 if we think it's going to be widely loved, too.

That's why I also wonder if we can try to find a softer landing place than a wholesale dump of 3.0 features back into production. If there were a 3.0.2/3.1.1/3.2 "strict" profile, for example, that validates like 3.1 would have, that might be one way to avoid future divisions. I'm not sure what the best way to do that is, though.

murata2makoto:
The response to the 3.0.2 proposal and draft from Japan has been very positive.  But this does not mean that Japanese users love switch.  To the contrary, I am pretty sure that switch has never been used in Japan.   We don't care even if it is gone.  But we are very strongly against version="3.1".

mattgarrish:
> I am pretty sure that switch has never been used in Japan.

I'd be surprised if it, bindings or triggers have legions of followers, either, but we did survey them and I recall at least one respondent was using them.

If we're going to retain validity for all existing content, we're still in the bind of not being able to do anything with these and also not change the version number.

I'm thinking we could define a strict profile as an appendix to the specification that lists all the things that should not be used anymore (old media types we have to add back, old vocabulary terms we have to add back, EPUB 2 legacy features, failed EPUB 3 features, etc.). That would give epubcheck a validatable set of criteria for implementing a "strict" flag. It would also be a clear signal to tool and RS developers that although these things can't be strictly deprecated/obsoleted, they're not worth the trouble of implementing.

dauwhe:
HTML and CSS are clear about how unknown elements should be handled. If a browser sees `epub:switch`, it doesn't crash, it just follows the rules. 

I like the idea of an appendix (graveyard?) of features that should not be used anymore. I don't think a new spec should define `epub:switch`, but epubcheck should not generate an error for some things. 

For this entire exercise, we must be very clear on how authors, reading systems, and validation tools should handle each feature. We might need to be more granular than the rather confusing current system of deprecation, superseding, and superseding and marked for removal.

mattgarrish:
> We might need to be more granular than the rather confusing current system of deprecation, superseding, and superseding and marked for removal.

It will destroy your sanity... :)

There are incompatibilities in what people want from validation and what is necessary to avoid problems in supply chains. That's where an "alert" that's not an error/warning but above an info message comes from. I'm still sceptical it wouldn't just get picked up the same way by vendors.

Plus, there was a desire to tell people that even though some things are still around, they're really, really not wanted in the next version. But not to tell them in a way they'll notice.

It would probably be better for the base specification to get out of the simple labeling (as much as possible, at least) and leave it to an appendix to handle the details. It's the only way I see out of being quiet sometimes and noisy others, at any rate: letting the user decide when they validate. epubcheck could emit a message recommending to validate in strict mode when it's not done, for example.

And taking a cue from 3.1, we don't need to add the full feature descriptions back into this next revision, but simply provide a link to the previous specification that defined them, with notes to the graveyard explanation of how these are (not) handled now. Whether something is obsolete, deprecated, or any flavour of superseded, there should be no prospects that it will live beyond 3.x.

RachelComerford:
Colleagues: This has been an educational conversation that has wandered across a number of different elements of this rough draft.  I am awed by the contributors ability to track the various conversations taking place but for myself, and the newer members of this group with a less robust background in these issues, I am closing this PR and reopening a number of different issue tickets so that we can discuss with greater focus and an eye towards resolution.

Apologies to anyone whose content I may have missed or mis-attributed. I tried to do this switch quickly as so to not hinder the conversation and to include everything that has been said so far. If you see something missing or in the wrong ticket, I encourage you to comment so we can get closer to resolution faster.

I will add the issue numbers shortly, as I push them live.

RachelComerford:
1. define a logical semantic versioning for EPUB #983
2. define exactly what backward/forward compatibility means for 3.0.2 #984
3. Update OCF in 3.0.2 #985
4. Do 3.1 attributes need to be dropped by 3.0.2 #986
5. Need document or wiki that explains why this revision is happening (3.0.2) #987
6. Should epub:switch be restored to EPUB? #988
7. Question of version numbering: 3.0.2 versus 3.1 #989
8. 3.0.1 etc not in the list of charter responsibilities #990

Issue #326, opened 2015-03-24
metadata for series name and volume number


GoogleCodeExporter:
```
Being able to indicate both the name of the series and the volume number in the 
metadata is critical for comics.

This would also be useful for genre fiction (SF, fantasy, crime novels) where 
series are predominant.

DublinCore doesn't seem to have such elements and I can tell from experience 
that ONIX won't work either.

This kind of metadata is every bit as important as the title for comics.


```

Original issue reported on code.google.com by `hadrien....@feedbooks.com` on 27 Mar 2013 at 3:03

elmimmo:
Where is this reflected in the draft of EPUB 3.1 of Jan 30th 2016?

Edit: Rereading the [metadata chapter of the draft](http://www.idpf.org/epub/31/spec/epub-packages.html#sec-metadata-elem) , I guess series metadata now goes out of the OPF and one is supposed to detail it into a separate ONIX XML file or some other metadata standard bundled within the EPUB archive.

> The Package Document is not designed to provide a comprehensive bibliographic record, and is not the correct location for such discovery information about the EPUB Publication. Metadata records, both that conform to international standards or that are designed for custom use, can instead be associated using the link element.

ghost:
Why was this removed in EPUB 3.1? ONIX XML seems like an overkill for a super common thing for books. (Series are after all one of the most purchased book types, no?)

Also, similar to NCX, ONIX is an entirely different beast of a format which isn't explained in EPUB itself, making it a lot more complicated to include what should be pretty straightforward information... (and making it less likely that reader systems actually implement checking for this info)

laudrain:
- Series almost never used by reading systems for the user library (as shown per survey to developers)
- Instead, general need for a simplier set of metadata
- Completeness and precision already achieved by other schemas as ONIX

ghost:
I don't see how optional additional meta tags that add simple fields make things complicated. And how can there ever be large developer adoption if it's just in one minor version of the standard and then taken out again immediately with the next iteration?

I'm aware ONIX does a lot more, but that's also why I think it's not a good match. (and if you push people to ONIX, I don't see how metadata gets "simpler" at all.)

Did you ever do a survey of the users if they'd be interested in series information being shown? I think that would be a much more helpful idea than to survey the current state of developer adoption.

mattgarrish:
[...]

PR #996, opened 2018-01-30
Rough Draft of EPUB 3.2

dauwhe:
Just as I did earlier in #982, but with the 3.2 version number, I've created a rough and incomplete version of EPUB 3.2, based on the Google Doc and using EPUB 3.1 as a starting point. My goal was to have something to look at, discuss, and file issues against.

I think there's a meta-issue on how to communicate to authors, reading systems, and validators the status of features restored only for backward compatibility. I also wonder if we need a more granular approach than the current categories of deprecated, superseded, and superseded and marked for removal. I know I found these confusing enough that I had to create a table :)

I hope we can also have a broad conversation within our entire community about the changes proposed.

Note that the "changes from 3.1" document has barely been started. The day job interferes with the good stuff!


mattgarrish:
> I think there's a meta-issue on how to communicate to authors, reading systems, and validators the status of features restored only for backward compatibility.

This has the potential to be unnecessarily confusing, since 3.1 won't see implementation.

It might be better to note here, as well, that 3.2 is a more direct successor to 3.0.1 that incorporates parts of 3.1 and that the changes reflect all of what differs from 3.0.1 (and only what differs).

Rather than talk about restoring and deprecating features, in other words, all the document will have to do is highlight the features from 3.0.1 that are being moved into the deprecation bin.

The 3.2 document will then also contain a list of all the new things that were technically added in 3.1, so we don't send people on a journey of discovery.

mattgarrish:
Has anyone gone over the vocabularies to figure out how they're going to be restored?

The meta/property vocab was emptied, for example, but that will invalidate any epub 3.0.1 file that makes use of the old properties for refines (funny enough, that may mean series metadata is coming back, as has been discussed in #326 today).

Are we still deprecating all the *-record link types in favour of record?

There's probably other stuff that isn't jumping immediately to mind. We were able to make these changes without affecting 3.0.1 content because the vocabularies were previously hard-coded into each version of the specification.

GarthConboy:
:-)

Issue #1025, opened 2018-04-12
Definition of superseded


dauwhe:
Our definition of superseded says,

> Reading Systems MUST NOT support the superseded feature in content that conforms to the current version of the specification.

This makes sense for things like ncx > nav and guide > landmarks. But what about something like bindings? Say we supercede epub:bindings in 3.2. What is a reading system's obligation when faced with a video element with bindings but native controls off?

And note we can now have content that is simultaneously valid 3.0, 3.0.1, and 3.2.

GarthConboy:
We likely need a "deprecated and RS's may not support it" (not "must not support it") -- per the likes of "Reintroduce the <bindings> element of <package>, but note that it is obsolete and deprecated.  Allow EPUB 3.2 Reading Systems to not support it." -- I'd think whatever we come up with should work for  binding, guide, NCX.

dauwhe:
How about an "obsolete"  or "legacy" category... name to be bikeshedded later:

1. Use by authors is "strongly discouraged"
2. Reading system support is optional. By the way, do we have spec language on how reading systems should respond to unknown elements in package? At least HTML and CSS are clear on how to parse unknown content. 
3. Validation tools should warn when these features are used (likely controversial because of how EpubCheck is used in practice, but I think authors must be notified if support for a feature is optional).
4. Such features will be removed in EPUB 4.X

GarthConboy:
I like it.  Matt?

I'd tend to leave off #4, as at least some of the these features are already clearly not OWP/EPUB4 compatible.

Perhaps "#4.  May be removed in a future version" or "#4. Marked for future removal" (which is close to something we've used before).

murata2makoto:
Very many features in EPUB 3 may be removed in EPUB 4.X.  If we mark only some features as "Marked for future removal", we implicitly say that other features will not be removed and send an incorrect message to users.

After all, EPUB 3.2 RSs will be used for existing EPUB 3 publications.  I think that any wording for controlling RS implementers will be ignored.

mattgarrish:
I oddly tend to side with Makoto on this one.

The only reason we introduced "marked for removal" was to really wag our finger at people that even though certain superseded features might have survived 3.1 (NCX only?) we were really, really going to toast them later. Sort of an early warning.

But predicting the future through specifications is a bad idea. What if the publishing working group decides for continued ease of translation they'll persist mechanisms to retain certain features, like the NCX? This group can't make decisions that are binding on another major revision.

[...]

PR #1029, opened 2018-04-18
reconfig of deprecated/superseded features

mattgarrish:
Here's a rewrite based on the google doc I opened in issue #1025.

I've removed the definitions of the deprecated and legacy features since we're not supporting them anymore. We did similar in 3.1, as making people work harder to figure out how to use the features should help in discouraging their use and/or highlighting that they are not features of EPUB 3.


GarthConboy (approved):
LGTM (on fairly quick review)

PR #1033, opened 2018-04-26
add clarifying note about foreign resource exception

mattgarrish:
As discussed on the 2018-04-26 telecon, this a possible clarifying note about why we allow some foreign resources without fallbacks. Text of the note is as follows:

> This exception allows Authors to include resources in the EPUB Container that are not for use by EPUB Reading Systems. The primary case for this exception is to allow data files to travel with an EPUB Publication, whether for use by scripts in its constituent EPUB Content Documents or for use by external applications (e.g., a scientific journal might include a data set with instructions on how to extract it from the EPUB Container).


GarthConboy:
I'm okay with this.

dauwhe (approved):

PR #1034, opened 2018-04-27
reading systems should treat deprecated rendition:spread "portrait" like "both"

mattgarrish:
As discussed on the 2018-04-26 telecon, this PR adds a reading system recommendation to treat portrait like spread:

> Reading Systems SHOULD treat the value "<code>portrait</code>" as a synonym of "<code>both</code>" and create spreads regardless of orientation.


GarthConboy (approved):
LGTM

PR #1035, opened 2018-05-03
remove references to alternate style tags and the non-core specs

mattgarrish:
As discussed on the 2018-04-26 telecon, this pull request removes the section on alternate style tags from the content documents specification, since the section is not enabling behaviour that is otherwise invalid.

It also removes the reference to the document being a core specification, as well as strips the specific list of IDPF non-core specifications. The existing paragraph that notes that extension specifications are valid for use continues to provide the option to use/support any of these technologies (similar to how EPUB CFI was removed but remains available for use).


dauwhe (approved):
This is great. I love seeing the spec get smaller and clearer.

Issue #1037, opened 2018-05-03
re-consolidate core media types


mattgarrish:
A prime motivator for removing the CMT list and making it updatable outside the specification was to deal with an uncertain future and long gaps between revisions. (We were looking at potentially adding Opus when it reached full support, as one example.)

But do these rationales still make sense? This group is now maintaining the specifications, and we technically have even less overhead to publishing updates.

There's a good argument that core media types are too central to the specifications to be updated outside of a full revision involving all interested parties, even if it might mean some inconvenience at times.

dauwhe:
I do think that **core** media types should be defined in the **core** spec :)

I've been annoyed dozens of times when I'm trying to look up the media type of fonts and ended up following a chain of links. We really can keep this in the core spec and update as needed. I also don't think new media types will arrive so often that this will become a maintenance headache.

mattgarrish:
Ya, it was a very late pull out from 3.1 when we got stuck on punting on the new CMT requests. A registry too far...

JayPanoz:
> I do think that core media types should be defined in the core spec :) I've been annoyed dozens of times when I'm trying to look up the media type of fonts and ended up following a chain of links.

(Kinda) Same. I lost myself too many times about that so I may be biased.

murata2makoto:
CMTs are so fundamental to EPUB 3.  When we want to create an ISO/IEC standard for EPUB 3.2, I hope to avoid the registry issue #999.   I thus support re-consolidation.

mattgarrish:
Closing with merge of #1038.

avneeshsingh:
Just want to document reason of separating core media type from core spec in EPUB 3.1. It was done when we were discussing inclusion of OPUS and other media types. We were not able to include them in EPUB 3.1 and we decided to wait for more adoption, and we separated core media from core spec so that we can easily update the list without incrementing EPUB version. Obviously at that time there was some uncertainty about future of EPUB 3.1 due to possibility of merger.
I do not have strong opinion for going in either way. If we keep it separate then complexity will be in ISO, and if we integrate then we will have to provide decimal point increment to EPUB specs for adding new media type.

PR #1038, opened 2018-05-03
Add core media type table

mattgarrish:
As identified in issue #1037 and discussed on the 2018-05-03 telecon, this PR adds the core media type table back to the 3.2 specification.


dauwhe (approved):
Hooray!

Issue #1045, opened 2018-05-14
Highlight "EPUB 3" over specific revisions


mattgarrish:
I've always found our use of version numbering within the specifications a bit confusing, as each revision often sounds like a new version rather than a continuation of EPUB 3 (this is true even for "3.0.1").

My thought is that we should look at revising some of the wording to highlight that all 3.x.x versions are "EPUB 3". To this end, I would change the overview to refer to EPUB 3 everywhere except for a line at the beginning that indicates that 3.2 is the latest definition of the standard.

We should similarly make "EPUB 3" the standard way of referring to the format, leaving 3.2 to document references and the changes document.

For example, this statement from the 3.2 document:

> This specification, EPUB 3.2, defines a distribution and interchange format for digital publications and documents.

Would be rewritten to:

> EPUB 3 defines a distribution and interchange format for digital publications and documents. This specification represents the second major revision of the standard.

Similarly, this:

> EPUB has been widely adopted as the format for digital books (ebooks), and EPUB 3.2 continues to increase the format's capabilities

Could be rewritten as:

> EPUB 3 has been widely adopted as the format for digital books (ebooks), and this revision continues to increase the format's capabilities

Thoughts welcome, though.

GarthConboy:
No objections.

RachelComerford:
Consensus from 5/17 meeting: highlight that all 3.x.x versions are "EPUB 3".  @GarthConboy and @dauwhe to review language

dauwhe:
Closing, as this was fixed by #1046. Thanks Matt!

PR #1046, opened 2018-05-16
Editorial fixes

mattgarrish:
This PR was initially going to address the issue I raised in #1045 about referring to EPUB 3 as the standard, but as I was going through I noticed a couple of other areas that could also be improved. So, in addition, to some text tweaking around version numbers, this PR also introduces the following changes:

- It moves the brief revision history of EPUB that had been in the changes document to the overview. The overview is the better location to talk generally about what we're tweaking as we go. The changes document is a reference to the specific changes made in one particular revision.

- It takes the "purpose and scope" sections and makes them the abstracts for each specification. These were always our version of an abstract in IDPF, and the current abstracts are mostly just ripoffs of the first sentence from the purpose and scope section.

### DIFF
--- a/epub32/spec/epub-mediaoverlays.html
+++ b/epub32/spec/epub-mediaoverlays.html
@@ -46,38 +46,29 @@
 	</head>
 	<body>
 		<section id="abstract">
-			<p>EPUB Media Overlays 3.2 specifies a usage of SMIL, the Package Document, CSS Style Sheets, and EPUB
-				Content Documents for representation of audio synchronized with the EPUB Content Document.</p>
+			<p>This specification defines a usage of [[SMIL]] (Synchronized Multimedia Integration Language), the <a
+					class="glossterm" href="epub-spec.html#gloss-package-document">Package Document</a>, CSS Style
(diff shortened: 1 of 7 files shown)


dauwhe (approved):
This is a PR of beauty!

epub32/spec/epub-mediaoverlays.html
```diff
@@ -46,38 +46,29 @@
...
+				who has difficulty following the text of a traditional book. Media Overlays also provide a continuous
+				listening experience for readers who are unable to read the text for any reason, something that
+				traditional audio embedding techniques cannot offer. They are even useful for purposes not traditionally
+				considered accessibility concerns (e.g., for language learning or reading of commercial audio
+				books).</p>
+
+			<p>This specification is one of a <a href="epub-spec.html#sec-epub-specs">family of specifications</a> that
+				compose [[EPUB32]], an interchange and delivery format for digital publications based on XML and Web
+				Standards. It is meant to be read and understood in concert with the other specifications that make up
+				EPUB 3.1.</p>
```
GarthConboy, line 64: The above wants to be "EPUB 3", right?

GarthConboy (approved):

PR #1083, opened 2018-08-02
EPUB OCF 3.2 sections A-E editorial fixes

mteixeira-wwn:
Implemented several editorial changes across appendix sections. Also fixed a few broken references in the technical examples, and applied uniform styles throughout. Some remaining questions:

- "Issue 1:" What are appropriate links for the schema section? What other content/copy still needs authoring?
- Do example implementations of the `application/epub+zip` media type need trademarks? I noticed EPUB uses it in the intro; unsure about "Apple iBooks", "Microsoft Edge", etc.
- "Issue 2:" List of general acknowledgments is still TK, correct?

I'm happy to revert any of my changes.

Thanks, and let me know how I can help further.

### DIFF
--- a/epub32/spec/acknowledgements.html
+++ b/epub32/spec/acknowledgements.html
@@ -7,20 +7,20 @@
 	<p>The EPUB 3.2 revision was led by: </p>
 
 	<ul>
-		<li>Rachel Comerford (Macmillan Higher Education) Chair</li>
-		<li>Dave Cramer (Hachette Livre) Chair</li>
+		<li>Rachel Comerford (Macmillan Higher Education), Chair</li>
+		<li>Dave Cramer (Hachette Livre), Chair</li>
 	</ul>
 
-	<p>In addition to the editors, this version of EPUB would not have been possible without the significant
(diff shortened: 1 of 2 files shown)


mattgarrish:
Are we planning to update the media type registration with IANA? If not, then although I'd normally be fine with your edits to the section, it probably adds confusion if it doesn't match the actual registration.

The only nitpick I have is that appendix B is a single example. It shows multiple files within the example container, but it's still only one example of an OCF.

To your questions: schemas are still TBD, epub's trademark is required for legal reasons but we don't typically acknowledge others, and, yes, acknowledgements are also to be completed.

mteixeira-wwn:
Thank you, @mattgarrish. I'll revise my edits, revert the changes to media type section (did not realize it had to match IANA's record), and update the example section.

mattgarrish:
Well, I was hoping someone else would answer the registration question as I missed what happened with that during my move.

But digging in deeper, I see Dave has already made some changes here: https://github.com/w3c/publ-epub-revision/pull/1055

So maybe leave those edits as is if we are re-registering. @dauwhe can you confirm?

PR #1125, opened 2018-09-05
tweak the prose introducing the properties of the meta vocab

rdeltour:
Previously the section said:

> Metadata `meta` properties
> These properties MUST…

I found it confusing as cursory readers could misunderstand that it
applies to all `meta` properties. This PR makes it clear that it applies
to the properties defined in the following sections, and aligns the
wording with the other vocabs.

Also, the PR clarifies a bit the language (properties can't "reference
an expression", subexpressions do; properties are not children of `meta`
elements, etc).

### DIFF
--- a/epub32/spec/epub-packages.html
+++ b/epub32/spec/epub-packages.html
@@ -1247,7 +1247,7 @@
 									Publication</a>. A <code>meta</code> element that omits a refines attribute defines
 								a primary expression.</li>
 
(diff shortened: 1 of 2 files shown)


dauwhe (approved):

rdeltour:
@dauwhe please hang on, something's wrong with this conformance requirement

rdeltour:
See further discussion in #1127

dauwhe:
Gonna merge and then edit a bit, as it seems simpler for my limited git skills :)

PR #1134, opened 2018-09-10
Update Reading System Conformance section for MathML

dauwhe:
This PR:

1. Rewords the reading system conformance requirement for MathML to better express our goals

2. Removes the section on fallback images via `altimg`, as we don't have much evidence that this is widely supported. Describing how to adapt content for non-conforming reading systems is also a tricky business to be in. 

3. Adds a note saying that reading systems may choose to support MathML via third-party libraries such as MathJax

4. Update links to MathML to point to the finished, undated MathML 3 spec rather than an outdated (and dated) version.

### DIFF
--- a/epub32/spec/epub-contentdocs.html
+++ b/epub32/spec/epub-contentdocs.html
@@ -684,12 +684,12 @@
 						<dl class="conformance-list">
 							<dt id="math-pres">Presentation MathML</dt>
 							<dd><p id="confreq-mathml-pres">The <code>math</code> element MUST contain only <a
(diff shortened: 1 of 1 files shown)


dauwhe:
@avneeshsingh @BCWalters does this work for you?

avneeshsingh:
Looks ok. Regarding mention of 3rd party libraries, most of software development projects involves use of 3rd party libraries, each and every thing is not coded natively. So, do we need to mention use of 3rd party libraries explicitly? This is not a disagreement, just want to highlight this fact.

PR #1148, opened 2018-09-24
Add non-normative schemas to repository

dauwhe:
I've copied the 31 schemas over to 32, and updated the link to the schemas in the package doc. 

I have not made any edits to the schema themselves.


mattgarrish:
I'd actually suggest starting from the 3.0.1 schemas. This revision is basically a reversal of most of the schema changes that were done in 3.1, so the 3.0.1 schemas will be closer to valid out of the box.

dauwhe:
@mattgarrish I've updated the PR to use the 30 schemas for packages. It seems the 31 schemas make sense for OCF so that they include compression?

mattgarrish:
Sure, makes sense. I was thinking mostly of all the changes to the packages rnc and sch files that now have to be walked back and schematron deprecation messages put in place instead.

PR #1157, opened 2018-10-27
add sample file that uses new epub32 features

dauwhe:
This sample file:

1. Uses a woff2 font
2. Uses a remote font
3. Uses a value for epub:type that is not in the EPUB SSV 
4. Includes an external resource that's not in the spine (a CSV file)


murata2makoto:
Where is the epub:type that is not in the SSV?

dauwhe:
> Where is the epub:type that is not in the SSV?

In content_001.xhtml we have:

```html
<section epub:type="foo">
```

PR #1181, opened 2018-11-09
More miscellaneous fixes

dauwhe:


murata2makoto:
I do not think "CSS has been an integral part of the Open Web Platform since nearly the beginning of the web " is true.  See ["Difficulty with adoption"](https://en.wikipedia.org/wiki/Cascading_Style_Sheets#Difficulty_with_adoption).

mattgarrish:
Well, in fairness, it was integral for authors even when it's been poorly supported.

But what about just moving it into the present and saying "CSS is an integral part of the Open Web Platform ..."? I don't think that can be disputed.

Issue #1198, opened 2019-01-03
"any other OCF ZIP Containers"


dauwhe:
https://w3c.github.io/publ-epub-revision/epub32/spec/epub-ocf.html#confreq-zip-enc says:

> OCF ZIP Containers MUST NOT use the encryption features defined by the ZIP format; instead, encryption MUST be done using the features described in Encryption File (encryption.xml). OCF Processors MUST treat any other OCF ZIP Containers that use ZIP encryption features as being in error.

Why are the words "any other" here? OK to delete?

dauwhe:
Closed via #1203

PR #1203, opened 2019-01-04
A few small clarifications in OCF (and an experiment)

dauwhe:
This PR:

1. Makes a tiny language fix per #1198 

2. Fixes an even tinier grammatical error in OCF.

3. Adds a very experimental MD file, which attempts to put the reading system conformance requirements for OCF in a separate doc.

### DIFF
--- a/epub32/spec/epub-ocf.html
+++ b/epub32/spec/epub-ocf.html
@@ -929,7 +929,7 @@
 						<p id="confreq-zip-enc">OCF ZIP Containers MUST NOT use the encryption features defined by the
 							ZIP format; instead, encryption MUST be done using the features described in <a
 								href="#sec-container-metainf-encryption.xml">Encryption File
-									(<code>encryption.xml</code>)</a>. OCF Processors MUST treat any other OCF ZIP
+									(<code>encryption.xml</code>)</a>. OCF Processors MUST treat OCF ZIP
 							Containers that use ZIP encryption features as being in error.</p>
 					</li>
 					<li>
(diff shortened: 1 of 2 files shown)


murata2makoto:
I think that the proposed MD requires a lot of discussions.  I am a bit worried since some requirements already present elsewhere might be normatively duplicated in this MD.  Duplication is always bad.

dauwhe:
> I think that the proposed MD requires a lot of discussions. I am a bit worried since some requirements already present elsewhere might be normatively duplicated in this MD. Duplication is always bad.

There is no intention of duplicating normative requirements. As I said, this is an experiment to see if we can gain clarity by separating requirements for reading systems from requirements for document authors. I'm also using the process to identify areas of OCF where the requirements on reading systems are unclear or absent.

murata2makoto:
"The first file in the OCF ZIP Container MUST be the mimetype file. " in the proposed MD is obviously a duplication of information.

dauwhe:
> "The first file in the OCF ZIP Container MUST be the mimetype file. " in the proposed MD is obviously a duplication of information.

There is an authoring requirement here, obviously. It's less clear if there's a requirement on the reading system. I hope this effort clarifies such issues.

Note that there is such duplication in the existing OCF spec.

> The OCF Abstract Container MUST include a directory named META-INF that is a direct child of the container's Root Directory.

> All OCF Abstract Containers MUST include a directory called META-INF in their Root Directory.

But as I've said repeatedly, this is experimental work.

mattgarrish:
I'm not sure I follow what the harm is in having a document that identifies just the reading system requirements in OCF.

If we're going to get into testing conformance, someone's going to have to go through and do this eventually.

We've discussed (editorially) whether to raise separating reading system processing requirements from the specifications as a goal for 3.0.1/3.1, but each time decided to stay the course due to the scope of such a change. Personally, I think it could make comprehension a lot easier if we weren't constantly mixing the authoring and RS concerns, and would make an interesting experiment for a future revision to see what it would look like.

PR #1230, opened 2019-02-09
Add EPUB CFI

mattgarrish:
Here's a conversion of the specification to respec. It also changes the one 3.1 reference to a generic latest EPUB 3 reference.


murata2makoto:
Should we reference this specification from EPUB 3.2 (the umbrella specification)?

mattgarrish:
We dropped it because the specifications make no reference to it anymore, as it's basically become a specification that reading systems might implement for their own internal use.

murata2makoto:
Thanks.  Understood.

Issue #1239, opened 2019-02-27
Limit the FXL SVG restrictions to top-level SVG content docs


rdeltour:
Currently section 6.5.2 Expressing in SVG seems to apply to all SVG Content Documents. That includes SVG that are embedded as images in XHTML, or in iframes, or referenced in other SVGs.

If I understand correctly this section is primarily meant for ensuring that the SVG scales properly in the initial viewport of Reading Systems, so I don't think it is relevant to embedded SVGs, only to top-level SVGs (i.e. SVGs in the `spine`).

mattgarrish:
> so I don't think it is relevant to embedded SVGs, only to top-level SVGs

Right, that's what the definition of a fixed-layout document states:

> Fixed-Layout Document
> An EPUB Content Document directly referenced from the spine that has been designated pre-paginated in the Package Document, as defined in The rendition:layout Property [Packages32].
> The dimensions to use for rendering Fixed-Layout Documents are defined in Fixed Layouts [ContentDocs32].

Would changing the first link from "SVG Content Documents" to the definition help? (e.g., "SVG Fixed-Layout Documents")

mattgarrish:
Fixed by #1245.

Issue #1236, opened 2019-02-22
SVG FXL restrictions on `height` and `width` may be too restrictive


rdeltour:
Looking at the rules on how the [Initial Containing Block dimensions are expressed in SVG](https://w3c.github.io/publ-epub-revision/epub32/spec/epub-contentdocs.html#sec-fxl-icb-svg)

1. regarding their combination with `viewBox`, the spec says:

The spec says:
> It is _RECOMMENDED_ to use only the `viewBox` attribute to ensure proper rescaling of the SVG Content Document, as needed.

As far as I understand, percentage values **do not negatively affect scalability**, so should be fine in combination with `viewBox`.

This could be reworded to something along these lines:

> It is _RECOMMENDED_ to use the `viewBox` attribute to ensure proper rescaling of the SVG Content Document, as needed.
> If `height` and `width` attribute are present, they values _SHOULD NOT_ be set to _absolute length_ when a `viewBox` attribute is present.

2. regarding the `height` and `width` pixel value

The spec says:
> If only the `x`, `y`, `height` and `width` attributes are present, they _SHOULD_ be defined in pixel values (establishing the ICB)

Given that all [_absolute length units_](https://www.w3.org/TR/css3-values/#absolute-lengths) are _compatible_ (they can all be converted to the canonical `px` unit), requiring a pixel value looks restrictive. The term "pixel value" itself might be a bit vague (does it mean "in pixel units" or "in units compatible with the pixel unit"?).

This could be open to any absolute length:
> If only the `x`, `y`, `height` and `width` attributes are present, their value _SHOULD_ be expressed in _user units_ or in _absolute length units_. (establishing the ICB).

3. regarding the impact of `x` and `y` on the ICB

The spec says:
> If only the `x`, `y`, `height` and `width` attributes are present, they _SHOULD_ be defined in pixel values (establishing the ICB)

I'm not sure the value of `x` and `y` is relevant there? The [intrinsic sizing](https://www.w3.org/TR/SVG/coords.html#SizingSVGInCSS) of SVG content is only impacted with `width`, `height`, and `viewBox`.

Suggested rewording:
>  If `height` and `width` attributes are present and there is `viewBox` attribute is not present, they _SHOULD_ be defined in…

For reference, this section was discussed in #860, #800, #463.
See also EPUBCheck issue w3c/epubcheck#902 and PR w3c/epubcheck#973

GeorgeKerscher:
Hello,
 
[...]

PR #1245, opened 2019-03-08
SVG fixed layout clarifications

mattgarrish:
This PR actually addresses two items that were resolved on the 2019-03-07 call:

- as raised in #1239, it changes the terminology links from "XHTML/SVG Content Documents" to "XHTML/SVG _Fixed-Layout_ Documents" to reduce any confusion that these rules apply to all content documents.
- as raised in #1236, it reverts the SVG ICB dimensions section to "MUST" use the viewBox attribute (noting that additional discussions about this next week may further change the section)

### DIFF
--- a/epub32/spec/epub-contentdocs.html
+++ b/epub32/spec/epub-contentdocs.html
@@ -112,11 +112,11 @@
 				<section id="sec-overview-relations-svg">
 					<h3>Relationship to SVG</h3>
 					<p>This specification does not reference a specific version of [[SVG]], but instead uses an undated
-						reference. Whenever there is any ambiguity in this reference, the latest recommended specification is the authoritative reference.</p>
-					<p>This approach ensures that EPUB
-						will always keep pace with changes to the SVG standard. Authors and Reading System developers
-						will need to keep track of changes to the SVG standard, and ensure that their processes and
-						systems are kept up to date.</p>
+						reference. Whenever there is any ambiguity in this reference, the latest recommended
+						specification is the authoritative reference.</p>
+					<p>This approach ensures that EPUB will always keep pace with changes to the SVG standard. Authors
+						and Reading System developers will need to keep track of changes to the SVG standard, and ensure
+						that their processes and systems are kept up to date.</p>
 
 					<div class="caution">
 						<p>As SVG evolves, it is possible that features that were valid in previous versions could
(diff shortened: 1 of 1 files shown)


dauwhe (approved):

rdeltour (approved):
Approving since the PR effectively reverts the language to what it was in 3.0.1, as was agreed in the CG call.

However, I still have an issue with:

> the ICB dimensions MUST be expressed using the `viewBox` attribute

Which I think is an impossible statement. My understanding is that `viewBox` –impacting the _intrinsic aspect ratio_– has an effect on the ICB only under some limited circumstances (e.g. absence of intrinsic width and height), so the above statement is confusing at best. But let's handle that in a separate issue.

mattgarrish:
> My understanding is that `viewBox` –impacting the _intrinsic aspect ratio_– has an effect on the ICB only under some limited circumstances

Isn't the point of fixed layouts that you have to specify the coordinate space, though? And for the reading system to be able to obtain that information to prepare the "page" prior to rendering?

That's what confuses me about saying viewbox is only a should. At what point is it no longer a fixed layout, or, conversely, is there really such a thing as a fixed-layout svg?

rdeltour:
> Isn't the point of fixed layouts that you have to specify the coordinate space, though? And for the reading system to be able to obtain that information to prepare the "page" prior to rendering?
>(…) conversely, is there really such a thing as a fixed-layout svg?

I don't know, and that's precisely where the spec isn't clear IMO. I think it's also a question of terminology, and I'm not convinced the spec is using the right one.

The [_initial containing block_](https://drafts.csswg.org/css2/visudet.html#containing-block-details) is defined as either the _viewport_ (for continuous media) or the _page area_ (for paged media). These are provided by the User Agent, so I don't understand how an FXL document can even define the ICB. I think what we're trying to say here is that we want to define the size of the [_actual viewport_](https://drafts.csswg.org/css-device-adapt/#the-viewport), or maybe how the document is constrained to the viewport (this is where the _intrinsic aspect ratio_ would play a role).

Anyhow. This is a complex issue, and after having dived a bit into it, I'm not convinced that we use the right terminology and that the spec is clear enough about both the intent and the way to reach it.

But I don't even know if SVG FXL documents are actually used in FXL EPUBs (aren't they all using an XHTML template, possibly embedding or referencing an SVG?). And I obviously couldn't claim to have enough expertise to recommend a better language. It's just that as a spec reader, following the spec links, with my understanding, this doesn't feel right 😊

mattgarrish:
> But I don't even know if SVG FXL documents are actually used in FXL EPUBs

I remember seeing a comment to this effect in one of the earlier threads. No issues about this have been raised by authors, so it's more a problem of theory. We should probably hammer out the other issues first.

And unfortunately this is one of a handful of IDPF specs I didn't have any involvement in, so I can't even provide any context.

PR #1247, opened 2019-03-14
update core media type requirements for picture and img

### LINKED ISSUE #1237: Fallback requirements for picture element
Now that the picture element is valid for use, there is a question for epubcheck about how to [determine when a manifest fallback needs to be applied](https://github.com/w3c/epubcheck/issues/781#issuecomment-466840604). The `media` attribute makes determining whether a core media type will be available complicated, since it all depends on the runtime environment which will apply.

If we require that a core media type be specified in the `img` element, though, then it doesn't matter what combination of formats and media queries are used in the source elements. The `img` element also represents the default, so a reading system will still have the option to choose the foreign resource(s). As the natural fallback, it seems like it should always specify the CMT to avoid manifest fallback requirements.

(It's probably also going to be best practice to have a CMT in the `img`, too, since there's no guarantee of support for picture/source yet.)

But, we need to know if the CG agrees with this conclusion, and we should make this requirement explicit in the fallback requirements, if so.

If not, how else can we reliably determine if a CMT will be available? Do we require that at least one source element specify a CMT and not specify a `media` attribute? Do we ignore `media` attributes and just hope for the best so long as a CMT is listed?

mattgarrish:
This PR fixes #1237 (again) as described in https://github.com/w3c/publ-epub-revision/issues/1237#issuecomment-472822055:

- when inside a `picture` the only time foreign resources can be referenced is from a `source` element that explicitly identifies that it references a foreign media type in its `type` attribute
- otherwise, the `img` element requires a core media type fallback whenever its `src` or `srcset` attributes reference foreign resources.

@rdeltour please verify that I have this right, and that the spec prose reflects your proposal.

### DIFF
--- a/epub32/spec/epub-contentdocs.html
+++ b/epub32/spec/epub-contentdocs.html
@@ -853,15 +853,24 @@
 								><code>img</code> element</a>, the following fallback conditions apply to its use:</p>
 
 					<ul>
-						<li>A Core Media Type Resource MUST be specified in its <code>src</code> attribute when it is
-							the child of a <a
-								href="https://www.w3.org/TR/html/semantics-embedded-content.html#the-picture-element"
-									><code>picture</code> element</a>.</li>
-						<li>A Core Media Type Resource MUST be specified in its <code>src</code> attribute when it
-							specifies a <code>srcset</code> attribute.</li>
-						<li>A <a href="epub-packages.html#sec-foreign-restrictions-manifest">manifest fallback</a>
-							[[Packages32]] MUST be specified in all other cases when a Foreign Resource is referenced
-							from its <code>src</code> attribute.</li>
+						<li><p>If it is the child of a <a
+									href="https://www.w3.org/TR/html/semantics-embedded-content.html#the-picture-element"
+										><code>picture</code> element</a>:</p>
+							<ul>
+								<li>it MUST reference Core Media Type Resources from its <code>src</code> and
+										<code>srcset</code> attributes, when those attributes are specified;</li>
+								<li>any child <a
+										href="https://www.w3.org/TR/html/semantics-embedded-content.html#the-source-element"
+											><code>source</code> elements</a> also MUST reference Core Media Type
+									Resources from their <code>src</code> and <code>srcset</code> attributes unless the
+									element specifies that it references a Foreign Resource media type in its
+										<code>type</code> attribute.</li>
+							</ul>
+						</li>
+						<li>Otherwise, it MAY reference Foreign Resources in its <code>src</code> and
+								<code>srcset</code> attributes provided a <a
+								href="epub-packages.html#sec-foreign-restrictions-manifest">manifest fallback</a>
+							[[Packages32]] is defined.</li>
 					</ul>
 
 					<p>The following [[!HTML]] elements can refer to <a href="epub-spec.html#sec-core-media-types"
--- a/epub32/spec/epub-contentdocs.html
+++ b/epub32/spec/epub-contentdocs.html
@@ -858,19 +858,18 @@
 										><code>picture</code> element</a>:</p>
 							<ul>
 								<li>it MUST reference Core Media Type Resources from its <code>src</code> and
-										<code>srcset</code> attributes, when those attributes are specified;</li>
-								<li>any child <a
+										<code>srcset</code> attributes, when those attributes are specified; and</li>
+								<li>each sibling <a
 										href="https://www.w3.org/TR/html/semantics-embedded-content.html#the-source-element"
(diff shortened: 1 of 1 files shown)


mattgarrish:
Updated prose can be read more easily here: https://raw.githack.com/w3c/publ-epub-revision/0e0781d8c5dacbcbd04b832ba6a542bca7b72230/epub32/spec/epub-contentdocs.html#confreq-resources-cd-fallback-img

epub32/spec/epub-contentdocs.html
```diff
@@ -853,15 +853,24 @@ <h3>Foreign Resource Restrictions</h3>
...
-						<li>A <a href="epub-packages.html#sec-foreign-restrictions-manifest">manifest fallback</a>
-							[[Packages32]] MUST be specified in all other cases when a Foreign Resource is referenced
-							from its <code>src</code> attribute.</li>
+						<li><p>If it is the child of a <a
+									href="https://www.w3.org/TR/html/semantics-embedded-content.html#the-picture-element"
+										><code>picture</code> element</a>:</p>
+							<ul>
+								<li>it MUST reference Core Media Type Resources from its <code>src</code> and
+										<code>srcset</code> attributes, when those attributes are specified;</li>
+								<li>any child <a
```
rdeltour: nitpick: "any child source elements" might need to explicit the parent context here, as the main subject is the `img` element.
We could say "any child `source` elements of this `picture` element" (referencing the wording introducing this list item) or "any sibling `source` elements of the `img` element".

Note btw that only the _preceding_ `source` siblings will be used to define the `img` source set. But we may not need to go to this level of detail…

epub32/spec/epub-contentdocs.html
```diff
@@ -853,15 +853,24 @@ <h3>Foreign Resource Restrictions</h3>
...
+										href="https://www.w3.org/TR/html/semantics-embedded-content.html#the-source-element"
+											><code>source</code> elements</a> also MUST reference Core Media Type
+									Resources from their <code>src</code> and <code>srcset</code> attributes unless the
+									element specifies that it references a Foreign Resource media type in its
```
rdeltour: nitpick: "specifies that it references a Foreign Resource media type in its `type` attribute" is understandable but might sound a little odd or hard to parse?
- we never explicitly specify what is a _Foreign Resource media type_ (is it any media type that is not a CMT? or the media type of a Foreign Resource that is correctly declared in the Package Doc?).
- also, a `source` element can't be said to "reference a (…) media type"
- it doesn't really take into account MIME type parameters, which are allowed in the `type` attribute value.

In HTML/WHATWG lingo, we could say something along the lines of:

> unless its `type` specifies a [MIME type](https://mimesniff.spec.whatwg.org/#mime-type) whose [essence](https://mimesniff.spec.whatwg.org/#mime-type-essence) is listed as a Core Media Type.

epub32/spec/epub-contentdocs.html
```diff
@@ -853,15 +853,24 @@ <h3>Foreign Resource Restrictions</h3>
...
+										<code>type</code> attribute.</li>
+							</ul>
+						</li>
+						<li>Otherwise, it MAY reference Foreign Resources in its <code>src</code> and
+								<code>srcset</code> attributes provided a <a
+								href="epub-packages.html#sec-foreign-restrictions-manifest">manifest fallback</a>
```
rdeltour: s/manifest fallback/manifest fallbacks/ ?

rdeltour (approved):
The changes do reflect my proposal accurately 👍 (I hope we got it right this time 😅).

I made a couple nitpicking comments inline, I'll let you decide whether they're worth an update or not.

epub32/spec/epub-contentdocs.html
mattgarrish: Blah, right, "child" was from an earlier drafting. I'll change it to sibling.

epub32/spec/epub-contentdocs.html
mattgarrish: If we use "essence", everyone has to take a trip to another specification... :)

How about we just keep this in epub parlance by restating it as:

> unless it specifies a media type in its `type` attribute that is not a Core Media Type.

rdeltour: > How about we just keep this in epub parlance by restating it as:
> 
> > unless it specifies a media type in its `type` attribute that is not a Core Media Type.

yes, works for me!

> If we use "essence", everyone has to take a trip to another specification... :)

meh, where's the fun when _not_ doing that? 😁🤓

dauwhe (approved):

PR #1249, opened 2019-03-25
remove restrictions on annotation-xml

### LINKED ISSUE #1212: clarify the content model of MathML `annotation-xml`
the current language is the _content conformance_ section for Embedded MathML says:

> The `math` element MUST contain only Presentation MathML, with the exception of the `annotation-xml` element as defined below.

and then it defines the constraints for how Content MathML and XHTML can be used in said `annotation-xml` element.

However, it is a bit vague as to whether other markup can appear in `annotation-xml`.

- can SVG be used as `annotation-xml`? with an SVG media type or only as embedded in XHTML?
- can presentation MathML be used as `annotation-xml`? (I don't know if it's common practice but in theory it might be used to store an equivalent markup of a formula with the `contentequiv` keyword?).

The Nu Html Checker allows both of the above. Current EPUBCheck rejects both.

Can the spec be clarified?

mattgarrish:
Fixes #1212 by removing the restrictions on the `annotation-xml` element.

### DIFF
--- a/epub32/spec/epub-contentdocs.html
+++ b/epub32/spec/epub-contentdocs.html
@@ -681,8 +681,8 @@
 						<dl class="conformance-list">
 							<dt id="math-pres">Presentation MathML</dt>
 							<dd><p id="confreq-mathml-pres">The <code>math</code> element MUST contain only <a
-										href="https://www.w3.org/TR/MathML3/chapter3.html">Presentation MathML</a>, with
-									the exception of the <code>annotation-xml</code> element as defined below.</p></dd>
+										href="https://www.w3.org/TR/MathML3/chapter3.html">Presentation
+								MathML</a>.</p></dd>
 							<dt id="math-cont">Content MathML</dt>
 							<dd><p id="confreq-mathml-annot-cont"><a href="https://www.w3.org/TR/MathML3/chapter4.html"
 										>Content MathML</a> MAY be included within MathML markup in XHTML Content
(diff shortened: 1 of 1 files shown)


dauwhe (approved):

epub32/spec/epub-contentdocs.html
```diff
@@ -681,8 +681,8 @@ <h4>Content Conformance</h4>
 						<dl class="conformance-list">
 							<dt id="math-pres">Presentation MathML</dt>
 							<dd><p id="confreq-mathml-pres">The <code>math</code> element MUST contain only <a
-										href="https://www.w3.org/TR/MathML3/chapter3.html">Presentation MathML</a>, with
-									the exception of the <code>annotation-xml</code> element as defined below.</p></dd>
```
rdeltour, line 685: shouldn't we keep something along the lines of 

> with the exception of annotation content within the `annotation-xml` element

?

mattgarrish: I don't think it's necessary to say, is it? annotation-xml is a presentation mathml element. Now that there are no restrictions on it, there's nothing more to say. The restriction is still required to deal with content mathml, as described in the next section, plus you can't just embed arbitrary xml inside the math element.

rdeltour: yeah it was just to avoid any confusion with:

> MUST contain only presentation MathML

If `annotation-xml` is used there can be non-Presentation content in the descendants. But maybe it's just clear enough as is…

mattgarrish: I honestly can't get worked up about this one. If you think it's clearer, let's just add it back.

Issue #1217, opened 2019-01-16
xml:base attribute


dauwhe:
So... EPUB Reading Systems [MUST support](https://w3c.github.io/publ-epub-revision/epub32/spec/epub-spec.html#confreq-rs-xml-base) the XML Base specification, which defines the `xml:base` attribute.

>  It MUST be a conformant application as defined by [XMLBase].

But... browsers have deprecated and removed support for `xml:base`. Blink removed it in 2015, Firefox a bit later. It's still normatively referenced in W3C HTML5, but only allowed in the XHTML serialization. Use counters on the web put its usage at essentially zero. It sounds like it was removed from SVG. It was removed from the DOM spec. It's not in WHATWG HTML5. 

Is this allowed by any of the XML docs used by EPUB?

murata2makoto:
SVG 2 (CR) does not have xml:base, but SVG 1.1 (REC) does.

dauwhe:
Should we add a non-normative note saying that support is required due to SVG 1.1, but that the feature is being removed from both html and svg, and advise people to avoid it?

murata2makoto:
>Should we add a non-normative note saying that support is required due to SVG 1.1, but that the feature is being removed from both html and svg, and advise people to avoid it?

I do not understand why such a statement is needed.  It is certainly allowed in SVG, and will probably be never removed.  Does XHTML5 really disallow this attribute?  (I opposed to xml:base, but it has been there for a long time.)

dauwhe:
> I do not understand why such a statement is needed. It is certainly allowed in SVG, and will probably be never removed. Does XHTML5 really disallow this attribute? (I opposed to xml:base, but it has been there for a long time.)

It has been removed from SVG2, WHATWG HTML5 (all serializations), and the DOM spec. Browsers don't support it. It remains in SVG 1.1 and in the XHTML serialization of W3C HTML5. If I was building a reading system today, I would not want to have to support this. I think a note is reasonable warning that this feature is unlikely to work as advertised, even if technically supported by the spec.

murata2makoto:
Has it been removed from WHATWG HTML?  The spec is silent about it.  One could argue that xml:base is allowed since the use of XML is allowed.  WHAG DOM is also silent.   Since resolution of relative references is not specified as part of WHATWG DOM, one could argue that xml:base is allowed.

iherman:
[...]

PR #1253, opened 2019-04-10
Add note discouraging use of xml:base

dauwhe:
Per #1217, add a note advising authors to avoid using XML base. It's been removed from blink and gecko. I have no information on its status in webkit.

### DIFF
--- a/epub32/spec/epub-spec.html
+++ b/epub32/spec/epub-spec.html
@@ -523,7 +523,7 @@
 						<p id="confreq-rs-xml-ns"> It MUST be a <a
 								href="https://www.w3.org/TR/2009/REC-XML-NAMES-20091208/#ProcessorConformance"
 								>conformant processor</a> as defined in [[!XML-NAMES]].</p>
-						<p id="confreq-rs-xml-base">It MUST be a conformant application as defined by [[!XMLBase]].</p>
+						<p id="confreq-rs-xml-base">It MUST be a conformant application as defined by [[!XMLBase]]. <span class="note">Note: browsers have removed support for [[!XMLBase]]. Authors should avoid using this feature.</span></p>
 					</dd>
 				</dl>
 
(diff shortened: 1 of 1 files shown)


epub32/spec/epub-spec.html
```diff
@@ -523,7 +523,7 @@ <h2>EPUB Reading Systems</h2>
 						<p id="confreq-rs-xml-ns"> It MUST be a <a
 								href="https://www.w3.org/TR/2009/REC-XML-NAMES-20091208/#ProcessorConformance"
 								>conformant processor</a> as defined in [[!XML-NAMES]].</p>
-						<p id="confreq-rs-xml-base">It MUST be a conformant application as defined by [[!XMLBase]].</p>
+						<p id="confreq-rs-xml-base">It MUST be a conformant application as defined by [[!XMLBase]]. <span class="note">Note: browsers have removed support for [[!XMLBase]]. Authors are advised to avoid using this feature.</span></p>
```
mattgarrish: There's just something weird about introducing the word browsers here when we normally use reading system. I know browser core -> reading system, but it's not a connection made in the specs. What about maybe making that clearer, like: "Note: Authors are advised to avoid using this feature as the browser cores Reading Systems are built on no longer support [[XMLBase]]." ?

dauwhe: I was wondering about that. I'll adjust the wording. 

What I should really do is try to make a test file. I'm doubting we have any support now.

dauwhe: Changed to 

> [[HTML]] and [[SVG]] are removing support for [[!XMLBase]]

Which is more accurate, and expresses the risk authors face much better.

mattgarrish (approved):

murata2makoto:
I think that "It MUST be a conformant application as defined by XMLBase" and "Note: HTML and SVG are removing support for XMLBase." are in conflict.  Shouldn't we remove the first sentence?

dauwhe:
A note is non-normative, and is just offering advice to authors that the use of xml:base in HTML and SVG is likely not going to work. I fear that some people may be using xml:base in other XML vocabularies that we don't know about, and so am wary of entirely dropping this requirement without further research.

murata2makoto:
Dropping the first sentence does not prohibit applications from supporting xml:base but rather allow them NOT to support it.   I think that the current wording is obviously conflicting.

Issue #1195, opened 2019-01-03
Reading system conformance for signatures


dauwhe:
OCF includes digital signatures. Are reading systems obligated to support this? Do they have to validate signatures? If the signature is invalid, must the reading system not allow the user to read the content? What if only some content is invalid?

GarthConboy:
My answers to the above: No, No, No, N/A

murata2makoto:
No for the first two questions.  For the 3rd question, I think that RSs should report errors but are not required to do so.

In the future, we might want to introduce a class of OPC processors that are required to reject EPUB publications containing invalid signatures.

dauwhe:
So our position is that reading systems can ignore digital signatures? We should at least make that clear in the spec.

llemeurfr:
> So our position is that reading systems can ignore digital signatures? 

It depends on what we really mean by that. The fact is that most reading systems don't handle digital signatures today. I even wonder if any RS supports that. 
But digital signatures are a way to stop undue modification of content, at any point in a distribution chain. If a provider takes the pain to add a signature to an EPUB file, there must be a reason. If the spec states that it's a proper behavior for reading systems to simply ignore the DS, we should rather suppress DS from the spec. 

I'd prefer the spec to state that RS SHOULD report an error if DS is erroneous. It MAY let the user read the publication after the user has been notified. 

Therefore,  RS may still decide that supporting DS is not needed depending their market, but at least the spec will be consistent.

GarthConboy:
I'd be largely okay with that, but

`I'd prefer the spec to state that RS SHOULD report an error if DS is erroneous. It MAY let the user read the publication after the user has been notified.`

Is really saying that the RS MUST display a notification before displaying a bad-DS EPUB.  I prefer:

`I'd prefer the spec to state that RS SHOULD report an error if DS is erroneous. It MAY let the user read the publication.`

dauwhe:
We should certainly allow signatures, as there might be some workflows and reading systems that actually use them. But I think we have an obligation to content authors to warn them that just adding a signature will have no effect unless a reading system chooses to implement something.  So in addition to Garth's text, I would add something like

> Note: Adding a digital signature is not a guarantee that an EPUB cannot be tampered with, since Reading Systems are not required to check signatures.

dauwhe:
Closed via #1254

PR #1254, opened 2019-04-11
add note about digital signatures per #1195

dauwhe:


mattgarrish (commented):
I'd just suggest stripping "Note:" from the beginning, as respec automatically adds a note header making it redundant.

Issue #1240, opened 2019-02-27
Clarify the definition of Content Documents


rdeltour:
I'm likely splitting hairs, but it doesn't feel right that the definition of [Content Documents](http://w3c.github.io/publ-epub-revision/epub32/spec/epub-spec.html#dfn-epub-content-document) depends on _conformance_. 

The def currently reads:
> A Publication Resource that conforms to one of the EPUB Content Document definitions (XHTML or SVG).

Taken literally, it implies that non-conforming XHTML is not a Content Document, problem solved 🙃.

Maybe this can be clarified a little to remove any ambiguity? This is low priority.

mattgarrish:
Since I suspect I know where you're going with this (svg fonts), what about:

> A Publication Resource with an XHTML or SVG media type that contains all or part of the content of an EPUB Publication (i.e., the textual, visual and/or audio content). These resources have to conform to their respective XHTML or SVG definitions to be used in the `spine` or be referenced from another EPUB Content Document.
>
> An EPUB Content Document is a Core Media Type Resource, so can be included in the EPUB Publication without the provision of fallbacks.

That would make conformance dependent on usage, and also note the required conformance when they are used as "content" documents.

rdeltour:
> That would make conformance dependent on usage,

and would incidentally make it difficult to differentiate in EPUBCheck, which would likely treat all the OPF items as Content Docs (as it does today), but I think we can live with it.

murata2makoto:
> Taken literally, it implies that non-conforming XHTML is not a Content Document,

Sure.  It is not a content document.  What's wrong about this?

Standards define conformance.  Certainly, non-conformant things exist in the real world.  But I do not believe that standards should be concerned with non-conformant things.

rdeltour:
> > Taken literally, it implies that non-conforming XHTML is not a Content Document,
> 
> Sure. It is not a content document. What's wrong about this?

Maybe it's just me… I feel like defining the concept in terms of whether it is conforming or not feels backwards to me. It basically means that terminology only applies to _conforming_ instance, making statements like "your Content Document is non-conforming to EPUB" actually contradictory.

I'd prefer if conformance was left to the related "Conformance" (or "Conformance classes") sections of the spec, rather than part of the terminology itself.

But again, it's really low priority, so let's not waste too many brain cycles on this 😊

murata2makoto:
@rdeltour 

[...]

PR #1255, opened 2019-04-11
update definition of EPUB Content Documents per #1240

dauwhe:


mattgarrish (approved):

murata2makoto:
Do we now have circular definitions?  The def of "content document" relies on "SVG content document" and "XHTML content document", both of which rely on "content document".

mattgarrish:
"epub content document" is really nothing more than a shorthand way of talking about XHTML and SVG Content Documents at the same time. There's always going to be an element of circularity to them, but I'd argue it's far less circular now since the revised text spells out in more detail what makes content documents unique kinds of publication resources.

murata2makoto:
I withdraw my comment.  When a subsequent version is created, I might raise a minor comment.

PR #1260, opened 2019-04-17
reword beginning of overview per #1186

dauwhe:


mattgarrish (approved):
I still wonder if we'll prove to be more responsible for warnings than other standards (or epubcheck) because of the newly-deprecated stuff, but this at least suggests that possibility.

PR #1261, opened 2019-04-18
Update epub-changes.html

nekennedy:
The display-seq and meta-auth properties *in the are* deprecated. For display sequence, precedence now follows document order.

### DIFF
--- a/epub32/spec/epub-changes.html
+++ b/epub32/spec/epub-changes.html
@@ -208,7 +208,7 @@
 					<li>The <a href="epub-packages.html#viewport"><code>rendition:viewport</code></a> property is
 						deprecated.</li>
 					<li>The <a href="epub-packages.html#sec-display-seq"><code>display-seq</code></a> and <a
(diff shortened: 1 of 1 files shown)


nekennedy:
Actually, I'm not sure this is the best re-wording, since Matt made this change to remove the official "deprecated" flag it might need to be more substantially re-written: https://github.com/w3c/publ-epub-revision/commit/5feb39383b96507a5732b58601cbaa28e8cebbd6

mattgarrish:
It would probably be good to split it into two bullets, like this:

```
<li>The <a href="epub-packages.html#sec-meta-auth"><code>meta-auth</code></a> property is deprecated.</li>
<li>The <a href="epub-packages.html#sec-display-seq"><code>display-seq</code></a> property no longer applies to <A href="epub-packages.html#sec-opf-dccreator">creators</a>. Display precedence now follows document order.</li>
```

dauwhe:
I'll merge this and then split into two bullets. Thanks @nekennedy !

nekennedy:
Thanks Dave! Credit goes to Alex Grover for catching that one :)

PR #1265, opened 2019-05-01
update Juli's affiliation

dauwhe:


murata2makoto:
Could you also change my affiliation?  It is now "Keio Advanced Publishing Laboratory".

dauwhe:
> Could you also change my affiliation? It is now "Keio Advanced Publishing Laboratory".

Fixed via #1266 :)

Thanks!

PR #1326, opened 2020-08-31
add old specs to archive directory

mattgarrish:
Here's a thought on how to declutter the repository root but keep everything together in one place. We've used an archive directory in the past for this purpose, so could be a useful dumping ground for the specs we're not anticipating taking up again (at least anytime soon) and any finished revisions. The root could then be reserved for active work.


mattgarrish:
See: https://github.com/w3c/publ-epub-revision/tree/archive-specs

wareid (approved):

iherman:
I presume that, eventually, 32 will also end up in the archives. But this may need a formal WG resolution whereby our publication will be 3.3. (and we would have a 33 directory at the root level).

iherman (approved):

shiestyle (approved):

mattgarrish:
Right, until we sort out the publishing of the next version I'm just leaving the 3.2 stuff where it is, but a future candidate, for sure... :)

dauwhe (approved):

PR #1330, opened 2020-10-01
EPUB 3.3 documents

mattgarrish:
This PR adds the following 3.3 documents:

- epub-core.html - the authoring requirements for publications, merging what was the core, packages, content documents, ocf and media overlays specifications
- epub-rs.html - basic extraction of reading system requirements, keeping the same heading structure to minimize initial confusion about what is coming from where (this will need more work as we go, obviously, but hopefully is enough to start showing the intended direction).
- epub-overview.html - the informative overview document (leaving any mention of changes until we're later in the process)
- epub-changes.html - placeholder changes doc


mattgarrish:
Forgot to note that it also moves the epub32 directory to the archive.

dauwhe:
OMG Matt this is so cool! To preview this, I switched to this branch locally so I could view the HTML directly. I'll take some time to look through it, but I think we should merge this as soon as possible. People can always look at 3.2 to see how it used to be.

dauwhe (approved):

mattgarrish:
> I switched to this branch locally so I could view the HTML directly.

Oops, ya, I probably could have given preview links for these:

- Core: https://raw.githack.com/w3c/publ-epub-revision/epub33/prep/epub33/core/
- RS: https://raw.githack.com/w3c/publ-epub-revision/epub33/prep/epub33/rs/
- Overview: https://raw.githack.com/w3c/publ-epub-revision/epub33/prep/epub33/overview/
- Changes: https://raw.githack.com/w3c/publ-epub-revision/epub33/prep/epub33/changes/

iherman:
I think the structuring of the documents is great; I greatly prefer this setup than what we had in 3.2. My comment goes exclusively to the organization of the overall repository. 

I have already talked to Philippe le Hégaret, and it is possible to run ECHIDNA from a repository that contains several documents. It will require some level of wizzardy, but he (and, I presume, Denis) will help us. That is good news. I am more worried right now about preview; I am not sure how it works, and maybe it is worth asking before we commit ourselves. (I see that you have added a .pr-preview file, did you find some description on how one can control what is previewed?)

Regardless, I wonder whether it is not better if we have separate directories for the separate parts. Ie, something like

```
epub33/core/
epub33/rs/
epub33/overview/
epub33/a11y/
```

Note that this makes it possible to put the a11y document alongside all the others, not keeping it aside from the rest. I understand there are reasons why its numbering is different (though it still bothers me), but keeping it totally separate, as it is now, sends the wrong message.

WDYT?

wareid (approved):

mattgarrish:
> I see that you have added a .pr-preview file, did you find some description on how one can control what is previewed?

Yes, you can specify the path to any one file in the repository to preview; it doesn't always have to be an index.html file in the root. But we can't preview multiple unless we move to separate repositories.

Using separate subdirectories is fine with me, and I presume would mean we could call each file index.html and avoid the pending publishing headache of unique names. That won't cause any publishing issues in terms of the /common directory, though, will it? (In particular, having to copy /common into each subdirectory because any common css/images can't be in a sibling folder when published.)

I'll defer to @avneeshsingh and @GeorgeKerscher on whether the repository structure of having the accessibility specification under the main epub33 folder is problematic. In one sense, it's just a directory, but it infers a stronger relationship to EPUB 3 than exists in the actual document. (The flip side is that in many ways, that's probably true given the progression to EPUB 3 and the limitations of EPUB 2.)

iherman:
> Yes, you can specify the path to any one file in the repository to preview; it doesn't always have to be an index.html file in the root. But we can't preview multiple unless we move to separate repositories.

Which means that preview becomes pretty much unusable for us; if we make a PR on a single document, we want to preview that one. 

We may have to use the poor-man's preview that I did: https://github.com/iherman/preview

mattgarrish:
> We may have to use the poor-man's preview that I did: https://github.com/iherman/preview

Ya, we're definitely going to have to come up with something.

If we can anticipate the pull request being opened, I suppose another option would be to flip the document referenced from the config file prior to opening the request. That would rotate the preview onto the wanted document each time.

murata2makoto:
I understand that this reorg is an attempt to make requirements on RSs clearer.  Hopefully, some implicit requirements will become clearer by creating a separate spec.  But I am not sure if it is a good idea to merge specs.  

When EPUB 3.1 was developed, there were lots of discussions about its organization.   Merging specs was proposed and I objected. 

One problem of EPUB 3.0 was too complicated interdependencies of specifications.  (I wrote a program for making this problem clear. )  Merging specs hides this problem.  I will use my program to the EPUB 3.3 core spec again.

mattgarrish:
> One problem of EPUB 3.0 was too complicated interdependencies of specifications.

That's arguably one of the benefits of merging the documents. With a single document, there's an opportunity to look at the structuring and see if some complexities can be refined. We don't have to stop with a basic merging of bodies into separate sections.

But there's also a process case for reducing the number of documents this time, too. We snuck through 3.2 without having to worry about W3C publishing requirements since they weren't on REC track. If we have separate documents for each core section, however, we're going to have a difficult time reviewing editor drafts. Links between documents will go to their respective latest published versions instead of the editor drafts, as you can't publish a group of documents under a single TR path with relative references like we've historically done.

murata2makoto:
I studied interdependencies of EPUB 3.2 modules by studying occurrences of some terms, file names, and spec names.  Here is [the result.](https://1drv.ms/x/s!An5Z79wj5AZBgtJsyMe0PIAwZ0Qs0A?e=fUdpAr)  If you compare it and [the same analysis of EPUB 3.0](https://drive.google.com/file/d/0B0R4fOPwdMGWVzlCUnJTRnR2ZUU/view?usp=sharing), you will see major improvements.  This is primarily becaues navigation documents were moved from ContentDocument 3.2 to Publication 3.2.  I am going to do the same analysis of sections of EPUB 33  Core.

shiestyle (approved):
I think it's good to simplify the documents!

murata2makoto:
Here is [my analysis](https://1drv.ms/x/s!An5Z79wj5AZBgtMxEqnJK_IAkVDA1g?e=bNDyqE) of interdependencies of sections in the EPUB 3.3 Core draft.  I see further improvements over 3.2.  

I highlighted some terms that occur in several clauses.  But they are not necessarily problematic.  I hope that my analysis helps people to further improve the EPUB 3.3 Core draft.

dauwhe:
@mattgarrish please feel free to merge when you feel this is ready.

mattgarrish:
There's more to do, of course, but the changes I want to get to are best done now in separate PRs.

PR #1334, opened 2020-10-04
Add switch property back to manifest item vocabulary

### LINKED ISSUE #1311: switch property gone awol
I just noticed that the `switch` property is missing from the [manifest item properties section](https://www.w3.org/publishing/epub3/epub-packages.html#sec-item-property-values).

It was stripped from 3.1, so looks like we missed it while restoring the removed features.

mattgarrish:
Fixes #1311

### DIFF
--- a/epub33/core/vocab/item-properties.html
+++ b/epub33/core/vocab/item-properties.html
@@ -1,272 +1,310 @@
-
-<section id="app-item-properties-vocab" class="vocab">
-	<h3>Manifest Properties Vocabulary</h3>
-	<section id="elemdef-opf-meta-overview">
-		<h4>Overview</h4>
-		<section id="item-properties-vocab-about" class="informative">
-			<h5>About this Vocabulary</h5>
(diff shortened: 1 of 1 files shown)


iherman:
Just for my understanding: was this a bug in the creation of 3.2 in the sense that there is still a backward incompatibility v.a.v. 3.01 (introduced in 3.1)? And this was not deliberate in the process of defining 3.2 but only an omission?

mattgarrish:
> was this a bug in the creation of 3.2

The bug goes all the way back to when we created 3.1. We were originally going to remove a number of features entirely, including epub:switch. In the end, we agreed to return the features and deprecate. In doing that, though, we forgot to put back this property for stating that a manifest item has switch elements in it. epubcheck has never stopped allowing the property, which has been valid since 3.0; it's absence from the vocabulary was just never noticed all these years as content switching is not widely used.

There's an argument the property should also be deprecated, since the element is, but I think it's enough to only warn about the use of the element. A double warning is probably a bit excessive. But that would be a separate issue.

iherman (approved):

PR #1339, opened 2020-10-09
Remove export dependency on the common css

mattgarrish:
This PR inlines in the common css file so that there are no references to the common folder in the exported html. Will this be sufficient for publishing using the current structure @iherman ?

(I've also removed the style sheet from the documents that don't actually employ any of the styles.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/publ-epub-revision/pull/1339.html" title="Last updated on Oct 10, 2020, 11:29 AM UTC (2adbe7d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/publ-epub-revision/1339/4b704cb...2adbe7d.html" title="Last updated on Oct 10, 2020, 11:29 AM UTC (2adbe7d)">Diff</a>


iherman (approved):
This is a neat trick :-) But it works, I also tested it by running a respec on the massaged file and the result is exactly fine!

PR #1340, opened 2020-10-11
Rewrite reading system sections to conformance lists

### LINKED ISSUE #903: A11y: MUST MAY, MAYBE SHOULDN'T?
In section 4.4.1.3: "must meet the following requirements" 

The list is introduced with a MUST, but has a MAY. MUST MAY is a bad conformance requirement. Also, repetition of the keywords is confusing. Suggest either dropping the "MUST" from the body text, or move the MAY requirement to a new list of one item, introduced by "In addition, it MAY include the following". I think dropping the MUST is cleaner - "A conformant EPUB Publication has the following requirements". Although still odd - a MAY is explicitly not a requirement. Can it just be dropped?

mattgarrish:
This is another iterative step on turning the reading system specification into a more readable document. It primarily just removes a lot of the excessive sectioning that was copied over from the original pull-out of requirements and replaces them with conformance lists.

A good example of this is the [metadata section](https://raw.githack.com/w3c/publ-epub-revision/rs/cleanup/epub33/rs/#metadata), which used to have subsections for every element and groups of elements.

In the interests of fixing some historical wrongs, it also fixes #903 by rewording the duplicative "XX MUST meet all of the following requirements" introductory paragraphs before the conformance lists (across all three major specifications).

We don't need to normatively assert that normative requirements have to be followed, that stands to reason, so the replacement wording is now simply "XX has to meet the following requirements". This reduces any confusion about why you MUST follow a requirement that is in fact only recommended or optional.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/publ-epub-revision/pull/1340.html" title="Last updated on Oct 11, 2020, 11:25 AM UTC (416c302)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/publ-epub-revision/1340/61ccd48...416c302.html" title="Last updated on Oct 11, 2020, 11:25 AM UTC (416c302)">Diff</a>

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -471,7 +471,7 @@
 						<p>Authors MAY include page navigation in reflowable EPUB Publications without statically
 							paginated equivalents.</p>
 
-						<p>A conformant EPUB Publication MUST meet the following criteria when it includes page
+						<p>A conformant EPUB Publication has to meet the following criteria when it includes page
 							navigation:</p>
 
 						<ul>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -20,23 +20,19 @@
 					name: "Garth Conboy",
 					company: "Google",
 					companyURL: "https://www.google.com"
-				},
-				{
+				}, {
 					name: "Dave Cramer",
 					company: "Hachette Livre",
 					companyURL: "https://www.hachettebookgroup.com",
-				},
-				{
+				}, {
 					name: "Marisa DeMeglio",
 					company: "DAISY Consortium",
 					companyURL: "https://www.daisy.org"
-				},
-				{
+				}, {
 					name: "Matt Garrish",
 					company: "DAISY Consortium",
 					companyURL: "https://www.daisy.org"
-				},
-				{
+				}, {
 					name: "Daniel Weck",
 					company: "DAISY Consortium",
 					companyURL: "https://www.daisy.org"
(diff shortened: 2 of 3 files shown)


iherman:
Comment that has nothing to do with the changes themselves: neither diff nor preview work. We may have to come up with an alternative...

mattgarrish:
> neither diff nor preview work

Was aws maybe down earlier? Links are working for me now.

Unfortunately, we only get the core spec, but that's why I linked to the metadata section. Oxygen's reformatting is getting flaky, so the github diff of the RS spec shows a lot of pointless formatting changes.

To be clear, though, I wasn't rewriting anything with the RS structure changes, just reducing sections to dt/dd tagging.

The only rewrites were the ones I mentioned above about normative musts leading into normative requirements.

iherman (approved):

iherman:
Yep, preview works now.

It is a pain that we can set only one source for preview... 

I would have to learn how to write a Github Webhook, an then turn my simple preview into a service that provides a URL for the display and diff for all our specs automatically. Going down the route of AWS is something I am not sure how to do (never did that) and I am not sure it is worth it. But it would be better if there was a simpler solution:-(

mattgarrish:
I gave your script a try, but it doesn't seem to work with a complex directory structure like we have. It expects a single index.html file in the root and in the pull request.

I could modify the URLs fed into the diff tool to get a link, though: [Reading Systems diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fw3c.github.io%2Fpubl-epub-revision%2Fepub33%2Frs%2F&doc2=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fraw.githack.com%2Fw3c%2Fpubl-epub-revision%2Frs%2Fcleanup%2Fepub33%2Frs%2Findex.html)

mattgarrish:
Maybe we can put template URLs into a wiki page for all the other specifications with instructions on how to insert the appropriate branch name?

The even poorer man's version... ?

Issue #645, opened 2016-01-15
Add Opus codec as core media type


mattgarrish:
A feature request for 3.1 is to add the Opus audio codec as a core media type:

Publishers that rely heavily on Media Overlays to provide synchronized text-audio publications would, in order to achieve reduced file sizes and increased audio quality, benefit from a speech-optimized format as a Core Media Type.

Opus would also represent a royalty-free option to MP3.

Discussions to date have been favourable, as Opus is supported, or indicated to be support, in all major browsers. The working group is not likely to come to a resolution prior to the first draft, however.

A note will be added to the core media type section that Opus is still under consideration and will be resolved in a future draft.

pettarin:
Why not including also OGG/Vorbis, Speex, SILK, FLAC then?

At least OGG/Vorbis and FLAC are widely supported.

caraya:
Matt,

Is there a technical reason why you chose Opus as the alternative to MP3 for audio? Has there been a comparison of formats that led to that conclusion?

mattgarrish:
We haven't actually added opus as a core media type at this point; I opened this issue so that everything under discussion is in the tracker.

A call was put out to the epub community to make feature requests prior to the start of the revision, and a speech-optimized audio format was one of the requests.[1] So far, the only interest expressed has been for opus.

A survey of DAISY consortium producers was done prior to the holidays and there's also outreach to mainstream audio publishers to determine what their needs are. What those efforts reveal is going to be important in making a decision on what to add, if anything. But we haven't returned to this issue in some time, as there are bigger items on the list of todos.

[1] https://docs.google.com/document/d/1QINuKE5Hsemll6G6tSQe6BaRuCHerFuwfXIRTgdaO7o/edit#heading=h.ar9g25b43eyd

pettarin:
For general-purpose, speech-optimized tasks, Opus has become the de-facto standard, merging and superseding e.g. SILK and Speex, so it is a reasonable choice.

On the other hand, for e.g. music files, having OGG/Vorbis and FLAC would be nice too.

lloydrasmussen:
[...]

PR #1341, opened 2020-10-12
Add audio/opus to audio core media types

mattgarrish:
Includes a link to issue 645, which we should keep open if we're not fully committed to inclusion yet.

A couple of issues I had with this PR, though:

- I've added audio/opus as the mime type, but I spotted the note about using audio/ogg for Opera. Since this is just the package document declaration, I assume we don't need to account for both.
- From what I've read, Opus is also supported in WebM and MP4 containers, but I assume we're only supporting Ogg since that's all that was discussed.

If I'm assuming wrong on either of these points, please chime in.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/publ-epub-revision/pull/1341.html" title="Last updated on Oct 13, 2020, 12:15 PM UTC (39ac4dc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/publ-epub-revision/1341/e8b20f4...39ac4dc.html" title="Last updated on Oct 13, 2020, 12:15 PM UTC (39ac4dc)">Diff</a>

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -24,13 +24,11 @@
 					"name": "George Kerscher",
 					"company": "DAISY Consortium",
 					"companyURL": "http://daisy.org"
-				},
-				{
+				}, {
 					"name": "Charles LaPierre",
 					"company": "Benetech",
 					"companyURL": "http://benetech.org"
-				},
-				{
+				}, {
 					"name": "Avneesh Singh",
 					"company": "DAISY Consortium",
 					"companyURL": "http://daisy.org"
@@ -39,8 +37,7 @@
 					name: "Romain Deltour",
 					company: "DAISY Consortium",
 					companyURL: "http://daisy.org"
-				},
-				{
+				}, {
 					name: "Jean Kaplansky",
 					company: "Invited Expert"
 				}, {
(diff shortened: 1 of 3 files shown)


dauwhe (commented):
Perhaps the note could provide a bit more context? 

> Core media types must be widely supported on all platforms used for EPUB Reading Systems. OPUS/OGG has good support in Android, MacOS, Windows, and Linux. Apple, starting with iOS 11, supports the OPUS codec, but only in a CAF container. The working group will monitor support for OPUS in iOS, and may remove OPUS as a core media type if we feel the level of support is inadequate.

iherman:
> Perhaps the note could provide a bit more context?
> 
> > Core media types must be widely supported on all platforms used for EPUB Reading Systems. OPUS/OGG has good support in Android, MacOS, Windows, and Linux. Apple, starting with iOS 11, supports the OPUS codec, but only in a CAF container. The working group will monitor support for OPUS in iOS, and may remove OPUS as a core media type if we feel the level of support is inadequate.

I agree with such a separate note; I believe we agreed on that on the call. If we do so, we can remove the reference to #645 from the table (which gets very distorted by this; we can refer to it from the note).

iherman (approved):
I believe this is the first clearly technical change on the spec (ie, not an editorial change). Maybe it is worth to set a changes' section in the document(s) right away and list this one there.

shiestyle (commented):
I just shared this topic to Japanese industry and I appreciate that this will be merged if nobody has negative feedback before the next call.

mattgarrish:
> Perhaps the note could provide a bit more context?

That's fine, but I think we should drop the first sentence. In part because if clarification about what core media types are is needed, we should raise that separately, but also because it's more nuanced (i.e., widely supported on reading systems that support the rendering of the resources, which is a mouthful for an issue).

> I agree with such a separate note

I'll disagree with this approach as it leads to issues being forgotten about. It's an open issue, so let's use the facilities that exist for them exactly so they're obvious to readers. If we move it to a note after the table, is anyone going to read that far on to find the additional context?

> Maybe it is worth to set a changes' section in the document(s) right away and list this one there.

Yes, I realized I forgot this late last night but didn't think anyone would notice before morning. :)

But it raised the question for me of whether we need a separate changes document anymore now that we only have one content specification and one reading system? The changes document was a place to collect all the major changes from across all the separate specifications. I'd much prefer to have a changes log in each document and have one less document to publish at the end.

mattgarrish:
Slightly branching off from the actual issue, but I've incorporated change logs for the three primary specs. We should have listed the spec re-org after formally adopting it, and for the accessibility document I've noted the ISO revisions have been incorporated.

iherman:
> 
> > I agree with such a separate note
> 
> I'll disagree with this approach as it leads to issues being forgotten about. It's an open issue, so let's use the facilities that exist for them exactly so they're obvious to readers. If we move it to a note after the table, is anyone going to read that far on to find the additional context?

You are right that the automatic Issue handling via respec is a good thing. We can use that (and be careful to document that issue properly!)

I think if it is put right after the table it is clearly visible, and it is editorially better in my view. That distortion of the table itself is ugly, and the issue is right after the table...
> 
> > Maybe it is worth to set a changes' section in the document(s) right away and list this one there.
> 
> Yes, I realized I forgot this late last night but didn't think anyone would notice before morning. :)
> 
> But it raised the question for me of whether we need a separate changes document anymore now that we only have one content specification and one reading system? The changes document was a place to collect all the major changes from across all the separate specifications. I'd much prefer to have a changes log in each document and have one less document to publish at the end.

I actually did not realize we have a separate changes document until you just said that:-) I absolutely agree that using the 'standard' approach of a change lot in each document is way better. Actually, I wonder if the pubrules checker would not complain if we do not have that section... So yes, i would think the old changes document could be just dumped as historical and start fresh with a changes section with 3.2 being the base line.

mattgarrish:
> I think if it is put right after the table it is clearly visible, and it is editorially better in my view.

Fair enough!

dauwhe (approved):

PR #1350, opened 2020-10-21
Clarify package version attribute value

### LINKED ISSUE #1349: Clarify version number is for EPUB 3
Although we've abandoned the idea of incrementing the package version number, the spec prose still makes it sound revision specific:

> The version attribute specifies the EPUB specification version to which the given EPUB Package conforms. The attribute MUST have the value "3.0" to indicate compliance with this version of the specification.

It might be helpful to change the last sentence more along the lines of:

>The attribute MUST have the value "3.0" to indicate conformance with EPUB 3.

mattgarrish:
Fixes #1349.

In addition to the change I mentioned in 1349, this PR also adds a note clarifying that new revisions are not new versions and that  we don't intend to modify the number, as this seems to always be a point of concern/confusion.

Comments on the approach welcome, as usual.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1350.html" title="Last updated on Oct 22, 2020, 3:20 PM UTC (ffe27c9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1350/a4dc3ea...ffe27c9.html" title="Last updated on Oct 22, 2020, 3:20 PM UTC (ffe27c9)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1292,7 +1292,14 @@
 
 						<p id="attrdef-package-version">The <code>version</code> attribute specifies the EPUB
 							specification version to which the given EPUB Package conforms. The attribute MUST have the
-							value "<code>3.0</code>" to indicate compliance with this version of the specification.</p>
+							value "<code>3.0</code>" to indicate conformance with EPUB 3.</p>
+
+						<div class="note">
+							<p>Updates to this specification do not represent new versions of EPUB 3 (i.e., each new 3.X
+								specification is a continuation of the EPUB 3 format). The working group is committed to
+								ensuring that changes do not invalidate existing content, allowing the
+									<code>version</code> atribute value to remain unchanged.</p>
+						</div>
 
 						<p id="attrdef-package-unique-identifier">The <code>unique-identifier</code> attribute takes an
 							IDREF [[!XML]] that identifies the <a class="codelink" href="#sec-opf-dcidentifier"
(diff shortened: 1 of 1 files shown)


iherman (approved):

wareid (approved):

dauwhe (approved):

shiestyle (approved):

epub33/core/index.html
```diff
@@ -1292,7 +1292,14 @@ <h5>The <code>package</code> Element</h5>
 
 						<p id="attrdef-package-version">The <code>version</code> attribute specifies the EPUB
 							specification version to which the given EPUB Package conforms. The attribute MUST have the
-							value "<code>3.0</code>" to indicate compliance with this version of the specification.</p>
+							value "<code>3.0</code>" to indicate conformance with EPUB 3.</p>
+
+						<div class="note">
+							<p>Updates to this specification do not represent new versions of EPUB 3 (i.e., each new 3.X
+								specification is a continuation of the EPUB 3 format). The working group is committed to
+								ensuring that changes do not invalidate existing content, allowing the
```
rdeltour: in practice, this is not strictly true. A few changes, sometimes to specifications not under the WG's control (like HTML), are not backward compatible. We had a few cases in 3.2 and it will likely happen again.
Could we maybe loosen this statement a bit?

mattgarrish: Being committed to something doesn't mean it can't happen. That's why I avoided strong wording that would suggest changes can't happen. Whether other specifications in other groups might change is getting a bit too deep into the weeds for a note.

But if the intention is not apparent enough, the latter sentence could maybe be modified to:

"The working group is committed to minimizing any changes that would invalidate existing content..."

rdeltour: FWIW, I prefer this latter wording 😊

PR #1363, opened 2020-10-26
Update prose about relationship to html to reflect change of standard

### LINKED ISSUE #1351: Editorial comments on the 'relationships to other specifications'
Two editorial comments on [§1.3 in Core](https://w3c.github.io/epub-specs/epub33/core/#sec-intro-relations) 

1. The text says, e.g., for HTML:

> This specification does not reference a specific version of W3C [HTML], but instead uses an undated reference that will always point to the latest recommendation.

However, a somewhat similar situation occurs with the CSS snapshot, which has a yearly update and which has an undated reference that we use in the spec. Should we use the same type of text, for the sake of consistency, for CSS?

2. There is no reference to MathML in this section. Discussions have also started to re-open the MathML WG; shouldn't we refer to the latest version of MatML the same way as we do for HTML, CSS, or SVG? I.e., shouldn't we add the same subsection in §1.3 for MathML?

mattgarrish:
This PR fixes #1351 as follows:

- removes the references to W3C HTML and versioning from the core and RS relationship sections
- mentions that the underlying technologies that HTML reference are also subject to change (MathML, SVG, CSS, JavaScript)

I discovered how we ended up with the MathML reference we have. "MathML" is already taken for referring to 1.0 and I can't override it in the biblio file (redefining gets ignored). "MathML3" uses the path for the latest MathML 3 version (/TR/MathML3) but specifically references 3.0 second edition because it's the current REC.

Looking at the HTML reference, it now goes to the draft spec for 3.0 third edition, so technically we have the right reference if that's how it's published.

What happens if there is a MathML 4, however, is beyond me, short of our coming up with some unique shorthand so we could make an even more generic reference to /TR/MathML and go back to what we were doing in IDPF.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1363.html" title="Last updated on Oct 26, 2020, 4:52 PM UTC (2bdac90)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1363/18dadfb...2bdac90.html" title="Last updated on Oct 26, 2020, 4:52 PM UTC (2bdac90)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -184,11 +184,11 @@
 
 					<p>The [[HTML]] standard is continuously evolving &#8212; there are no longer versioned releases of
 						it. That standard, in turn, references various technologies that continue to evolve, such as
-						MathML, SVG, CSS and JavaScript.</p>
+						MathML, SVG, CSS, and JavaScript.</p>
 
 					<p>The benefit of this approach for EPUB is that EPUB Publications always keep pace with changes to
-						the web without the need for new revisions. Authors, however, will need to keep track of the
-						various changes to HTML, and the technologies it references, and ensure that their processes are
+						the Web without the need for new revisions. Authors, however, will need to keep track of the
+						various changes to HTML and the technologies it references, and ensure that their processes are
 						kept up to date.</p>
 
 					<div class="warning">
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -181,23 +181,27 @@
 
 				<section id="sec-overview-relations-html">
 					<h4>Relationship to HTML</h4>
-					<p>This specification does not reference a specific version of W3C [[HTML]], but instead uses an
-						undated reference that will always point to the latest recommendation. This approach ensures
-						that EPUB will always keep pace with changes to the HTML standard. Authors will need to keep
-						track of changes to HTML, and ensure that their processes are kept up to date.</p>
 
-					<div class="caution">
+					<p>The [[HTML]] standard is continuously evolving &#8212; there are no longer versioned releases of
+						it. That standard, in turn, references various technologies that continue to evolve, such as
+						MathML, SVG, CSS and JavaScript.</p>
+
+					<p>The benefit of this approach for EPUB is that EPUB Publications always keep pace with changes to
+						the web without the need for new revisions. Authors, however, will need to keep track of the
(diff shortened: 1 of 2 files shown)


epub33/core/index.html
```diff
@@ -181,23 +181,27 @@ <h3>Relationship to Other Specifications</h3>
...
-					<p>This specification does not reference a specific version of W3C [[HTML]], but instead uses an
-						undated reference that will always point to the latest recommendation. This approach ensures
-						that EPUB will always keep pace with changes to the HTML standard. Authors will need to keep
-						track of changes to HTML, and ensure that their processes are kept up to date.</p>
 
-					<div class="caution">
+					<p>The [[HTML]] standard is continuously evolving &#8212; there are no longer versioned releases of
+						it. That standard, in turn, references various technologies that continue to evolve, such as
+						MathML, SVG, CSS and JavaScript.</p>
+
```
iherman, line 188: "SVG, CSS and Javascript" -> "SVG, CSS, and Javascript"

(In general, the W3C house style favors Oxford commas. This may require a more thorough editorial review...)

epub33/core/index.html
```diff
@@ -181,23 +181,27 @@ <h3>Relationship to Other Specifications</h3>
...
+					<p>The benefit of this approach for EPUB is that EPUB Publications always keep pace with changes to
+						the web without the need for new revisions. Authors, however, will need to keep track of the
+						various changes to HTML, and the technologies it references, and ensure that their processes are
```
iherman: the web -> the Web

(Again the W3C house style requires a capitalization of the 'Web')

iherman (approved):

epub33/core/index.html
mattgarrish: Heh, ISO rules picked the other way at these and "Web" in the a11y doc. Round and round...

We never bothered with serial commas in IDPF. Unless there's some particular ambiguity the lack of one introduces, there's no harm. It will take some review if we need to make usage consistent, as I certainly wasn't adding and I wasn't fixing in any other contributions.

Web is probably capitalized, though, as that was more a slip-up from still thinking in ISO terms.

iherman: > Heh, ISO rules picked the other way at these and "Web" in the a11y doc. Round and round...

One standard is good, more is better :-)

PR #1366, opened 2020-10-26
Fix uses of "attached" for properties

### LINKED ISSUE #1365: (Purely editorial) describing refines for meta
While reading through the chain of descriptions/references in, say, the [`<dc:creator>` element](https://w3c.github.io/epub-specs/epub33/core/#sec-opf-dccreator) I hit the sentence:

> The `file-as` property MAY be attached to include a normalized form of the name, and the alternate-script property to represent a creator's name in another language or script.

The problem is that the text does not say (only the example does) that 'attaching a property' means using the `<meta>` element via the `refines` attribute using a, well, property. There is reference to the [definition of `file-as`](https://w3c.github.io/epub-specs/epub33/core/#file-as) which has then a reference somewhere to the `<meta>` element, but it is bit up in the section hierarchy.

There is a description for the `<meta>` element [§4.2.3.3.4](https://w3c.github.io/epub-specs/epub33/core/#sec-meta-elem) but I did not find the expression "attach a property to an element" there either. This usage of `<meta>` is all over the place in the section on metadata, so I believe this terminology should be made more explicit.

mattgarrish:
Fixes #1365 - replaces "attached" with "associated with the element" and adds linking to where subexpressions are defined.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1366.html" title="Last updated on Oct 27, 2020, 4:28 PM UTC (26d1569)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1366/18dadfb...26d1569.html" title="Last updated on Oct 27, 2020, 4:28 PM UTC (26d1569)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1977,9 +1977,9 @@
 
 								<p>The [[!DC11]] <code>creator</code> element represents the name of a person,
 									organization, etc. responsible for the creation of the content of the
-										<a>Rendition</a>. The <a href="#role"><code>role</code> property</a> can be
-									attached to the element to indicate the function the creator played in the creation
-									of the content.</p>
+										<a>Rendition</a>. The <a href="#role"><code>role</code> property</a> can be <a
+										href="#subexpression">associated with the element</a> to indicate the function
+									the creator played in the creation of the content.</p>
 
 								<aside class="example">
 									<p>The following example shows how to represent a creator as an author using a <a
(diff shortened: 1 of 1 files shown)


iherman (changes requested):
I am fine with the changes, but I believe that is only part of what would be needed. The term "associated with", in conjunction with the `<meta>` element and the `refines` attribute, should be made explicit, e.g., at the definition of the `refines` attribute in the list of shared attributes and in the definition of the `<meta>` element.

mattgarrish:
I've added "association" into both the subexpression and refines attribute definitions.

iherman (approved):

PR #1368, opened 2020-10-26
Revised restriction on external identifiers

### LINKED ISSUE #1338: Problem with XML content requirements and external entities: make MathML and SVG1.1 invalid
The current text on [XML conformance](https://w3c.github.io/epub-specs/epub33/core/#sec-xml-constraints) says that an XML-based media _MUST_ meet:

> ...  
> External identifiers MUST NOT appear in the document type declaration [XML]  
> ...

This constraint apply regardless of whether the given Publication Resource is a Core Media Type Resource or a Foreign Resource.

However, that creates problems with SVG 1.1. The [conformance rules of SVG1.1](https://www.w3.org/TR/SVG11/conform.html) require the presence of: 

> `<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" SYSTEM "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">`

The [DTD itself](http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd), as described in the [relevant appendix](https://www.w3.org/TR/SVG11/svgdtd.html), is making a heavy use of external entities (via its usage of DTD Modularization). As a consequence, per the current document, _no conform SVG 1.1 file_ is valid for EPUB 3.3.

Although the upcoming SVG 2 removed this requirement, there are lots of perfectly valid SVG 1.1 out there (SVG 2 is not yet a Recommendation, b.t.w., only a CR). Most notably, Adobe Illustrator generates SVG 1.1 files with those DTD-s in them. 

A similar problem may occur with MathML 3, whose [official DTD](https://www.w3.org/Math/DTD/mathml3/mathml3.dtd) relies on the same approach (although MathML is more [permissive](https://www.w3.org/TR/MathML3/appendixa.html) and it does not _require_ the usage of that DTD).

This problem was originally raised in https://github.com/w3c/epubcheck/issues/1114.

mattgarrish:
Without getting into whether we should list other allowed doctypes, is there at least the possibility for agreement on this approach to updating the restriction @iherman @murata2makoto?

Fixes #1338.

[Reading Systems preview](https://raw.githack.com/w3c/epub-specs/fix/issue-1338/epub33/rs/)
[Reading Systems diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fw3c.github.io%2Fepub-specs%2Fepub33%2Frs%2F&doc2=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fraw.githack.com%2Fw3c%2Fepub-specs%2Ffix%2Fissue-1338%2Fepub33%2Frs%2Findex.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1368.html" title="Last updated on Nov 6, 2020, 5:25 PM UTC (2f80ee4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1368/378a906...2f80ee4.html" title="Last updated on Nov 6, 2020, 5:25 PM UTC (2f80ee4)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1016,12 +1016,11 @@
 					<li>
 						<p id="confreq-xml-identifiers">It MAY only include a <a
 								href="https://www.w3.org/TR/2008/REC-xml-20081126/#dt-doctype">document type
-								declaration</a> that references the <a
-								href="https://www.w3.org/TR/2008/REC-xml-20081126/#dt-pubid">public identifier</a> and
-								<a href="https://www.w3.org/TR/2008/REC-xml-20081126/#dt-sysid">system identifier</a>
-							for a format referenced in <a href="#app-identifiers-allowed"></a>, or that omits <a
+								declaration</a> that references an <a
 								href="https://www.w3.org/TR/2008/REC-xml-20081126/#NT-ExternalID">external
-								identifiers</a> [[!XML]].</p>
+								identifier</a> appropriate for its media type &#8212; as defined in <a
+								href="#app-identifiers-allowed"></a> &#8212; or that omits external identifiers
+							[[!XML]].</p>
 					</li>
 					<li>
 						<p id="confreq-xml-entities">It MUST NOT declare external <a
@@ -8767,20 +8766,27 @@ html.-epub-media-overlay-playing * {
 		<section id="app-identifiers-allowed" class="appendix">
 			<h2>Allowed External Identifiers</h2>
 
-			<p>The following table lists the public and system identifiers <a href="#confreq-xml-identifiers">allowed in
-					document type declarations</a>.</p>
+			<p>The following table lists the <a aria-label="public identifiers"
+					href="https://www.w3.org/TR/2008/REC-xml-20081126/#dt-pubid">public </a> and <a
+					href="https://www.w3.org/TR/2008/REC-xml-20081126/#dt-sysid">system identifiers</a> [[!XML]] allowed
+				in <a href="https://www.w3.org/TR/2008/REC-xml-20081126/#dt-doctype">document type declarations</a>.
+				[[!XML]]</p>
+
+			<p>These external identifiers MAY be used only in <a>Publication Resources</a> with the listed media types
+				specified in their <a href="#sec-manifest-elem">manifest</a> declarations. (Refer to <a
+					href="#confreq-xml-identifiers"></a> for more information.)</p>
 
 			<table>
 				<thead>
 					<tr>
-						<th>Format</th>
+						<th>Media Type</th>
 						<th>Public Identifier</th>
 						<th>System Identifier</th>
 					</tr>
 				</thead>
 				<tbody>
 					<tr>
-						<td>SVG 1.1</td>
+						<td>application/svg+xml</td>
 						<td>
 							<code>-//W3C//DTD SVG 1.1//EN</code>
 						</td>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7821,7 +7821,7 @@ html.-epub-media-overlay-playing * {
 
 			<p>These external identifiers MAY be used only in <a>Publication Resources</a> with the listed media types
 				specified in their <a href="#sec-manifest-elem">manifest</a> declarations. (Refer to <a
-					href="#confreq-xml-identifiers"></a> for more information.)</p>
+					href="#sec-xml-constraints"></a> for more information.)</p>
 
 			<table>
 				<thead>
@@ -7930,7 +7930,6 @@ html.-epub-media-overlay-playing * {
 				<aside class="example" id="ex.epubtype.note">
 					<p>The following example shows how a preamble could be marked up with the <code>epub:type</code>
 						attribute on its containing [[!HTML]] <code>section</code> element.</p>
-
 					<pre>
 &lt;html … xmlns:epub="http://www.idpf.org/2007/ops"&gt;
     …
@@ -7944,7 +7943,6 @@ html.-epub-media-overlay-playing * {
 				<aside class="example" id="ex.epubtype.gloss">
 					<p>The following example shows the <code>epub:type</code> attribute used to add glossary semantics
 						on an [[!HTML]] definition list.</p>
-
 					<pre>
 &lt;html … xmlns:epub="http://www.idpf.org/2007/ops"&gt;
     …
@@ -7958,7 +7956,6 @@ html.-epub-media-overlay-playing * {
 				<aside class="example" id="ex.epubtype.pg">
 					<p>The following example shows the <code>epub:type</code> attribute used to add pagebreak
 						semantics.</p>
-
 					<pre>
 &lt;html … xmlns:epub="http://www.idpf.org/2007/ops"&gt;
    …
@@ -8926,11 +8923,15 @@ EPUB/images/cover.png</pre>
 					3.2</a></h3>
 
 				<ul>
-					<li>5-Nov-2020: Generalized the restriction on custom attribute namespace URIs to exclude all of
-						idpf.org and w3.org. See <a href="https://github.com/w3c/epub-specs/issues/1388">issue
-						1388</a>.</li>
-					<li>12-Oct-2020: Added OPUS to the audio core media types with a warning that it is still subject to
-						review depending on support in Mac/iOS. See <a
+					<li>6-Nov-2020: A <a href="#app-identifiers-allowed">restricted set of external identifiers</a> are
+						now allowed in publication resources. <a href="#sec-xml-constraints">References to external
+							entities</a> from the internal DTD subset remain restricted, however. See <a
+							href="https://github.com/w3c/epub-specs/issues/1338">issue 1338</a>.</li>
+					<li>5-Nov-2020: Generalized the restriction on <a href="#sec-xhtml-custom-attributes">custom
+							attribute namespace URIs</a> to exclude all of idpf.org and w3.org. See <a
+							href="https://github.com/w3c/epub-specs/issues/1388">issue 1388</a>.</li>
+					<li>12-Oct-2020: Added OPUS to the <a href="#cmt-grp-audio">audio core media types</a> with a
+						warning that it is still subject to review depending on support in Mac/iOS. See <a
 							href="https://github.com/w3c/epub-specs/issues/645">issue 645</a>.</li>
 					<li>30-Sept-2020: The structure of the EPUB core specifications has been simplified to ease
 						understanding and access to information. This specification now consolidates the authoring
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1014,9 +1014,20 @@
 								Documents</a> [[!XML-NAMES]].</p>
 					</li>
 					<li>
-						<p id="confreq-xml-extmarkupdecl">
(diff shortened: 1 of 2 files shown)


murata2makoto:
I think that the direction is right.  It remains to create a list of public identifiers and system identifiers for document type declarations and to write some text about how RSes handle external DTD subsets.

Strictly speaking, the original text further disallows declarations of external unparsed entities and notations in internal subsets.  Although I don't think that they are used, I would like to mention them in the third bullet.

iherman:
@mattgarrish that works for me. Two remarks

1. Because this PR is a discussion that is based on overwriting a WG resolution, merging this PR should be subject of an explicit and documented WG resolution on a call. Cc @shiestyle @wareid @dauwhe 
2. As an overall reaction (without going into the details) on the content of Appendix B. my fundamental stand is that if there is any chance now or in the future that a well-known, _stable_ and, possibly standard DTD _might_ be used by any resource that _might_ be part of the EPUB instance, then we should allow it (even if it is not in use today). Adding a DTD to that table today is painless and harmless, a process of adding a DTD later may become more complicated. (That is why I am in favour of, for example, adding MathML that can be used as an exchange format in educational or scholarly publications even if not embedded directly in the HTML content.)

iherman:
Just an additional comment that came to my mind after I pushed the 'comment' button: we already have a bunch of registries and we will have to have these in future. Maybe the list of the DTD-s could be migrated to such a registry; this would make it easier to maintain them later. We can populate them with 1-2 DTD-s today and we can then move from there.

mattgarrish:
> 1. Because this PR is a discussion that is based on overwriting a WG resolution, merging this PR should be subject of an explicit and documented WG resolution on a call.

Oh, absolutely. I just wanted to start creating something concrete that we can take to the working group, as the original issue has grown a little too immense to wade through.

epub33/core/index.html
```diff
@@ -1014,9 +1014,20 @@ <h3>XML Conformance</h3>
...
+								<a href="https://www.w3.org/TR/2008/REC-xml-20081126/#dt-sysid">system identifier</a>
+							for a format referenced in <a href="#app-identifiers-allowed"></a>, or that omits <a
+								href="https://www.w3.org/TR/2008/REC-xml-20081126/#NT-ExternalID">external
+								identifiers</a> [[!XML]].</p>
+					</li>
+					<li>
+						<p id="confreq-xml-entities">It MUST NOT declare external <a
+								href="https://www.w3.org/TR/2008/REC-xml-20081126/#dt-parsedent">parsed entities</a> or
+							external <a href="https://www.w3.org/TR/2008/REC-xml-20081126/#dt-PE">parameter entities</a>
+							in an internal DTD subset [[!XML]].</p>
```
rdeltour: I think "[general entities](https://www.w3.org/TR/2008/REC-xml-20081126/#gen-entity)" is more appropriate for the former, instead of just "parsed entities", which is broader. 
Parameter entities are a kind of parsed entities. Here we want to disallow any external parsed entity, so we disallow both external general and parameter entities. 

Note that we would still allow external unparsed entities, FWIW.

mattgarrish:
> Maybe the list of the DTD-s could be migrated to such a registry

I thought of that, too, but it's been controversial because of validation. The remaining registries generally avoid defining aspects that require changes to epubcheck (or are tied to features that have largely died off, like specialized publication types). The only exception is the structural semantics vocabulary, but even for that we ultimately had to remove strict validation of the terms from the specification because they were causing problems with vendor ingestion.

I like the approach of being as thorough as we can now knowing what we know about what people actually use. My impression is most new specifications aren't defining doctypes/DTDs for their content anymore, and we can always loosen epubcheck in advance of another revision if it absolutely becomes important to add a new one (the wait for vendors to implement will always be outside our control, though).

epub33/core/index.html
```diff
@@ -1014,9 +1014,20 @@ <h3>XML Conformance</h3>
...
-							<a href="https://www.w3.org/TR/2008/REC-xml-20081126/#NT-ExternalID">External
-								identifiers</a> MUST NOT appear in the document type declaration [[!XML]].</p>
+						<p id="confreq-xml-identifiers">It MAY only include a <a
+								href="https://www.w3.org/TR/2008/REC-xml-20081126/#dt-doctype">document type
+								declaration</a> that references the <a
+								href="https://www.w3.org/TR/2008/REC-xml-20081126/#dt-pubid">public identifier</a> and
+								<a href="https://www.w3.org/TR/2008/REC-xml-20081126/#dt-sysid">system identifier</a>
+							for a format referenced in <a href="#app-identifiers-allowed"></a>, or that omits <a
+								href="https://www.w3.org/TR/2008/REC-xml-20081126/#NT-ExternalID">external
+								identifiers</a> [[!XML]].</p>
```
rdeltour: I find the wording a bit confusing: doesn't it mean that any XML document, regardless of what it contains or how it's declared in the Package Document, can contain any of the public/system ID defined in the list?

For instance, a non-SVG XML document declaring the typical SVG doctype would conform to this statement.

Can we somehow better tie the doctype declaration to the media type of the document it is declared in?

rdeltour (changes requested):
I think the direction is the right one, with the following caveats:

- editorial: s/parsed entity/general entity/
- editorial: I wonder how to better tie the list (or registry) of allowed doctype to the related formats
- general: while we're discussing that aspect of the spec, I think we should try to identify the conformance statements for RS processing. That can be done in a separate PR of course.

epub33/core/index.html
mattgarrish: Ya, I was thinking of that more in the table, but even here we could add something like "that is allowed for its media type" and then also add a media type column to the table.

epub33/core/index.html
murata2makoto: @rdeltour 

The taxonomy of entities in XML is extremely difficult to understand.  I believe that I read every sentence in the first edition of XML at least 20 times.  But this taxonomy is still confusing to me.

You correctly pointed out that parameter entities are parsed entities referenced within DTDs.  Stand corrected.  But general entities include unparsed entities as well.  So, "parsed entities" are not broader than "general entities".

IMHO, internal DTD subsets must not declare

1. external parameter entities
2. external general parsed entities
3. unparsed entities (which are always external and general)

The sum of the first three is "external entities".  We can simply say that declarations of external entities are disallowed in internal DTD subsets.  But this is just too difficult for everybody.

The original wording in EPUB 3 is quite good since it is free from the confusing taxonomy.

We might want to slightly change the original wording.  How about this?

- External identifiers MUST NOT appear in the internal DTD subset [XML].

mattgarrish: Is there terminology that could cover "all things external" so we don't have to get into a list of technical terms?

For example:

> "It MUST NOT reference external resources from the internal DTD subset [[!XML]].

mattgarrish: Ah, we were writing at the same time! I'm fine with any wording that encompasses all we need to restrict without being overly detailed.

mattgarrish:
So looking at [the current XML conformance requirements](https://w3c.github.io/epub-specs/epub33/rs/#sec-epub-rs-conf-xml), reading systems are already required to be non-validating processors (i.e., the current resolution is already handled).

Since external identifiers are largely still banned in content, and we only allow select safe doctypes, is there actually anything new we need to restrict on the RS side? Do we mention not resolving external identifiers just to be on the safe side?

iherman:
> Do we mention not resolving external identifiers just to be on the safe side?

I think so. This is kind of a obscure part of the spec, emphasizing can only improve it.

mattgarrish:
I've added preview/diff links for the RS spec to the first comment. Please review the current state of both documents everyone and let me know if we've finally reached a happy place on this.

(Let's take up the other DTDs to add in the existing open issues so we don't block this part trying to agree on those.)

mattgarrish:
Purely for github automation purposes, this PR fixes #1338.

rdeltour:
> I've added preview/diff links for the RS spec to the first comment. Please review the current state of both documents everyone and let me know if we've finally reached a happy place on this.

Looks good to me. If I were to nitpick, I would only reword:

>  External identifiers MUST NOT appear in the internal DTD subset

to:
> It  MUST NOT contain [external entity](https://www.w3.org/TR/2008/REC-xml-20081126/#dt-extent) declarations in the internal DTD subset

The only reason being that "external entity" has a proper definition in the XML spec, whereas "external identifier" has not (it's only kinda inferred from the combined definitions of _system identifier_ and _public identifier_). So we can point to that definition instead of pointing to the non-terminal symbol of the grammar 😉.

murata2makoto:
If we disallow external entities in the DTD subset, I  think that the conformance section does not have to disallow fetching of external entities.  But if we disallow it on the safe side, we need a note saying that external entities are already banned.

murata2makoto:
@rdeltour 

You are correct.  Another possibility is

- Public identifiers and system identifiers MUST NOT appear in the internal DTD subset.

This is technically equivalent.  To me, this is easier to understand.  Every user of XML understands SYSTEM and PUBLIC since they type these words often.  But different guys think different things easy.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2020-11-06-epub#section3).

- `RESOLVED:  Merge PR #1368 to address outstanding DTD issues, and close GH issues 1369-1373`
<details><summary><i class="differentiate">View the transcript</i></summary>
Wendy Reid: we had resolutions at the F2F, and further discussions on github<br/>
… and came to a happy place<br/>
Matt Garrish: <a href="https://github.com/w3c/epub-specs/pull/1368">https://github.com/w3c/epub-specs/pull/1368</a><br/>
Matt Garrish: where we ended up was…<br/>
… we put in an allowance for a specific set of external identifiers that we have put in an appendix<br/>
… we have SVG and MathML that are allowed to be used in content docs or in separate files<br/>
… and we made a restriction against external entities in the internal DTD subset<br/>
… so it prevents some security issues but eases authoring<br/>
… so we’ll no longer force people to remove SVG DTDs from tool-generated files<br/>
… I’m hoping this is it :)<br/>
Ivan Herman: tech comment<br/>
… in fact, the changes are such that<br/>
… makes possible something that I’m not sure we really use<br/>
… I can define as part of an internal entity something that won’t go out to the network<br/>
… I’m not sure if this feature is in use<br/>
… formal comment<br/>
… there was a formal resolution on the previous version; this PR slightly changes that<br/>
… can we get a formal resolution to merge, and also close a bunch of issues which were examples of the problem?<br/>
Proposed resolution: Merge PR #1368 to address outstanding DTD issues, and close GH issues 1369-1373 (Wendy Reid)<br/>
Garth Conboy: +1<br/>
Matt Garrish: +1<br/>
Ivan Herman: +1<br/>
Charles LaPierre: +1<br/>
Matthew Chan: +1<br/>
Wendy Reid: +1<br/>
Brady Duga: +1<br/>
George Kerscher: +1<br/>
Laura Brady: +1<br/>
Bill Kasdorf: +1<br/>
Ben Schroeter: +1<br/>
Resolution #1: Merge PR #1368 to address outstanding DTD issues, and close GH issues 1369-1373<br/>
</details>

PR #1391, opened 2020-10-30
Alternative conformance configuration

mattgarrish:
Here's a second possibility for dealing with all the conformance sections:

- it adds a new "EPUB Publications" section after the introduction that incorporates publication conformance and publication resources (it's kind of odd now that we've merged everything not to have such a section)
- only the three key conformance statements are listed: at least one package, accessibility conformance and ocf (there is no need to say that a publication has to contain resources, as that's required elsewhere, if not patently obvious). Everything else follows from the rest of the specification.
- the package requirements are similarly minimized since we don't need to list the various resources that you may include but are not required to (specific requirements like at least one resource in the spine are already covered)
- after that, I've tried to reduce the need for conformance sections for each technology, but it's not possible in all cases. But where we have to note requirements, I've tried to strip any redundant bullets

Maybe this is a happier medium, as it avoids the need for a huge conformance criteria section, but still identifies the core requirements while keeping technology-specific requirements where you expect/need to find them.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1391.html" title="Last updated on Nov 1, 2020, 11:55 PM UTC (e18ae7f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1391/af19628...e18ae7f.html" title="Last updated on Nov 1, 2020, 11:55 PM UTC (e18ae7f)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7332,6 +7332,8 @@ store destination as source in ocf
 				<section id="sec-overlay-req">
 					<h4>Media Overlay Document Requirements</h4>
 
+					<p>A <a>Media Overlay Document</a> has to meet the following requirements:</p>
+
 					<ul class="conformance-list">
 						<li>
 							<p id="confreq-mo-docprops-schema">It MUST be valid to the Media Overlays schema as defined
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3248,7 +3248,7 @@ Manifest:
 				</section>
 
 				<section id="sec-package-enc">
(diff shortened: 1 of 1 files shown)


mattgarrish:
I'd like to merge this now that it's less dramatic a change and more of an editorial shuffle/cleanup. Due to the number of changes involved, it makes merging with other issues more complicated.

iherman (approved):

wareid (approved):

Issue #1387, opened 2020-10-29
prefix attribute clash?


iherman:
There is a bit of a clash between EPUB and RDFa on the `prefix` attribute.

[§5.1.3.1.3.3 The prefix Attribute](https://w3c.github.io/epub-specs/epub33/core/#sec-contentdocs-prefix-attr) says:

> The prefix attribute definition is unchanged, but the attribute is defined to be in the namespace `http://www.idpf.org/2007/ops` when used in EPUB Content Documents.

The definition is in a different [section](https://w3c.github.io/epub-specs/epub33/core/#sec-prefix-attr), hence the "definition is unchanged".

However. EPUB also allows for RDFa, and refers to [HTML+RDFa](https://www.w3.org/TR/html-rdfa/). _RDFa also defines and uses the `prefix` attribute_. The above definition means that the EPUB prefix is _not_ the same as RDFa's prefix...

More pragmatically: any tool that processes an XHTML content document and also handles RDFa will use the `prefix` attribute, if used, from RDFa. Also, RDFa allows `prefix` to be used anywhere, whereas EPUB specifies it as being usable on the HTML element only, which may lead to false error messages.

Luckily, the definition of the EPUB `prefix` and the RDFa `prefix` are identical. Which may lead to the possibility of solving this by saying that EPUB uses the RDFa notion of `prefix` but reuses it for its own purposes (my understanding is that the only reason is to use it as property values), which is not entirely o.k. but we can get by. It should be made clear, however, that any EPUB usage of `prefix` would also influence a possible RDFa processor...

iherman:
Actually... there may be no problem after all because, I presume, the IDPF version of `prefix` must be used as `epub:prefix`. Ie, the syntactically correct way of using this is:

```html
<html xmlns="http://www.w3.org/1999/xhtml" 
    xmlns:epub="http://www.idpf.org/2007/ops"
    epub:prefix="pref_foo: http://example.org/foo/"
    prefix="pref_bar: http://example.org/bar/"
>
```

In which case `pref_foo:term` would be meaningful for epub purposes only and RDFa would ignore and the opposite would happen for `pref_bar:term`.

I got confused; I guess others will get confused, too. Maybe examples on the correct usage of `epub:prefix` would help (at the moment all occurrences of `prefix` is without CURIE) and a special note in the RDFa section drawing attention on this apparent clash could be added...

PR #1395, opened 2020-11-03
Re-org of vocabularies and mechanisms

### LINKED ISSUE #1377: (Editorial) placement of the rendering metadata
There is some sort of an inconsistency of the placement of the rendering metadata vocabulary in [§4.3.3](https://w3c.github.io/epub-specs/epub33/core/#sec-package-metadata-rendering) whereas all the other vocabularies are defined as part of [Appendix B](https://w3c.github.io/epub-specs/epub33/core/#app-vocabs). I believe the rendering metadata should be moved to Appendix B.

### LINKED ISSUE #1386: Vocabulary association regrouping?
This is just (editorial) musing...

At the moment, we have a vocabulary association section as [§4.3.2](https://w3c.github.io/epub-specs/epub33/core/#sec-metadata-assoc), as part of the section on publishing metadata and then we have a somewhat similar section [§5.1.3.1.3](https://w3c.github.io/epub-specs/epub33/core/#sec-contentdocs-vocab-association) as part of the content document spec. The latter refers back to the former, in fact but with different defaults and namespaces.

This is of course the result of the original structure. I wonder whether it is not possible/better to factor this out in one place, that is in its own section. It may make the overall structure cleaner.

(I am not sure this is necessary, but just raising the question...)

### LINKED ISSUE #1392: Remove the index
The index was added in 3.1 to try and help people make sense of information spread across five specifications, but now there's not much it adds that the table of contents doesn't.

Recommend we give it an early retirement...

mattgarrish:
This PR attempts to address the following vocabulary re-use issues:

- it moves the vocabulary association mechanisms section to an appendix and generalizes it to not be so specific to the package document
- the default vocabularies are defined in appropriate attribute to which they apply - the section in the appendix only describes their purpose
- it moves the epub:type attribute definition to an appendix because it is not strictly an HTML extension but used by HTML, SVG and media overlays. It is also generalized. (I tried to find a way to incorporate this under the vocabularies, but it just doesn't logically fit there.)

It also addresses some peripheral issues:

- it integrates the fixed layout re-org into this PR as it is also part of normalizing the vocabularies
- it adds the structure vocab to the vocabularies (this may be controversial, but it is getting to be a strange outlier now)
- it deletes the index - which, in addition to being unnecessary, also helps avoid too much bloat from adding the structure vocab

Fixes #1377 
Fixes #1386 
Fixes #1392

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1395.html" title="Last updated on Nov 4, 2020, 11:34 PM UTC (4c54db4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1395/e18ae7f...4c54db4.html" title="Last updated on Nov 4, 2020, 11:34 PM UTC (4c54db4)">Diff</a>

### DIFF
--- a/epub33/common/css/common.css
+++ b/epub33/common/css/common.css
@@ -273,3 +273,40 @@ table.prefix caption {
     font-size: 90%
 }
 
+
+/* Table zebra style... */
+
+table.zebra {
+    font-size:inherit;
+    font:90%;
+    margin:1em;
+}
+
+table.zebra td {
+    padding-left: 0.3em;
+}
+
+table.zebra th {
+    font-weight: bold;
+    text-align: center;
+    background-color: rgb(0,0,128) !important;
+    font-size: 110%;
+    background: hsl(180, 30%, 50%);
+    color: #fff;
+}
+
+table.zebra th a:link {
+  color: #fff;
+}
+
+table.zebra th a:visited {
+  color: #aaa;
+}
+
+table.zebra tr:nth-child(even) {
+    background-color: hsl(180, 30%, 93%) !important;
+}
+
+table.zebra th{border-bottom:1px solid #bbb;padding:.2em 1em;}
+table.zebra td{border-bottom:1px solid #ddd;padding:.2em 1em;}
+
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -2194,8 +2194,8 @@
 								></a>.</p>
 
 							<aside class="example">
-								<p>The following example shows various property declarations using reserved
-									prefixes.</p>
+								<p>The following example shows various property declarations using <a
+										href="#sec-reserved-prefixes">reserved prefixes</a>.</p>
 
 								<pre>&lt;metadata xmlns:dc="http://purl.org/dc/elements/1.1/"&gt;
     …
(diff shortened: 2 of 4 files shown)


iherman:
I very much like it. I also note that it covers #1387 insofar as it makes it clear that the EPUB version of `prefix` MUST be used with the `epub` CURIE prefix and not by itself, thereby avoiding the clash with the RDFa usage of a similar term.

Huge +1 from me, thus...

Minor points:

-  the `media` prefix is also used in media overlay documents, and not only in the Package document. (ref to the new §C.1.5.). Maybe worth adding there (although it is mentioned in the separate media overlay part)
- The various schemas seem all to be hosted in the epubcheck repository. We may have to discuss this; I believe it would be better to store those (authoritative) schemas as part of the group's specs repository and serve them via a w3c.github.io URL.
- What is the role, if any, of the EPUB-SSV reference now? It looks to me as if C.8 has superseded the separate document. If that is the case, that reference should be removed. If this is not the case, we have to have a clear separation between the two, which definition is authoritative, etc.

iherman (approved):

iherman:
Oops, my minor comment

> the media prefix is also used in media overlay documents, and not only in the Package document. (ref to the new §C.1.5.). Maybe worth adding there (although it is mentioned in the separate media overlay part)

doesn't hold. It is used in the package document for media overlay, not in the smil file. Withdrawn... :-)

mattgarrish:
> The various schemas seem all to be hosted in the epubcheck repository. We may have to discuss this; I believe it would be better to store those (authoritative) schemas as part of the group's specs repository and serve them via a w3c.github.io URL.

Right, we had this discussion in 3.1 or 3.2. We used to host them with the specifications, but that meant this group had to maintain them separately from epubcheck, even though epubcheck is where the schema development is primarily undertaken. The schemas are only part of the checks required to validate epubs, so they're only ever incomplete at best when pulled out of a real implementation scenario.

Plus it also raised questions about the point of creating static schemas. Once authored, any fixes would be implemented in epubcheck, making the versions pointed to by the specification out of date.

mattgarrish:
> What is the role, if any, of the EPUB-SSV reference now?

I'd imagine none. I took the one out of the epub:type section, but forgot to hunt for others. I'll switch them to internal links now.

mattgarrish:
I made a few minor clarifications/generalizations going back over the changes, but am good now with integrating this.

PR #1401, opened 2020-11-05
Delete invalid scheme attribute from example

### LINKED ISSUE #1396: xsd:string example as identifier scheme
[Example 83](https://w3c.github.io/epub-specs/epub33/core/#example-83) includes the following:

```xml
    <dc:identifier id="pub-id">urn:uuid:A1B0D67E-2E81-4DF5-9E67-A64CBE366809</dc:identifier>
    <meta refines="#pub-id" property="identifier-type" scheme="xsd:string">uuid</meta>
 ```

I presume the intention is that the `uuid` is some sort of a `xsd:string` derived type and serves as a constraint or indication on what the `pub-id` value is. However,

1. `uuid` is not a [primitive datatype](https://www.w3.org/TR/xmlschema-2/#built-in-datatypes) in XSD; `anyURI` is
2. Formally, `anyURI` is "just" another primitive datatype, that does not have any 'derived' relationship to `xsd:string`

I would expect what would work is something like:

```xml
    <dc:identifier id="pub-id">urn:uuid:A1B0D67E-2E81-4DF5-9E67-A64CBE366809</dc:identifier>
    <meta refines="#pub-id" property="identifier-type" scheme="https://www.w3.org/TR/xmlschema-2/">anyURI</meta>
 ```

mattgarrish:
Fixes #1396

### DIFF
--- a/epub33/core/vocab/meta-property.html
+++ b/epub33/core/vocab/meta-property.html
@@ -796,7 +796,6 @@
 				<pre>&lt;metadata xmlns:dc="http://purl.org/dc/elements/1.1/">
     
     &lt;dc:identifier id="pub-id">urn:uuid:A1B0D67E-2E81-4DF5-9E67-A64CBE366809&lt;/dc:identifier>
-    &lt;meta refines="#pub-id" property="identifier-type">uuid&lt;/meta>
     
     &lt;dc:identifier id="isbn-id">urn:isbn:9780101010101&lt;/dc:identifier>
     &lt;meta refines="#isbn-id" property="identifier-type" scheme="onix:codelist5">15&lt;/meta>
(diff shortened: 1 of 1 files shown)


iherman (commented):
The spec says:

> When the identifier-type value is drawn from a code list or other formal enumeration, the scheme attribute SHOULD be attached to identify its source.

it is a SHOULD not a MUST, so syntactically it is o.k., but without the scheme the term 'uuid' is not very meaningful :-(

mattgarrish:
Ya, the whole tag can be removed come to think of it. Without a controlled vocabulary, and ONIX doesn't define UUIDs, there's nothing it adds that the URN doesn't already identify.

iherman (approved):

PR #1409, opened 2020-11-06
Note that data blocks are not scripted content

### LINKED ISSUE #1352: Definition of a scripting context may be overzealous
[§5.4 of the core](https://w3c.github.io/epub-specs/epub33/core/#sec-scripted-content) defines a scripting context base on the presence of the `<script>` element in, say, HTML. However, the [HTML `<script>`](https://www.w3.org/TR/html/semantics-scripting.html#the-script-element) element is not restricted to active 'scripts', i.e., javascript code. It can hold any type of data: a typical case is schema.org metadata in JSON-LD format. 

The [HTML spec](https://www.w3.org/TR/html/semantics-scripting.html#the-script-element) says of the `type` attribute:

> Omitting the attribute, setting it to the empty string, or setting it to a [JavaScript MIME type essence match](https://mimesniff.spec.whatwg.org/#javascript-mime-type-essence-match) [...] to be interpreted according to the JavaScript [Script](https://tc39.es/ecma262/#prod-Script) top-level production.

which probably gives us the way we should restrict the case when we consider the scripting context: it does not have a type or the value of the type is a [JavaScript MIME type essence match](https://mimesniff.spec.whatwg.org/#javascript-mime-type-essence-match).

mattgarrish:
Fixes #1352 by adding a note to the scripting section that HTML data blocks do not represent scripted content.

The SVG case is a little murkier, as it says the element is the same as HTML but goes on to define things slightly differently. A script without a `type` is treated as javascript, otherwise the block is executed or not depending on whether the media type is recognized as a supported scripting language. That's not the same as HTML where everything outside of JS and 'module' types are data blocks.

I added a sentence noting the same data block concept does not apply, but could in the future. We might just have to live with json in svg having to be set as scripted in the package for now. It's probably not a common case, though.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1409.html" title="Last updated on Nov 15, 2020, 2:17 PM UTC (6085350)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1409/358b499...6085350.html" title="Last updated on Nov 15, 2020, 2:17 PM UTC (6085350)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -8881,6 +8881,9 @@ EPUB/images/cover.png</pre>
 					3.2</a></h3>
 
 				<ul>
+					<li>6-Nov-2020: Clarified that HTML <code>script</code> elements that contain <a
+							href="#sec-scripted-context">data blocks are not instances of scripting</a>. See <a
+							href="https://github.com/w3c/epub-specs/issues/1352">issue 1352</a>.</li>
 					<li>5-Nov-2020: Generalized the restriction on custom attribute namespace URIs to exclude all of
 						idpf.org and w3.org. See <a href="https://github.com/w3c/epub-specs/issues/1388">issue
 						1388</a>.</li>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7846,9 +7846,15 @@ html.-epub-media-overlay-playing * {
 					<tr>
 						<td>
 							<ul class="flat">
-								<li><code>application/mathml+xml</code></li>
-								<li><code>application/mathml-presentation+xml</code></li>
-								<li><code>application/mathml-content+xml</code></li>
+								<li>
+									<code>application/mathml+xml</code>
+								</li>
+								<li>
+									<code>application/mathml-presentation+xml</code>
+								</li>
+								<li>
+									<code>application/mathml-content+xml</code>
+								</li>
 							</ul>
 						</td>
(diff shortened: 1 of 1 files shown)


epub33/core/index.html
```diff
@@ -4445,10 +4445,17 @@ <h4>Scripting Contexts</h4>
...
 							<code>script</code> element or referenced via its <code>src</code> attribute makes no
 						difference to the executing context.</p>
 
+					<div class="note">
+						<p>When an [[HTML]] <code>script</code> element contains a <a
+								href="https://html.spec.whatwg.org/multipage/scripting.html#data-block">data block</a>
+							[[HTML]], it does not represent scripted content. [[SVG]] does not define data blocks as of
+							publication, but the same exclusion would apply to a future update.</p>
+					</div>
+
```
iherman, line 4469: To be extremely picky (sorry...) the first sentence should not be in a Note, which is informational. That should be part of the normative text imho...

iherman (approved):

epub33/core/index.html
mattgarrish: Ya, it felt kind of icky when I was adding it. No normative keywords, but it's not fully a note, either.

wareid (approved):

PR #1423, opened 2020-11-12
Remove file extension recommendations

### LINKED ISSUE #1294: Remove advice on HTML file extension
I thought we had discussed this, but I can't find an issue.

[Content Docs](https://www.w3.org/publishing/epub32/epub-contentdocs.html#sec-xhtml-conf-content) says

> The XHTML Content Document filename SHOULD use the file extension .xhtml

I think we should delete this sentence, and remove the corresponding test from EPUBCheck. I think this annoys users without helping anyone else. And if you use file extensions to tell you anything about a file, you're going to have a bad time.

mattgarrish:
Fixes #1294

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1423.html" title="Last updated on Feb 17, 2021, 12:03 PM UTC (767d9f2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1423/af82824...767d9f2.html" title="Last updated on Feb 17, 2021, 12:03 PM UTC (767d9f2)">Diff</a>


dauwhe (approved):
Woo!

iherman (commented):
I am a little bit less enthusiastic about this... 

Removing the _SHOULD_ on the `.xhtml` extension is clearly a good thing. But giving an authoring advice somewhere, saying something like 

> authors _SHOULD_ use well-known and well-deployed extensions for the various content, possibly abiding to the corresponding extensions in the relevant media types

may be useful to keep somewhere in the doc. A bunch of tools may go strange if I, for example, decide to publish my XHTML content in a file called `just_to_be_a_pain.svg`... Of course, tools will usually dump their XHTML content into files with a suffix of `.xhtml`, `.html`, or `.htm`, but some other tools, or indeed those manually editing their EPUB, may do strange things if there is no warning...

mattgarrish:
It doesn't make a lot of sense to pick out a few formats and say it's a warning if authors don't use a prescribed extension but then ignore everything else you can include. Especially when the resources where the file extensions might actually influence the rendering (embedded in html) are the ones not being enforced.

Reading systems can identify the resource from the media type declaration in the manifest if they're using these particular resources.

> authors _SHOULD_ use well-known and well-deployed extensions for the various content, possibly abiding to the corresponding extensions in the relevant media types

Change it to "are advised", and add to the publication resources section, and I'd be more agreeable. Otherwise, we're just making what has to be warned about that much broader and more difficult to pin down.

iherman:
> > authors _SHOULD_ use well-known and well-deployed extensions for the various content, possibly abiding to the corresponding extensions in the relevant media types
> 
> Change it to "are advised", and add to the publication resources section, and I'd be more agreeable. 

Yes, that is actually much better.

dauwhe:
Isn't it a general principle to not trust file extensions?

iherman:
> Isn't it a general principle to not trust file extensions?

True. But there is a general principle not to trust tools either :-)

iherman:
E.g.: if I create an HTML file `just_to_be_a_pain.svg`, and load it to my text editor, I am sure it will display it in SVG mode...

dauwhe:
> E.g.: if I create an HTML file `just_to_be_a_pain.svg`, and load it to my text editor, I am sure it will display it in SVG mode...

I don't see this as a problem for the EPUB 3 Specification.

mattgarrish:
It's a stretch, but doesn't IE do file extension mapping to the registry at some point in its determination algorithm, although I think it forces a download at that point.

It does feel we like we're not giving authors much credit to assume they can't figure out the file extension for their content, though. :)

wareid (approved):

iherman:
I do not want to hold this PR up, so forget about my remark (it was only an editorial issue, we can always come back to it if we really hit a problem, e.g., during CR phase).

Approving this PR

iherman (approved):

murata2makoto:
I am going to raise new issues about this merged change.  It is a bit inconsistent (e.g., *.epub and *.opf) and also drop a normative requirement on the use of the application/oebps-package+xml media type.  In my understanding, H.1 does not require that package documents specify this media type.

mattgarrish:
> In my understanding, H.1 does not require that package documents specify this media type.

No, but neither does this statement:

> Package Documents have the MIME media type application/oebps-package+xml

That's just a repetition of what H.1 already lays out. We currently take it on a bit of trust that the media type will be used in the container.xml file by only requiring that package documents be listed. If we need to require the media type, that would be the place to do it.

mattgarrish:
I've added the package doc media type as a requirement of rootfile elements. Any reason not to merge this?

iherman:
> I've added the package doc media type as a requirement of rootfile elements. Any reason not to merge this?

Not that I could see...

PR #1445, opened 2020-12-15
Review of the Overview document

iherman:
This is a possible revision of the Overview document. Here are the important changes (note that I rely on the fact that PR #1438 is due to be merged!):

- the abstract referred to a separate 'EPUB-CHANGES-33' document, which does not exist anymore (removed it and replaced by a reference to the changes' section) 
- rearranged the list of editors, with Matt remaining as the only active editor
- added the w3cid values for all editors in the respec preamble
- removed the "Rendition" term in favor of publication everywhere, except for one place where the text explicitly refers to multiple renditions (and added a reference to the new multi-rend document)
- removed the reference to the EPUB 2 NCX document which is now real history and I do not think we should continue referring to it at this point
- refreshed the cross-references (using the references of #1438 )
- added the new rendition document to the biblio.js file
- I added a new diagram for EPUB 3, that does not refer to Renditions
- minor stylistic changes

I also ran into the following questions and I decided not to touch those yet:

- The last paragraph of §1.1 refers to the `collection` element. Is that element widely used? If not, maybe we can remove it from the overview, it looks like a distraction
- §1.3 includes a reference to the release identifier, but, afaik, this feature is very rarely used and its usage is not required. I wonder if this should not be removed altogether, leaving only the first sentence (on unique identifiers) in the document.

---
* [Preview](https://raw.githack.com/w3c/epub-specs/overview-review/epub33/overview/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fepub-specs%2Fepub33%2Foverview%2F&doc2=https%3A%2F%2Fraw.githack.com%2Fw3c%2Fepub-specs%2Foverview-review%2Fepub33%2Foverview%2Findex.html)


iherman:
A minor stylistic question: is it intentional that all lists are without any bullet? This is a bit unusual... (e.g., §2.3)

shiestyle (approved):

mattgarrish:
> A minor stylistic question: is it intentional that all lists are without any bullet?

What browser are you using? I'm seeing bullets in firefox/edge/chrome.

mattgarrish:
> The last paragraph of §1.1 refers to the `collection` element. Is that element widely used?

It's used in a number of the extension specs: indexes, dictionaries, distributable objects, previews, etc.

Don't know how much use it has in actual practice, though.

mattgarrish:
> includes a reference to the release identifier, but, afaik, this feature is very rarely used and its usage is not required.

It's the only justification for the required dcterms:modified property. I don't personally think it's terribly useful, but as long as we have that requirement we probably shouldn't bury the rationale for it too much.

mattgarrish:
It looks like there's overlap with PR #1448 regarding the renditions terminology, so I'm going to hold off on reviewing until that one is done. I'll probably have to do some merging to make them work together.

iherman:
> > A minor stylistic question: is it intentional that all lists are without any bullet?
> 
> What browser are you using? I'm seeing bullets in firefox/edge/chrome.

Wow. I used Brave, and it is wrong, but now I checked with Safari, Edge, Chrome, and Vivaldi, and it is all right. This means Brave does something wrong, which is strange because they use the same rendering engine as Vivaldi or Edge.

Incidentally, I ran into another issue yesterday evening with Brave related again to list handling... :-(

Forget this comment...

iherman:
> Incidentally, I ran into another issue yesterday evening with Brave related again to list handling... :-(

... and actually I located the problem (but not sure of the reason). I had, in my Brave browser, installed the epub reader (chrome) extension. Well, that extension clashed and affected some list handling (this problem included). Go figure... :-)

wareid (approved):

PR #1463, opened 2021-01-12
First setup for .travis.yml

iherman:
This is (hopefully) the last step to set up ECHIDNA.

@deniak I would appreciate if you made a last minute check on this before we really use it. (I hope the secure values are fine, see my separate email.)

As a specific point: is the check on `push` in the lines

```
 - if: branch = publish-content AND type = push
```

really necessary? (The protocol we would use is when, say, the content document is ready to go, then the whole repository is merged with the publish-content branch (on, say, my or Matt's machine) and is pushed on github.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,44 @@
+language: node_js
+
+branches:
+ only:
+    - publish-content
+    - publish-rs
+    - publish-overview
+    - publish-multi-rend
+
+jobs:
+ include:
+ - if: branch = publish-content AND type = push
+   env:
+    - URL="https://w3c.github.io/epub-specs/epub33/snapshot/epub-33/ECHIDNA"
+    - DECISION="https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2020-11-20-epub#resolution2"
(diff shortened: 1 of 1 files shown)


deniak (approved):
The 4 manifest URLs you are passing are served by Github pages, which can only be linked to a specific branch (`master` here) so you'll need to make sure that branch is rebased regularly.
Otherwise it looks good to me, knowing this is really the first time we experiment with that kind of setup.

iherman:
> The 4 manifest URLs you are passing are served by Github pages, which can only be linked to a specific branch (`master` here) so you'll need to make sure that branch is rebased regularly.

I try to check/understand this: the `master` branch is our "main" branch, which is also the github.io branch Any update on the editor's draft is done through that branch being updated.

The `publish-content` branch is never directly edited. It should be changed only by merging (at publication time) the `master` branch into `publish-content` which should trigger the publication.

Do you see any other steps we would have to do?

deniak:
> Do you see any other steps we would have to do?

Nope, if the manifests on the master branch are up-to-date, it should work as expected

PR #1466, opened 2021-01-13
clarify requirement for id attribute on unique identifier element

### LINKED ISSUE #1461: Possible ambiguity in section 3.4.3.2.1 The identifier Element
In EPUB Packages 3.2 in section 3.4.3.2.1 the id element is said to be optional, however the language later is:
"The metadata section MUST include an identifier element that contains an unambiguous identifier for the Rendition. This identifier MUST be marked as the Unique Identifier via the package element unique-identifier attribute."

So the id is not optional, it is the id marked as unique-idnetifier.

I guess the meaning is that if there are more then one dc:identifier elements, the id is optional.
So I think the language should be clearer.

mattgarrish:
Fixes #1461

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1466.html" title="Last updated on Jan 13, 2021, 4:23 PM UTC (105d9b6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1466/714b5e5...105d9b6.html" title="Last updated on Jan 13, 2021, 4:23 PM UTC (105d9b6)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1564,7 +1564,7 @@
 													<a href="#attrdef-id">
 														<code>id</code>
 													</a>
-													<code>[optional]</code>
+													<code>[conditionally required]</code>
 												</p>
 											</li>
 										</ul>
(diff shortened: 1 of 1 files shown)


epub33/core/index.html
```diff
@@ -1575,14 +1575,13 @@ <h6>The <code>identifier</code> Element</h6>
...
 
-								<p>The <a>Author</a> MUST provide a primary identifier that is unique to one and only
-									one <a>EPUB Publication</a> in the <a href="#sec-opf-dcidentifier"
-											><code>dc:identifier</code> element</a>. This identifier MUST be referenced
-									as the Unique Identifier in the <a href="#elemdef-opf-package"><code>package</code>
-										element's</a>
+								<p>The <a>Author</a> MUST provide an identifier that is unique to one and only one
+										<a>EPUB Publication</a> &#8212; its <a>Unique Identifier</a> &#8212; in an
+										<code>identifier</code> element. This <code>identifier</code> element MUST
+									include an <code>id</code> attribute whose value is referenced from the <a
```
iherman: Shouldn't it be "contain" rather than "include"? I would not want the content of the element to be something like "https://identify.er obladi oblada" :-)

iherman (approved):

epub33/core/index.html
mattgarrish: Don't think so, but maybe we're not reading it the same way. An `id` attribute only ever has a single value, and the phrasing is "must include an id attribute whose value is referenced from". I don't see a suggestion there that the id can have more than one value or that only part of the value is matched.

"Contain" is typically used to refer to the value / child content space of the attribute or element it refers to.

Looking around at some other specs for guidance, I'm not finding a consistent phrasing for referring to attributes of elements. The XML specification likes to use "attach", HTML prefers "specify", SVG uses "include" and MathML seems to mix "include" and "specify".

Of the options, I'd probably go with "specify" if we want to harmonize on something other than including. But I'm not too concerned either way.

iherman: Let us go with specify. Just to be in line with HTML :-)

dauwhe (approved):

PR #1467, opened 2021-01-13
Change the requirement for progressive enhancement with spine-level scripting

### LINKED ISSUE #1444: Normative restrictions on scripting: is it necessary?
I was wondering about this [paragraph in the EPUB spec](https://w3c.github.io/epub-specs/epub33/core/#sec-scripted-spine):

> EPUB Content Documents that include spine-level scripting MUST utilize the progressive enhancement technique, which for the purposes of this specification has the following definition: when the document is rendered by a Reading System without scripting support or with scripting support disabled, the Top-level Content Document MUST retain its integrity, remaining consumable by the user without any information loss or other significant deterioration.

I indeed ran into the following Web page:

https://ciechanow.ski/cameras-and-lenses/

this is a (in my view) superbly done page on photography basics, their physics, etc, which has tons of interactive graphics examples that are essential for understanding the subject. Looking at the source all the graphics are done via scripts; as far as I could see via a canvas of some sort and using GL (I must admit I did not go into the details). It is all 'local', in the sense of there is no external reference to major frameworks: it is all a bunch of clever scripts and CSS.  I.e., this content could be turned into EPUB 3.3 and it would be a nice example of what an educational EPUB would look like.

Except that... it violates the normative restriction above. In this very case, I just do not see how the author could fulfill the requirement above. This page, and the text, is inherently bound to the interactive possibilities in the graphics; the text is simply not understandable without it. As a consequence, and in spite of what was said in the previous paragraph, it does not seem to be possible to turn this into a compliant EPUB 3.3.

Is it necessary to keep the _MUST_ here? Would it be enough to use a _SHOULD_?

mattgarrish:
This PR fixes #1444 as follows:

- changes the requirement for progressive enhancement to a recommendation that the top-level content documents remain consumable
- progressive enhancement is moved to a parenthetical as one way of accomplishing this
- adds context for why it is important that scripted top-level content documents be consumable

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1467.html" title="Last updated on Jan 20, 2021, 1:59 PM UTC (55db9eb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1467/5839683...55db9eb.html" title="Last updated on Jan 20, 2021, 1:59 PM UTC (55db9eb)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -4417,13 +4417,14 @@ Manifest:
 				<section id="sec-scripted-spine">
 					<h4>Spine-Level Scripts</h4>
 
-					<p id="confreq-cd-scripted-spine">EPUB Content Documents that include <a
-							href="#sec-scripted-content-type-spine-level">spine-level scripting</a> MUST utilize the
-							<em>progressive enhancement technique</em>, which for the purposes of this specification has
-						the following definition: when the document is rendered by a Reading System without scripting
-						support or with scripting support disabled, the <a>Top-level Content Document</a> MUST retain
-						its integrity, remaining consumable by the user without any information loss or other
-						significant deterioration.</p>
+					<p id="confreq-cd-scripted-spine"><a>Top-level Content Documents</a> that include <a
+							href="#sec-scripted-content-type-spine-level">spine-level scripting</a> SHOULD remain
+						consumable by the user without any information loss or other significant deterioration when
+						scripting is disabled (e.g., by employing progressive enhancement techniques).</p>
+
+					<p>Since Reading Systems are not required to support scripting, failing to account for non-scripted
+						environments in Top-level Content Documents can result in EPUB Publications being
+						unreadable.</p>
 				</section>
 
 				<section id="sec-scripted-a11y">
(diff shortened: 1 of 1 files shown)


epub33/core/index.html
```diff
@@ -4417,13 +4417,14 @@ <h4>Container-Constrained Scripts</h4>
...
-							<em>progressive enhancement technique</em>, which for the purposes of this specification has
-						the following definition: when the document is rendered by a Reading System without scripting
-						support or with scripting support disabled, the <a>Top-level Content Document</a> MUST retain
-						its integrity, remaining consumable by the user without any information loss or other
-						significant deterioration.</p>
+					<p id="confreq-cd-scripted-spine"><a>Top-level Content Documents</a> that include <a
+							href="#sec-scripted-content-type-spine-level">spine-level scripting</a> SHOULD remain
+						consumable by the user without any information loss or other significant deterioration when
+						scripting is disabled (e.g., by employing progressive enhancement techniques).</p>
+
```
iherman, line 4424: I do not feel strongly about this, but we may consider adding an (informative) reference to https://developer.mozilla.org/en-US/docs/Glossary/Progressive_Enhancement.

I leave to you to decide whether that is necessary.

iherman (approved):

epub33/core/index.html
mattgarrish: How would we create a cite for it? An informative "MDN" entry in the local biblio that just states the site name?

Or does W3C allow hyperlinking to external sources without citations?

dauwhe (approved):
This is much better. I don't think we necessarily need a reference for "progressive enhancement." It's very easily searchable, and a term quite familiar to the web community.

epub33/core/index.html
iherman: Good point. I have spent some time getting around existing documents, but hyperlinking to external sources does occur but not in a normative section (and this one is). We can either drop the idea, or put the reference into a separate (small) note. But that may not worth it...

mattgarrish: I think we're safe leaving it as is. As @dauwhe says, it's a widely known term and searching returns only relevant results, amusingly including the same definition in adjoining boxes:

wikipedia definitions

iherman: O.k. Go for it!

iherman (approved):

shiestyle (approved):

wareid (approved):

Issue #1456, opened 2021-01-04
xml:base in package document and elsewhere


dauwhe:
EPUBCheck seems unhappy with any use of `xml:base` in the package document. EPUB requires support for `xml:base.` I don't even want to think about using `xml:base` in `container.xml`. 

Is `xml:base` only for use with EPUB content documents? Can we be clearer in the spec about where it is and isn't allowed?

mattgarrish:
The note about HTML and SVG dropping support is probably misplaced, as prior to that being added the only mention of xml:base was that reading systems had to support it. But that was only required for completeness of xml processing, not because the attribute is actually allowed in every xml format. It's never been supported in the package document.

It might be less confusing to copy the note into both the HTML and SVG definitions, since those are the places where it is relevant. Commenting on xml:base for any format someone might include seems unnecessary.

It might also help to move this sentence from the item element definition to a more obvious location:

>  In the case of relative IRIs, the IRI of the Package Document is used as the base when resolving to absolute IRIs.

This is true for all href attributes and anything else that takes a relative IRI (refines, ?).

But this begins to tie into #1374.

iherman:
@murata2makoto is the expert here. But it is interesting to note that the [core XML specification](https://www.w3.org/TR/xml/) does _not_ include `xml:base`, which is defined by a [separate specification](https://www.w3.org/TR/xmlbase/). It is therefore, spec-wise, perfectly fine to disallow its usage, or restrict it explicitly to some content. @murata2makoto should tell us whether this is an acceptable practice spec-wise.

Note, however, that "disallowing" it in HTML is not that simple, because HTML also has the equivalent `<base>` element...

murata2makoto:
&lt;rant>I opposed to xml:base when it was invented and I still do.&lt;/rant>

I think that disallowing the xml:base attribute in some XML vocabulary is perfectly fine.  It is certainly possible to write a schema (in both RELAX NG and W3C XML Schema) so that this attribute is disallowed or restricted.  The latest version of WHATWG HTML does not mention xml:base.  In my understanding, this attribute is thus disallowed.  We only have to use the base element of HTML.

dauwhe:
Proposal: specify that `xml:base` is allowed only in EPUB Content Documents (XHTML and SVG), and even then is a bad idea.

mattgarrish:
> specify that `xml:base` is allowed only in EPUB Content Documents

I'm a bit concerned we're creating an issue where none exists.

[...]

Issue #1374, opened 2020-10-27
"IRI of the Package Document": what is this exactly?


iherman:
We may not be in position of providing an absolutely clean definition (so maybe some editorial hand waving would be necessary) but…

[§4.2.3.4.2 The item Element](https://w3c.github.io/epub-specs/epub33/core/#sec-item-elem) says:

> Each item element in the manifest identifies a Publication Resource by the IRI [RFC3987] provided in its href attribute. The IRI MAY be absolute or relative. In the case of relative IRIs, the IRI of the Package Document is used as the base when resolving to absolute IRIs. The resulting absolute IRI MUST be unique within the manifest scope.

The intention is clear but what is the "IRI of the Package Document"? After all, the package document is part of a ZIP file, it is not really on the Web, ie, it is not clear what its IRI is. 

Can we say something more precise about this?

llemeurfr:
This issue is why we introduced the notion of "root directory" in the [LPF format](https://www.w3.org/TR/lpf/). We still used urls (we decided not using the IRI term) in "Contents within the Package MUST reference these resources [those in the package] via relative-URL strings [url]."

iherman:
Ah yes, indeed. We may want to get inspired and bring this over to the EPUB spec...

mattgarrish:
Is there a difference from the OCF abstract container's root directory? Doesn't it already establish the virtual directory structure within the zip?

iherman:
I think the difference is not in the terms/concept but the surrounding explanation. I find the few extra terms in the LPF spec on "virtual in nature", etc, helpful. And a reference to this from the item element definition may also be helpful.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-04-01-epub#section2) on 2021-04-01

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-04-01-epub#resolution1):  Merge PR 1468***

<details><summary><i>View the transcript</i></summary>

### 2. Clarify base IRI 

_See github pull request [#1468](https://github.com/w3c/epub-specs/pull/1468)._




**Dave Cramer:** this is a PR about base IRI in package documents  

_See github issue [#1374](https://github.com/w3c/epub-specs/issues/1374), [#1456](https://github.com/w3c/epub-specs/issues/1456)._




[...]

PR #1468, opened 2021-01-14
Clarify base iri of package document

mattgarrish:
Maybe not a complete fix for #1374 (and not for #1456), but I've taken the prose about resolving relative IRIs from the item element definition and created a section in the RS spec on resolving relative IRIs to cover this (resolving to absolute IRIs isn't an authoring concern).

See if there's enough of an explanation of the base IRI when the package document is zipped. I don't think we want to go too deep into resolving files within a zip container, as I can't find there's a single way of doing this. It looks like a "!" (jar scheme), "#", or OS-specific path separator followed by the relative path of the file in the container are viable.

Reading Systems: [preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1456-packagedoc/epub33/rs/index.html) [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1456-packagedoc/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1468.html" title="Last updated on Apr 1, 2021, 8:57 PM UTC (8884e0d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1468/4da0947...8884e0d.html" title="Last updated on Apr 1, 2021, 8:57 PM UTC (8884e0d)">Diff</a>


iherman (approved):
I think it would be worth adding a note to the change section of the RS spec. It may be an important clarification relative to the previous version.

llemeurfr:
In the EPUB3.3 Core spec, [in the `item` section](https://www.w3.org/TR/epub-33/#sec-item-elem), we still read: "Each item element identifies a Publication Resource by the IRI [RFC3987] provided in its href attribute. The IRI MAY be absolute or relative. In the case of relative IRIs, the IRI of the Package Document is used as the base when resolving to absolute IRIs. The resulting absolute IRI MUST be unique within the manifest scope."

In this Core spec, the "IRI of the Package Document" is not defined: this was the original issue #1374. Tips must be given to EPUB authors on how to solve this issue, therefore something must be done first in the EPUB 3.3 Core document.

More: 

a) In Zip packages, such "relative" strings must in fact be seen as relative paths. As it happens, both relative IRIs and relative Unix paths have the same syntax; but they're not the same thing.  It is weird but still real: internal resource are referenced via relative paths, external resources are referenced via absolute URLs, and both can be values of the same href attribute. 

The [Zip AppNote](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) uses the term path, never URL/URI/IRI.  Section 4.4.17.1 of the Zip AppNote gives clear indications about the syntax, but does not define "relative path"; I suppose it was clear for the authors that this path was relative to the "root directory" (which is not a notion defined in the Zip AppNote).  

b) I don't think RS developers need to _resolve_ (if resolving means converting to absolute IRIs) relative IRIs/paths in Zip packages. They simply need to find and fetch these files using their relative path, which is a standard Zip API feature. 

In consequence, I'd be in favor of 
1) changing the wording in the Core spec first, replacing "relative IRIs" by "relative paths" (relative to the Root Directory which is already defined there).
2) reviewing sections 3.2.1 and 7.2.1 in the RS spec and thinking about removing them entirely.

llemeurfr (changes requested):
see my previous comment.

mattgarrish:
> In the EPUB3.3 Core spec, [in the `item` section](https://www.w3.org/TR/epub-33/#sec-item-elem), we still read:

I changed that if you look at the preview/diff links at the bottom of the first comment (pr-preview always creates those automatically):

> Each item element identifies a Publication Resource by the IRI [RFC3987] provided in its href attribute. The IRI MAY be absolute or relative, but each IRI MUST be unique within the manifest scope after resolution to an absolute IRI [EPUB-RS-33].

The last few words link across the the RS section so authors interested in how resolution works can read more without overly complicating that section.

> 1\. changing the wording in the Core spec first, replacing "relative IRIs" by "relative paths"

But isn't this why we have the abstract container defined separately from the zip container? It allows us to talk about the paths consistently as IRIs so we don't get into these kinds of technicalities. This sounds like it leads us to redefining all the standards we rely on to also be paths instead of IRIs simply because an EPUB may be zipped when consumed

I'd argue we're getting into too much detail with the package document base by delving into zip paths, but can live with it for thoroughness.

mattgarrish:
> I don't think RS developers need to _resolve_ (if resolving means converting to absolute IRIs) relative IRIs/paths in Zip packages.

Yes, I've fixed this in the latest commit as it's not a requirement in all cases. The intention is only that if you require an absolute you must use the base iri.

dauwhe:
How would you use an absolute IRI in the `href` attribute of an `item`? What scheme would you use?

mattgarrish:
> How would you use an absolute IRI in the `href` attribute of an `item`?

Which case are you referring to?

Absolute IRIs in general would be critical for any external resources.

Resolved from a relative one might be useful if you're serving it off the web, I suppose.

> What scheme would you use?

We've avoided trying to list out allowed and disallowed protocols leaving it to whatever makes sense to use. The only requirement for relative IRIs is that they must resolve within the container. But I believe it's valid to use file: URLs if you want to write out a full path.

dauwhe:
> Which case are you referring to?

an `item` in the spine.

mattgarrish:
> an `item` in the spine.

Still not sure I understand.

The spine itself only has `itemref` elements and the `ref` attributes are IDREFs.

What would be unique about an `itemref` referencing an `item` that is an external resource, if that's what you mean? You'd need a fallback, but otherwise there's nothing exceptional about it, is there?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-04-01-epub#section2) on 2021-04-01

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-04-01-epub#resolution1):  Merge PR 1468***

### 2. Clarify base IRI 

_See github pull request [#1468](https://github.com/w3c/epub-specs/pull/1468)._




**Dave Cramer:** this is a PR about base IRI in package documents  

_See github issue [#1374](https://github.com/w3c/epub-specs/issues/1374), [#1456](https://github.com/w3c/epub-specs/issues/1456)._




**Matt Garrish:** basically all the PR does is define base IRI for package because it wasn't clear how that was to be calculated  
… it is defined for container.xml, but for package doc there was just a stray sentence that absolute IRIs are to be calculated from base IRI of document  
… Ivan wanted some clarification  
… PR pulls out that statement and elaborates on it  
… Laurent suggested that maybe we define everything in core docs as paths rather than IRIs  
… not sure why we'd want to do that since we already define abstract container to allow us to use IRI language  
… how far do we want to get into relative paths vs absolute paths  
… can we just clean up what we already have, or do we want to take up this IRI vs path question at this point?  

**Dave Cramer:** i'm happy with PR  
… worried about Laurent's idea because not sure we want to start talking about paths when all the other specs that we rely upon are already happy about how we define things  
… also, this issue didn't come out of a concrete problem with a RS or similar, it came from abstract issue about spec language  

**Matt Garrish:** the PR seemed to make Ivan happy  
… from the perspective of what we need to describe here, I think we've done enough  
… the question about changing to path language leads off into other areas  

**Dave Cramer:** i think we should accept the PR, and then if Laurent wants to raise the other question, maybe he can come back with a more detailed rationale  

**Matt Garrish:** there was an issue about whether relative IRIs MUST be resolved, but it was only ever the intention that it be possible if you need to do it  

> **Proposed resolution: Merge PR 1468** *(Wendy Reid)*

> *Dan Lazin:* +1

> *Ben Schroeter:* +1

> *Matt Garrish:* +1

> *Toshiaki Koike:* +1

> *Wendy Reid:* +1

> *Matthew Chan:* +1

> *Masakazu Kitahara:* +1

> *Brady Duga:* +1

> *Shinya Takami (高見真也):* +1

> ***Resolution #1: Merge PR 1468***

</details>

PR #1469, opened 2021-01-15
Clarify WCAG conformance thresholds

### LINKED ISSUE #1454: Require AA conformance for accessibility
Should we consider requiring level AA conformance to meet the accessibility specification to avoid having too much disparity in content?

As it stands, we could end up allowing conformance to multiple versions of WCAG along with either A or AA within any version. AA is the more generally required standard for accessibility.

### LINKED ISSUE #1459: Approaches for elevating EPUB accessibility to new versions of WCAG
As discussed in Accessibility task force calls, migrating EPUB accessibility specifications to WCAG 2.1/2.2 is important. The main question is how should we do it.

Option 1:
Hard wire EPUB Accessibility 1.1 with WCAG 2.1
Advantage: that It is simple and straight forward, no confusions. If publishers find it difficult to conform to WCAG 2.1 then they can conform to EPUB Accessibility 1.0 which conforms to WCAG 2.0.
Disadvantage: WCAG 2.2 will be released soon, and in such a case will will need to revise EPUB Accessibility as WCAG progress.

Option 2:
Allow EPUB Accessibility 1.1 to refer to different versions of WCAG (version 2.0 onwards).
i.e. Publication can conform either to EPUB accessibility 1.1 (WCAG 2.0), EPUB Accessibility 1.1 (WCAG 2.1), and also EPUB Accessibility 1.1 (WCAG 2.2)
Advantage: High flexibility and future proof approach.
Disadvantage: Reporting will become highly complex.

Option 3:
EPUB Accessibility 1.1 has date-less or version-less reference to WCAG.
i.e. it refers to the latest version of WCAG. Right now it is version 2.1, in future we would have version 2.2, 3.0 etc.
Advantage: Conceptually simple and future proof.
Disadvantage: Some segments of publishing industry may feel that it is very aggressive approach, because publishing industry moves at a slower pace than web. It is possible that most of the industry remains at EPUb Accessibility 1.0, and some highly progressive publishers move to version 1.1. Moreover there is a practical problem, publications are not updated like web pages, which means that reporting will still need to include reference to WCAG version met by the publication at the time of production.

Thoughts are welcome, especially from the publishers and supply chain actors who need to deal with complexities of conformance reporting.

mattgarrish:
Rewrites the section on WCAG conformance to better explain why we have the baseline of 2.0 A and why authors should, and in many cases will have to, meet a higher standard.

- [Accessibility preview](https://raw.githack.com/w3c/epub-specs/a11y-11/conformance-issues/epub33/a11y/)
- [Accessibility diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fw3c.github.io%2Fepub-specs%2Fepub33%2Fa11y%2F&doc2=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fraw.githack.com%2Fw3c%2Fepub-specs%2Fa11y-11%2Fconformance-issues%2Fepub33%2Fa11y%2Findex.html)

Fixes #1459 
Fixes #1454

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -351,8 +351,7 @@
 				<section id="sec-wcag-conf">
 					<h4>WCAG Conformance Requirements</h4>
 
-					<p>An EPUB Publication has to meet to the following requirements to conform to this
-						specification</p>
+					<p>An EPUB Publication has to meet the following requirements to conform to this specification</p>
 
 					<ul class="conformance-list">
 						<li>
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -351,17 +351,47 @@
 				<section id="sec-wcag-conf">
 					<h4>WCAG Conformance Requirements</h4>
 
-					<p>EPUB Publications MUST meet WCAG 2.0 <a href="https://www.w3.org/TR/WCAG20/#conformance-reqs"
-							>Level A</a> to be conformant with this document, but it is RECOMMENDED that they meet <a
-							href="https://www.w3.org/TR/WCAG20/#conformance-reqs">Level AA</a> [[!WCAG20]].</p>
+					<p>An EPUB Publication has to meet to the following requirements to conform to this
+						specification</p>
 
-					<div class="note">
-						<p>Although this document only requires Level A conformance, local and national laws can
-							influence the level of conformance an EPUB Publication has to meet to be considered
-							accessible. Level AA conformance is often cited as the benchmark for accessibility in legal
-							frameworks and policies, for example. Additionally, any procurer or distributor of EPUB
-							Publications can demand higher conformance requirements than the baseline defined here.</p>
-					</div>
+					<ul class="conformance-list">
+						<li>
+							<p id="confreq-wcag-20">It MUST meet the requirements of WCAG 2.0 [[!WCAG20]] but it is
+								strongly RECOMMENDED that it conform to the latest recommended version [[WCAG]].</p>
+						</li>
+
+						<li>
+							<p id="confreq-wcag-a">It MUST meet the requirements of <a
+									href="https://www.w3.org/TR/WCAG/#cc1_a">Level A</a> but it is strongly RECOMMENDED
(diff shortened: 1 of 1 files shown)


clapierre (approved):
Looks awesome as usual Matt.

epub33/a11y/index.html
```diff
@@ -351,17 +351,47 @@ <h3>WCAG conformance</h3>
...
-
-					<div class="note">
-						<p>Although this document only requires Level A conformance, local and national laws can
-							influence the level of conformance an EPUB Publication has to meet to be considered
-							accessible. Level AA conformance is often cited as the benchmark for accessibility in legal
-							frameworks and policies, for example. Additionally, any procurer or distributor of EPUB
-							Publications can demand higher conformance requirements than the baseline defined here.</p>
-					</div>
+					<p>An EPUB Publication has to meet to the following requirements to conform to this
+						specification</p>
```
GeorgeKerscher: There is an extra "to" in this line.

mattgarrish: Oops, ya. I've deleted the second one.

GeorgeKerscher:
It looks good. I left one comment on a typo of an extra “to” on one line.

gregalchin:
I really like how this is reading. I would like to suggest a small change. 

An EPUB Publication has to meet to the following requirements to conform to this specification:

It MUST conform to the latest recommended version [WCAG] at the time of publication with any subsequent edition to meet the latest recommended version [WCAG] at the time of publication of the new edition. 

---This is similar to the wording in Australian building code about the application of accessibility standards when undertaking new builds and or major renovations

mattgarrish:
> with any subsequent edition to meet the latest recommended version [WCAG] at the time of publication of the new edition

I agree with this in principle, but we didn't have time to get into this issue on the last call. We did touch on the point-in-time nature of EPUB publications compared with the web, where you have to keep re-evaluating your content as it evolves, but agreed it was something we need to come back to.

I'm going to hold off on crafting a statement until we can return to the issue, as I think this topic may warrant its own subsection. I'll open a specific issue for it shortly.

PR #1472, opened 2021-01-19
Clarifications about epub:type, structure vocab and accessibility

### LINKED ISSUE #1437: Clarify pagebreak in spec
I think the specification around pagebreak is ambiguous. I would like to clarify it for readers and publishers.

It currently says:

> **pagebreak**
>  *  A separator denoting the position before which a break occurs between two contiguous pages in a statically paginated version of the content.
>  *  HTML usage context: phrasing and flow content, where the value of the carrying elements title attribute takes precedence over element content for the purposes of representing the pagebreak value

https://github.com/w3c/epub-specs/blob/8b7e0966b14d0ab65a693fc7d1c63e822ed1879b/epub33/core/vocab/structure.html#L2355

I recently purchased an eBook which has this markup:
```html
<span id="pg5" epub:type="pagebreak">‘But</span> of course!’
```

Throughout the book, the first word, or first letter, on a page is wrapped with ` epub:type="pagebreak"`.  When I tried to view it in a popular eReader, the words in the span didn't render. This is expected behaviour because, normally, the contents of that element would be a page number.  See https://github.com/koreader/koreader/issues/6894

A few other rendering engines (like Kindle) did display the text.

Judging from the example ePubs, I suspect that the intent of this part of the spec is that the content of that element must *only* contain a page number.  That won't necessarily be an integer, it could be `Page 42` or `ⅩⅬⅡ`. But the default assumption should be that it is *metadata* and may not be displayed to a user.

I think that the spec should say something like this:

"Element content *should* be empty. If element content is not empty, the content *must* be text relating to the page number. Not all clients will display pagebreak metadata."

mattgarrish:
I was initially just going to fix #1437 with this PR, but expanded to include some additional informative explanations:

- it includes a new note in the epub:type definition to clarify that the `role` attribute addresses accessibility;
- it adds "dpub-aria role" fields for the structural semantics entries that have matching roles to simplify finding the alternatives - with an explanation about the field that mentions that usage of roles may be more restrictive

Also includes the new pagebreak text from dpub-aria 1.1, but without the normative requirement and stripping the text about what assistive technologies might do with the value.

While we explain these issues outside the specification, it can only help to be clearer about usage inside, especially now that the structure vocabulary is integrated.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1472.html" title="Last updated on Jan 19, 2021, 2:07 PM UTC (fed1074)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1472/9702689...fed1074.html" title="Last updated on Jan 19, 2021, 2:07 PM UTC (fed1074)">Diff</a>

### DIFF
--- a/epub33/common/css/common.css
+++ b/epub33/common/css/common.css
@@ -319,3 +319,17 @@ table.zebra tr:nth-child(even) {
 
 table.zebra th{border-bottom:1px solid #bbb;padding:.2em 1em;}
 table.zebra td{border-bottom:1px solid #ddd;padding:.2em 1em;}
+
+/* structure defs */
+
+section#structure-vocab dd > p:first-child {
+	margin-top: 1rem !important;
+}
+
+section#structure-vocab dd + dd {
+	font-size: 90%;
+}
+
+.subproplabel {
+	font-style: italic;
+}
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7814,6 +7814,19 @@ html.-epub-media-overlay-playing * {
 					already conveyed by an existing element (e.g., that a <code>div</code> represents a paragraph or
 					section).</p>
 
+				<div class="note">
+					<p>Although the <code>epub:type</code> attribute is similar in nature to the <a
+							href="https://html.spec.whatwg.org/multipage/infrastructure.html#attr-aria-role"
+								><code>role</code> attribute</a> [[HTML]], the attributes serve different purposes. The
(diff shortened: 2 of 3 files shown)


GeorgeKerscher:
I think this is a good addition.

iherman (approved):

clapierre (approved):

wareid (approved):

PR #1479, opened 2021-01-20
Clarify MO user classes must be defined in package document

### LINKED ISSUE #1319: Media overlay playback class names only a should?
Also ran across these somewhat contradictory statements:

> Visual rendering information for the currently-playing EPUB Content Document element MAY be expressed in the CSS Style Sheet using author-defined classes. These author-defined class names SHOULD be declared in the Package Document metadata using the metadata properties active-class and playback-active-class.

https://www.w3.org/publishing/epub/epub-mediaoverlays.html#sec-docs-assoc-style

The first sentence is clear that you don't have to supply the classes if you don't want, but the "should" in the second sentence is unclear. It could be read as contradicting the first (you should specify them), being non-committal about where to declare them (should be in the package document, but could be elsewhere?), or non-committal about the names (you should use these property names, but could use others?).

I expect, of course, it's none of the above and this could be reworded as a "must":

"When used, author-defined class names MUST be defined in the Package Document using the active-class and playback-active class properties."

### LINKED ISSUE #1320: Rules for active-class and playback-active-class
Looking at validation, I see a couple of underspecified areas around the use of the properties:

- should there be a requirement that every content document include the style sheet with the definitions so that the classes can be applied? The spec isn't clear what will happen if you put the classes in their own style sheet but don't declare it in the content document(s). Does the reading system harvest the styles and apply them into the documents, and how would it go about doing that?
- should we care about multiple declarations? For example, should there be a warning/info message to the user that they potentially duplicated their definitions, or do we let them silently cascade like any other CSS?
- also, if you declare the properties but don't include the named css classes in your style sheet, is that an error/warning/info?

mattgarrish:
Fixes the ambiguous wording noted in https://github.com/w3c/epub-specs/issues/1319#issue-663689940

I also noticed that the paragraph before the example and the two after are part of the example -- they're explanatory text. I've grouped them all together as one big example to make this clearer. I also repositioned the note about the class names just being examples immediately after their declarations.

Fixes #1319
Fixes #1320

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1479.html" title="Last updated on Jan 25, 2021, 11:49 AM UTC (9ae6dce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1479/90a703b...9ae6dce.html" title="Last updated on Jan 25, 2021, 11:49 AM UTC (9ae6dce)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -2131,7 +2131,7 @@
     …
     &lt;meta property="dcterms:modified"&gt;2016-02-29T12:34:56Z&lt;/meta&gt;
     &lt;meta property="rendition:layout"&gt;pre-paginated&lt;/meta&gt;
-    &lt;meta property="media:active-class"&gt;-epub-media-overlay-active&lt;/meta&gt;
+    &lt;meta property="media:active-class"&gt;my-active-item&lt;/meta&gt;
     …
 &lt;/metadata&gt;
 </pre>
@@ -7266,6 +7266,16 @@ store destination as source in ocf
 							href="#active-class"><code>active-class</code></a> and <a href="#playback-active-class"
 								><code>playback-active-class</code></a> properties.</p>
 
+					<p>Authors MUST define exactly one CSS class name in each property they define. Each property MUST
+						define a <a href="https://www.w3.org/TR/CSS2/syndata.html#characters">valid CSS class name</a>
+						not including any <a href="https://www.w3.org/TR/CSS2/selector.html">selectors</a> [[!CSS2]].
+						This specification <strong>does not</strong> reserve names for use with these properties.</p>
+
+					<p>Authors MAY define any CSS properties for the specified CSS classes. Authors only need to ensure
+						that each EPUB Content Document with an associated Media Overlay Document includes a link to the
+						CSS style sheet that contains the class definitions. Reading Systems might provide their own
+						styling, or no styling at all, in the absence of a linked definition.</p>
+
 					<p>The <code>active-class</code> and <code>playback-active-class</code> properties MUST NOT be used
 						in conjunction with a <a href="#attrdef-refines"><code>refines</code> attribute</a> as they are
 						always considered to apply to the entire <a>EPUB Publication</a>.</p>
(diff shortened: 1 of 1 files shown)


wareid (approved):

marisademeglio (approved):

iherman (approved):
Overall, +1 to this.

Reflecting on the comment of @GathComboy https://github.com/w3c/epub-specs/issues/1319#issuecomment-662644061 I wonder whether the fact of using `-epub-media-overlay-playing` as an example class name is a good choice. Using a `-epub` at the beginning does sound as this was some sort of official class name. Maybe using something like `my-active`, `my-playback-active` may look more generic and less official.

But this a tiny point.

mattgarrish:
> Maybe using something like `my-active`, `my-playback-active` may look more generic and less official.

This seems fine to me. What do you think @marisademeglio ?

marisademeglio:
I like this idea way better. How about:

`my-active-item` and `my-document-playing`

for "active specific item" and "whole document is playing" states?  It's easy to get the two mixed up, and as a bonus, these names are just verbose enough that hopefully no one will use them as-is.

mattgarrish:
I've fixed the examples as noted in the most recent commit, but have also made normative requirements out of what was implied by the note.

Have a look and make sure these are correct intepretations:

- each property must define exactly one class name with no selectors (was ambiguous about whether multiple are allowed, but assuming not since we don't define separators, etc.)
- each class name must be valid to css rules, but otherwise we don't require any specific names
- authors may define any properties (this was stated informatively)

I've also possibly fixed #1320, but informatively. I've noted that each content document with an associated media overlay needs to include a link to the css file that defines the classes otherwise no styling may be applied. I'm open to a should or must, but at least in theory an author might only want to style playback specially for specific content and let the RS control the rest??? Not sure what is an overreach in this case.

marisademeglio:
> I've fixed the examples as noted in the most recent commit, but have also made normative requirements out of what was implied by the note.
> 
> Have a look and make sure these are correct intepretations:
> 
>     * each property must define exactly one class name with no selectors (was ambiguous about whether multiple are allowed, but assuming not since we don't define separators, etc.)
> 
>     * each class name must be valid to css rules, but otherwise we don't require any specific names
> 
>     * authors may define any properties (this was stated informatively)
> 
> 

Yes agree on all

> I've also possibly fixed #1320, but informatively. I've noted that each content document with an associated media overlay needs to include a link to the css file that defines the classes otherwise no styling may be applied. I'm open to a should or must, but at least in theory an author might only want to style playback specially for specific content and let the RS control the rest??? Not sure what is an overreach in this case.

I don't know that it's feasible for validation to inspect style information to see that the actual class declaration is present, so I am comfortable leaving this as a "needs to" and not a should or must.

PR #1485, opened 2021-01-27
Clarify nav content rules and RS processing of elements

### LINKED ISSUE #975: Nav doc: access to all nav elements?
In section 5.3 we say:

> When requested by a user, it must provide access to the links and link labels in the nav elements of the EPUB Navigation Document in a fashion that allows the user to activate the links.

http://www.idpf.org/epub/31/spec/epub-packages.html#confreq-nav-access

Is this more wishful thinking than reality? Does any reading system present all the nav elements?

I also can't make this statement jibe with the landmarks nav, which is supposed to be there only to provide enhanced behaviours.

Should we try and make the statement more realistic, like a reading system must provide access to the table of contents and page list (when available) and may provide access to other navs?

### LINKED ISSUE #976: Nav doc: when is epub:type necessary
Section 5.4.1 currently reads:

> When a nav element carries the epub:type attribute [Content Docs 3.1] in an EPUB Navigation Document, this specification restricts the content model of the element and its descendants as follows:

http://www.idpf.org/epub/31/spec/epub-packages.html#sec-package-nav-def-model

But then in 5.4.2.5 it says:

> EPUB Navigation Documents MAY include one or more nav elements in addition to the toc, page-list and landmarks nav elements defined in the preceding sections. These additional nav elements SHOULD have an epub:type attribute to provide a machine-readable semantic, and MUST have a human-readable heading as their first child.

http://www.idpf.org/epub/31/spec/epub-packages.html#sec-package-nav-def-types-other

The problem is that if you omit epub:type to avoid the content model restrictions, you're supposed to get a warning. For most people, that will mean the content model restrictions cannot be avoided.

The navigation document is also supposed to be a content document that allows any kind of additional content within it, but these requirements prevent that from truly happening.

Since machine-processing of non-standardized nav elements seems unlikely, can we drop the "should" requirement from other navs and instead shift it so that reading systems should ignore nav elements that don't have an epub:type element or carry a semantic they don't recognize? (Somewhat overlapping here with #975.)

mattgarrish:
This PR fixes two problems with the current nav doc definition:

- it exempts nav elements without an `epub:type` attribute from the content model restrictions since their use by reading systems is generally unrealistic anyway.
- it sets out specific requirements for reading system handling of nav elements (must for toc, should for page-list when present, and may for everything else) as our current requirement to provide access to all the nav elements is sure to fail all reading systems when it comes time for testing

Reading Systems spec: [preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-976/epub33/rs/index.html) [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-976/epub33/rs/index.html)

Fixes #976 
Fixes #975

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1485.html" title="Last updated on Feb 15, 2021, 10:21 AM UTC (9128cd7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1485/133e2a9...9128cd7.html" title="Last updated on Feb 15, 2021, 10:21 AM UTC (9128cd7)">Diff</a>


epub33/core/index.html
```diff
@@ -4882,9 +4882,10 @@ <h5>Other <code>nav</code> Elements</h5>
 
 						<p>EPUB Navigation Documents MAY contain one or more <code>nav</code> elements in addition to
 							the <code>toc</code>, <code>page-list</code>, and <code>landmarks</code>
-							<code>nav</code> elements defined in the preceding sections. These additional
-								<code>nav</code> elements SHOULD have an <code>epub:type</code> attribute to provide a
-							machine-readable semantic and MUST have a human-readable heading as their first child.</p>
+							<code>nav</code> elements defined in the preceding sections. If these <code>nav</code>
+							elements are intended for Reading System processing, they MUST have an
+								<code>epub:type</code> attribute with a machine-readable semantic and are subject to the
```
wareid: `a machine-readable semantic` word missing here?

wareid (approved):
See comment, but otherwise looks good.

epub33/core/index.html
mattgarrish: Hm, it's actually kind of redundant to say, since the attribute by definition has machine-readable semantics and is required to have at least one value. I think I'll strip that out entirely.

iherman (approved):
I am fine with the changes.

However: the RS says something about the page-list and toc. It seems to be entirely silent on what it should do with the landmark nav element, although that is defined in the content document, alongside with page-list and toc. I must admit I do not know what a RS is expected to do with it... Should we give at least some guidance for RS-s?

mattgarrish:
> I must admit I do not know what a RS is expected to do with it...

Nothing specifically, so I consider it part of the "MAY" section.

It's supposed to help RSes implement their own interface -- like providing a mapping for a dedicated toc button. That's the "efficient access" part of the definition. But other than skipping to the body matter when a book opens I don't know how much the links are even used. 

We can't mandate reading systems implement behaviours or buttons, and otherwise there isn't a specific user-facing expectation. I haven't seen a reading system that lets users pull up the landmarks, for example, so it'd probably be hard to bump it up to a recommendation and also find implementations.

iherman:
> > I must admit I do not know what a RS is expected to do with it...
> 
> Nothing specifically, so I consider it part of the "MAY" section.
> 

O.k. I may be very picky, but I just may be the average reader: maybe it is worth adding (e.g., as a remark in parentheses) that this the "MAY" remark also applies to the landmark nav element, although that value of epub:type is explicitly defined in the spec.

But, again, I may just be too picky.

GeorgeKerscher:
Thorium has in the navigation area landmarks in addition to TOC, bookmarks, annotations and go to page.

 

I hear students say they wish publishers had a landmark at the beginning of a chapter that identifies all the essentials they should learn in this chapter. They would read this before they start reading.

mattgarrish:
> maybe it is worth adding ... that this the "MAY" remark also applies to the landmark nav element

I'm fine either way. I've added this for now:

> It MAY provide users access to the links in the landmarks nav element [EPUB-33] and MAY use the links to enable functionality in the application.

If there exists another implementation that gives users direct access to the links we could always consider moving it to a recommendation.

> I hear students say they wish publishers had a landmark at the beginning of a chapter that identifies all the essentials they should learn in this chapter.

You could probably do this with the `learning-objectives` or `learning-outcomes` semantics. It could also be done with ARIA within each chapter if there were better access to ARIA landmarks.

epub33/core/index.html
wareid: Sounds good! Just thought it read oddly.

mattgarrish:
Merged this as we need a clarification. We can always review the requirements if testing shows they're unrealistic.

PR #1498, opened 2021-02-02
Add auto value to dir attribute

### LINKED ISSUE #1490: (i18n) Assuming direction from language
The [i18n self test](https://github.com/w3c/epub-specs/wiki/Internationalization-self-test-for-the-EPUB-3.3.-spec/_edit) says:

> Do not assume that direction can be determined from language information. <a class="local" href="https://w3c.github.io/bp-i18n-specdev/#bidi_lang">more</a>

Maybe worth adding such a warning in the package specification. At this moment there is none.

### LINKED ISSUE #1491: (i18n) 'auto' value for the 'dir' attribute?
The [i18n self test](https://github.com/w3c/epub-specs/wiki/Internationalization-self-test-for-the-EPUB-3.3.-spec/_edit) says:

> Values for the default base direction should include left-to-right, right-to-left, and auto. <a class="local" href="https://w3c.github.io/bp-i18n-specdev/#bidi_values">more</a>

We do not have this for `dir`, there is a `default` for `page-progression-direction`. We could be inspired by the [HTML specification](https://html.spec.whatwg.org/multipage/dom.html#the-dir-attribute).  Note that the behavior of `dir` can also be modeled (or referred to?) the HTML spec.

### LINKED ISSUE #1494: (i18n) Relationship between dir and the content
The [i18n self test](https://github.com/w3c/epub-specs/wiki/Internationalization-self-test-for-the-EPUB-3.3.-spec/_edit) says:

> Specify that consumers of strings should use  heuristics, preferably based on the Unicode Standard first-strong algorithm, to detect the base direction of a string _**except where metadata is provided**_. <a href="https://w3c.github.io/string-meta/#best-practices-recommendations-and-gaps">more</a> (emphasis is mine, I.H.)

This requirement seems to be contradicted by the spec. At the moment, the [EPUB 3.3 spec says](https://w3c.github.io/epub-specs/epub33/core/#attrdef-dir)

> Inherent directionality specified using [Unicode] takes precedence over this attribute.

The way I can read this is that the first-strong algorithm in Unicode wins over the value of the `dir` attribute. 

That would certainly be wrong; the [Publication Manifest example](https://www.w3.org/TR/pub-manifest/#example-2-set-the-language-and-the-base-direction-of-a-string) shows the situation where the value of `dir` must be followed in contrast to the first strong algorithm.

Maybe that was not the intention in the EPUB spec, but then it needs clarification.

mattgarrish:
This pull request incorporates the changes suggest in #1491 and #1494.

Let me know if this better now or if there's more that needs to be done.

Fixes #1490
Fixes #1491 
Fixes #1494 

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/editorial/issue-1494/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/editorial/issue-1494/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1498.html" title="Last updated on Feb 9, 2021, 3:17 PM UTC (2842a92)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1498/2ad405b...2842a92.html" title="Last updated on Feb 9, 2021, 3:17 PM UTC (2842a92)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1244,12 +1244,22 @@
 								<code>dir</code>
 							</dt>
 							<dd>
-								<p>Specifies the base text direction of the content and attribute values of the carrying
-									element and its descendants.</p>
-								<p>Inherent directionality specified using [[!Unicode]] takes precedence over this
-									attribute.</p>
-								<p>Allowed values are <code>ltr</code> (left-to-right) and <code>rtl</code>
-									(right-to-left).</p>
+								<p>Specifies the base direction [[!BIDI]] of the carrying element and its
+									descendants.</p>
+								<p>Allowed values are:</p>
+								<ul>
+									<li><code>ltr</code> &#8212; left-to-right base direction;</li>
+									<li><code>rtl</code> &#8212; right-to-left base direction; and</li>
+									<li><code>auto</code> &#8212; base direction is determined using the Unicode Bidi
+										Algorithm [[BIDI]].</li>
+								</ul>
+								<p>The value <code>auto</code> is assumed when the attribute is not specified or
+									contains an invalid value.</p>
+								<div class="note">
+									<p>The base direction does not override the inherent direction of text runs within
+										an element but specifies how those runs are ordered relative to each other for
+										display.</p>
+								</div>
 								<div class="example">
 									<pre>&lt;package … dir="ltr"&gt;</pre>
 								</div>
@@ -8869,6 +8879,10 @@ EPUB/images/cover.png</pre>
 				-->
 
 				<ul>
+					<li>2-Feb-2021: Added <code>auto</code> value for <code>dir</code> attribute and clarified the
+						precedence of the attribute. See <a href="https://github.com/w3c/publ-epub-revision/issues/1491"
+							>issue 1491</a> and <a href="https://github.com/w3c/publ-epub-revision/issues/1494">issue
+							1494</a>.</li>
 					<li>20-Jan-2021: Clarified that user-defined media overlay style classes must be declared in the
 						Package Document metadata. See <a href="https://github.com/w3c/publ-epub-revision/issues/1319"
 							>issue 1319</a>.</li>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1256,9 +1256,9 @@
 								<p>The value <code>auto</code> is assumed when the attribute is not specified or
 									contains an invalid value.</p>
 								<div class="note">
-									<p>The base direction does not override the inherent direction of text runs within
-										an element but specifies how those runs are ordered relative to each other for
-										display.</p>
+									<p>The base direction does not override the logical order [[Unicode]] of text runs
+										within an element but specifies how those runs are ordered relative to each
+										other for display.</p>
 								</div>
 								<div class="example">
 									<pre>&lt;package … dir="ltr"&gt;</pre>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1256,9 +1256,9 @@
 								<p>The value <code>auto</code> is assumed when the attribute is not specified or
 									contains an invalid value.</p>
 								<div class="note">
-									<p>The base direction does not override the logical order [[Unicode]] of text runs
-										within an element but specifies how those runs are ordered relative to each
-										other for display.</p>
+									<p>The base direction specified in the dir attribute does not affect the ordering of
+										characters within directional runs, only the relative ordering of those runs and
+										the placement of weak directional characters such as punctuation.</p>
 								</div>
 								<div class="example">
 									<pre>&lt;package … dir="ltr"&gt;</pre>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1256,9 +1256,10 @@
 								<p>The value <code>auto</code> is assumed when the attribute is not specified or
 									contains an invalid value.</p>
 								<div class="note">
-									<p>The base direction specified in the dir attribute does not affect the ordering of
-										characters within directional runs, only the relative ordering of those runs and
-										the placement of weak directional characters such as punctuation.</p>
+									<p>The base direction specified in the <code>dir</code> attribute does not affect
+										the ordering of characters within directional runs, only the relative ordering
+										of those runs and the placement of weak directional characters such as
+										punctuation.</p>
 								</div>
 								<div class="example">
 									<pre>&lt;package … dir="ltr"&gt;</pre>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -306,6 +306,20 @@
 			<section id="sec-package-def">
 				<h3>Package Document</h3>
 
+				<section id="dir">
+					<h4>Base Direction</h4>
+
+					<p>If the <code>dir</code> attribute is set and indicates a base direction of <code>ltr</code> or
+							<code>rtl</code>, the Reading System MUST override the bidi algorithm per <a
+							data-cite="bidi#Higher-Level_Protocols">the higher-level protocols</a> defined in [[!BIDI]],
+						setting the paragraph embedding level to 0 if the base direction is <code>ltr</code>, or 1 if
+						the base direction is <code>rtl</code>.</p>
+
+					<p>Otherwise the base direction is <code>auto</code>, in which case the Reading System MUST
+						determine the text's direction by applying the Unicode Bidi Algorithm, beginning with <a
+							data-cite="bidi#P2">Rule P2</a> of [[!BIDI]].</p>
+				</section>
+
 				<section id="metadata">
 					<h4>Metadata</h4>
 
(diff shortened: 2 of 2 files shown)


iherman (approved):
Thanks!!

xfq:
Maybe it has no effect on EPUB, but the default value for [HTML](https://html.spec.whatwg.org/multipage/dom.html#the-dir-attribute), [CSS](https://drafts.csswg.org/css-writing-modes-3/#direction) and [TTML](https://www.w3.org/TR/ttml2/#style-attribute-direction) are all `ltr` (maybe for historical/compat reasons).

epub33/core/index.html
```diff
@@ -1244,12 +1244,22 @@ <h5>Shared Attributes</h5>
...
+									<li><code>rtl</code> &#8212; right-to-left base direction; and</li>
+									<li><code>auto</code> &#8212; base direction is determined using the Unicode Bidi
+										Algorithm [[BIDI]].</li>
+								</ul>
+								<p>The value <code>auto</code> is assumed when the attribute is not specified or
+									contains an invalid value.</p>
+								<div class="note">
+									<p>The base direction does not override the inherent direction of text runs within
+										an element but specifies how those runs are ordered relative to each other for
+										display.</p>
```
xfq: I'm not sure what "inherent direction" means here. Do you mean the [logical order](https://www.w3.org/TR/charmod/#sec-LogicalOrder) of the characters?

iherman:
> Maybe it has no effect on EPUB, but the default value for [HTML](https://html.spec.whatwg.org/multipage/dom.html#the-dir-attribute), [CSS](https://drafts.csswg.org/css-writing-modes-3/#direction) and [TTML](https://www.w3.org/TR/ttml2/#style-attribute-direction) are all `ltr` (maybe for historical/compat reasons).

Our constituent content files are in HTML or SVG, and they use CSS and TTML. It is probably confusing if the default value is not 'ltr', so I would support switching to that.

mattgarrish:
> It is probably confusing if the default value is not 'ltr', so I would support switching to that.

I'd want to hear more from reading system devs before making a change to a specific directionality as the default. It could break existing implementations. Can you open this as a separate issue?

epub33/core/index.html
mattgarrish: Thanks, yes, that's looks like the better technical phrasing for it.

r12a: I'm not sure that's correct either – depending on what you really wanted to say.

The **logical** order of Unicode text describes the order of characters in storage, which is unidirectional, and ordered essentially as the user would speak or type the text.  There is no bidi text or RTL text in storage. Everything is unidirectional.

However, to render the text you can imagine that it goes through two steps: 
- first, sequences of characters with the same _strong_ Unicode direction property are arranged into 'runs' in which the characters progress according to that direction, eg. a word in Hebrew containing only Hebrew characters has characters that proceed from right to left.
- second, the ordering of those directional runs relative to each other and the positioning of characters with only weak directional properties are determined by the base direction that is set.

For examples, see [Unicode Bidirectional Algorithm basics](https://www.w3.org/International/articles/inline-bidi-markup/uba-basics.en).

So to say that the dir attribute or base direction doesn't affect the logical order of text kind of goes without saying, because that's a fundamental Unicode rule.  But the intention might have been to say that the `dir` attribute or base direction doesn't affect the ordering of characters within directional runs, but does affect the relative ordering of those runs and the placement weak directional characters such as punctuation.

Does that help?

r12a:
The way i see it is that there is a difference between setting direction for a 'structure' (eg. a page, document, or subsection) and setting base direction for a string or inline run of text that is inserted into a document.

When setting direction for an inline run of text, you are probably best using first-strong heuristics as the default, but allowing that to be modified by explicit instructions where needed (eg. using the dir attribute).

For a structural item, you'd typically expect the structure to be one of LTR or RTL, rather than switching paragraph alignment, punctuation placement, etc. according to what just happens to begin each block.  (There may be rare exceptions for block-level elements, but this is nearly always the case for sections or pages.)  For that reason, i think a default value should be one of LTR or RTL, and of course if we pick one it's going to be LTR because it has wider applicability.  

So apart from the fact that the various specs mentioned originated long before the `auto` concept existed, i think that it still makes sense to choose LTR as the default direction for structural things, such as pages.  For handling strings, however, wrapping them in auto makes sense unless we know for sure what the appropriate base direction should be when they are displayed.

iherman:
> The way i see it is that there is a difference between setting direction for a 'structure' (eg. a page, document, or subsection) and setting base direction for a string or inline run of text that is inserted into a document.

@r12a, in the case of the EPUB specification, the only area of direction (and other) settings are metadata items in a specialized (and very simple) XML file. There are no pages, sections or subsections in that XML document, only individual text fields (without any internal structure). The real content is in HTML/SVG files within the EPUB file, and the EPUB specification refers to the I18 features of HTML/SVG.

We have begun to go through the I18n self-test, and we have added an introduction to describe the characteristics of the EPUB Specification. The self-test is not yet completed (we will obviously contact you when we are ready), but it may help if you look at that intro part: https://github.com/w3c/epub-specs/wiki/Internationalization-self-test-for-the-EPUB-3.3.-spec

iherman:
> > It is probably confusing if the default value is not 'ltr', so I would support switching to that.
> 
> I'd want to hear more from reading system devs before making a change to a specific directionality as the default. It could break existing implementations. Can you open this as a separate issue?

Done in #1499.

epub33/core/index.html
mattgarrish: > So to say that the dir attribute or base direction doesn't affect the logical order of text kind of goes without saying, because that's a fundamental Unicode rule.

I don't get the feeling any of us are completely sure what this statement was trying to get at:

> Inherent directionality specified using [Unicode] takes precedence over this attribute.

I think it may have been saying what you write in the former statement, but what you've expressed in the latter makes more sense. Any objections to adapting that as the note text, so modifying it to:

> The base direction specified in the `dir` attribute does not affect the ordering of characters within directional runs, only the relative ordering of those runs and the placement of weak directional characters such as punctuation.

r12a: Sounds good to me.

mattgarrish: Are you okay with the new wording @xfq ?

xfq: Looks good to me.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-02-04-epub#section1) on 2021-02-04

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-02-04-epub#resolution1):  Add the default value of `auto` to the `dir` element, resolved 1491 and 1499***

### 1. What should be the default value for the dir attribute 

_See github issue [#1499](https://github.com/w3c/epub-specs/issues/1499), [#1491](https://github.com/w3c/epub-specs/issues/1491)._




_See github pull request [#1498](https://github.com/w3c/epub-specs/pull/1498)._




**Dave Cramer:** Also going to talk about 1491 because they're related  
… there is also a PR addressing this (#1498)  
… most of the i18n issues we face are in HTML, SVG, CSS, where we rely on those specs  
… our area of influence is the package doc  
… the `dir` attribute  
… it doesn't have a default value  
… some guidance from i18n says we should have a default of `auto`  
… lets the system decide what to do, which depends on them, and can be overwritten by `rtl` or `ltr`  
… we have a PR open to address this  
… Matt has added this  
… is this a good idea and will it break anything  
… `ltr` = left to right  
… `rtl` = right to left  

**Brady Duga:** I think there is some history we want to keep with HTMLs LTR by default  
… for our case it makes sense to use `auto` because that's how things work now  
… the default is `auto`  
… HTML doesn't have a default, but the root is LTR, but it also has more complex inheritance  

**Dave Cramer:** I don't want to worry about the HTML situation  
… there are complexities  
… it seems to me that adding auto is risk free as a default  
… I can't imagine writing a test where something would happen  

**Brady Duga:** The scarier thing is the inherent directionality of the text winning over the `dir` attribute  

**Dave Cramer:** I wonder that too  
… went down a rabbit hole  
… stared into the abyss...  

**Brady Duga:** If there's an inherent directionality, it wins over `dir`, `dir` might mean nothing  

**Dave Cramer:** base paragraph direction over the bits of text within  
… atomic elements of the package file that have no instructions  
… instinct to have auto as a nice to have  
… doesn't change existing behaviour  
… the only person who needs to worry is the tester  
… I would propose we resolve to add the auto value  
… does anyone object  

> **Proposed resolution: Add the default value of `auto` to the `dir` attribute, resolved 1491 and 1499** *(Wendy Reid)*

> ***Resolution #1: Add the default value of `auto` to the `dir` element, resolved 1491 and 1499***

**Dave Cramer:** We got it! Moving on to 1494  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-02-04-epub#section4) on 2021-02-04

List of resolutions:

- ***[Resolution No. 4](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-02-04-epub#resolution4):  close 1490 because mooted by PR 1498***

### 4. Assuming direction from language 

_See github issue [#1490](https://github.com/w3c/epub-specs/issues/1490)._




**Dave Cramer:** 1490, do not assume that direction can be assumed from language information  
… but then we have PR 1498  

_See github pull request [#1498](https://github.com/w3c/epub-specs/pull/1498)._




**Brady Duga:** If we accept 1498, we don't care about this, according to the comments  

**Dave Cramer:** Not to mention the i18n folks who have commented  
… this one is kind of moot  

**Brady Duga:** Will not fix, obsolete  

> **Proposed resolution: close 1490 because mooted by PR 1498** *(Dave Cramer)*

> ***Resolution #4: close 1490 because mooted by PR 1498***

</details>

wareid (approved):

Issue #1156, opened 2018-10-20
[Discussion] Explicitly handle origin in further revisions?


JayPanoz:
Please allow me to open and use this issue as food for thought. 

This is much much longer term but it seems to me that not explicitly handling the `origin` concept has actually been an underlying issue creating compat, interoperability and security issues for all parties involved in the EPUB ecosystem – so users, authors, distributors (???), and Reading Systems. 

First things first, I’d vastly prefer not to be familiar with the origin concept and its security models, which can be painful to learn, but they do exist for a lot of reasons so it’s kinda worth putting ~~some~~ outstanding effort into understanding how it works. 

That’s to say I know this is a complex issue, but it’s already dealt with ± implicitly in the spec, albeit partially i.e. remote resources, JS guidance, etc. 

JS guidance is the most obvious case, as issue #873 might demonstrate. The thing is it’s been 2 years now, and the situation has not improved that much. 

To be fair, kudos to the people who did put effort into that; to name a few: iBooks, to which I personally reported a security issue, Daniel Weck @ Readium, or Mantano in Bookari. I’m pretty sure I’m forgetting some people so please be assured this is much appreciated if you did too. 

However, EPUB files still share the same origin in an awful lot of apps – i.e. something that is considered a security/authoring issue by other apps. It also has practical issues for authors because if say one file `localStorage.clear()` for instance, all EPUB files will lose the items they previously set (cf. education → quizzes, etc.). Web Storage also has quotas, so having a lot of books using it may create other issues for authors.

This isn’t the only example though. My goal isn’t to make a super extensive list, obviously, but to point out how handling origin could help quite significantly. 

By extension, origin could be unique but not persistent i.e. the Reading App assigns a random port @ launch, in which case you’ll have Web Storage spread across `127.0.0.1:3000` and `127.0.0.1:3001` for instance. 

Another issue is when the Reading App doesn’t use a (local) server behind the scenes so the `file://` scheme is used, and it comes with severe restrictions in some underlying rendering engines as the whole origin is consequently opaque (e.g. no Web Storage, restricted access to images, audio, video, iframes, etc.).

Speaking of iframes, we have issue #1061, and there’s once again little interop in RSs there, as they decide how to manage external links (e.g. opening the platform’s default browser, emptying the iframe’s content, etc.). 

[...]

PR #1511, opened 2021-02-10
Draft privacy & security section

iherman:
Just to have a starting point for a further discussion...

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1511.html" title="Last updated on Feb 17, 2021, 12:40 PM UTC (fedb99e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1511/4062eb1...fedb99e.html" title="Last updated on Feb 17, 2021, 12:40 PM UTC (fedb99e)">Diff</a>


mattgarrish:
I pushed a few minor formatting/grammar fixes, but otherwise no strong opinions on this.

What I think will ultimately be complex is separating out what are authoring concerns from what are reading systems concerns. Issues with xml and zip security probably belong in the RS spec, for example.

Are there authoring security concerns at all, beyond maybe trusting third-party content? Do we need to delve into privacy issues related to tracking user behaviour, which could be both an authoring and reading system concern?

But I suppose these topics are for another day.

iherman:
> I pushed a few minor formatting/grammar fixes, but otherwise no strong opinions on this.
> 
> What I think will ultimately be complex is separating out what are authoring concerns from what are reading systems concerns. Issues with xml and zip security probably belong in the RS spec, for example.
> 
> Are there authoring security concerns at all, beyond maybe trusting third-party content? Do we need to delve into privacy issues related to tracking user behaviour, which could be both an authoring and reading system concern?
> 

You are absolutely right; I simply forgot to look at the RS document. You are also right that the xml and zip things are for the RS; the relevant parts of the media type specification are clearly talking about the reading system.

The other two items are a bit of a catch-all; I think they can simply be kept in small. This section is not on providing specification but, rather, guidelines, so repeating does not harm. (For now at least.)

So I have (1) copied the content to the RS spec and (2) removed the xml and zip references from the core spec.

mattgarrish (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-02-11-epub#section3) on 2021-02-11

- no resolutions were taken

### 3. Security and Privacy Questionnaire 

_See github pull request [#1511](https://github.com/w3c/epub-specs/pull/1511)._




**Ivan Herman:** we have to put something about security and privacy in our recs, both in RS piece and content piece  
… i've made a start  
… we had this discussion in the context of publication manifest before  
… i've made a PR  
… what we do say right now is that the various security and privacy issues are mostly related to the content documents  
… so whatever is implementers and content providers need to be concerned about are the same things that would apply to HTML, etc.  
… just like with i18n, there are self-tests that the WG needs to provide  
… so we should take these self-tests, and see whether they reveal anything that we need to consider  
… like I did with i18n  
… we would need someone with some expertise in this area to do the self-test  

**Tzviya Siegman:** a lot of security questionnaire will be hard for us to answer  

**Dave Cramer:** dealing with origin is kind of way of expressing our security concerns in a way that the larger web understands  

**Ivan Herman:** i think many of the potential security and privacy issues are already handled by CSS and HTML  
… we are not supposed to redo these things  
… the question is whether the fact of packing things, adding metadata, etc. Do these things open new security and privacy issues?  

**Dave Cramer:** Jiminy has found epub specific security bugs  
… as a consequence of the package, and our uncertain relation to the web, there are additional issues  
… this may need to remain an open question for now  
… as we don't have obvious experts within our own group  

</details>

iherman:
@dauwhe @wareid @shiestyle, can we merge this? After all, it is a placeholder only...

iherman:
A pair of presentation given by @lrosenthol a few years ago is very much relevant for this discussion:

- [Web Security](https://lists.w3.org/Archives/Public/public-epub-wg/2021Feb/att-0007/Web_Security-Intro.pdf)
- [(P)WP	Security](https://lists.w3.org/Archives/Public/public-epub-wg/2021Feb/att-0007/PWP_Security.pdf)

Both were given at a meeting for the (now defunct) Publishing Interest Group.

(also relevant to #1156)

iherman:
@mattgarrish took care of merge conflicts.

PR #1512, opened 2021-02-11
Add clarifying note about legislated a11y requirements

mattgarrish:
As requested on the 2021-02-11 telecon, adds a note that meeting level a is not sufficient for conforming to European accessibility act or section 508


gregoriopellegrino (commented):
Fine for me!

PR #1518, opened 2021-02-19
Change security best practice to unique origin per publication

### LINKED ISSUE #873: What is origin in epub context?
OK so this might be a security issue to some extent. 

As far as I know, there’s nothing about “origin” in the EPUB spec. 

Why is this an issue? 

Because `localStorage`. See https://html.spec.whatwg.org/multipage/webstorage.html#the-localstorage-attribute and https://html.spec.whatwg.org/multipage/browsers.html#concept-origin

In other words, Reading System as the origin is valid, which means you can retrieve every item stored in the RS and not only the local storage area for one EPUB file.

Now, at the moment, it appears you can get items set in other EPUB files in some RS. See following screenshot (width was set in one file, we `getItem` using JavaScript in another file)

getitem

Here, we actually retrieve the whole storage using a loop (every item set in different files before running the script can be accessed)

capture d ecran 2016-09-11 a 20 00 39

I must admit I would be much more comfortable if origin = each EPUB file and not the whole RS. 

If someone set sensitive data in localStorage at some point, you could theoretically access it from another file and it would be valid per spec.

### LINKED ISSUE #1156: [Discussion] Explicitly handle origin in further revisions?

mattgarrish:
This PR updates the security section so that it discusses assigning a unique origin to each publication rather than assigning a unique domain per content document.

Let me know if the new note captures the issues that were raised on the 2021-02-18 telecon relating to how web-based reading systems cannot always enforce this kind of restriction and/or if there's anything else missing or that needs fixing.

Fixes #1156 
Fixes #873 

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1156/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1156/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2197,7 +2197,7 @@ alert("Feature " + feature + " supported?: " + conformTest);</pre>
 
 				<ul>
 					<li>19-Feb-2021: Updated the <a href="sec-scripted-content-security">scripting security
-							considerations</a> so that the text discusses assigning a unique origin to each EPUB
+							considerations</a> so that the text recommends assigning a unique origin to each EPUB
 						Publication for security rather than domain isolation at the Content Document level. See <a
 							href="https://github.com/w3c/epub-specs/issues/1156">issue 1156</a>.</li>
 				</ul>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -957,20 +957,18 @@
 
 					<ul>
 						<li>
-							<p>Reading Systems need to behave as if a unique domain were allocated to each Content
-								Document, as browser-based security relies heavily on document URLs and domains.
-								Adopting this approach will isolate documents from each other and from other Internet
-								domains, thereby limiting access to external URLs, cookies, DOM storage, etc.</p>
+							<p>Reading Systems need to behave as if a unique origin were allocated to each <a>EPUB
+									Publication</a>. Adopting this approach will isolate publications from each other,
+								thereby limiting access to cookies, DOM storage, etc.</p>
+							<div class="note">
+								<p>Although domain isolation is the most effective way to secure EPUB Publications in
+									Web-based Reading Systems, it is not always a realistic option. Web-based Reading
+									Systems still need to maintain isolation in such cases, which may require limiting
+									access to scripting APIs that allow client-side data storage and access.</p>
+							</div>
 							<p>Reading Systems that enable scripting and network access also need to consider including
 								methods to notify the user that network activity is occurring and/or that allow them to
 								disable it.</p>
-							<div class="note">
-								<p>In practice, Reading Systems might share domains across documents, but they still
-									need to maintain isolation between documents.</p>
-								<p>If parts of a document are encrypted and parts are not, or if different encryption
-									keys are used for different parts of the document, a unique per-document domain
-									might not provide sufficient protection.</p>
-							</div>
 						</li>
 						<li>
 							<p>If a Reading System allows persistent data to be stored, that data needs to be treated as
@@ -2188,8 +2186,8 @@ alert("Feature " + feature + " supported?: " + conformTest);</pre>
 					>Working Group's issue tracker</a>.</p>
 
 			<section id="changes-latest">
-				<h3>Substantive Changes since the <a href="https://www.w3.org/TR/2021/WD-epub-rs-33-20210112/">First
-						Public Working Draft</a></h3>
+				<h3>Substantive Changes since the <a href="https://www.w3.org/TR/2021/WD-epub-rs-33-20210223/">Previous
+						Working Draft</a></h3>
 
 				<!--
 					After each working draft is published:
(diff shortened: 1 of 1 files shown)


epub33/rs/index.html
```diff
@@ -2188,15 +2186,26 @@ <h2>Change Log</h2>
...
+						Working Draft</a></h3>
 
 				<!--
 					After each working draft is published:
 						- change the link/text in the section heading to refer to the published draft (use the dated URL)
 						- move all changes down to the next section
 				-->
 
+				<ul>
+					<li>19-Feb-2021: Updated the scripting security recommendations so that the text discusses assigning
```
dauwhe: Maybe "recommends" rather than "discusses"? Not sure of the ideal way to mention a non-normative change.

dauwhe (approved):

wareid (approved):

epub33/rs/index.html
mattgarrish: That's okay. W3C specs include a reference to RFC 8174, which makes it formal that non-capitalized uses of 2119 keywords only have their plain English meaning. I've been meaning to give the docs a once over to undo some of the more egregious uses of "might" and "need"... :)

epub33/rs/index.html
```diff
@@ -957,20 +957,18 @@ <h4>Security Considerations</h4>
 
 					<ul>
 						<li>
-							<p>Reading Systems need to behave as if a unique domain were allocated to each Content
-								Document, as browser-based security relies heavily on document URLs and domains.
-								Adopting this approach will isolate documents from each other and from other Internet
-								domains, thereby limiting access to external URLs, cookies, DOM storage, etc.</p>
+							<p>Reading Systems need to behave as if a unique origin were allocated to each <a>EPUB
+									Publication</a>. Adopting this approach will isolate publications from each other,
```
iherman: I think it is worth putting a reference to the 'formal' origin (ie, to https://url.spec.whatwg.org/#origin) emphasize the Web Security aspect. This can be as simple as making the word an active link or a formal reference to the URL spec (probably the best is to do both).

iherman: I wonder whether 'must' is not a better term. I know this is not a normative section, ie, a MUST is not an option, but the usage of the word 'must' may give more emphasis.

iherman (approved):
Wouldn't it be better to add a similar text to the content document as well? Authors should know that this restriction is in the RS because, if they want to do scripting, they have to plan accordingly...

epub33/rs/index.html
mattgarrish: Yes, no reason not to start fixing some of these now.

shiestyle (approved):

mattgarrish:
> Wouldn't it be better to add a similar text to the content document as well?

Yes, but I'm thinking of doing that in a separate PR. I want to look at the authoring document and ensure it still makes sense. It looks like not all the reading system requirements were reflected in the prose, like which contexts we recommend scripting support in and which are only optional.

iherman:
I just realized... we may want to add a link from §11 to §4.4.2. Or merge §4.4.2 to §11? I am tempted to go with the latter, but I am not sure what was the rationale for §4.4.2 originally.

PR #1519, opened 2021-02-22
Consolidate and clean up the scripting section

mattgarrish:
This PR makes the following changes:

- creates a new section called "script inclusion" that includes the text about adding scripts to content documents (removes the scripting context definitions)
- adds an explanation about origins and the effect on scripting (no data sharing) to the inclusion section
- creates a new "scripting contexts" section with a short intro followed by subsections that cover each of the contexts in detail (basically consolidates prose that was spread across sections before)
- creates a new appendix for detailed examples and moves the context example to it (adding together with the example of a packaged epub and wall clock values so we avoid a proliferation of top-level appendixes)

Mostly shuffling with some rewriting, in other words, but hopefully a little easier to read now.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1519.html" title="Last updated on Feb 24, 2021, 2:23 PM UTC (bd562a3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1519/b9642ea...bd562a3.html" title="Last updated on Feb 24, 2021, 2:23 PM UTC (bd562a3)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -4295,8 +4295,8 @@ Manifest:
 			<section id="sec-scripted-content">
 				<h3>Scripting</h3>
 
-				<section id="sec-scripted-context">
-					<h4>Scripting Contexts</h4>
+				<section id="sec-scripted-support">
+					<h4>Script Inclusion</h4>
 
 					<p><a>EPUB Content Documents</a> MAY contain scripting using the facilities defined for this in the
 						respective underlying specifications ([[!HTML]] and [[!SVG]]). When an EPUB Content Document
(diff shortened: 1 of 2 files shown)


epub33/core/index.html
```diff
@@ -4362,123 +4320,118 @@ <h4>Scripting Contexts</h4>
 							a future update adds the concept.</p>
```
iherman, line 4318: This is purely aesthetic: I would merge the two notes in one, and let the sentence on the data block follow the initial paragraph. The green boxes for notes break the flow...

epub33/core/index.html
```diff
@@ -4362,123 +4320,118 @@ <h4>Scripting Contexts</h4>
...
 					</div>
 
-					<p>Which context a script is used in determines the rights and restrictions that a Reading System
-						places on it. Refer to <a href="#sec-scripted-container-constrained">the following sections</a>
-						and <a href="https://www.w3.org/TR/epub-rs-33/#sec-scripted-content-rs-reqs">Scripting
-							Conformance</a> [[EPUB-RS-33]] for some specific requirements that have to be adhered to
-						(not all Reading Systems will provide the same scripting functionality).</p>
+					<p>Authors should note that Reading Systems are expected to behave as though a unique <a
+							href="https://url.spec.whatwg.org/#origin">origin</a> [[URL]] has been assigned to each EPUB
```
iherman, line 4322: I know this is in line with the change on the origin, but I would prefer to see a stronger term than "expected", ie, "required".

epub33/core/index.html
```diff
@@ -4362,123 +4320,118 @@ <h4>Scripting Contexts</h4>
...
 						</ul>
-					</aside>
-				</section>
 
-				<section id="sec-scripted-container-constrained">
-					<h4>Container-Constrained Scripts</h4>
+						<p id="confreq-cd-scripted-container">A container-constrained script MUST NOT contain
+							instructions for modifying the DOM of the EPUB Content Document that embeds it (i.e., the
+							one that contains the <code>iframe</code> element). It also MUST NOT contain instructions
+							for manipulating the size of its containing rectangle.</p>
```
iherman, line 4384: The problem with these statements is that I do not see how this can be tested. I presume epubcheck would be unable to do that, it would require to understand the behavior of the script.

I am completely out of my comfort zone here, maybe what I say is very different. Looking at the [iframe `sandbox` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) isn't it possible to replace this (or make it more precise) by applying constraints on the attributes the `<iframe>` element should or should not use? _That_ can be tested by epubcheck.

I am just wondering; maybe @rdeltour, @danielweck, or @bduga can comment on this...

epub33/core/index.html
```diff
@@ -4362,123 +4320,118 @@ <h4>Scripting Contexts</h4>
...
+									><code>script</code></a> or [[!SVG]] <a
+								href="https://www.w3.org/TR/SVG/interact.html#ScriptElement"><code>script</code></a>
+							element contained in a <a>Top-level Content Document</a>.</p>
+
+						<p>Authors should note that <a
+								href="https://www.w3.org/TR/epub-rs-33/#sec-scripted-content-rs-reqs">support for
+								spine-level scripting in Reading Systems</a> is only recommended in fixed layout
+							documents and reflowable documents when they are set to scroll [[EPUB-RS-33]]. Furthermore,
+							Reading System support in all other contexts is optional.</p>
+
```
iherman, line 4408: I wonder whether a more direct reference to [§4.4.1](https://w3c.github.io/epub-specs/epub33/rs/#sec-scripted-content-rs-reqs) in the RS document, that is much more precise about by the properties that must be used. Ie, statements like:

> reflowable EPUB Content Documents that use the "scrolled-doc" or "scrolled-continuous" [EPUB-33] presentation modes defined by the rendition:flow property.

are much more concrete and enforceable, ie, the author should use those constraints.

That being said, I am a bit bothered by this restriction (I know, that may be an inherited feature that we cannot change). For example, a script that puts a line number to each line in a `<pre>` element does not fall into this category if the content is displayed in paginated mode, although that may make a lot of sense. I understand that the danger is to interfere with the pagination algorithm of the RS, but I still wonder whether we can be, somehow, more liberal...

epub33/core/index.html
mattgarrish: Can you open a separate issue for this as it would change existing normative requirements?

I don't want to open a new discussion in a PR.

epub33/core/index.html
mattgarrish: > I understand that the danger is to interfere with the pagination algorithm of the RS, but I still wonder whether we can be, somehow, more liberal...

Again, this is probably one for a new issue.

epub33/core/index.html
mattgarrish: Merging doesn't really make sense in this case as the notes refer to two separate statements. It would jumble the discussion.

The first note could probably be regular spec prose, though.

epub33/core/index.html
iherman: You are right. I will.

pinging @dauwhe and @wareid because we just had some discussions about this...

epub33/core/index.html
iherman: True for the last remark, and maybe we should simply ignore this for now...

epub33/core/index.html
iherman: done in #1523

iherman (approved):

PR #1520, opened 2021-02-23
Extending ECHIDNA to the A11y documents

iherman:
Just a copy and paste for a11y and a11-tech. The TOKEN_A11Y and TOKEN_A11Y_TECH were used when creating the secure value in the relevant branch via, e.g.:

```
travis encrypt --pro TOKEN_A11Y="<token>"
```


deniak (approved):
LGTM!

PR #1531, opened 2021-02-25
Update conformance reporting

### LINKED ISSUE #1455: Reporting accessibility conformance
The old IDPF conformance URLs have to be updated for the next release and we also need to figure out how to ensure a more stable reporting method going forward.

The base URL to use for the epub accessibility specification should be fairly straightforward as we'll have a versioned /TR path to use this time (no need for a dated specification).

The complexity comes in also having to report the WCAG version and level, assuming we are allowing flexibility in conformance. We should figure out how important this information is to break out. Do vendors care about the specifics of conformance, for example, or only that content conforms to the accessibility standard?

In a world where the details aren't that critical, one option would be to drop the WCAG conformance level from the `conformsTo` URL entirely so only the accessibility specification URL is used. The WCAG version and level could be required in the accessibility summary, for example, so users can still understand the quality. For example:

```
<link rel="dcterms:conformsTo" href="https://www.w3.org/TR/epub-a11y-11/"/>
<meta property="schema:accessibilitySummary">... This publication conforms to WCAG 2.1 Level AA. ... </meta>
```

One problem with this approach, however, is that the information won't be easily parseable by tools like Ace. But these tools may also always assume conformance to the latest version (making them account for variances between releases could prove a maintenance headache).

Alternatively, we could require (or recommend) a second `conformsTo` field that specifies the WCAG version:

```
<link rel="dcterms:conformsTo" href="https://www.w3.org/TR/epub-a11y-11/"/>
<link rel="dcterms:conformsTo" href="https://www.w3.org/TR/WCAG21/"/>
```

The problem remains here is that WCAG doesn't have URLs for each conformance level, so unless we require AA conformance we're still going to be missing information. We could require that context in the summary, though.

Looking at the Dublin Core definitions again, we might also simply consider using a natural language descriptor for WCAG conformance. The vocabulary now says this:

> Creators of non-RDF metadata can use the terms in contexts such as XML, JSON, UML, or relational databases by disregarding both the global identifier and the formal implications of the RDF-specific aspects of term definitions. Such users can take domain, range, subproperty, and subclass relations as usage suggestions and focus on the natural-language text of definitions, usage notes, and examples.

https://dublincore.org/specifications/dublin-core/dcmi-terms/#section-1

This sounds to me like the rigidity of dcterms properties for linked data has been relaxed. In that case, we could require authors to follow a pattern like "WCAG 2.X Level AA":

```
<link rel="dcterms:conformsTo" href="https://www.w3.org/TR/epub-a11y-11/"/>
<meta property="dcterms:conformsTo">WCAG 2.1 Level AA</meta>
```

We could also take this one step further and incorporate everything in one common pattern like:

```
<meta property="dcterms:conformsTo">EPUB Accessibility 1.1 + WCAG 2.1 Level AA</meta>
```

That's as far as my thought experiments have taken me with this, at any rate. Other ideas welcome.

/cc @avneeshsingh @clapierre @GeorgeKerscher

mattgarrish:
This PR makes the following changes per our discussion on the 2021-02-25 telecon:

- adds and explains the new text string pattern for reporting
- it splits the section in two so that publication reporting is separated from evaluator reporting - to make reading the relevant sections simpler
- it adds a short intro for some prose that was left hanging, although I'm not sure we really need this (can be convinced to drop it)

I formatted the new text string pattern with placeholder variables where each moving piece is filled in and then put a definition list beneath explaining what each value must be.

Let me know if this makes sense, or if anyone has other thoughts on how we might approach this. I can't think of a similar instance like this I've encountered before in a specification. (Have you run into this kind of a pattern before @iherman ?)

- [Accessibility preview](https://raw.githack.com/w3c/epub-specs/fix/issue-1455/epub33/a11y/)
- [Accessibility diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fw3c.github.io%2Fepub-specs%2Fepub33%2Fa11y%2F&doc2=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fraw.githack.com%2Fw3c%2Fepub-specs%2Ffix%2Fissue-1455%2Fepub33%2Fa11y%2Findex.html)

Fixes #1455

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -57,6 +57,14 @@
 				preProcess:[inlineCustomCSS],
 				postProcess:[addConformanceLinks]
 			};</script>
+		<style>
+			.conf-pattern {
+				margin-left: 3rem;
+				font-style: italic;
+			}
+			.varlist {
+				margin-left: 3rem;
+			}</style>
 	</head>
 	<body>
 		<section id="abstract">
@@ -115,8 +123,9 @@
 					them. This approach allows the guidelines to remain stable even as the format evolves.</p>
 
 				<p>To facilitate this approach, the companion <a href="https://www.w3.org/TR/epub-a11y-tech-11/">EPUB
-						Accessibility Techniques</a> [[EPUB-A11Y-TECH-11]] document outlines conformance techniques. These
-					techniques explain how to meet the requirements of this document for different versions of EPUB.</p>
+						Accessibility Techniques</a> [[EPUB-A11Y-TECH-11]] document outlines conformance techniques.
+					These techniques explain how to meet the requirements of this document for different versions of
+					EPUB.</p>
 			</section>
 
 			<section id="sec-application" class="informative">
@@ -310,10 +319,10 @@
 					Publications. (The relationship to WCAG is explained for each requirement in its respective
 					section.)</p>
 
-				<p>The same is true of the techniques in the EPUB Accessibility Techniques document [[EPUB-A11Y-TECH-11]].
-					It provides coverage of techniques that are unique to EPUB Publications, or that need clarification
-					in the context of an EPUB Publication. It does not mean that the rest of the WCAG techniques are not
-					applicable.</p>
+				<p>The same is true of the techniques in the EPUB Accessibility Techniques document
+					[[EPUB-A11Y-TECH-11]]. It provides coverage of techniques that are unique to EPUB Publications, or
+					that need clarification in the context of an EPUB Publication. It does not mean that the rest of the
+					WCAG techniques are not applicable.</p>
 
 				<p>As a result, although this section can be read without deep knowledge of WCAG conformance, to
 					implement the accessibility requirements of this document will require an understanding of WCAG.</p>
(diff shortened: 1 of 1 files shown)


clapierre:
I thought we discussed 3 pieces of data this seems there are only two EPUB Accessibility and WCAG version and combined Level.

So instead of
EPUB Accessibility 1.1 + WCAG 2.0 Level A
this would be
EPUB Accessibility 1.1 + WCAG 2.0 +  Level A

mattgarrish:
There are three pieces of information that are variable, but two of them are specific to WCAG.

For me, adding another plus before the level makes it seem like "Level A" is distinct from either the epub or wcag specs, but it's not something I feel all that strongly about.

mattgarrish:
Another option might be to put a dash before the level rather than a plus?

clapierre:
I see your point Matt, and I guess it doesn't really matter I am fine with the way it is currently or with the - as you suggest, one would probably not parse out the level from the WCAG version anyways so having them as one entity seems reasonable.  As long as others agree.

clapierre:
Just making sure this still works with future versions of WCAG
EPUB Accessibility A11Y-VER + WCAG WCAG_VER Level WCAG-LVL
IE:
EPUB Accessibility 1.1 + WCAG 3.0 Level Bronze

iherman:
> Have you run into this kind of a pattern before @iherman ?

No, but that may not be relevant. The pattern, i.e., to store the result of some conformance check of a resource as part of the resource in the form of metadata, is not one that I have seen before outside the WAI area. Which is not a judgement at all, simply that I do not have the experience to offer...

mattgarrish:
> to store the result of some conformance check of a resource as part of the resource in the form of metadata

I wasn't wondering that specifically, but only whether you'd seen in any W3C specs this sort of a string pattern where you interpolate variables.

Putting placeholder variables in the pattern with a "where" list to define them isn't unusual, but maybe there are better/clearer ways of doing this? (e.g., put quotes around the text strings with + to indicate the variables are concatenated together, although I opted not to do that as it looked much more complex to read.)

iherman:
> > to store the result of some conformance check of a resource as part of the resource in the form of metadata
> 
> I wasn't wondering that specifically, but only whether you'd seen in any W3C specs this sort of a string pattern where you interpolate variables.
> 

You could use ABNF… But please don't :-) It is really not easy to read.

> Putting placeholder variables in the pattern with a "where" list to define them isn't unusual, but maybe there are better/clearer ways of doing this? (e.g., put quotes around the text strings with + to indicate the variables are concatenated together, although I opted not to do that as it looked much more complex to read.)

I would leave it as is.

Issue #1528, opened 2021-02-25
Space trimming in dc:title and DCMES optional elements.


emuller-amazon:
For dc:title (https://www.w3.org/TR/2021/WD-epub-33-20210224/#sec-opf-dctitle), it is only at the very end of this section that one finds "This specification imposes no additional restrictions or requirements on the title except that it MUST be at least one character in length after white space has been trimmed.."

For DCMES optional elements (https://www.w3.org/TR/2021/WD-epub-33-20210224/#sec-opf-dcmes-optional), the text is just after the formal definition of the elements (the blue box) and slightly different: "The value of all OPTIONAL [DC11] elements MUST be at least one character in length after white space has been trimmed."

Suggestion 1 is to adopt the position and wording of "DCMES optional elements" for "dc:title".

Suggestion 2: I think the intent is that consumers of those elements should trim the white space. Strictly speaking, that is not implied by the text. May be change it to: "Initial and final white space in the value of {this element/all OPTIONAL [DC11] elements} is not significant and there MUST be at least one character in length after white space has been trimmed."

Also: do you intend only trimming of initial and final whitespace, or something more like CSS (e.g. collapsing multiple spaces in the middle)?

dauwhe:
See also https://github.com/w3c/epub-specs/issues/1295

iherman:
> See also #1295

Yep, indeed. There is a longer discussion on how to define this properly in that issue which we should probably follow.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-02-26-epub#section3-2) on 2021-02-26

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 3.2. Trimming leading and following whitespace in package metadata 

_See github issue [#1295](https://github.com/w3c/epub-specs/issues/1295), [#1528](https://github.com/w3c/epub-specs/issues/1528)._




**Dave Cramer:** Let's discuss 1295  
… package metadata, according to the spec we need to trim leading and trailing whitespace  
… the INFRA spec does have instructions on this  
… it would be a step in the right direction to define the trimming as described in the INFRA spec  
… this is separate to a discussion on whitespace within the strings  
… there is a major RS that ignores this requirement  

**Garth Conboy:** What is the proposed change?  

**Dave Cramer:** Could it be as simple as linking the word "trim" to the INFRA spec?  

[...]

Issue #1295, opened 2019-09-09
Trimming leading and following whitespace in package metadata


dauwhe:
The packages spec says the following many, many times about almost all the metadata in the OPF file:

> Reading Systems MUST trim all leading and trailing white space [XML] before processing the value.

Inconveniently, the word "trim" does not appear in the XML 1.0 spec. 

Conveniently, it turns out that WHATWG has a infrastructure spec (who knew?) that actually has a [definition](https://infra.spec.whatwg.org/#strip-leading-and-trailing-ascii-whitespace) of this

> To strip leading and trailing ASCII whitespace from a string, remove all ASCII whitespace that are at the start or the end of the string. 

I wonder if we should improve our definition. 

Incidentally, it appears many reading systems ignore this. I wrote a test this afternoon...

danielweck:
... and this should probably apply consistently to metadata attribute values, as well as element content (text child node), right?

Another interesting parsing rule is whitespace *collapsing* (space characters, tabs, etc. merged into a single space) *inside* strings, after leading and trailing whitespace has been removed. This would typically be needed for long textual descriptions, which may span across multiple lines, with indentation (i.e. "insignificant" whitespace).

So, how much of that is purely down to de-facto XML string normalization rules (in other words, reading system implementations just use whatever text the XML parser returns), vs. an additional layer of parsing rules, defined by the EPUB specification?

dauwhe:
At least one major reading system does not collapse whitespace within metadata, even just plain runs of U+0020.

rdeltour:
@dauwhe 
> Inconveniently, the word "trim" does not appear in the XML 1.0 spec.

I think the XML reference is for the definition of _white space_, not for _trim_.

> it turns out that WHATWG has a infrastructure spec (who knew?) that actually has a definition of this (…) I wonder if we should improve our definition.

Yes, it would be a good clarification IMO.

Note that the [_ASCII whitespace_](https://infra.spec.whatwg.org/#ascii-whitespace) definition in the Infra standard differs from the [_white space_](https://www.w3.org/TR/xml/#NT-S) definition in XML in that the former includes the form feed character (U+000C) when the latter does not. But this is OK since [control characters are disallowed in well-formed XML 1.0 documents](https://www.w3.org/TR/xml/#wf-Legalchar) anyways.

@danielweck 
> this should probably apply consistently to metadata attribute values, as well as element content (text child node), right?

[...]

PR #1539, opened 2021-02-26
Update whitespace handling

mattgarrish:
As discussed on the 2021-02-26 telecon, this PR adds a cross-reference to infra spec definition for stripping leading and trailing whitespace from metadata elements in the reading system spec definition.

I've also created a new section in the core spec under the package metadata section for common metadata value requirements. These are the only elements that accept text values, so I've moved the floating paragraph that used be at the start of the package doc definition into it. I've also added a single statement to cover the requirement for at least one character after whitespace stripping so we don't have to maintain these statements in every definition.

This only partially solves #1295 and #1528 since we haven't yet addressed whitespace collapsing within elements.

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/editorial/issue-1295/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/editorial/issue-1295/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1539.html" title="Last updated on Mar 1, 2021, 4:34 PM UTC (8ee4fd0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1539/7a60ccf...8ee4fd0.html" title="Last updated on Mar 1, 2021, 4:34 PM UTC (8ee4fd0)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1561,12 +1561,15 @@
 							<h5>Metadata Values</h5>
 
 							<p>The Dublin Core elements [[!DC11]] and <a href="#sec-meta-elem"><code>meta</code>
-									element</a> have mandatory text content. That content is referred to as the
-									<dfn>value</dfn> of the element in their descriptions.</p>
+									element</a> have mandatory <a
+									href="https://dom.spec.whatwg.org/#concept-child-text-content">child text
+									content</a> [[!DOM]]. That content is referred to as the <dfn>value</dfn> of the
+								element in their descriptions.</p>
 
-							<p>These elements MUST have non-empty values after Reading Systems <a
-									href="https://www.w3.org/TR/epub-rs-33/#conf-meta-ws">strip leading and trailing
-									whitespace</a> [[!EPUB-RS-33]].</p>
+							<p>These elements MUST have non-empty values after <a
+									href="https://infra.spec.whatwg.org/#strip-leading-and-trailing-ascii-whitespace"
+									>leading and trailing ASCII whitespace</a> [[!Infra]] is stripped (i.e., they must
+								consist of at least one non-whitespace character).</p>
 						</section>
 
 						<section id="sec-opf-dcmes-required">
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1100,9 +1100,6 @@
 							<code>http://www.idpf.org/2007/opf</code> namespace [[!XML-NAMES]] unless otherwise
 						specified.</p>
 
-					<p>When an element defined in this section has mandatory text content, that content is referred to
-						as the value of the element in the explanatory descriptions.</p>
-
 					<section id="sec-package-elem">
 						<h5>The <code>package</code> Element</h5>
 
@@ -1560,6 +1557,18 @@
 
 						</section>
 
+						<section id="sec-package-elem-value">
+							<h5>Metadata Values</h5>
+
+							<p>When a child element of the <a href="#sec-metadata-elem"><code>metadata</code>
+									element</a> has mandatory text content, that content is referred to as the value of
+								the element in the explanatory descriptions.</p>
+
+							<p>These values MUST be at least one character in length after <a
+									href="https://www.w3.org/TR/epub-rs-33/#conf-meta-ws">stripping leading and trailing
+									ASCII whitespace</a> [[!EPUB-RS-33]].</p>
+						</section>
+
 						<section id="sec-opf-dcmes-required">
 							<h6>DCMES Required Elements</h6>
 
@@ -1634,7 +1643,8 @@
 									NOT be issued when making minor revisions such as updating metadata, fixing errata,
 									or making similar minor changes.</p>
 
-								<p>Authors MAY specify additional identifiers.</p>
+								<p>Authors MAY specify additional identifiers. It is strongly advised that identifiers
+									be fully qualified URIs.</p>
 
 								<p>The <a href="#identifier-type"><code>identifier-type</code> property</a> is used to
 									indicate that an <code>identifier</code> conforms to an established system or has
@@ -1651,11 +1661,6 @@
     …
 &lt;/metadata></pre>
 								</aside>
-
-								<p>This specification imposes no additional restrictions or the requirements of the
-									identifier except that it MUST be at least one character in length after white space
-									has been trimmed. It is strongly advised that the identifier be a fully qualified
-									URI, however.</p>
 							</section>
 
 							<section id="sec-opf-dctitle">
@@ -1751,10 +1756,6 @@
 &lt;/metadata&gt;
 </pre>
 								</aside>
-
-								<p>This specification imposes no additional restrictions or requirements on the title
-									except that it MUST be at least one character in length after white space has been
-									trimmed.</p>
 							</section>
 
 							<section id="sec-opf-dclanguage">
@@ -1885,9 +1886,6 @@
 										<p>Text</p>
 									</dd>
 								</dl>
-
-								<p>The value of all OPTIONAL [[!DC11]] elements MUST be at least one character in length
-									after white space has been trimmed.</p>
 
 								<p>This specification does not modify the [[!DC11]] element definitions except as noted
 									in the following sections.</p>
@@ -2185,10 +2183,6 @@
 									value of the <code>meta</code> tag is drawn from ONIX code list 5.</p>
 								<pre>&lt;meta refines="#isbn-id" property="identifier-type" scheme="onix:codelist5">15&lt;/meta></pre>
 							</aside>
-
-							<p>Every <code>meta</code> element MUST express a value that is at least one character in
-								length after white space normalization.</p>
-
 						</section>
 
 						<section id="sec-metadata-last-modified">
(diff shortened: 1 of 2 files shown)


iherman (approved):

dauwhe (approved):

epub33/core/index.html
```diff
@@ -1560,6 +1557,18 @@ <h6>The <code>metadata</code> Element</h6>
 
 						</section>
 
+						<section id="sec-package-elem-value">
+							<h5>Metadata Values</h5>
+
+							<p>When a child element of the <a href="#sec-metadata-elem"><code>metadata</code>
+									element</a> has mandatory text content, that content is referred to as the value of
+								the element in the explanatory descriptions.</p>
```
rdeltour: The term being defined could be marked up to stand out and so that it can be referenced from the other spec?

Also, "value of the element" can be confusing. A better terms might be "value of the metadata expression" or something along these lines (
("element" is markup terminology, here we're defining metadata terminology).

Finally, it would perhaps be easier to define the "value" in the general case (empty or not), and then state case-by-case when the value must be non-empty. In the current wording, the term is only defined when the value is mandatory (i.e. not empty).

epub33/core/index.html
```diff
@@ -1560,6 +1557,18 @@ <h6>The <code>metadata</code> Element</h6>
...
+
+							<p>These values MUST be at least one character in length after <a
+									href="https://www.w3.org/TR/epub-rs-33/#conf-meta-ws">stripping leading and trailing
+									ASCII whitespace</a> [[!EPUB-RS-33]].</p>
```
rdeltour: I would put the "stripping leading and trailing" in the definition of the value, and put "value MUST be non-empty" statements case-by-case in the metadata property definitions.

If all metadata expressions mandate non-empty values, then the above definition of value can be simplified (no "When"), and the "value MUST be non-empty statement" can be defined for all.

Also, shouldn't "stripping leading and trailing" refer to INFRA rather than EPUB-RS here?

epub33/rs/index.html
```diff
@@ -341,13 +341,15 @@ <h4>Metadata</h4>
...
-									<code>meta</code> element values before further processing them.</p>
+							<p>Reading Systems MUST <a
+									href="https://infra.spec.whatwg.org/#strip-leading-and-trailing-ascii-whitespace"
+									>strip leading and trailing ASCII whitespace</a> [[!Infra]] from Dublin Core element
+								values before processing.</p>
+							<p>Unless an individual property explicitly defines a different whitespace normalization
+								algorithm, Reading Systems MUST <a
+									href="https://infra.spec.whatwg.org/#strip-leading-and-trailing-ascii-whitespace"
+									>strip leading and trailing ASCII whitespace</a> [[!Infra]] from <code>meta</code>
+								element values before further processing them.</p>
```
rdeltour: I was wondering what "element values" meant (when reading the preview), before seeing that this PR defined that term in the core spec. I suggest (see comment above) a more specific term is used (e.g. "metadata value" or "metadata expression value"), and that its usage here points to its definition in the Core spec.

rdeltour (changes requested):
Looks good generally, but I suggest the "value" concept could be better defined, be referenceable, and use less markupy and more metadatay terminology (e.g. "metadata value" instead of "element value"). See inline comments.

epub33/core/index.html
mattgarrish: I think what you're proposing might make the text more confusing. The element definitions talk in places about their "value". The meta element has a value, for example, but to get to a "metadata expression" you need to also be talking about what is expressed in its property attribute (or the inherent semantics of the DCMES elements).

Looking again at this though, what might help is to be even more specific. We know exactly what elements this applies to - the DCMES elements and meta - so we can list those.

I'm fine with setting value up as a definition that we can link to, though. That makes sense.

> Finally, it would perhaps be easier to define the "value" in the general case (empty or not)

Note completely sure what you mean about defining the "value"? The elements already have a content model of text in their definitions. Are you suggesting we try to further define what that means?

epub33/core/index.html
mattgarrish: > If all metadata expressions mandate non-empty values, then the above definition of value can be simplified (no "When"), and the "value MUST be non-empty statement" can be defined for all.

Right, this is what being more specific might help with. There is a finite set of elements and they all require a non-empty value. That's why I'd rather define this once than list it for each. It's always been cumbersome.

> Also, shouldn't "stripping leading and trailing" refer to INFRA rather than EPUB-RS here?

I considered that, but it's technically the RS that does the stripping. If you want to find out more, the first hop should be to the RS spec.

rdeltour: > > Also, shouldn't "stripping leading and trailing" refer to INFRA rather than EPUB-RS here?
> 
> I considered that, but it's technically the RS that does the stripping. I

Even if the RS will effectively do the stripping at some point, the stripping is a fundamental part of the MUST statement (or of the definition of "value", alternatively).

As an example (out of thin air 😇😁),  a checker would need to do the stripping too, it shouldn't have to go through the RS spec indirection to interpret this MUST statement 😊.

epub33/core/index.html
rdeltour: > I think what you're proposing might make the text more confusing. The element definitions talk in places about their "value". The meta element has a value, for example, but to get to a "metadata expression" you need to also be talking about what is expressed in its property attribute (or the inherent semantics of the DCMES elements).

OK fair enough. I'm not too hung-up on _not_ using "element value", as long as the term is a proper definition that can be referenced. In fact, I came here from the RS spec preview, when I found "element value" and was confused by that terminology which didn't sound familiar (from HTML or XML). So it's probably OK if uses of the term can be linked to the definition.

> Note completely sure what you mean about defining the "value"? The elements already have a content model of text in their definitions. Are you suggesting we try to further define what that means?

I meant that the "value" of an element is not defined by XML, so we should be more explicit about what we mean here. DOM defines a [`nodeValue`](https://dom.spec.whatwg.org/#dom-node-nodevalue) attribute for nodes, which for elements is unambiguously defined as the [descendant text context](https://dom.spec.whatwg.org/#concept-descendant-text-content). I think we should use a similar unambiguous definition of what is exactly meant by "value" here.

The current text says:
> "When a child element of the metadata element has mandatory text content, that content is referred to as the value of the element in the explanatory descriptions."

As all the elements require a non-empty value, why make a case ("when") for mandatory text content when defining what we refer to when we say "the value of the element"?

What I suggest is something along these lines:

- add a definition for "the value of the element" (not my favorite term, but OK), based on a concept similar to DOM's _descendant text content_, without the "when".
- add the MUST not be empty statement for that term just defined, based on INFRA.

So a lot like what is already there, but more explicit :)

epub33/core/index.html
mattgarrish: > a checker would need to do the stripping too, it shouldn't have to go through the RS spec indirection to interpret this MUST statement

I'd argue it does, though, as the stripping is conditional.

For DC elements its required, but there's a loophole for the meta element that properties can override the default behaviour. This is underspecified, in my opinion, as I believe the intention was only that properties _we_ define can override, not that any property defined in any vocabulary can change our rules. That would be a nightmare of inconsistency to deal with.

But given that I'm not aware of any properties we define that change the whitespace handling rules, I wonder if we can just do away with that rule entirely?

I'd then be more at ease with each spec separately referencing across to infra.

epub33/core/index.html
mattgarrish: > I'm not too hung-up on _not_ using "element value"

For context, this complaint came a couple of revisions back that we were using "value" for element text content when the XML spec doesn't define such a concept. The prime reason I want to stick with it is that changing to something else like "metadata expression" means we probably also have to pick through the whole spec and try to rewrite the references so that they also make sense about the text content being the expression. This made people happy at the time, so I'd rather not try to reinvent it, not that I necessarily love the terminology either. 😄 

> * add a definition for "the value of the element" (not my favorite term, but OK), based on a concept similar to DOM's _descendant text content_, without the "when".

That's probably too complex as EPUB doesn't have element nesting. Even the DOM definition of child text content is confusing in our context, as there's only one instance of child text content so nothing to concatenate.

An element has text content and that's it. I don't know what else we can compare that with, but I don't have time today to skim the usual specs for definitions.

I did make a few changes before signing off last night, like ditching the "when", adding a specific list of elements this applies to, and changing the wording to non-empty. If you have any thoughts on what we can do with text content, though, let me know.

mattgarrish: > An element has text content and that's it.

I wasn't thinking you could put anything but text in an element, but I suppose comments, processing instructions and even cdata blocks are all valid. Epubcheck isn't complaining when I add them.

Given that, the child text content definition is correct since there could be a concatenation in some extreme cases.

mattgarrish: I've made some more changes to add the child text content definition and infra whitespace handling to the core spec, and also removed the exception for meta properties defining their own whitespace handling from the RS. Let me know if you're happier with the sections now (I'm going to close these issues just because they reference out of date prose now).

rdeltour: > Let me know if you're happier with the sections now

Looks better to me now! Thanks for the changes (and for bearing with my nitpicking 😅).

rdeltour (approved):

PR #1549, opened 2021-03-01
Reformulated the beginning of §4.1

### LINKED ISSUE #1536: Something is missing in "EPUB Navigation Document"
https://www.w3.org/TR/2021/WD-epub-33-20210224/#sec-nav

The pieces we have:

> The EPUB Navigation Document is an XHTML Content Document

> restrictions on <nav> elements with epub:type

What I don't see is any description of the content of the `<body>`. Can it have something else than one or more `<nav epub:type='...'>`? Must it have at least one? Must it have a `<nav epub:type='toc'>`? exactly one? at most one? Can the `<nav epub:type='...'>` be inside other elements?

iherman:
I used mostly the text of @dauwhe (https://github.com/w3c/epub-specs/issues/1536#issuecomment-786868837) for the second paragraph, but reformulated it to follow the same style as the other paragraphs.

After all, I did not touch paragraph 1; I was not sure that this was necessary to answer to the issue.

Fixes #1536

cc @emuller-amazon

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1549.html" title="Last updated on Mar 1, 2021, 5:19 PM UTC (07f701e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1549/7960d10...07f701e.html" title="Last updated on Mar 1, 2021, 5:19 PM UTC (07f701e)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -4498,11 +4498,14 @@ Manifest:
 						Publication</a>. It allows <a>Authors</a> to include a human- and machine-readable global
 					navigation layer, thereby ensuring increased usability and accessibility for the user.</p>
 
-				<p>The EPUB Navigation Document is an <a>XHTML Content Document</a>, but with additional restrictions on
-					its structure to facilitate the machine-processing of its contents. [[HTML]] <a
-						href="https://html.spec.whatwg.org/multipage/sections.html#the-nav-element"><code>nav</code></a>
-					elements contain the specialized navigational information, which remains human-readable as well as
-					allowing Reading Systems to generate navigational interfaces.</p>
+				<p>The EPUB Navigation Document is an <a>XHTML Content Document</a>, which must include a single 
+					[[HTML]]&nbsp;<a href="https://html.spec.whatwg.org/multipage/sections.html#the-nav-element"><code>nav</code></a>
+					element with the <a href="#sec-epub-type-attribute"
+					><code>epub:type</code> attribute</a> set to the value 
+					<a href="#sec-nav-toc"><code>toc</code></a> for table of contents. It may also include other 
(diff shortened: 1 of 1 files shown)


mattgarrish:
This still sounds a bit heavy on technical details for an intro. Could we maybe further simplify it by linking across to details, so turn paras 2 and 3 into:

The EPUB Navigation Document is a special type of `<a>`XHTML Content Document`</a>` that defines the `<a href="https://html.spec.whatwg.org/multipage/sections.html#sec-nav-toc">`table of contents`</a>` for `<a>`Reading Systems`</a>`. It may also include other specialized navigation elements, such as a `<a href="#sec-nav-pagelist">`page list`</a>` and a list of key `<a href="#sec-nav-landmarks">`landmarks`</a>`. These navigation elements have `<a href="#sec-nav-def-model">`additional restrictions`</a>` on their content in order to facilitate their processing.

The EPUB Navigation Document is not exclusively for machine processing, however. There are no restrictions on the structure or content of the EPUB Navigation Document outside of the specialized navigation elements (i.e., Authors can mark the rest of the document up like any other XHTML Content Document). As a result, it can also be part of the linear reading order, avoiding the need for duplicate tables of contents. Navigation elements that are only destined for machine processing, such as the page list, can be hidden from visual rendering with the `<a href="#sec-nav-def-hidden">`hidden`</a>` attribute.

iherman:
I have made those changes (it is indeed much more readable), but I also added a small remark on the introductory section making it clear up front that, for each type of navigation aid, there can be at most one of those. This was the issue that started #1549 after all...

mattgarrish:
> I also added a small remark on the introductory section making it clear up front that, for each type of navigation aid, there can be at most one of those

I much prefer it there. It's not true for all navigation aids, which is why I dropped it from the broader intro. We only restrict the three we define.

mattgarrish (approved):

dauwhe (approved):
I think we addressed the concerns of the OP. Looks good.

PR #1554, opened 2021-03-03
Changed the Refs for EPUB 3 and A11y

### LINKED ISSUE #1552: EPUB 3 references in the A11y documents
Both the a11y and the a11y-techiques document refer to:

```
EPUB 3. W3C. URL: https://www.w3.org/publishing/epub3/
```

I believe this should be changed for the EPUB 3.3 references now...

iherman:
I have changed the references for EPUB 3 and EPUB Accessibility from the `/publishing/` to `/TR` equivalents.

Two comments:

- we may have to go through the full list at some point, maybe prune the unnecessary ones
- we may want to re-publish the Overview document to get the editors right...

This fixes #1552

### DIFF
--- a/epub33/common/js/biblio.js
+++ b/epub33/common/js/biblio.js
@@ -33,12 +33,8 @@ var biblio = {
 		"href": "http://www.daisy.org/guidelines/epub/navigable-audio-only-epub3-guidelines"
 	},
 	"EPUB-3": {
-		"title": "EPUB 3 Overview",
-		"href": "https://www.w3.org/TR/epub-overview/",
-		"authors" : [
-			"Matt Garrish",
-			"Ivan Herman"
-		],
+		"title": "EPUB 3",
+		"href": "https://www.w3.org/TR/epub/",
 		"publisher": "W3C"
 	},
 	"EPUB-31": {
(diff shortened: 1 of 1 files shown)


epub33/common/js/biblio.js
```diff
@@ -33,8 +33,12 @@ var biblio = {
 		"href": "http://www.daisy.org/guidelines/epub/navigable-audio-only-epub3-guidelines"
 	},
 	"EPUB-3": {
-		"title": "EPUB 3",
-		"href": "https://www.w3.org/publishing/epub3/",
+		"title": "EPUB 3 Overview",
```
mattgarrish: This should be a reference to the latest version of the EPUB core spec, not the overview (/TR/epub). It's used, for example, so the accessibility specification can always refer to the latest release if they're not always updated in sync. Any links the ref is associated with are also supposed to be version independent (but need to go back and check those).

I don't think any specification except the core has ever referenced the overview, so we don't need a "latest" link for it.

mattgarrish (changes requested):

epub33/common/js/biblio.js
iherman: I changed it. I also removed the list of editors (for now) if we refer to EPUB 3 in general and to its 3.3 incarnation... Maybe we can change that later.

mattgarrish (approved):
SGTM

PR #1558, opened 2021-03-05
Restructure the accessibility specification requirements

### LINKED ISSUE #1458: Media Overlays reading order
The MO spec currently says in [3.2.1 Structure](https://www.w3.org/publishing/epub32/epub-mediaoverlays.html#sec-media-overlays-structure).

> The ordering of [Media Overlays] elements MUST match the default reading order of the EPUB Content Document.

Can this conformance requirement be clarified? Should it be moved to the accessibility spec instead?

Someone from a DAISY member organization recently brought up to us an interesting use case: an old Chinese poem where the reading order differs from the visual order. In display, the text has three paragraphs A, B, C. The reading order is A, C, B.

The issue with the requirement above is that "reading order" isn't defined anywhere. 

As @mattgarrish said (in an internal discussion around this):
> Looking at it again, the wording is kind of ambiguous, as it doesn't say it must match the order of elements in the file like we had for the table of contents. I suppose you could read the requirement that if the default reading order is not sequential then the order of the media overlay also must be non-sequential
>(…)
> The accessibility requirement for elements to be ordered as you'd expect to read them is a content issue.

@marisademeglio then pointed out that this is somewhat linked to [WCAG SC 1.3.2](https://www.w3.org/TR/WCAG21/#meaningful-sequence) which says:

> When the sequence in which content is presented affects its meaning, a correct reading sequence can be programmatically determined

In my understanding, and to simply, the DOM order provides the reading order. For the use case above, a solution would be to have the DOM and MO respect the meaningful reading sequence, and to use CSS to visually re-order the paragraphs.

@marisademeglio says that footnotes are an interesting use case too. They can be at the bottom of a text document, but inline and skippable in the SMIL file.

I'm hoping we can clarify all this in 3.3.

### LINKED ISSUE #1556: Restructure epub accessibility requirements
Looking at the two requirements we have, they're actually a mix of different requirements under a common heading.

I'm thinking it would make sense to restructure these sections to better pull out the individual requirements, especially if we intend to expand on the requirements (whether we assign them as levels or keep as must/should/may).

For example, instead of one big "page navigation" section, we could restructure it as follows:

- Page Navigation
  - Overview - this would be the current explainer text in the "understanding" section
  - Applicability - this would be the prose explaining when we recommend including page navigation
  - Page Source - we'd define a summary, explainer and requirements for this and the next two criteria
  - Page List
  - Page Markers

The same could then be done for the media overlays section, where skippability, escapability and the nav doc requirements could be detailed separately.

Before I get into the actual restructuring of the document, though, does this make sense?

/cc @avneeshsingh @GeorgeKerscher @clapierre @gregoriopellegrino

mattgarrish:
This PR generally matches what I wrote in #1556 but I made a few additional changes:

- I merged the "relationship to wcag" sections into the general overviews.
- I grouped the requirements for each section under an "objectives" heading.
- I used definition lists within each objective as sectioning them seemed a bit excessive now that they're much smaller.

I also had to write "understanding" sections for all six objectives from scratch, so please take some time to review those.

And, finally, I integrated the page list and break requirements I had suggested in the other issues but only using normative keywords for now (no mention of levels). See if these make sense, too.

Otherwise, let me know if you find anything amiss or have any concerns with the new structure.

[Accessibility preview](https://raw.githack.com/w3c/epub-specs/restructure/a11y-spec/epub33/a11y/)
[Accessibility diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fw3c.github.io%2Fepub-specs%2Fepub33%2Fa11y%2F&doc2=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fraw.githack.com%2Fw3c%2Fepub-specs%2Frestructure%2Fa11y-spec%2Fepub33%2Fa11y%2Findex.html)

Fixes #1556 
Fixes #1458

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -513,7 +513,7 @@
 						<h5>Objectives</h5>
 
 						<section id="sec-page-src">
-							<h4>Pagination Source</h4>
+							<h6>Pagination Source</h6>
 
 							<dl>
 								<dt id="sec-page-src-obj">Objective</dt>
@@ -543,7 +543,7 @@
 						</section>
 
 						<section id="sec-page-list">
-							<h5>Page List</h5>
+							<h6>Page List</h6>
 
 							<dl>
 								<dt id="sec-page-list-obj">Objective</dt>
@@ -578,7 +578,7 @@
 						</section>
 
 						<section id="sec-page-breaks">
-							<h4>Page Breaks</h4>
+							<h6>Page Breaks</h6>
 
 							<dl>
 								<dt id="sec-page-breaks-obj">Objective</dt>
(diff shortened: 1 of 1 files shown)


mattgarrish:
I've also updated this PR to add the requirement for media overlays to follow the default reading order. For now, this:

- recommends that the order of par and seq elements match the spine and element order of the publication
- otherwise if there are deviations the result must still be a logical reading experience

Does this make sense?

PR #1453, opened 2020-12-24
Remove release identifier

mattgarrish:
This PR fixes #1440 as follows:

- Removes references to the release identifier from the core specification leaving only a section labelled "last modification date" with the same requirements. Adds a note for future generations about the release identifier to explain why these requirements exist as there's no other context for them.
- Removes all reference to creating a release identifier from the reading system specification.
- Removes explanation of release identifier from the overview.
- Removes the sections on the release identifier from the multiple renditions document and moves the requirements for the unique identifier and last mod date into the publication metadata section.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1453.html" title="Last updated on Dec 24, 2020, 2:39 PM UTC (218b7ff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1453/2ee308e...218b7ff.html" title="Last updated on Dec 24, 2020, 2:39 PM UTC (218b7ff)">Diff</a>

PR #1560, opened 2021-03-08
Change in RS conformance on unique identifier

### LINKED ISSUE #1310: Non-unique unique identifiers and reading system conformance
[The spec says](https://www.w3.org/publishing/epub32/epub-packages.html#sec-metadata-elem-identifiers-uid)

> Reading Systems MUST NOT depend on the Unique Identifier being unique to one and only one EPUB Publication. Determining whether two EPUB Publications with the same Unique Identifier represent different versions of the same publication (see Release Identifier), or different publications, might require inspecting other metadata, such as the titles or authors.

I'm trying to figure out how to write a test for this—create two different books with the same unique ID and make sure that two books show up in the RS UI?  Should this be normative text? In some sense this reads more like non-normative advice to reading systems. In general we say nothing about how reading systems present different publications to users, or how reading systems respond to updated versions of the same content.

iherman:
Per [WG resolution](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2020-12-18-epub#resolution2) the statement on unique identifiers should not be normative.

I also propose an editorial change: after PR https://github.com/w3c/epub-specs/pull/1453 only the `unique-identifier` property is relevant as an identifier, so I created a new section on the package conformance to handle only that case and removed the old one that was a leftover of a time when we also had release identifiers.

Fixes #1310

See preview:

    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1310/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1310/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -330,6 +330,15 @@
 							data-cite="bidi#P2">Rule P2</a> of [[!BIDI]].</p>
 				</section>
 
+				<section id="sec-pub-identifiers">
+					<h3>Unique Identifier</h3>
+
+					<p>Reading Systems SHOULD NOT depend on the <code>unique-identifier</code> attribute being unique to one and only one
+						EPUB Publication. Determining whether two EPUB Publications with the same Unique Identifier
+						represent different versions of the same publication, or different publications, might
+						require inspecting other metadata, such as the titles or authors.</p>
+				</section>
+	
 				<section id="metadata">
 					<h4>Metadata</h4>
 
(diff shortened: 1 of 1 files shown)


mattgarrish:
I'm still unsure if this addresses the resolution. The meeting notes say the requirement should be turned into a note and made shorter, but the only change is "must not" to "should not" (fine with the restructuring, though). I'll have to take a pass on approving.

dauwhe (approved):

iherman:
@mattgarrish the merge conflict is around the change section. I never know how to use the Web editor to solve these and I am out of time to handle this properly... can you take care of this and merge this? If not, I can try to do something tomorrow.

PR #1563, opened 2021-03-08
Remove requirement for overlay order to match reading order

### LINKED ISSUE #1458: Media Overlays reading order

mattgarrish:
This PR only addresses the media overlays part of the issue. I'm going to add this as a requirement to the accessibility specification, but I want to make it fit with the pending re-org in #1558 so am going to update that PR with the requirement.

Fixes #1458

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1563.html" title="Last updated on Mar 8, 2021, 7:06 PM UTC (bf09ce2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1563/6f49938...bf09ce2.html" title="Last updated on Mar 8, 2021, 7:06 PM UTC (bf09ce2)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7064,9 +7064,7 @@ store destination as source in ocf
 							of two elements: the <a href="#elemdef-smil-par"><code>par</code> element</a> and the <a
 								href="#elemdef-smil-seq"><code>seq</code> element</a>. The ordering of these elements
 							represents how the content in the corresponding EPUB Content Documents is rendered during
-							playback. In general, the order of these elements will match the order of the elements in
-							each EPUB Content Document (i.e., the logical reading order), but other rendering orders are
-							possible.</p>
+							playback.</p>
 
 						<p>The <code>par</code> element represents phrases. Each element identifies a text and audio
 							component to synchronize during playback.</p>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7063,7 +7063,10 @@ store destination as source in ocf
 						<p>The <a href="#elemdef-smil-body"><code>body</code></a> of a Media Overlay Document consists
 							of two elements: the <a href="#elemdef-smil-par"><code>par</code> element</a> and the <a
 								href="#elemdef-smil-seq"><code>seq</code> element</a>. The ordering of these elements
-							MUST match the default reading order of the EPUB Content Document.</p>
+							represents how the content in the corresponding EPUB Content Documents is rendered during
+							playback. In general, the order of these elements will match the order of the elements in
+							each EPUB Content Document (i.e., the logical reading order), but other rendering orders are
+							possible.</p>
 
 						<p>The <code>par</code> element represents phrases. Each element identifies a text and audio
 							component to synchronize during playback.</p>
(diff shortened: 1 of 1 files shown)


epub33/core/index.html
```diff
@@ -7063,7 +7063,10 @@ <h5>Overlay Structure</h5>
 						<p>The <a href="#elemdef-smil-body"><code>body</code></a> of a Media Overlay Document consists
 							of two elements: the <a href="#elemdef-smil-par"><code>par</code> element</a> and the <a
 								href="#elemdef-smil-seq"><code>seq</code> element</a>. The ordering of these elements
-							MUST match the default reading order of the EPUB Content Document.</p>
+							represents how the content in the corresponding EPUB Content Documents is rendered during
+							playback. In general, the order of these elements will match the order of the elements in
+							each EPUB Content Document (i.e., the logical reading order), but other rendering orders are
```
rdeltour: If the a11y spec requires that the MO order matches the Content Document reading order, maybe we shouldn't explicitly mention the other possibility here? I don't know.

rdeltour (approved):

epub33/core/index.html
mattgarrish: I'm actually concerned that this may not be something we should be trying to require. Media overlays are a rendering of the content, and I'm not sure if matching the default element order is always the best thing.

For example, what if you had a table that made more sense to read column by column instead of row by row?

Or, in the absence of skippability, what if footnotes were interspersed in the main flow of a book but in media overlay playback they were come back to after reading the primary narrative of a section first?

I know what we're after here (that there be a logical progression), but I'm not sure there's an obvious way of pinning that down.

mattgarrish: Or, in other words, I think this wording is fine but I'm struggling with what exactly to require in the accessibility specification.

rdeltour: Yeah, I agree that we should word carefully whatever we require in the a11y spec. It's not as simple as "DOM order", and the examples you mention are good case in point.

Here I'm just wondering if that entire sentence could not be left out entirely, or simplified to something like "The order of the elements typically match the logical reading order of the Content Document", which is open enough and doesn't add explicit restrictions 🤔

mattgarrish: > It's not as simple as "DOM order", and the examples you mention are good case in point.

What I'm thinking is that we make that a "should" (basically the same requirement prose in the page list) but if there is deviation it must still provide a logical reading experience. Only a human will be able to check the latter, but that's not uncommon for accessibility.

> I'm just wondering if that entire sentence could not be left out entirely

Ya, I'm going to end up covering this in the accessibility spec where it makes more sense.

marisademeglio (approved):

PR #1577, opened 2021-03-16
Change Author to EPUB Creator

### LINKED ISSUE #1575: Author definition question
The various specifications often refer to the concept of "Author", the definition of which can be found at the link https://www.w3.org/TR/epub/#dfn-author

- Perhaps it might make sense to make it explicit that a publisher is an author (as understood in the spec)?
- Maybe adding "For example, the publisher" to the end of the definition.

mattgarrish:
Changes all instance of "Author" to "EPUB Creator". Can always be switched to another term if people prefer something else (it's not as clunky sounding as I thought it might be, at least). Take a look at the definition, too.

Otherwise, only a couple of other minor content moves as I found some author guidance hiding in the reading system prose:

- added some of the relevant video note text to the core media types section so you don't have to link across
- moved some note text for authoring into the relevant scripting sections
- added the paragraph on authoring in scripting event model to the core spec

The only bit of authoring guidance left in the reading system specification, at least that this change surfaced, was in the readingsystemobject appendix, but leaving it where it is for now.

Fixes #1575 

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1577.html" title="Last updated on Mar 16, 2021, 3:18 PM UTC (99b7ef6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1577/2577dab...99b7ef6.html" title="Last updated on Mar 16, 2021, 3:18 PM UTC (99b7ef6)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -117,7 +117,7 @@
 					functionality normatively referenced as part of the standard. New functionality is also added
 					periodically through the development of extension specifications. Features and functionality defined
 					outside of core revisions to the standard, while not formally recognized in this specification, are
-					nonetheless available for use by <a>Authors</a> and Reading System developers.</p>
+					nonetheless available for use by <a>EPUB Creators</a> and Reading System developers.</p>
 
 				<p>The informative <a href="https://www.w3.org/TR/epub-overview-33/">EPUB 3 Overview</a>
 					[[EPUB-OVERVIEW-33]] provides a general introduction to EPUB 3. A list of technical changes from the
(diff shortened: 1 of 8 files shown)


epub33/core/index.html
```diff
@@ -315,6 +308,19 @@ <h3>Terminology</h3>
...
+					</dt>
+					<dd>
+						<p>The person(s) or organization responsible for the creation of an <a>EPUB Publication</a>.</p>
+						<p>Depending on how EPUB Publications are produced, EPUB Creator may sometimes refer to
+							responsibilities of the organization (e.g., the publisher) or the individuals preparing the
+							publication (e.g., technical editors).</p>
+						<div class="note">
+							<p>Previous versions of this specification referred to the EPUB Creator as the <dfn
+									id="dfn-author">Author</dfn>.</p>
+						</div>
```
iherman, line 322: Wouldn't `<dfn id="dfn-author">Author</dfn>` lead to a link error (or simply a respec error)? That term is not defined any more...

iherman (approved):

epub33/core/index.html
iherman: (I have not checked the other documents, the same problem may be presen in those, too.

iherman: Oops, sorry. It is ok and valid as is, so there is no error (I do not know whether we would ever need this definition, but it does not do any harm)

mattgarrish: > I do not know whether we would ever need this definition, but it does not do any harm

Right, I was worried there might be some old satellite specs using "latest" epub links for their definitions, so having a destination wouldn't hurt if they still redirect their way forward to this version. May well prove useless but harmless.

wareid (approved):

PR #1573, opened 2021-03-15
Removed normative dependencies on schemas

iherman:
Fixes #1566.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1573.html" title="Last updated on Mar 17, 2021, 2:00 PM UTC (d55c203)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1573/332c414...d55c203.html" title="Last updated on Mar 17, 2021, 2:00 PM UTC (d55c203)">Diff</a>

PR #1578, opened 2021-03-16
Add missing element definitions

mattgarrish:
This merges on top of #1573 to add the missing element definitions for the container.xml, encryption.xml and signature.xml files.

This ensures there's no reliance on the schemas to normatively understand what is necessary.

The only tweak I made to the schemas section was to remove the text about removing foreign namespace elements and attributes from the container.xml file. It's back in the body plus the NVDL schema should already take care of this before dispatching to the RNC.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1578.html" title="Last updated on Mar 16, 2021, 5:12 PM UTC (6ea0595)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1578/9729102...6ea0595.html" title="Last updated on Mar 16, 2021, 5:12 PM UTC (6ea0595)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -6341,8 +6341,8 @@ Spine:
 									applied to an EPUB Publication as a whole or to its parts and can specify the
 									signing of any kind of data (i.e., not just XML).</p>
 
-								<p class="note"> The content of the <code>signatures.xml</code> file is also defined,
-									informally, through an XML schema; see <a href="#app-schema-signatures"></a> for
+								<p class="note"> The content of the <code>signatures.xml</code> file is also defined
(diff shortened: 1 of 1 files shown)


iherman (approved):
Shame on me that I did not realize the missing boxes for the XML elements in the normative chains... :-(

mattgarrish:
> I did not realize the missing boxes for the XML elements in the normative chains

OCF didn't get the same level of rewrite during the transition from 2 to 3 that we put into publications and content documents. That's why the reliance on the schemas to explain the container file, in particular, has always bugged me.

I'm going to merge this into the other PR as I guess we don't really need a double review.

PR #1579, opened 2021-03-17
Add non characters at end of supplementary planes to file naming restrictions

### LINKED ISSUE #1538: Non characters in file names
https://www.w3.org/TR/2021/WD-epub-33-20210224/#sec-container-filenames

If you exclude "Non characters in Arabic Presentation Forms-A (U+FD00 ... U+FDEF)", then you ought exclude all non characters. <http://www.unicode.org/faq/private_use.html#nonchar4>

mattgarrish:
Kind of an awkward fit since it overlaps with the private use area restrictions. Rather than duplicate the code points, I've left off those last two planes and ended at U+EFFFE and U+EFFFF, but open to living with a bit of duplication.

Fixes #1538

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1579.html" title="Last updated on Mar 19, 2021, 2:18 PM UTC (ed8cf8f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1579/42671d8...ed8cf8f.html" title="Last updated on Mar 19, 2021, 2:18 PM UTC (ed8cf8f)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -965,9 +965,8 @@
 						</li>
 					</ul>
 
-					<p>Authors are encouraged to locate these resources inside the EPUB Container
-						whenever feasible to allow users access to the entire presentation regardless of
-						connectivity status.</p>
+					<p>Authors are encouraged to locate these resources inside the EPUB Container whenever feasible to
+						allow users access to the entire presentation regardless of connectivity status.</p>
 
 					<aside class="example">
 						<p>The following example shows a reference to an audio file in an <a>XHTML Content Document</a>
@@ -5692,6 +5691,10 @@ Spine:
 									<p>Specials (<code>U+FFF0 … U+FFFF</code>)</p>
 								</li>
 								<li>
+									<p>Non characters at the end of the Supplementary Planes (<code>U+1FFFE, U+1FFFFF …
+											U+EFFFE, U+EFFFF</code>)</p>
+								</li>
+								<li>
 									<p>Tags and Variation Selectors Supplement (<code>U+E0000 … U+E0FFF</code>)</p>
 								</li>
 								<li>
@@ -9021,6 +9024,9 @@ EPUB/images/cover.png</pre>
 				-->
 
 				<ul>
+					<li>17-Mar-2021: Include non characters at the end of the supplementary planes in list of characters
+						not allowed in file names. See <a href="https://github.com/w3c/epub-specs/issues/1538">issue
+							1538</a>.</li>
 					<li>10-Mar-2021: Require that resources referenced from an EPUB Publication not be located in the
 							<code>META-INF</code> directory. See <a href="https://github.com/w3c/epub-specs/issues/1205"
 							>issue 1205</a>.</li>


iherman:
I must admit that I have never looked into the details of all this, but I am worried about such lists, because they always incur the danger of becoming incomplete (which was the case, it seems, for 3.2). I wonder whether there is some other organization or institution that we could refer to instead...

mattgarrish:
> because they always incur the danger of becoming incomplete (which was the case, it seems, for 3.2)

Sure, but is there a greater harm in being incomplete than leaving it to authors to figure out for themselves?

But what's the reality that people are out there trying to make EPUBs with non characters or private use characters in the first place? And if they are, so what? We've gone beyond what characters operating systems actually disallow in filenames while also ignoring actual file names that are not allowed (e.g., Windows disallows various names, like AUX and NUL).

The list in OCF 2 was much smaller and more focused on what are actually invalid characters.

> I wonder whether there is some other organization or institution that we could refer to instead...

That would be optimal, but I personally don't know of anything we could normatively refer to. All I've found is informative stuff, like [wikipedia](https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations).

iherman:
> But what's the reality that people are out there trying to make EPUBs with non characters or private use characters in the first place? And if they are, so what? We've gone beyond what characters operating systems actually disallow in filenames while also ignoring actual file names that are not allowed (e.g., Windows disallows various names, like AUX and NUL).
> 
> The list in OCF 2 was much smaller and more focused on what are actually invalid characters.
> 

So what is the history of this list? How and why did it land into this spec? Indeed, is this really necessary?

The LPF spec is much more cavalier in this (see https://github.com/w3c/epub-specs/issues/1538#issuecomment-789773707) but, of course, that is not a Rec. But I nevertheless wonder whether the LPF model is not enough for the EPUB case, too.

mattgarrish:
> So what is the history of this list? How and why did it land into this spec? Indeed, is this really necessary?

I don't know the history or why it was expanded to the character sets it now includes. Maybe @GarthConboy or @bduga or someone with a longer memory of epub can speak to that.

It's the sort of warning I can imagine publishers like having, though, as it avoids them having to check this manually. I can also understand it from the user perspective of wanting to ensure that if you purchase an epub it will work on your operating system.

That's why I'd be wary of removing the list entirely. It's a little too easy a solution.

According the [Unicode FAQ](http://www.unicode.org/faq/private_use.html#nonchar6), non characters are not intended for interchange, so adding this group at least makes sense from that perspective.

I'm not sure why we have some of the others, though, like the private use areas. From a public interchange perspective it makes sense, but is that being overly restrictive?

bduga:
I am not sure what the history is exactly, other than a general abuse of file names. (I have a pretty good story about customers using 0x00 to sort filenames on the Mac) Being overly restrictive here doesn't seem to be a huge problem, since none of this changes the actual content, and having some general guidance is helpful.

dauwhe (approved):

iherman:
Ok, I yield:-) and approve...

iherman (approved):

iherman:
@mattgarrish I took care of the merge conflicts; I let you merge at your time:-)

PR #1581, opened 2021-03-19
Clean up of open issues for multiple renditions

### LINKED ISSUE #584: old linking pattern referenced in MR spec
Our original idea for adding link elements to the container was that they would follow the rootfiles element, but during the 3.0.1 revision the AHL proposal was changed to add a links element wrapper.

Romain has noted that the prose in section 4.4.4 wasn't updated to reflect this new reality:

The location of the Rendition Mapping Document is identified in the Container Document using an [OCF301] link element child of the root container element, where:[1]

We should either remove mention of it being a child of the container or switch "root container element" to "links element".

I'll add this to the errata for the document. We can perhaps address this oversight when we look at consolidation during 3.1.

[1] http://www.idpf.org/epub/renditions/multiple/#sec-4.4.4

/cc @mgylling @murata0204 @rdeltour

### LINKED ISSUE #619: clarify package document location within container
A problem that came up during the ESC work is that if you locate the package document in a sibling directory from the components, some major reading systems won't recognize them (the package location becomes like a domain root and resources outside it aren't available).

For example, with this directory structure:

/EPUB/package.opf
/components/whatever

The resources under the components directory aren't accessible.

This is contradictory behaviour to the specifications, as there is no requirement that the package document be at the root and not be able to access resources in a sibling directory, but given the reality is it worth making a similar recommendation to storing all resources under a dedicated directory that all package documents be located at the root of that structure? (when there is more than one opf, at least)

If resources are shared between renditions, then a structure like this is going to become problematic:

/EPUB/fxl/package.opf
/EPUB/reflow/package.opf
/EPUB/video
/EPUB/audio

As you won't be able to reach across directories. Similarly, if the resources go into one of the fxl/reflow directories, the one they don't go into still won't have access.

The multiple-renditions specification doesn't say anything about this issue.

### LINKED ISSUE #1443: Multiple renditions and obfuscation
The document doesn't explain how obfuscation is supposed to work when there are multiple renditions.

Since unique identifiers are allowed per-Rendition, and there's also a unique identifier required in the metadata.xml file, which of these are supposed to be used? (Is obfuscation done per-Rendition?)

If the unique identifier is the one in the metadata.xml file, then that seemingly requires that the identifier be the same as the one in the first rendition for reading systems that don't process multiple renditions.

mattgarrish:
This PR fixes the remaining open issues for the multiple renditions spec:

- reference to link being a child of the container element is removed and replaced by a link to the element's definition
- noted the inability to share resources when package documents aren't stored together
- require unique identifier in metadata.xml for obfuscation key

Also adds notes in the text pointing to the schemas and fixes normative references that had been translated to informative.

For the other specifications, I noticed that the multiple renditions specification was referencing DCTERMS not DC11. Following their links, DCTERMS is now the proper reference so I updated all instances.

- [MR preview](https://raw.githack.com/w3c/epub-specs/multi-rend/fixes/epub33/multi-rend/)
- [MR diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fw3c.github.io%2Fepub-specs%2Fepub33%2Fmulti-rend%2F&doc2=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fraw.githack.com%2Fw3c%2Fepub-specs%2Fmulti-rend%2Ffixes%2Fepub33%2Fmulti-rend%2Findex.html)

Fixes #584 
Fixes #619 
Fixes #1443 

/cc @murata2makoto

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1581.html" title="Last updated on Apr 16, 2021, 5:25 PM UTC (1f67161)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1581/4bc32a3...1f67161.html" title="Last updated on Apr 16, 2021, 5:25 PM UTC (1f67161)">Diff</a>


iherman:
Just trying to be nasty.... I create an EPUB instance

- It has only one rendition, whose container metadata contains a unique id
- I also add a `metadata.xml` file, which also contains a unique id (I know, it is stupid, why would people do that for a non-multiple rendition file). This unique id is different from the previous one (the multiple rendition spec only say that this _SHOULD_ be avoided)

Question: which id do I use for obfuscation?

1. per the EPUB Core spec, `metadata.xml` is ignored, so it is the one in `container.xml`
2. per the multiple rendition spec (o.k., it is only a note) the one in the `metadata.xml` file.

A solution may be to say that `metadata.xml` becomes relevant only if there are at least two renditions? Or should we move (part of) the `metadata.xml` specification into the core document, meaning that even in the case of single rendition the identifier of `metadata.xml` wins?

mattgarrish:
> This unique id is different from the previous one (the multiple rendition spec only say that this _SHOULD_ be avoided)

This is one of the failings of multiple renditions and the reason why I put it as a "should". The `metadata.xml` file is supposed to allow you to provide an identifier for the overall publication that is different from any given rendition, but we know reading systems don't use this file.

To keep some kind of consistency we kind of have to give authors the choice between whether they want separate identifiers or whether they care if their publication works everywhere. That's why obfuscation isn't particularly workable as the specification was envisioned.

The other alternative we can try to avoid any breakage is that you always have to use the unique identifier of the default rendition. It forces reading systems to parse the first package document whether it's the one they intend to display or not, but maybe it's slightly less nonsensical overall? You can still have separate identifiers that way.

iherman:
> The other alternative we can try to avoid any breakage is that you always have to use the unique identifier of the default rendition. It forces reading systems to parse the first package document whether it's the one they intend to display or not, but maybe it's slightly less nonsensical overall? You can still have separate identifiers that way.

I have personally no problem with that (although the usage of metadata.xml seems the cleaner choice but only if this is also reflected in the overall document). But I have no real knowledge about RS behaviour around on this, so others should chime in whether this is the right solution indeed.

mattgarrish:
> so others should chime in whether this is the right solution indeed.

As far as I'm aware, we're only working in theories with this so even if we get it wrong we're not breaking any content in the real world.

I expect we just have to pick one approach and see what happens. Maybe we have to correct course in the future, but this approach at least is consistent with how reading systems work right now for single renditions.

iherman (approved):

iherman:
I've approved it but I think this is something we should run through a WG call at least...

@wareid @dauwhe @shiestyle

mattgarrish:
> I think this is something we should run through a WG call at least

That's fine. I just wanted to knock these issues off so they're not lingering around.

The more I think about it, the more I think this option has to be the one we use if we want to remain backwards compatible.

We had to accept when we made multiple renditions that we couldn't enforce all publications follow the metadata requirements. The file had gone unstandardized for too long and we couldn't know for sure how it might be getting used privately. So it's always possible that there is obfuscated content (in a single rendition publication) that has a metadata.xml in the container that has nothing to do with the multiple renditions specification. If we introduce a requirement for all publications to prefer the metadata.xml file when present, it could break existing content.

We also can't say to use the unique identifier for each publication because that would break the ability to share a resource between two renditions with different unique identifiers.

It's too bad the obfuscation key isn't stored in the encryption file (not sure if it's a technical limitation or was just seen as too easy to access), but too late for anything like that, either.

wareid (approved):

shiestyle (approved):

dauwhe (approved):

iherman:
@mattgarrish I believe this can be merged...

PR #1582, opened 2021-03-22
Add new section clarifying the requirements for data urls

### LINKED ISSUE #1564: Are data URLs always disallowed?
The question of whether a data URL is allowed in an `a` tag came up in the [epubcheck tracker](https://github.com/w3c/epubcheck/issues/1213).

My understanding is this is disallowed because you can't navigate to non-spine resources.

I've also understood that embedding resources using data URLs is forbidden since they're also not listed in the manifest.

Assuming both these impressions are correct, we should probably explicitly note somewhere for extra clarity that the data: scheme cannot be used.

mattgarrish:
The issue appears a little more nuanced than we've discussed so far - top-level navigation is being restricted, but you can still link to data URLs within an iframe, for example. I've tried to craft the new section to also make this distinction.

It's also possible to use data URLs in the package document along with the various content formats, so I put the new section under publication resources. It's either forbidden or pointless to use them in the manifest, but you could, for example, embed a resource in a `link` element.

I've also noted they are subject to the foreign resource restrictions, so that limits what you can use and where.

The one complexity is how reading systems restrict top-level navigation to data URLs. I don't know that this is something we can, or should, try to be too specific about. While it's possible you might disable hyperlinks with a data scheme, that wouldn't stop scripts from opening such content. I was reading some of the bug trackers for this and it sounds like browsers may only prevent the resource from loading in the spawned window, not restrict the event that triggered it.

As a result, I've only given a very general requirement to prevent data URLs from loading, but open to alternative wording if someone knows a better way to phrase this.

Fixes #1564 

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1564/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1564/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1582.html" title="Last updated on Mar 26, 2021, 3:25 PM UTC (3913551)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1582/5e73f5b...3913551.html" title="Last updated on Mar 26, 2021, 3:25 PM UTC (3913551)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -440,16 +440,25 @@
 							not render as intended by the <a>EPUB Creator</a>. Examples of Publication Resources include
 							the <a>Package Document</a>, <a>EPUB Content Document</a>, CSS Style Sheets, audio, video,
 							images, embedded fonts, and scripts.</p>
-						<p>Publication Resources are listed in the Package Document <a href="#sec-manifest-elem"
-								>manifest</a> and bundled in the <a>EPUB Container</a> file unless specified otherwise
-							in <a href="#sec-resource-locations"></a>.</p>
+						<p>Publication Resources are typically listed in the Package Document <a
+								href="#sec-manifest-elem">manifest</a> and bundled in the <a>EPUB Container</a>, with
+							two exceptions:</p>
+						<ul>
+							<li>
+								<p>resources encoded as <a href="#sec-data-urls">data URLs</a> are not required to be
+									listed in the manifest; and</p>
+							</li>
+							<li>
+								<p>resources listed in <a href="#sec-resource-locations"></a> may be located outside the
+									EPUB Container.</p>
+							</li>
+						</ul>
 						<p>Examples of resources that are not Publication Resources include those identified by the
 							Package Document <a href="#sec-link-elem"><code>link</code> element</a> and those identified
 							in outbound hyperlinks that resolve to <a>Remote Resources</a> (e.g., referenced from the
 								<code>href</code> attribute of an [[!HTML]] <a
 								href="https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element"
-									><code>a</code></a> element). Resources represented by <a href="#sec-data-urls">data
-								URLs</a> are also not Publication Resources.</p>
+									><code>a</code></a> element).</p>
 					</dd>
 					<dt>
 						<dfn id="dfn-remote-resource" data-lt="Remote Resources">Remote Resource</dfn>
@@ -1010,14 +1019,38 @@
 						allows a resource to be embedded within another, avoiding the need for an external file.</p>
 
 					<p><a>EPUB Creators</a> MAY use data URLs in EPUB Publications provided their use does not result in
-						a <a>Top-Level Content Document</a> or open a new browser instance (e.g., they cannot be used in
-						manifest <code>item</code> elements that are referenced from the <a>spine</a>, in [[HTML]] or
-						[[SVG]] <code>a</code> elements (except when inside an <code>iframe</code>), or in calls to
-						[[ECMASCRIPT]] <code>window.open</code> or <code>document.open</code>).</p>
+						a <a href="https://html.spec.whatwg.org/multipage/browsers.html#top-level-browsing-context"
+							>top-level browsing context</a> [[!HTML]]. This restriction applies to data URLs used in the
+						following scenarios:</p>
 
-					<p>This restriction on their use is to prevent security issues (i.e., phishing scams) and also to
-						ensure that <a>Reading Systems</a> can determine where to take a user next (i.e., because such
-						resources will typically not be listed in the spine).</p>
+					<ul>
+						<li>
+							<p>in manifest <a href="#sec-item-elem"><code>item</code> elements</a> referenced from the
+									<a>spine</a>;</p>
+						</li>
+						<li>
+							<p>in the <code>href</code> attribute on [[!HTML]] or [[!SVG]] <code>a</code> elements
+								(except when inside an <a
+									href="https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element"
+										><code>iframe</code> element</a> [[!HTML]]);</p>
+						</li>
+						<li>
+							<p>in the <code>href</code> attribute on [[!HTML]] <code>area</code> elements (except when
+								inside an <code>iframe</code> element);</p>
+						</li>
+						<li>
+							<p>in calls to [[!ECMASCRIPT]] <code>window.open</code> or <code>document.open</code>.</p>
+						</li>
+					</ul>
+
+					<div class="note">
+						<p>The list of prohibited uses for data URLs is subject to change as the respective standards
+							that allow their use evolve.</p>
+					</div>
+
+					<p>This restriction on their use is to prevent security issues and also to ensure that <a>Reading
+							Systems</a> can determine where to take a user next (i.e., because these resources are not
+						be listed in the spine).</p>
 
 					<p>Resources represented as data URLs are not Publication Resources so are exempt from the
 						requirement to be listed in the <a>manifest</a>.</p>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1019,7 +1019,8 @@
 						allows a resource to be embedded within another, avoiding the need for an external file.</p>
 
 					<p><a>EPUB Creators</a> MAY use data URLs in EPUB Publications provided their use does not result in
-						a <a href="https://html.spec.whatwg.org/multipage/browsers.html#top-level-browsing-context"
+						a <a>Top-level Content Document</a> or <a
+							href="https://html.spec.whatwg.org/multipage/browsers.html#top-level-browsing-context"
 							>top-level browsing context</a> [[!HTML]]. This restriction applies to data URLs used in the
 						following scenarios:</p>
 
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -2893,8 +2893,8 @@ Spine:
 									href="#confreq-cd-scripted-flbk">fallbacks for scripted content</a>.</p>
 
 							<div class="note">
-								<p>As it is not possible to use manifest fallbacks for resources represented using <a
-										href="#sec-data-urls">data URLs</a>. Foreign resources can only be represented
+								<p>As it is not possible to use manifest fallbacks for resources represented in <a
+										href="#sec-data-urls">data URLs</a>, Foreign Resources can only be represented
 									as data URLs where an intrinsic fallback mechanism is available.</p>
 							</div>
 						</section>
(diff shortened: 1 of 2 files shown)


epub33/core/index.html
```diff
@@ -2867,6 +2891,12 @@ <h6>Manifest Fallbacks</h6>
 							<p>Fallbacks MAY also be provided for <a>Top-Level Content Documents</a> that are EPUB
 								Content Documents. An example of when this feature can be utilized is when providing <a
 									href="#confreq-cd-scripted-flbk">fallbacks for scripted content</a>.</p>
+
+							<div class="note">
+								<p>As it is not possible to use manifest fallbacks for resources represented using <a
+										href="#sec-data-urls">data URLs</a>. Foreign resources can only be represented
+									as data URLs where an intrinsic fallback mechanism is available.</p>
+							</div>
```
iherman, line 2933: I presume:

> using `<a href="#sec-data-urls">`data URLs`</a>`. Foreign 

should be

> in a `<a href="#sec-data-urls">`data URLs`</a>`, foreign

iherman (approved):
I cannot judge whether the text you have put for RS-s is enough or something more specific should be said; deferring to the RS experts on this... But, from where I stand, really +1 to what you propose!

epub33/core/index.html
mattgarrish: Oops, yes!

epub33/core/index.html
```diff
@@ -448,7 +448,8 @@ <h3>Terminology</h3>
 							in outbound hyperlinks that resolve to <a>Remote Resources</a> (e.g., referenced from the
 								<code>href</code> attribute of an [[!HTML]] <a
 								href="https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element"
-									><code>a</code></a> element).</p>
+									><code>a</code></a> element). Resources represented by <a href="#sec-data-urls">data
+								URLs</a> are also not Publication Resources.</p>
```
bduga: This seems to allow data URLs as a mechanism to bypass the fallback chain when specifying non-core media types. So an `img` element referencing a file that is not supported would need to provide a fallback, but a data URL would not. And if the aren't Publication Resources, what are they? They aren't foreign resources, since those are defined to be publication resources. I am a little worried about introducing a new type of entity. Is it possible to leave them as publication resources, but remove the requirement to be listed in the manifest in 2.2.2?

epub33/core/index.html
```diff
@@ -1001,6 +1002,30 @@ <h4>Resource Locations</h4>
...
+
+					<p>The <a href="https://tools.ietf.org/html/rfc2397"><code>data:</code> URL scheme</a> [[!RFC2397]]
+						is used to encode resources directly into a URL string. The advantage of this scheme is that it
+						allows a resource to be embedded within another, avoiding the need for an external file.</p>
+
+					<p><a>EPUB Creators</a> MAY use data URLs in EPUB Publications provided their use does not result in
+						a <a>Top-Level Content Document</a> or open a new browser instance (e.g., they cannot be used in
+						manifest <code>item</code> elements that are referenced from the <a>spine</a>, in [[HTML]] or
+						[[SVG]] <code>a</code> elements (except when inside an <code>iframe</code>), or in calls to
+						[[ECMASCRIPT]] <code>window.open</code> or <code>document.open</code>).</p>
```
bduga: This list is exhaustive enough that it may need clarification that it is not exhaustive (in addition the "e.g." at the start). Actually might be better to introduce a new sentence here, since you have a parenthetical in the parenthetical.

epub33/core/index.html
```diff
@@ -1001,6 +1002,30 @@ <h4>Resource Locations</h4>
...
+
+					<p>This restriction on their use is to prevent security issues (i.e., phishing scams) and also to
```
bduga: I don't know if we need to list any specific security issues here, since the terms can be vague.

epub33/rs/index.html
```diff
@@ -172,6 +172,13 @@ <h2>EPUB Publications</h2>
 									href="https://www.w3.org/TR/epub-33/#sec-resource-locations">Resource Locations</a>
 								[[!EPUB-33]].</p>
 						</li>
+						<li>
+							<p id="confreq-rs-data-urls">It MUST prevent attemps at top-level navigation to data URLs
```
bduga: Is there a more precise term for this? Or is `top-level navigation` defined somewhere we can reference?

bduga: Is this overly restrictive? If a book has an `img` element with a data url, and a reading system has a "open in new window" option for images, would that be disallowed for the data url case?

epub33/core/index.html
mattgarrish: Ew, ya, that's not what I was intending at all. It was only about an exclusion from the manifest. I'll fix.

epub33/core/index.html
mattgarrish: It's nearly exhaustive in that I can only see that the area element is missing from it. I'll add that case in and add a note that it's still subject to change as html, svg, etc. evolve.

epub33/rs/index.html
mattgarrish: I saw something similar to this in one of the bug trackers cited in the issue. They were going to exempt it from the open option in context menus, so no harm in doing the same here.

mattgarrish: Can we consider any top-level content document a [top-level browsing context](https://html.spec.whatwg.org/multipage/browsers.html#top-level-browsing-context)?

If so, we don't need to separate the concepts and can use only the HTML term.

I wasn't sure if they're considered equal all the time because SVG doesn't get into this level of detail. But if we consider all EPUB rendering in terms of html user agent rendering then it works.

mattgarrish: I've made a change to this effect with the latest commit. Worst case we keep working at massaging this text, I guess... 😄

epub33/rs/index.html
```diff
@@ -173,11 +173,10 @@ <h2>EPUB Publications</h2>
...
 						</li>
 						<li>
-							<p id="confreq-rs-data-urls">It MUST prevent attemps at top-level navigation to data URLs
-								[[!RFC2397]]. (Examples of top-level navigation include loading resources not in the
-								spine as though they are <a>Top-Level Content Documents</a> and opening <a
+							<p id="confreq-rs-data-urls">It MUST prevent data URLs [[!RFC2397]] from opening in <a
 									href="https://html.spec.whatwg.org/multipage/browsers.html#top-level-browsing-context"
-									>top-level browsing contexts</a> [[HTML]])</p>
+									>top-level browsing contexts</a> [[HTML]], except when initiated through the user
+									interface of the Reading System (e.g., a context menu).</p>
```
bduga: Yeah, this language is tricksy :( Isn't clicking on something to make something happen part of the user interface of the RS? Not sure how to best phrase this.

bduga (commented):
Changes look good!

epub33/rs/index.html
dauwhe: > Can we consider any top-level content document a [top-level browsing context](https://html.spec.whatwg.org/multipage/browsers.html#top-level-browsing-context)?

I worry about that, because in implementations the spine item might not be a top-level browsing context.

epub33/rs/index.html
mattgarrish: > Isn't clicking on something to make something happen part of the user interface

Ya, it took a few tries to even come up with this. I wouldn't consider clicking in the content the same as the clicking a part of the user interface, but it is awkward.

Would "affordance" improve it any: "except when initiated through a Reading System affordance such as a context menu"?

epub33/rs/index.html
mattgarrish: > because in implementations the spine item might not be a top-level browsing context

But aren't most (all) reading systems using an html core to render content in the spine?

The problem, of course, is that we don't require this to be true, but it's probably true from a practical perspective. Maybe we could note the assumption and that an RS that doesn't use a browser core still needs to restrict access?

epub33/rs/index.html
mattgarrish: I've updated this in the latest commit.

epub33/rs/index.html
mattgarrish: I've added the following in the latest commit: "If a Reading System does not use a top-level browsing context for Top-level Content Documents, it MUST also prevent data URLs from opening as though they are Top-level Content Documents."

That's the best I can come up with since we lack any alternative name for the rendering context.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-03-26-epub#section2) on 2021-03-26

List of resolutions:

- ***[Resolution No. 2](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-03-26-epub#resolution2):  Merge PR 1582, open issue for TAG review***

### 2. Data URLs PR 

_See github pull request [#1582](https://github.com/w3c/epub-specs/pull/1582)._




_See github issue [#1564](https://github.com/w3c/epub-specs/issues/1564)._




**Matt Garrish:** this is from what we discussed last week's call  
… where do we allow data URLs, etc.  
… browsers are starting to disallow navigation to data URLs (to circumvent security issues)  
… so last week we said we would follow that path, but still allow embedding content via data URLs  
… plus we were going to say that we wouldn't allow data URLs as spine docs  
… how to phrase this was kind of complicated  
… the language uses "top level browsing"  
… but some RSes don't use a browser based model  
… so we say that in those cases, they still have to treat the data URLs similar to the way that a browser would  

**Brady Duga:** its hard to get the language right so that we don't disallow things by mistake, or allow things by mistake  
… e.g. when an RS wants to provide the ability to let user pop out content embedded via data URL in a separate window  
… wonder if there is another group working on specifying how this should work?  
… or are we on the bleeding edge here? Would rather not be the ones who set this standard  

**Matt Garrish:** this very much seems to be raised and discussed in the browser realm  
… potentially a part of the HTML spec  
… but nothing I can point to specifically  

**Ivan Herman:** this PR does a lot that is not contentious, i think  
… i propose to merge PR as it is, subject to the nitty gritty about the exact terminology  
… but then open an issue saying that this terminology still has to be checked with TAG and HTML community  
… and link to our language  
… that way we specifically flag this for TAG input  
… the new issue would be more focused  
… and in the meantime we capture everything else in the PR  

**Matt Garrish:** in that light, do we maybe want to keep the authoring side requirements, but leave out the RS expectations side?  

**Ivan Herman:** pref trying to do something now, even with it being incomplete, raise a separate issue about the terminology, link that issue to the text and ask for external help, e.g., from the TAG.  

**Brady Duga:** would refer this to TAG, yes  

**Ivan Herman:** also, let's mark issues as explicitly "referred to TAG for input"  

**Dave Cramer:** i'm okay with merging for now, on understanding that this is not final final  

> **Proposed resolution: Merge PR 1582, open issue for TAG review** *(Wendy Reid)*

> *Ivan Herman:* +1

> *Wendy Reid:* +1

> *Matt Garrish:* +1

> *Charles LaPierre:* +1

> *Matthew Chan:* +1

> *Brady Duga:* +`

> *Brady Duga:* +1

> *Bill Kasdorf:* +1

> ***Resolution #2: Merge PR 1582, open issue for TAG review***

</details>

PR #1584, opened 2021-03-23
Clarify that non-linear content can be skipped over not suppressed completely

### LINKED ISSUE #1480: Require that linear=no content be accessible via links?
This is related to #632. 

The RS spec says:

> When rendering an EPUB Publication, a Reading System MAY either suppress non-linear content so that it does not appear in the default reading order, or ignore the linear attribute to provide users access to the entire content of the EPUB Publication. This specification does not mandate which model Reading Systems have to use. A Reading System MAY also provide the option for users to toggle between the two models.

Do we need a requirement that `linear="no"` content MUST nevertheless be reachable via links? "Suppress" is a strong word.

mattgarrish:
Implements the prose in https://github.com/w3c/epub-specs/issues/1480#issuecomment-764842519

Fixes #1480 

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/editorial/issue-1480/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/editorial/issue-1480/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2159,6 +2159,9 @@ alert("Feature " + feature + " supported?: " + conformTest);</pre>
 				-->
 
 				<ul>
+					<li>23-Mar-2021: Changed "suppressing" of non-linear content to "skipping" when traversing the spine
+						to clarify that the intention is not to remove all access to such content. See <a
+							href="https://github.com/w3c/epub-specs/issues/1480">issue 1480</a>.</li>
 					<li>10-Mar-2021: Changed restriction against using resources not listed in the Package Document to a
 						recommendation not to use. See <a href="https://github.com/w3c/epub-specs/issues/810">issue
 							810</a>.</li>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -477,12 +477,12 @@
 
 						<dt id="sec-itemref-elem">The <code>itemref</code> element</dt>
 						<dd>
-							<p>When rendering an EPUB Publication, a Reading System MAY either suppress non-linear
-								content so that it does not appear in the default reading order, or ignore the
-									<code>linear</code> attribute to provide users access to the entire content of the
-								EPUB Publication. This specification does not mandate which model Reading Systems have
-								to use. A Reading System MAY also provide the option for users to toggle between the two
-								models.</p>
+							<p>When a user traverses the default reading order defined in the <a
+									href="https://www.w3.org/TR/epub-33/#sec-spine-elem">spine</a> [[!EPUB-33]], a
+								Reading System MAY automatically skip <code>itemref</code> elements marked as non-linear
+								(excluding when a user specifically activates a hyperlink to such items). Reading
+								Systems MAY also provide the option for users to select whether non-linear content is
+								skipped by default or not.</p>
 							<p>Reading Systems MUST ignore all metadata properties expressed in the
 									<code>properties</code> attribute that they do not recognize.</p>
 						</dd>


iherman:
I do not want to ignite some old discussion... but the current text suggests that

- the RS _MAY_ skip a non-linear spine item and that is it; or
- the RS _MAY_ offer the user the choice of displaying those spine items explicitly

Because both of these are _MAY_-s, the RS can also decide to ignore `linear=no` altogether and happily display all items one after the other. Is that really acceptable?

mattgarrish:
> and happily display all items one after the other. Is that really acceptable?

Yes, that's what some reading systems do right now. We've tried to bridge this divide without success. (One of the oldest and most divisive issues.)

wareid (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-03-26-epub#section1) on 2021-03-26

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-03-26-epub#resolution1):  Close PR 1584 with the new language around linear="no", close issue 1480***

### 1. PR Clarifying non-linear items 

_See github pull request [#1584](https://github.com/w3c/epub-specs/pull/1584)._




_See github issue [#1480](https://github.com/w3c/epub-specs/issues/1480)._




**Dave Cramer:** much of what we are talking about is in the comment to this issue  
… there is some history to `linear=no`  
… items are `linear=yes` when they are in reading order  
… but we don't say what to do when linear=no  
… most common implementation seems to have `linear=no` content only appear when link is clicked  
… belief that that content can be suppressed, but we wanted to clarify that that content must be reachable if explicitly clicked  

**Brady Duga:** `linear=no` was put in for publishers, to better control what RSes was doing with order of their content  

**Dave Cramer:** one of the issues was with cover images right? sometimes in the spine, sometimes in metadata?  

**Brady Duga:** no, this was more about content at the end, e.g. footnotes  
… where each footnote is in its own document, but only linked to the main book, not in reading order  
… so without linear the footnotes would sometimes appear in random order at the end of the book  

**Dave Cramer:** i think some RS will also just move all linear=no content at the end regardless  

**Ivan Herman:** just to clarify, e.g. i have a complex illustration that i don't want in the text, like an svg, and there is a reference to it, is this an example of `linear=no`?  

**Brady Duga:** if you have a link to that from something in your spine, then yes, that would be a use for linear=no  

**Dave Cramer:** and in some RS clicking that sort of link would open that content in a window in front of the text  

**Brady Duga:** its essentially metadata about this piece of content that enables RS to make informed decision about how to process this piece of content in its UI  

**Charles LaPierre:** we just had a example where extended image descriptions were marked as linear=no and but some RSes were still showing it where it appeared in the spine  
… this was kind of counter intuitive  

**Brady Duga:** if its a link then it has to show up in the spine, so RSes that want to show it know where to do so  
… this kind of goes back to the ideological debate of whether ebooks should replicate limitations of print  
… some RS are going to show all content, and some won't  
… linear doesn't change that  
… it just allows publishers to have some control over whether it appears at the back of the book vs some other location  

**Matt Garrish:** from a RS "make it render consistently" point of view, I'm not sure we can solve that  
… from a user perspective, that's why we wanted to give users an option to skip linear=no content or not  

**Dave Cramer:** there's so much existing content that i'm not sure there is a path towards uniform behaviour of the linear attribute  
… that said, i think the clarifying language of the PR is good  
… just underlining that RS must still show linear=no content if the user clicks a link  
… and not just suppress that content  

**Ivan Herman:** the current text uses a lot of MAYs  
… i just want to be sure that this is as far as we can go  

**Hadrien Gardeur:** in Readium community (Thorium etc.), we decided 3 years ago that if something is marked as linear=no then you won't encounter if you're just moving forward through a pub, but that you will see it if accessed via a link  
… this was not a hard decision  
… but how do you present these resources?  
… some people think that linear=no resource should behave like the rest of the content (i.e. if everything is normally paginated, then it should also be paginated)  
… but it could also be scrolled etc.  
… it may not be wg's role to deal with this, but it is a problem lacking a solution  
… even when we simply the scenario, by excluding desktop user case, for example, there is still disagreement about what to do with it  

**Ivan Herman:** so, the maximum we can do at this point in terms of the PR is to say yes, but i don't know whether we are in position to put the various alternatives of what can happen (along the lines of what Hadrien just discussed)  
… but that can be a separate issue  
… and maybe have it somewhere in spec as note  
… as it currently stands, you are in for a surprise, not even sure what behaviour to expect  

**Wendy Reid:** we get very similar comments about how we handle footnotes  
… need clarity about what to do with content that is linked from main text, but not necessarily required for main text  
… e.g. is it a pop-up, or does the link take you somewhere else, etc.  

**Dave Cramer:** can we resolve on accepting PR?  

> **Proposed resolution: Close PR 1584 with the new language around linear="no", close issue 1480** *(Wendy Reid)*

> *Brady Duga:* +1

> *Matt Garrish:* +1

> *Gregorio Pellegrino:* +1

> *Masakazu Kitahara:* +1

> *Wendy Reid:* +1

> *Toshiaki Koike:* +1

> *Juliette McShane:* +1

> *Matthew Chan:* +1

> *Bill Kasdorf:* +1

> *Dan Lazin:* +1

> *Hadrien Gardeur:* +1

> *Charles LaPierre:* +1

> *Ivan Herman:* +1

> ***Resolution #1: Close PR 1584 with the new language around linear="no", close issue 1480***

</details>

Issue #1500, opened 2021-02-03
Page list ordering


mattgarrish:
The page list definition currently includes the old requirements of the table of contents -- that the list has to be ordered to reflect the occurrence of page breaks by spine entry and within each content document.

Is this actually what we want, or should the page list be allowed to (recommended to) reflect the original numbering of the source document?

gregoriopellegrino:
I think it should stay, otherwise it becomes very difficult for RS to put the pages in the correct order in the UI, as discussed in #1471

mattgarrish:
> it becomes very difficult for RS to put the pages in the correct order in the UI

But this is essentially the question. Does the user want to access the page numbers all out of sequence because that's the way the digital content is ordered, or do they want to access a sequential list that matches the print?

If they're trying to follow the print edition, in other words, wouldn't they want to follow the print sequence and not the digital?

My understanding from @danielweck's response is that thorium at least doesn't re-order the list; it just uses whatever is provided for manual access. In the case of going to a specific page, it doesn't sound like it matters.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-03-11-epub-a11y#section4) on 2021-03-11

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

### 4. pages related issues 

_See github issue [#1503](https://github.com/w3c/epub-specs/issues/1503), [#1502](https://github.com/w3c/epub-specs/issues/1502), [#1501](https://github.com/w3c/epub-specs/issues/1501), [#1500](https://github.com/w3c/epub-specs/issues/1500), [#1542](https://github.com/w3c/epub-specs/issues/1542)._



**Matt Garrish:**: Updated EPUB accessibility for addressing a part of the following static pages issues. Some requirements can be addressed by EPUB Accessibility while others should go to best practices.

> *Avneesh Singh:* [https://w3c.github.io/epub-specs/epub33/a11y/index.html#sec-page-nav](https://w3c.github.io/epub-specs/epub33/a11y/index.html#sec-page-nav)

**Matt Garrish:** we had different questions about pagelists  
… one propose is to structure the requirements for pagelists  
… e.g. level A: some pages in the pagelist  
… level AA: all content pages linked in pagelist  
… level AAA: all pages (no exclusion, even blank pages)  
… I didn't move over with my requirements  

[...]

Issue #1471, opened 2021-01-16
Questions about page-list


gregoriopellegrino:
Hi,
working in accessible EPUB production I have some questions with respect to [page-list](https://www.w3.org/publishing/epub3/epub-packages.html#sec-nav-pagelist):

- do content creators have to list **all** pages, or can they list only some of them (e.g. the page number at the beginning of the chapter)?
- how to deal with empty pages?
- what should RSs do if the page-list has jumps in numbering? Interpolate?
- are there any rules about how RSs should deal with non-Arabic page numbers?

mattgarrish:
> * do content creators have to list **all** pages, or can they list only some of them (e.g. the page number at the beginning of the chapter)?

Technically we don't require every page, as digital versions may not include all the same content as the print, but the example you give renders the feature mostly useless. It doesn't provide any value, as the table of contents already gets users to the start of chapters. Users typically want to reach other pages.

We might want to address this in the accessibility specification, but it could be tricky to word (e.g., something like "the publication must include page break markers for all print pages represented in the content")

> * how to deal with empty pages?

I believe best practice is to include these for completeness so that users don't get confused that there are pages missing.

I'd offer an alternative of omitting them and stating in the summary that blank pages are not included. Users aren't going to jump to blank pages except by accident. (But this could conflict with a requirement to represent all page breaks.)

> * what should RSs do if the page-list has jumps in numbering? Interpolate?

We generally give reading systems flexibility, but I wouldn't expect them to add missing pages. If the publication only consists of pieces of a work, the user will only want to know which pages are actually available. I'd expect if the user enters a page number that doesn't exist, the reading system would simply inform them of the fact.

> * are there any rules about how RSs should deal with non-Arabic page numbers?

Not sure what you mean here.

/cc @GeorgeKerscher @avneeshsingh @clapierre

danielweck:
In Thorium, there are 2 distinct implementation use cases:

1) the `nav@epub:type="pageList"` in the XHTML Navigation Document => used to implement the "goto page" feature based on user input / "fuzzy" string matching (i.e. not accurate number comparisons), and of course also based on the exposed list of authored "pages" in the GUI (e.g. some kind of selector control for predefined "as-is" values). In both "goto page" cases, a link from the NavDoc is activated and followed into a specific XHTML Content Document at a specific anchor location (the expectation is that this lands on a `page-break`, see below).
[...]

PR #1588, opened 2021-03-26
Remove requirement for page list to match the order of page break markers

mattgarrish:
As currently recommended in #1500, this PR removes the requirement for the page list to match the order of page break markers in the content.

Assuming this is accepted by the WG, a future PR will address recommending the page list match the print sequencing in the accessibility spec.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1588.html" title="Last updated on Mar 28, 2021, 3:20 PM UTC (6d0b46e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1588/f3879d4...6d0b46e.html" title="Last updated on Mar 28, 2021, 3:20 PM UTC (6d0b46e)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1794,7 +1794,7 @@
 										combine them in unexpected ways.</p>
 
 									<p>For example, the following example shows a basic multipart title:</p>
-									
+
 									<pre>&lt;metadata xmlns:dc="http://purl.org/dc/elements/1.1/"&gt;
     &lt;dc:title&gt;THE LORD OF THE RINGS&lt;/dc:title&gt;
     &lt;dc:title&gt;Part One: The Fellowship of the Ring&lt;/dc:title&gt;
@@ -1803,7 +1803,7 @@
 </pre>
 									<p>The same title could instead be expressed using a single <code>dc:title</code>
 										element as follows:</p>
-									
+
 									<pre>&lt;metadata xmlns:dc="http://purl.org/dc/elements/1.1/"&gt;
     &lt;dc:title&gt;
         THE LORD OF THE RINGS, Part One: The Fellowship of the Ring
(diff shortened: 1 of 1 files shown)


iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-04-01-epub#section3) on 2021-04-01

List of resolutions:

- ***[Resolution No. 2](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-04-01-epub#resolution2):  Merge PR 1588***

### 3. Pagelist Requirements 

_See github pull request [#1588](https://github.com/w3c/epub-specs/pull/1588)._




_See github issue [#1471](https://github.com/w3c/epub-specs/issues/1471)._




**Dave Cramer:** we had required that the order of the nav elements had to match the order of the elements in the spine  
… then when this was implemented in epubcheck, it turned out that it made a bunch of epubs invalid  
… we eventually got that sorted out, but we never went back to address the corresponding issue in the pagelist  
… mgarrish is this going to be covered in epub a11y instead of core now  

**Matt Garrish:** yes. The issue is that this forces the pagelist to match the order of the digital version, which is not at all helpful when there is an expectation that the pagelist correspond to print  

**Brady Duga:** why are we removing this requirement?  
… the toc is not without controversy because it breaks UIs  
… we didn't want to make publishers go back and fix their content  
… is this similar?  

**Matt Garrish:** here we're just removing the strict requirement that the pagelist match  
… so we wouldn't be invalidating anything  
… not the same as the toc because the toc is being used by RSes in various ways  
… but we don't know that RSes are using pagelist in the same way  
… allowing authors to reorder pagelist might actually make it align with expectations  

**Brady Duga:** so, does this mean, for e.g., that page 7 could come after page 10?  

**Matt Garrish:** yes, but you could still do that right now if that was the order of content in the spine  

**Wendy Reid:** with the current requirements we might accidentally be forcing the page 7 after page 10 thing  
… with the PR that might still happen, but it would be by choice, as opposed to being forced by spec  

**Ben Schroeter:** you could also have a case where the print book has a sidebar that in the digital version comes after the body text  
… in cases like that you might want the pagelist to come out of order intentionally  

**Matt Garrish:** i'm not sure if that would be an example of where we'd want that for a11y purposes or not, but there are situations where you'd want the pagelist to match print content rather than spine order  

**Dave Cramer:** i'm seeing several cases where we're making the choice that these "best practices" are better off being in the a11y spec when the directly impact a11y rather than try to bake good practice into the epub spec itself, especially when there is such a diversity of content  

**Ben Schroeter:** but i've always wonder why pagelist is an a11y feature  
… it is, but it is also useful for everyone  

**Matt Garrish:** that's a possibility we could take up, i.e. just recommend something about page sequencing  

> **Proposed resolution: Merge PR 1588** *(Wendy Reid)*

> *Ben Schroeter:* +1

> *Matthew Chan:* +1

> *Matt Garrish:* +1

> *Shinya Takami (高見真也):* +1

> *Wendy Reid:* +1

> *Dan Lazin:* +1

> *Toshiaki Koike:* 0

> *Brady Duga:* +0 as I don't fully understand all the implications

> *Masakazu Kitahara:* 0

> ***Resolution #2: Merge PR 1588***

**Wendy Reid:** i understand your concern duga  
… and this is something we can take a look at when we do virtual locators as well  
… as there is crossover  

**Brady Duga:** it may be fine, but i'd just like to think about it some more first  

</details>

PR #1594, opened 2021-03-27
Clarify that a primary itemref is one with linear="yes"

mattgarrish:
This is a minor edit from the call yesterday. There's an assumption that anyone reading the RS spec will understand that a "primary" itemref is one whose linear attribute is "yes". This just makes that point explicit in a parenthetical.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1594.html" title="Last updated on Mar 27, 2021, 8:03 PM UTC (4dd1fd8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1594/493b1ce...4dd1fd8.html" title="Last updated on Mar 27, 2021, 8:03 PM UTC (4dd1fd8)">Diff</a>

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -179,7 +179,6 @@
 								System affordance such as a context menu. If a Reading System does not use a top-level
 								browsing context for <a>Top-level Content Documents</a>, it MUST also prevent data URLs
 								from opening as though they are Top-level Content Documents.</p>
-
 							<p class="issue" data-number="1592"></p>
 						</li>
 						<li>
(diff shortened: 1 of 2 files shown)


epub33/rs/index.html
```diff
@@ -471,8 +471,9 @@ <h3>Spine</h3>
 						<dd>
 							<p>Reading Systems MUST provide a means of rendering the EPUB Publication in the order
 								defined in the <code>spine</code>, which includes: 1) recognizing the first primary
-									<code>itemref</code> as the beginning of the default reading order; and, 2)
-								rendering successive primary items in the order given in the <code>spine</code>.</p>
+									<code>itemref</code> (i.e., whose <code>linear</code> attribute value is
+									"<code>yes</code>") as the beginning of the default reading order; and, 2) rendering
+								successive primary items in the order given in the <code>spine</code>.</p>
```
iherman: I know that per the standard the default value of `linear` is `true` and it is not required. Nevertheless, maybe to avoid confusion, wouldn't it be better to say 

> whose <code>linear</code> attribute is missing or its value is "<code>yes</code>"

WDYT?

iherman (approved):

epub33/rs/index.html
mattgarrish: The more I look at this, the more I wonder why a new term like "primary itemref" was even introduced.

The latest commit changes the term to "linear itemrefs" and links across to where the requirement for one linear item is defined in the core spec.

iherman: That looks good to me!

Issue #1516, opened 2021-02-15
Clarify what support for "presentation logic" means


mattgarrish:
The reading system spec has this high-level requirement for the package document:

> [A Reading System] MUST honor all presentation logic expressed through the Package Document [EPUB-33] (e.g., the reading order, fallback chains, page progression direction and fixed layouts).

https://w3c.github.io/epub-specs/epub33/rs/#sec-package-rs-conformance

The inclusion of fixed layouts in the examples is confusing, though, as it there is a separate bullet for fixed layout handling in the same list.

I believe the intention of this bullet was only to require support for the "presentation logic" expressed directly through the elements and attributes of the package document, not indirectly through metadata expressed in meta tags.

On top of the general ambiguity about what "presentation logic" refers to, "honor" isn't exactly a technical term. The bullet seems more like an indirection to actual requirements/recommendations than something that can be tested.

In any case, it needs work.

mattgarrish:
I think this "requirement" is just a regular old pointer wrapped up in some serious-sounding handwaving.

If there is "logic" that reading systems have to "honor", then it should be written as normative requirements not left for developers to vaguely give deference to.

Looking at the examples cited:

- the reading order - we already define requirements for processing in the [spine section](https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-spine)
- fallback chains - we define fallback processing in the [foreign resources](https://w3c.github.io/epub-specs/epub33/rs/#sec-epub-rs-conf-foreign-res) and [manifest](https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-manifest) sections
- page progression direction - also defined in the spine section
- fixed layouts - aren't inherent logic of the package markup but at any rate have a [dedicated section](https://w3c.github.io/epub-specs/epub33/rs/#sec-fixed-layouts) of requirements

So what else is there? General rendering metadata is like fixed layouts - not really applicable but already covered.

If we're missing requirements for any of the above, or there's something I didn't spot, we should be adding those in not hoping they'll be "honored".

mattgarrish:
So, in case it needs saying, my proposal is to drop this statement/section.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-04-15-epub#section6) on 2021-04-15

List of resolutions:

[...]

PR #1597, opened 2021-03-30
Restructuring of the reading system specification

mattgarrish:
This PR also sprung out of last week's telecon as I noticed that the spine discussion was split between elements only because that's the way the core spec was structured. There's no need to break topics like that along element lines.

From there, I noticed other quirks of pulling out the RS requirements that could stand for improvement:

I've removed all the "conformance" sections that were embedded within each subsection. I've either integrated these without the headings if they're the only thing in the section and contain actual requirements, or deleted them if all they contained were pointers to other sections (most of these lists were just helpful locators to where actual requirements were listed when the specs were merged so devs didn't have to read everything). Some of these statements were confusingly half-true (e.g., you "must" support scripting as defined in the scripting section when the scripting section doesn't actually require support for scripting).

Some of the other notable changes include:

- I've grouped three bullets in the main conformance section that all began with "if it has a viewport".
- I've noted that support for fixed layouts and media overlays is optional in the main conformance section (we used to say that support for fixed layout metadata is required, but this was never true as reading systems default to conformance, so I removed that statement)
- The support for "presentation logic" bullet in the package document conformance I've added as a subsection since I couldn't find any other place for it. It still needs to be returned to for #1516 
- The only bullet in the xhtml conformance section that wasn't redundant was support for ARIA. I've added this to the extensions section as a new section (HTML allows ARIA attributes but I can't find that it requires exposing them).
- I merged the requirements for handling author/user css changes with the conformance bullets and dropped the subsectioning.
- I've moved the informative section on scripting security to the security section at the end. These will likely need better integration in the future, though.

Also, the media overlays conformance section had one bullet that wasn't just a pointer to a subsection, but it was a bit confusing in that it says that xhtml content documents have to be supported and SVG may. I've made the obvious assumption this refers to playback, not support for the formats and worked this bullet into the loading section.

Long story short, the intention is only to simplify the spec by removing the unnecessary pointers and subsectioning. Actual requirements that have to be conformed to should all remain.

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -124,9 +124,9 @@
 
 					<p>This specification does not require <a>EPUB Reading Systems</a> to support scripting, HTML forms
 						or the HTML DOM. Reading Systems conformant with this specification are only expected to be able
-						to process a conforming <a>EPUB Content Document</a>. As <a href="#sec-scripted-content-rs-reqs"
-							>support for scripting and HTML forms</a> is not compulsory, a conformant Reading System
-						might not be a fully-conformant HTML user agent.</p>
+						to process a conforming <a>EPUB Content Document</a>. As <a href="#sec-scripted-content">support
+							for scripting and HTML forms</a> is not compulsory, a conformant Reading System might not be
+						a fully-conformant HTML user agent.</p>
 				</section>
 
 
@@ -205,7 +205,7 @@
 					<ul class="conformance-list">
 						<li>
 							<p id="confreq-rs-epub3-css">It MUST support visual rendering of <a>XHTML Content
-									Documents</a> as defined in <a href="#sec-css-rs-conf"></a>.</p>
+									Documents</a> as defined in <a href="#sec-css"></a>.</p>
 						</li>
 						<li>
 							<p id="confreq-rs-epub3-images">It MUST support the <a
@@ -221,7 +221,7 @@
 					<p>Support for the following features is OPTIONAL:</p>
 					<ul>
 						<li>
-							<p id="confreq-fxl"><a href="#sec-fixed-layouts">Fixed-Layout Documents</a></p>
+							<p id="confreq-fxl"><a href="#sec-fxl">Fixed-Layout Documents</a></p>
 						</li>
 						<li>
 							<p id="confreq-mo"><a href="sec-media-overlays">Media Overlays</a></p>
@@ -695,7 +695,7 @@
 					</li>
 					<li>
 						<p id="confreq-svg-rs-scrpt">It MUST meet the Reading System conformance criteria defined in <a
-								href="#sec-scripted-content-rs-reqs"></a>.</p>
+								href="#sec-scripted-content"></a>.</p>
 					</li>
 					<li>
 						<p id="confreq-svg-rs-css">If it has a <a>Viewport</a>, it MUST support the visual rendering of
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -308,9 +308,10 @@
 			<section id="pkg-doc-presentation">
 				<h4>Presentation Logic</h4>
 
-				<p id="confreq-rendition-rs-package">It MUST honor all presentation logic expressed through the <a
-						href="https://www.w3.org/TR/epub-33/#sec-package-def">Package Document</a> [[!EPUB-33]] (e.g.,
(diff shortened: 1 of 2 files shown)


mattgarrish:
- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/editorial/rs-reorg/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/editorial/rs-reorg/epub33/rs/index.html)

epub33/rs/index.html
```diff
@@ -280,280 +302,215 @@ <h2>EPUB Publications</h2>
...
-					</li>
-					<li>
-						<p id="confreq-rendition-rs-manifest">It SHOULD NOT use non-<a>Publication Resources</a> in the
-							rendering of an EPUB Publication due to the inherent limitations and risks involved (e.g.,
-							lack of information about the resource and how to process it, security risks from
-							remotely-hosted sources, lack of fallbacks, etc.).</p>
-					</li>
-				</ul>
+				<p id="confreq-rendition-rs-package">It MUST honor all presentation logic expressed through the <a
+						href="https://www.w3.org/TR/epub-33/#sec-package-def">Package Document</a> [[!EPUB-33]] (e.g.,
```
iherman: This is a new section; maybe it reads better if "an EPUB Reading System" is used instead of "It"...

epub33/rs/index.html
```diff
@@ -1230,20 +1050,32 @@ <h4>The <code>rendition:page-spread-*</code> Properties</h4>
...
+				<h3>Initial Containing Block Dimensions</h3>
 
-				<p>In this version of this specification, only the width and height expressions as defined in <a
-						href="https://www.w3.org/TR/epub-33/#sec-fxl-icb-html">Expressing in HTML</a> [[!EPUB-33]] MUST
-					be recognized by <a>Reading Systems</a>.</p>
+				<dl>
+					<dt>XHTML</dt>
+					<dd>
+						<p>In this version of this specification, <a>Reading Systems</a> MUST recognize only the width
+							and height expressions as defined in <a
```
iherman: Why this "In this version of this specification"? I have not seen this type of reference elsewhere...

epub33/rs/index.html
```diff
@@ -1930,6 +1683,89 @@ <h2>Security and Privacy</h2>
 				<p>The Working Group will have to address the particularity of cross-origin by addressing the question
 					of what the "origin" is for an EPUB Publication.</p>
 			</div>
+
+			<section id="sec-scripted-content-security">
+				<h4>Scripting Considerations</h4>
```
iherman, line 1684: I wonder whether the editor's note preceding this section on the origin is still relevant. I believe that has been solved.

iherman: Actually, I wonder whether the editor's note referring to the pub manifest should still be kept, too. Maybe just to say that this section needs further work overall.

iherman (approved):
An additional comment: although, strictly speaking, this is not a substantive change, I think this editorial change should also be mentioned in the change log.

mattgarrish:
> I think this editorial change should also be mentioned in the change log.

Ya, no harm. I was thinking being a new creation that structural reshuffling is par for the course, but now that we're past fpwd it is a notable change.

epub33/rs/index.html
mattgarrish: > Why this "In this version of this specification"?

I think that was added because there wasn't a firm definition of the viewport meta when we first introduced this. We were still citing the apple developer docs. Sort of a subtle warning that there might be additional requirements in the future.

I agree it's not really needed, as everything is subject to change in the future.

epub33/rs/index.html
mattgarrish: Oops, ya, missed that one.

epub33/rs/index.html
mattgarrish: I'm also going to tweak the wording so that it's the same as the SVG ICB text.

mattgarrish:
I'm going to merge this so it doesn't become a blocker. I'm sure we'll continue to work at optimizing this document.

Issue #1601, opened 2021-04-02
Contradiction between "must" and "advisory"?


dlazin:
https://w3c.github.io/epub-specs/epub33/rs/#sec-opf-dclanguage says `The language(s) of the EPUB Publication specified in language elements are advisory. Reading Systems MUST use the language information associated with a resource to determine its language.`

This seems contradictory. In normal use, "advisory" means non-binding, whereas MUST is binding.

dauwhe:
My understanding is that "advisory" applies to the `dc:language` element in the package file. The "resource" here refers to (for example) HTML files. The reading system MUST obey the language in `<html lang="en-CA">`.

dlazin:
Ah! That makes sense upon explanation, but is unclear to a non-expert. Can we rephrase?

dauwhe:
> Ah! That makes sense upon explanation, but is unclear to a non-expert. Can we rephrase?

Yes we can!

dlazin:
Related (and a little clearer): `The language identified in an hreflang attribute is not authoritative. Language information defined in a linked resource determines its language.`
https://w3c.github.io/epub-specs/epub33/rs/#conf-metadata-link

iherman:
> The reading system MUST obey the language in `<html lang="en-CA">`.

I guess this is also true for the `xml:lang` attribute in the package file, isn't it?

mattgarrish:
> I guess this is also true for the `xml:lang` attribute in the package file

Yes, I believe the original prose cited was lifted from the `hreflang` definition in the HTML spec when we were looking at this during the i18n review.

It might be clearer to have a section on language and publication resources so that we're not having to say the same thing in every case.

iherman:
> > I guess this is also true for the `xml:lang` attribute in the package file
> 
> Yes, I believe the original prose cited was lifted from the `hreflang` definition in the HTML spec when we were looking at this during the i18n review.
> 
> It might be clearer to have a section on language and publication resources so that we're not having to say the same thing in every case.

Yes. Furthermore, looking at the original problem of the issue (which is in a section on metadata only, ie on dc: language, the second sentence should simply removed. It is irrelevant in the given context.

PR #1609, opened 2021-04-04
Further deconstruct the epub publications section

mattgarrish:
Opening this PR for review from anyone who has an opinion -- it's not finished or ready to be integrated yet, so not asking for formal reviews.

These edits are another pass at trying to improve the support statements that reference out to entire sections. Instead of grouping at the top, I've moved the requirements to the beginning of their respective sections so it's obvious when you reach each section what the expectation is (the boxes I've formatted can no doubt be improved on, but are there to add extra emphasis to the top-level support nature of the statements).

This is mostly an exercise in shuffling, as unlike the last PR none of the requirements were redundant and could be removed.

One thing this did highlight was that we were missing specific requirements for some sections. In particular, there wasn't a top-level requirement to process the navigation document. The vocabulary sections also tended to lack any clear support statements, so I've added those as best I understand them (epub:type is optional but the vocabulary handling mechanisms in the package document are required).

I had to make up a "must process publication resources" requirement for the group of resource sections that were embedded in the old section 2. I'm not sure this is the best phrasing, but I can't think of a better statement (saying RSes must support them is misleading, and saying they must handle them seems to make less sense than process).

Anyway, before I do any more clean up on this PR, I want to make sure it's going in the right direction, so any general comments welcome.

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/editorial/rs-pub-req/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/editorial/rs-pub-req/epub33/rs/index.html)

### DIFF
--- a/epub33/common/css/common.css
+++ b/epub33/common/css/common.css
@@ -333,3 +333,18 @@ section#structure-vocab dd + dd {
 .subproplabel {
 	font-style: italic;
 }
+
+p.support-req, p.support-rec, p.support-opt {
+	border: 0.1rem solid cornflowerblue;
+	border-left-width: 0.4rem;
+	border-right-width: 0.4rem;
+	padding: 1.5rem;
+	background-color: rgb(245,245,245);
+	font-weight: bold;	
+}
+
+p.support-rec {
+}
+
+p.support-opt {
+}
--- a/epub33/common/css/common.css
+++ b/epub33/common/css/common.css
@@ -10,11 +10,6 @@ ul.flat {
 	padding-left: 0;
 }
 
-
-ul.conformance-list {
-	padding-left: 0;
-}
-
 ul.conformance-list > li > ul.conformance-list {
 	padding-left: 2em;
 }
@@ -334,17 +329,8 @@ section#structure-vocab dd + dd {
 	font-style: italic;
 }
 
-p.support-req, p.support-rec, p.support-opt {
-	border: 0.1rem solid cornflowerblue;
-	border-left-width: 0.4rem;
-	border-right-width: 0.4rem;
-	padding: 1.5rem;
-	background-color: rgb(245,245,245);
-	font-weight: bold;	
-}
-
-p.support-rec {
-}
-
-p.support-opt {
+p.support {
+	border-left: 0.4rem solid rgb(255,255,0);
+	background-color: rgb(255,255,230);
+	padding: 0.7rem;
 }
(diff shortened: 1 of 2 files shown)


iherman:
I like it. A lot. It makes the whole document so much clearer (what a journey since 3.2!).

I have only one specific comment and this is related to #1601 or, more specifically, to your remark in https://github.com/w3c/epub-specs/issues/1601#issuecomment-812859449. It may be a good idea to add a general i18n section in (the new) §2, much like we have a section on Data URLs. The section would say (this is just a rough text, should be wordsmithed):

> RS-s _MUST_ process language information provided by `xml:lang` and HTML's `lang` attribute values, as well as the base direction values expressed by HTML's or the Package document's 'dir'  for the natural language content in all content documents, when appropriate

I am not sure whether §3.3 should remain, or should be folded into this (except maybe specifying the `auto` default values).

At the moment, there is nothing about language in the spec, except for the section mentioned in #1601. One could of course argue that the XHML conformance covers that, and that is of course correct, but @dlazin's comment shows that there is a possible source of confusion.

iherman:
(B.t.w., once this gets to an equilibrium point, I am happy to try to draft such a section, unless you beat me into it!)

mattgarrish:
> It may be a good idea to add a general i18n section in (the new) §2, much like we have a section on Data URLs.

Yes, I had to hold myself back from trying to solve some of the other issues at the same time. I figure there are enough moving parts in this PR as it is, so better to focus on one thing.

Once this is done -- which should be shortly as I only want to check some of the lead-in sentences after the highlighted requirements -- feel free to jump on #1601 or the others.

mattgarrish:
Okay, I think what's in the PR now is reasonably good to go. The last commit was mostly minor editorial stuff: standardizing some language, making sure all the requirements link across to their respective sections in the core and not just definitions, etc.

The only mildly notable change I had to do was in the xml processing section. Three of the four bullets described the type of processor a reading system has to be, so I separated the bullet about not resolving external identifiers.

Still not sure about the colour scheme for the top-level requirements. All the colour shadings that aren't too offensive are in use (blue for definitions, green for notes, etc.). I settled on yellow as it's not too hard on the eyes and there aren't many examples in the reading system specification to collide with. Suggested improvements welcome.

Given that this is less of a change than the last pass, it might be good to also get this out in tomorrow's release. I don't mind putting a new snapshot together.

dauwhe (approved):
Yeah, I like this approach, too. It provides context for the broad requirements, which helps understanding. And keeping all the CSS stuff together, for example, is lovely.

iherman (approved):

PR #1613, opened 2021-04-06
Added an I18N section to §2 of the RS 

### LINKED ISSUE #1601: Contradiction between "must" and "advisory"?

iherman:
Fix #1601

See:

    * [Preview](https://raw.githack.com/w3c/epub-specs/rs-i18n/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/rs-i18n/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -230,12 +230,12 @@
 			<section id="sec-epub-rs-i18n">
 				<h2>Internationalization</h2>
 				<p>
-					A Reading System MUST process the attributes setting the primary language for the documents' contents for all <a href="https://www.w3.org/TR/epub-33/#dfn-publication-resource">Publication Resources</a>, when applicable. This includes:    
+					A Reading System MUST process, for all <a href="https://www.w3.org/TR/epub-33/#dfn-publication-resource">Publication Resources</a>, the attributes that set the primary language for the documents' contents, when applicable. This includes:    
 				</p>
 				<ul>
 					<li>
 						the <code>xml:lang</code> attribute for all XML documents like the 
-						<a  href="https://www.w3.org/TR/epub-33/#attrdef-xml-lang">Package Document</a> 
+						<a href="https://www.w3.org/TR/epub-33/#attrdef-xml-lang">Package Document</a> 
 						or the <a href="https://www.w3.org/TR/epub-33/#sec-overlay-docs">Media Overlay 
 						Document</a> [[!EPUB-33]].
 					</li>
@@ -248,9 +248,8 @@
 				</ul>
 
 				<p>
-					Similarly, a Reading System MUST process the attributes setting the base direction 
-					for the documents' contents for all <a href="https://www.w3.org/TR/epub-33/#dfn-publication-resource">Publication Resources</a>, 
-					when applicable. This includes:
+					Similarly, a Reading System MUST process, for all <a href="https://www.w3.org/TR/epub-33/#dfn-publication-resource">Publication Resources</a>, the attributes that set the base direction 
+					for the documents' contents, when applicable. This includes:
 				</p>
 
 				<ul>
@@ -266,7 +265,6 @@
 						the <code>direction</code> attribute for <a href="https://www.w3.org/TR/SVG2/text.html#DirectionProperty">SVG</a> [[!SVG]].
 					</li>
 				</ul>
-
 			</section>
 		</section>
 		<section id="sec-package-doc">
@@ -352,8 +350,11 @@
 
 					<dt id="sec-opf-dclanguage">The <code>language</code> element</dt>
 					<dd>
-						<p>The language(s) of the <a>EPUB Publication</a> specified in <code>language</code> elements
-							are advisory.</p>
+						<p>
+							The language(s) of the <a>EPUB Publication</a> specified in <code>language</code> elements
+							are informational. This information MAY be exposed to the user through the Reading System specific
+							user interface.
+						</p>
 					</dd>
 
 					<dt id="dc-creator">The <code>creator</code> element</dt>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -226,6 +226,45 @@
 						href="https://www.w3.org/TR/2008/REC-xml-20081126/#NT-ExternalID">external identifiers</a>
 					[[!XML]].</p>
 			</section>
+
+			<section id="sec-epub-rs-i18n">
+				<h2>Internationalization</h2>
+				<p>
+					A Reading System MUST process the attributes setting the primary language for the documents' contents for all <a href="https://www.w3.org/TR/epub-33/#dfn-publication-resource">Publication Resources</a>, when applicable. This includes:    
+				</p>
+				<ul>
+					<li>
+						the <code>xml:lang</code> attribute for all XML documents like the 
+						<a  href="https://www.w3.org/TR/epub-33/#attrdef-xml-lang">Package Document</a> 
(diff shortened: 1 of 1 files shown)


epub33/rs/index.html
```diff
@@ -311,8 +350,7 @@ <h4>Metadata</h4>
 					<dt id="sec-opf-dclanguage">The <code>language</code> element</dt>
 					<dd>
 						<p>The language(s) of the <a>EPUB Publication</a> specified in <code>language</code> elements
-							are advisory. Reading Systems MUST use the language information associated with a resource
-							to determine its language.</p>
+							are advisory.</p>
```
dlazin: Can we pick something clearer than "advisory"? It feels like we should be saying what ought to be done, probably normative, for example "Reading Systems SHOULD NOT use language elements to blah blah blah because the language is controlled by the [Package Document's language]..." (But if that's the case, what's even the point of the language element? I might be confused.)

epub33/rs/index.html
```diff
@@ -226,6 +226,45 @@ <h4>XML Processing</h4>
 						href="https://www.w3.org/TR/2008/REC-xml-20081126/#NT-ExternalID">external identifiers</a>
 					[[!XML]].</p>
 			</section>
+
+			<section id="sec-epub-rs-i18n">
+				<h2>Internationalization</h2>
+				<p>
+					A Reading System MUST process the attributes setting the primary language for the documents' contents for all <a href="https://www.w3.org/TR/epub-33/#dfn-publication-resource">Publication Resources</a>, when applicable. This includes:
```
dlazin: Participles like "setting" can be unclear. Could this be "process the attributes that set the primary language" or "to set" or similar, depending on what we mean? I think "that set" is best unless it isn't what we mean to imply.

epub33/rs/index.html
```diff
@@ -226,6 +226,45 @@ <h4>XML Processing</h4>
...
+
+			<section id="sec-epub-rs-i18n">
+				<h2>Internationalization</h2>
+				<p>
+					A Reading System MUST process the attributes setting the primary language for the documents' contents for all <a href="https://www.w3.org/TR/epub-33/#dfn-publication-resource">Publication Resources</a>, when applicable. This includes:    
+				</p>
+				<ul>
+					<li>
+						the <code>xml:lang</code> attribute for all XML documents like the 
+						<a  href="https://www.w3.org/TR/epub-33/#attrdef-xml-lang">Package Document</a>
```
```suggestion
						<a href="https://www.w3.org/TR/epub-33/#attrdef-xml-lang">Package Document</a> 
```

epub33/rs/index.html
```diff
@@ -226,6 +226,45 @@ <h4>XML Processing</h4>
...
+
+					<li>
+						the <code>xml:lang</code> attribute for 
+						<a href="https://html.spec.whatwg.org/multipage/dom.html#the-lang-and-xml:lang-attributes">XHTML</a> [[!HTML]] 
+						or <a href="https://www.w3.org/TR/SVG/struct.html#LangSpaceAttrs">SVG</a> [[!SVG]].
+					</li>
+				</ul>
+
+				<p>
+					Similarly, a Reading System MUST process the attributes setting the base direction
```
dlazin: Same comment — "that set" or "to set" instead of "setting".

dlazin (approved):

epub33/rs/index.html
iherman: The point is: it _is_ advisory indeed. Put it another way, it is metadata about the publication proper, but there is no processing requirement added to it.

The only thing it may influence is the user interface behaviour of the reading system, of the book catalogues, etc. It does _not_ control the language of the individual textual items in the metadata (that is what `xml:lang` is for) or the language of the individual content document (which have their own way of setting the language, encoded in the respective standards.

I made some adjustments, though, here is how it looks now:

> The language(s) of the <a>EPUB Publication</a> specified in <code>language</code> elements are informational. This information MAY be exposed to the user through the Reading System specific user interface.
	
Is that better?

epub33/rs/index.html
iherman: Yep. Using 'that set'

epub33/rs/index.html
iherman: Yep.

epub33/rs/index.html
iherman: Removed the unnecessary space

epub33/rs/index.html
dlazin: Yes, much clearer, thanks. (Sorry, it's hard to ask good directional questions sometimes because I only half-understand what these attributes do :)

iherman:
@mattgarrish can we merge this?

mattgarrish:
Aren't we still missing an important piece about what to do when language information isn't provided in the resource? I was expecting something like this at the end of the section:

> In the absence of this information in a Publication Resource, Reading Systems MUST NOT assume either language or direction from information expressed in the Package Document (i.e., in `xml:lang` and `dir` attributes, in `hreflang` attributes on `link` elements, or from `dc:language` elements). The handling of documents without explicit language or direction is defined in their respective specifications.

HTML, for example, has this to say:

> In the absence of any such language information, and in cases where the higher-level protocol reports multiple languages, the language of the node is unknown, and the corresponding language tag is the empty string.

SVG isn't quite as clear, but nodes don't have a language if one isn't set.

The statement about using language for display seems to suggest that's the only use, but a reading system can also use `dc:language` for any internal needs it may require. The usual example that comes to mind is there's a reading system that uses these languages to pre-load TTS languages so they're available as soon as the user begins to read rather than having to delay playback for each document to see what it requires. It could also be used to categorize books by language in a bookshelf. And so on.

It might help to make the intended uses a little more open-ended and non-normative:

> The language(s) of the <a>EPUB Publication</a> specified in <code>language</code> elements are informational. Some uses of this information include:
> - exposing it to the user through the Reading System specific user interface
> - using it to enhance functionality in a bookshelf (e.g., sorting by language)
> - using it to optimize the reading interface (e.g., to preload text-to-speech languages)

And possibly end with a note that refers back to the internationalization section for how to determine the language of resources.

iherman:
I do like your additions, @mattgarrish. I have committed the changes.

epub33/rs/index.html
```diff
@@ -310,9 +354,16 @@ <h4>Metadata</h4>
...
 					<dd>
-						<p>The language(s) of the <a>EPUB Publication</a> specified in <code>language</code> elements
-							are advisory. Reading Systems MUST use the language information associated with a resource
-							to determine its language.</p>
+						<p>
+							The language(s) of the <a>EPUB Publication</a> specified in <code>language</code> elements
+							are informational. Some uses of this information include:
+						</p>
+						<ul>
+							<li>exposing it to the user through the Reading System specific user interface</li>
```
```suggestion
							<li>exposing it to the user through the Reading System user interface</li>
```

dlazin (approved):

mattgarrish:
I'm going to merge this but it needs a change log entry. Do you want to push one directly to main @iherman or should I?

Issue #1610, opened 2021-04-05
epubReadingSytem methods should be marked as required


dlazin:
https://w3c.github.io/epub-specs/epub33/rs/#app-ers-properties (`epubReadingSystem` properties) says `The following properties MUST be made available for retrieving information about the Reading System.`

However, the equivalent section for `epubReadingSystem` methods, https://w3c.github.io/epub-specs/epub33/rs/#app-ers-methods, does not explicitly say that the methods must be implemented. Can we copy the same sentence, swapping "properties" for "methods"?

iherman:
There is a more general issue about the role and fate of Appendix A in general: #1420. I would propose to postpone this issue until #1420 is resolved.

dlazin:
Fine by me if that's what you want, but we *could* just add the one sentence (mirroring the Properties section) and then close this issue, just so we have fewer option issues. #1420 could still track the bigger question. To me, adding this one sentence now seems uncontroversial by itself because Properties already has it.

iherman:
You are probably right. But we should sync with what we decide related to #1615

dlazin:
That's merged now — want to get this one out of the way?

iherman:
Well... looking at it again, I am not sure that there is anything to change. The real meat is in [§A.4.1.2](https://w3c.github.io/epub-specs/epub33/rs/#app-ers-hasFeature-features) and that says essentially the same thing already.

dlazin:
Yes, you're right. Closing!

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-04-08-epub#section3) on 2021-04-08

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-04-08-epub#resolution1):  Flag appendix A as at-risk, keep as a normative requirement and test implementation, remove the WebIDL***

<details><summary><i>View the transcript</i></summary>

### 3. WebIDL/epubReading System 

_See github issue [#1610](https://github.com/w3c/epub-specs/issues/1610), [#1420](https://github.com/w3c/epub-specs/issues/1420), [#716](https://github.com/w3c/epub-specs/issues/716)._




> *Wendy Reid:* See [relevant section in the core spec](https://www.w3.org/TR/epub-rs-33/#app-epubReadingSystem)

[...]

Issue #1420, opened 2020-11-11
Usage of WebIDL, Appendix A in RS.


iherman:
The RS document includes a separate section [§A JavaScript epubReadingSystem Object](https://w3c.github.io/epub-specs/epub33/rs/#app-epubReadingSystem), and this section is normative (it is not defined as non-normative). WebIDL is considered (these days) as _defining browser features_. In the present situation what that would mean is that to pass the CR criteria _**we would have to prove at least two browser implementations using this interface**_.

We probably do not want to do that, we cannot expect reading systems to implement a RS. This section must therefore be marked as non-normative. 

(We should also check with browser implementers whether the interface definition is valid and useful.)

mattgarrish:
> we would have to prove at least two browser implementations using this interface

Reading systems have to extend JS for this, yes, so we'd have to show that you can access the object within html documents within a reading system. That is the normative requirement.

But we have a deferred issue about dropping RSO: https://github.com/w3c/epub-specs/issues/716

iherman:
We used WebIDL in the JSON-LD 1.1 spec. It was used to define an interface to JSON-LD processing in a programming language independent way, but there is no requirement in JSON-LD11 to have any processing implemented in browsers. On the other hand, the current view for many is that WebIDL _must be used for browser interfaces only_. This view of the world led to formal objections at the time of our Proposed Recommendation of JSON-LD1.1, and we had to do some last minute kludge to get the documents published.

If indeed the resolution of #716 is to keep the interface, either in its current or some restricted form, I believe we should do an editorial change and define the interface using some other formalism. We should simply keep away from WebIDL just for our own sanity.

mattgarrish:
> I believe we should do an editorial change and define the interface using some other formalism

But it's a javascript interface? It's to allow authors to query the reading system from within javascript. What other than WebIDL would you use for that?

Have a look at example 2:

```
var feature = "dom-manipulation";
var conformTest = navigator.epubReadingSystem.hasFeature(feature);
alert("Feature " + feature + " supported?: " + conformTest);
```

iherman:
> But it's a javascript interface? It's to allow authors to query the reading system from within javascript

I do not have a satisfactory answer :-(

The interface is very small. We can simply define it in Javascript, or Typescript to get the type annotations, or something similar.

mattgarrish:
Well, my preference would be to drop it if we can.

iherman:
Just to move ahead with this:

[...]

PR #1615, opened 2021-04-06
Fixing issue in 1611

### LINKED ISSUE #1611: OPTIONAL should not be capitalized in epubReadingSystem.hasFeature()
https://w3c.github.io/epub-specs/epub33/rs/#app-ers-hasFeature-desc says `The OPTIONAL `version` parameter is included for querying custom features that could change in incompatible ways over time. The return value indicates support only for the specified version of the feature.`

But this is referring to optionality from the caller's perspective, not the implementer's perspective. I believe that all-caps OPTIONAL should be reserved for features that a Reading System may or may not implement; in this case, I believe the Reading System MUST implement the `version` parameter, and an EPUB author may or may not use the parameter.

Therefore (assuming my understanding is correct), I think OPTIONAL should not be capitalized in this case.

iherman:
Turning terms into lower case, see #1611.

Fix #1611

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1900,11 +1900,11 @@ partial interface Navigator {
 							boolean value indicating whether any version of the specified feature is supported, or
 								<code>undefined</code> if the specified feature is not recognized.</p>
 
-						<p>The OPTIONAL <code>version</code> parameter is included for querying custom features that
+						<p>The optional <code>version</code> parameter is included for querying custom features that
 							could change in incompatible ways over time. The return value indicates support only for the
 							specified version of the feature.</p>
 
-						<p><a>EPUB Creators</a> SHOULD NOT include the <code>version</code> parameter when querying <a
+						<p><a>EPUB Creators</a> should not include the <code>version</code> parameter when querying <a
 								href="#app-ers-hasFeature-features">features defined in this specification</a> — these
 							features are considered versionless. If a Reading System supports a feature defined in this
 							specification, it MUST ignore any supplied <code>version</code> parameter and return a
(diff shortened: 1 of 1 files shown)


dlazin (approved):

mattgarrish:
Sorry, this is unfortunately a bigger problem than this PR gets into.

A quick recap is that there was no easy way to break up this section at the time I was pulling the reading system requirements out from the content, so I took the lazy route and brought it all over to the RS spec since there were far fewer authoring requirements embedded.

We also have the lingering questions of whether this interface should be deprecated or made informative to address. I was sort of hoping the whole section would go away before having to figure out what to do with the authoring statements... :)

So the solution could be
- keeping the interface and finding some way of writing the requirements into the authoring specification
- making the interface informative, in which case all normative keywords can go away
- deprecating the interface and dropping the section except for a link back to 3.2 where it would be last defined

iherman:
> Sorry, this is unfortunately a bigger problem than this PR gets into.

Indeed, this is https://github.com/w3c/epub-specs/issues/1420 which I referred to from in #1610. But (although I did not produce a PR for #1610 yet) I tend to agree with the comment of @dlazin in https://github.com/w3c/epub-specs/issues/1610#issuecomment-814298887 that making this change is easy, we can at least improve what is already there and come back to #1420 at some point (which we will have to do).

mattgarrish:
Can we at least drop "EPUB Creators" from the sentence?

I'd find this less problematic if it didn't bring in creators at all:

> Features defined in this specification are considered versionless. If a Reading System supports a feature defined in this specification, it MUST ignore any supplied version parameter and return a true value.

iherman:
> 
> I'd find this less problematic if it didn't bring in creators at all:
> 
> > Features defined in this specification are considered versionless. If a Reading System supports a feature defined in this specification, it MUST ignore any supplied version parameter and return a true value.

I have made the change.

Issue #1313, opened 2020-04-03
The default value of rendition:flow and testing


dauwhe:
[We say](https://www.w3.org/publishing/epub32/epub-packages.html#layout-property-flow-usage):

> The default value auto MUST be assumed by Reading Systems as the global value if no meta element carrying this property occurs in the metadata section. Reading Systems MAY support only this default value.

I'm trying to figure out if it's possible to violate this rule. The [definition of auto](https://www.w3.org/publishing/epub32/epub-packages.html#layout-property-flow-values) is:

> The Author does not have a preference for overflow handling. The Reading System MAY render overflow content using its default method or a user preference, whichever is applicable.

So... say a book has no `rendition: flow` element. The reading system is therefore obligated to... use the auto value, which means it can choose any of the possible values? It could choose to `paginate`. It could choose to do `scrolled-continuous`. It could choose to do `scrolled-doc`. 

Is the original MUST statement thus untestable?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-01-07-epub#section3) on 2021-01-07

List of resolutions:

- ***[Resolution No. 2](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-01-07-epub#resolution2):  Change the language for `rendition:flow` to define the default is auto, but remove the MUST statement/ untestable assertion.***

<details><summary><i>View the transcript</i></summary>

### 3. The default value of `rendition:flow`.

_See github issue [#1313](https://github.com/w3c/epub-specs/issues/1313)._




**Wendy Reid:** this came up during testing.  
… `rendition:flow` is a property by which an author can determine if pagination style is scrolled, paginated, or auto.  
… if the RS supports it.  
… the issue is that the spec says that the default value is auto, and that the RS may choose to support only the default.  
… since auto means that its basically the RS's choice, does this make the assertion untestable?.  

**Brady Duga:** i think this is solved by saying that the default is auto, period..  
… with the understanding that if authors are not explicit, then they get default functionality.  

**Wendy Reid:** to clarify, W3C says any time we make a MUST statement, the assertion must be accompanied by a test.  
… or, the statement shouldn't be MUST.  
… if this results in something untestable, we're going to run into issues later.  

**Ben Schroeter:** or change MUST to SHOULD?.  

[...]

PR #1616, opened 2021-04-06
Handled the mega-issue 1612

### LINKED ISSUE #1612: Some apparent normative statements do not have all-caps control words
This is a megabug for apparent normative statements in the 3.3 spec that do not have all-caps control words like MUST. These look to me like they are *supposed* to be normative statements and thus should gain such a word. Conversely, anything here that is truly not supposed to be a normative statement should probably be rephrased to make clear that it's non-normative.

1. `The language identified in an hreflang attribute is not authoritative. Language information defined in a linked resource determines its language.` (also discussed in https://github.com/w3c/epub-specs/issues/1601). This seems like two normative statements — hreflang SHOULD NOT be used? and resource language MUST be used? https://w3c.github.io/epub-specs/epub33/rs/#conf-metadata-link
2. `Reading System [note: should be plural] do not have to use or present linked resources, even if they recognize the relationship defined in the rel attribute.` Same section as above.
3. `When the default value of the page-progression-direction attribute is specified, the Reading System can choose the rendering direction. The default value MUST be assumed when the attribute is not specified.` I think the second statement only makes sense if the first is also MUST. (Otherwise we're requiring RSes to use `default`, but not requiring that `default` do anything.) https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-spine
4. `In addition to the requirements in § 9. Processing Structural Semantics, a Reading System has to process the epub:type attribute as follows in XHTML Content Documents: It MUST ignore semantics expressed on the [HTML] head element or its descendants.` Should "has to" become MUST? If so, there are ~10 instances of "has to" in the spec that should change in sync ... but that suggests this wording may be intentional. https://w3c.github.io/epub-specs/epub33/rs/#sec-xhtml-structural-semantics
5. `The default global value is reflowable if no meta element carrying this property occurs in the metadata section [EPUB-33].` This feels as if it should be rephrased to address the RS: "If no meta element... the RS MUST apply a global default..." If so, there are ~4 intances of "default global value" in the spec that should change in sync. https://w3c.github.io/epub-specs/epub33/rs/#layout
6. `The rendition:page-spread-* properties take precedence over whatever value of the page-break-before property [CSSSnapshot] has been set for an XHTML Content Document.` Seems like it could have a MUST. https://w3c.github.io/epub-specs/epub33/rs/#page-spread
7. `For the rendition:flow-scrolled-continuous property, the scroll direction is defined relative to the block flow direction of the root element of the XHTML Content Document referenced by the itemref element [EPUB-33]. The scroll direction is vertical if the block flow direction is downward (top-to-bottom). It is horizontal if the block flow direction of the root element is rightward (left-to-right) or leftward (right-to-left).` This seems like maybe 2 or 3 normative statements. https://w3c.github.io/epub-specs/epub33/rs/#flow

iherman:
Quoting from #1612

> 1. `The language identified in an hreflang attribute is not authoritative. Language information defined in a linked resource determines its language.` (also discussed in https://github.com/w3c/epub-specs/issues/1601). This seems like two normative statements — hreflang SHOULD NOT be used? and resource language MUST be used? https://w3c.github.io/epub-specs/epub33/rs/#conf-metadata-link

I change the text, mostly using the way the HTML standard defines the similar attribute. Hopefully that is now clearer

> 2. `Reading System [note: should be plural] do not have to use or present linked resources, even if they recognize the relationship defined in the rel attribute.` Same section as above.

Set to plural.

> 3. `When the default value of the page-progression-direction attribute is specified, the Reading System can choose the rendering direction. The default value MUST be assumed when the attribute is not specified.` I think the second statement only makes sense if the first is also MUST. (Otherwise we're requiring RSes to use `default`, but not requiring that `default` do anything.) https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-spine

I have re-written the text, hopefully it is clearer and also accurate...

> 4. `In addition to the requirements in § 9. Processing Structural Semantics, a Reading System has to process the epub:type attribute as follows in XHTML Content Documents: It MUST ignore semantics expressed on the [HTML] head element or its descendants.` Should "has to" become MUST? If so, there are ~10 instances of "has to" in the spec that should change in sync ... but that suggests this wording may be intentional. https://w3c.github.io/epub-specs/epub33/rs/#sec-xhtml-structural-semantics

I am not sure. The real normative statement is in the bullet item. I leave this to @mattgarrish, this is a generic styling issue...

> 5. `The default global value is reflowable if no meta element carrying this property occurs in the metadata section [EPUB-33].` This feels as if it should be rephrased to address the RS: "If no meta element... the RS MUST apply a global default..." If so, there are ~4 intances of "default global value" in the spec that should change in sync. https://w3c.github.io/epub-specs/epub33/rs/#layout

I agree it is more readable. For now, I have re-written it in this section; if reviewers agree this is a better choice, the other instances will have to be changed, too. 

> 6. `The rendition:page-spread-* properties take precedence over whatever value of the page-break-before property [CSSSnapshot] has been set for an XHTML Content Document.` Seems like it could have a MUST. https://w3c.github.io/epub-specs/epub33/rs/#page-spread

Yep.

> 7. `For the rendition:flow-scrolled-continuous property, the scroll direction is defined relative to the block flow direction of the root element of the XHTML Content Document referenced by the itemref element [EPUB-33]. The scroll direction is vertical if the block flow direction is downward (top-to-bottom). It is horizontal if the block flow direction of the root element is rightward (left-to-right) or leftward (right-to-left).` This seems like maybe 2 or 3 normative statements. https://w3c.github.io/epub-specs/epub33/rs/#flow

Changed.

Fix #1612 

See:

    * [Preview](https://raw.githack.com/w3c/epub-specs/editorial/issue-1612/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/editorial/issue-1612/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -337,9 +337,10 @@
 					<dt id="conf-metadata-link">The <code>link</code> element</dt>
 					<dd>
 						<p>Retrieval of Remote Resources is OPTIONAL.</p>
-						<p>The language identified in an <code>hreflang</code> attribute is not authoritative. Language
-							information defined in a linked resource determines its language.</p>
-						<p>Reading System do not have to use or present linked resources, even if they recognize the
+						<p>The language identified in an <code>hreflang</code> is purely advisory.
+							Upon fetching the resource, a Reading System must only use the language information associated with the resource to determine its language, not metadata included in the link to the resource.
+						</p>
+						<p>Reading Systems do not have to use or present linked resources, even if they recognize the
 							relationship defined in the <code>rel</code> attribute.</p>
 						<p id="sec-linked-records">In the case of a <a href="https://www.w3.org/TR/epub-33/#record"
 								>linked metadata record</a> [[!EPUB-33]], Reading Systems MUST NOT skip processing the
@@ -409,9 +410,9 @@
 					specifically activates a hyperlink to such items). Reading Systems MAY also provide the option for
 					users to select whether non-linear content is skipped by default or not.</p>
 
-				<p>When the <code>default</code> value of the <code>page-progression-direction</code> attribute is
-					specified, the Reading System can choose the rendering direction. The <code>default</code> value
-					MUST be assumed when the attribute is not specified. In this case, the reading system SHOULD choose
+				<p>If the <code>page-progression-direction</code> is not specified, the value of <code>default</code>
+					MUST be assumed. When the value of the <code>page-progression-direction</code> is set to 
+					<code>default</code> (either explicitly or implicitly) the Reading System SHOULD choose
 					a default <code>page-progression-direction</code> value based on the first <code>language</code>
 					element.</p>
 
@@ -898,9 +899,9 @@
 				<section id="layout">
 					<h4>The <code>rendition:layout</code> Property</h4>
 
-					<p>The default global value is <code>reflowable</code> if no <code>meta</code> element carrying this
-						property occurs in the <a href="https://www.w3.org/TR/epub-33/#elemdef-opf-metadata"
-								><code>metadata</code> section</a> [[!EPUB-33]].</p>
+					<p>If no <code>meta</code> element carrying this
+						property occurs in the <a href="https://www.w3.org/TR/epub-33/#elemdef-opf-metadata"><code>metadata</code> 
+						section</a> [[!EPUB-33]], then the Reading System MUST assume the value <code>reflowable</code>.</p>
 
 					<p>When the <code>rendition:layout</code> property is set to <code>pre-paginated</code>, Reading
 						Systems MUST NOT include space between the adjacent content slots when rendering <a>Synthetic
@@ -982,7 +983,7 @@
 						reflowable content, and they only apply when the Reading System is creating Synthetic
 						Spreads.</p>
 
-					<p>The <code>rendition:page-spread-*</code> properties take precedence over whatever value of the <a
+					<p>The <code>rendition:page-spread-*</code> properties MUST take precedence over whatever value of the <a
 							href="https://www.w3.org/TR/CSS2/page.html#propdef-page-break-before"
 								><code>page-break-before</code> property</a> [[!CSSSnapshot]] has been set for an
 							<a>XHTML Content Document</a>.</p>
@@ -1614,11 +1615,11 @@
 					</dd>
 				</dl>
 
-				<p>For the <code>rendition:flow-scrolled-continuous</code> property, the scroll direction is defined
+				<p>For the <code>rendition:flow-scrolled-continuous</code> property, the scroll direction MUST be defined
 					relative to the block flow direction of the root element of the XHTML Content Document referenced by
 					the <a href="https://www.w3.org/TR/epub-33/#elemdef-spine-itemref"><code>itemref</code> element</a>
-					[[!EPUB-33]]. The scroll direction is vertical if the block flow direction is downward
-					(top-to-bottom). It is horizontal if the block flow direction of the root element is rightward
+					[[!EPUB-33]]. The scroll direction MUST be vertical if the block flow direction is downward
+					(top-to-bottom). It MUST be horizontal if the block flow direction of the root element is rightward
 					(left-to-right) or leftward (right-to-left).</p>
 			</section>
 
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -336,23 +336,32 @@
 
 					<dt id="conf-metadata-link">The <code>link</code> element</dt>
 					<dd>
-						<p>Retrieval and support of Remote Resources is OPTIONAL.</p>
+						<p>Retrieval and support of linked resources is OPTIONAL.</p>
 
-						<p>The language identified in an <code>hreflang</code> is purely advisory.
+						<!-- <p>Reading System may use support linked resources, even if they recognize the
+							relationship defined in the <code>rel</code> attribute.
+						</p> -->
+
+						<p>
+							The language identified in an <code>hreflang</code> is purely advisory.
 							Upon fetching the resource, a Reading System must only use the language information associated with the resource to determine its language, not the metadata included in the link to the resource.
 						</p>
-						<p id="sec-linked-records">In the case of a <a href="https://www.w3.org/TR/epub-33/#record"
-								>linked metadata record</a> [[!EPUB-33]], Reading Systems must not skip processing the
+
+						<p id="sec-linked-records">
+							In the case of a <a href="https://www.w3.org/TR/epub-33/#record">linked metadata 
+							record</a> [[!EPUB-33]], Reading Systems MUST NOT skip processing the
 							metadata expressed in the Package Document and only use the information expressed in the
 							record. Reading Systems may compile metadata from multiple linked records; they do not have
 							to select only one record.</p>
+
 						<p>When it comes to resolving discrepancies and conflicts between metadata expressed in the
-							Package Document and in linked metadata records, Reading Systems must use the document order
+							Package Document and in linked metadata records, Reading Systems MUST use the document order
 							of <code>link</code> elements in the Package Document to establish precedence (i.e.,
 							metadata in the first linked record encountered has the highest precedence and metadata in
 							the Package Document the lowest, regardless of whether the <code>link</code> elements occur
 							before, within or after the package metadata elements).</p>
-						<p>Reading Systems must ignore any instructions contained in linked resources related to the
+
+						<p>Reading Systems MUST ignore any instructions contained in linked resources related to the
 							layout and rendering of the EPUB Publication.</p>
 					</dd>
 
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -337,9 +337,9 @@
 					<dt id="conf-metadata-link">The <code>link</code> element</dt>
 					<dd>
 						<p>Retrieval of Remote Resources is OPTIONAL.</p>
-						<p>The language identified in an <code>hreflang</code> is purely advisory.
-							Upon fetching the resource, a Reading System must only use the language information associated with the resource to determine its language, not metadata included in the link to the resource.
-						</p>
+						<p>The language identified in an <code>hreflang</code> is purely advisory. Upon fetching the
+							resource, a Reading System must only use the language information associated with the
+							resource to determine its language, not metadata included in the link to the resource. </p>
 						<p>Reading Systems do not have to use or present linked resources, even if they recognize the
 							relationship defined in the <code>rel</code> attribute.</p>
 						<p id="sec-linked-records">In the case of a <a href="https://www.w3.org/TR/epub-33/#record"
@@ -411,9 +411,9 @@
 					users to select whether non-linear content is skipped by default or not.</p>
 
 				<p>If the <code>page-progression-direction</code> is not specified, the value of <code>default</code>
-					MUST be assumed. When the value of the <code>page-progression-direction</code> is set to 
-					<code>default</code> (either explicitly or implicitly) the Reading System SHOULD choose
-					a default <code>page-progression-direction</code> value based on the first <code>language</code>
+					MUST be assumed. When the value of the <code>page-progression-direction</code> is set to
+						<code>default</code> (either explicitly or implicitly) the Reading System SHOULD choose a
+					default <code>page-progression-direction</code> value based on the first <code>language</code>
 					element.</p>
 
 				<p>Reading Systems MUST ignore the page progression direction defined in <a href="#layout"
@@ -459,17 +459,11 @@
 					<section id="sec-xhtml-structural-semantics">
 						<h5>Structural Semantics</h5>
 
-						<p>In addition to the requirements in <a href="#sec-structural-semantics"></a>, a Reading System
-							has to process the <code>epub:type</code> attribute as follows in XHTML Content
-							Documents:</p>
-
-						<ul class="conformance-list">
-							<li>
-								<p id="confreq-rs-epubtype-head">It MUST ignore semantics expressed on the [[!HTML]] <a
-										href="https://html.spec.whatwg.org/multipage/semantics.html#the-head-element"
-											><code>head</code></a> element or its descendants.</p>
-							</li>
-						</ul>
+						<p id="confreq-rs-epubtype-head">In addition to the requirements in <a
+								href="#sec-structural-semantics"></a>, a Reading System MUST ignore semantics expressed
+							on the [[!HTML]] <a
+								href="https://html.spec.whatwg.org/multipage/semantics.html#the-head-element"
+									><code>head</code></a> element or its descendants.</p>
 					</section>
 
 					<section id="sec-xhtml-rdfa">
@@ -570,22 +564,22 @@
 					<section id="sec-xhtml-mathml">
 						<h3>MathML</h3>
 
-						<p>A Reading System has to support MathML embedded in XHTML Content Documents as follows:</p>
+						<p>To support MathML embedded in XHTML Content Documents, a Reading System:</p>
 
 						<ul class="conformance-list">
 							<li>
-								<p id="confreq-mathml-rs-behavior">It MUST be an input-compliant renderer for <a
+								<p id="confreq-mathml-rs-behavior">MUST be an input-compliant renderer for <a
 										href="https://www.w3.org/TR/MathML3/chapter3.html">Presentation MathML</a>, as
 									defined in the [[!MATHML3]] specification.</p>
 							</li>
 							<li>
-								<p id="confreq-mathml-rs-anno">It MAY support rendering of <a
+								<p id="confreq-mathml-rs-anno">MAY support rendering of <a
 										href="https://www.w3.org/TR/MathML3/chapter4.html">Content MathML</a> found in
 										<code>annotation-xml</code> elements.</p>
 							</li>
 							<li>
-								<p id="confreq-mathml-rs-render">If it has a <a>Viewport</a>, it MUST support visual
-									rendering of Presentation MathML.</p>
+								<p id="confreq-mathml-rs-render">MUST support visual rendering of Presentation MathML if
+									it has a <a>Viewport</a>.</p>
 							</li>
 						</ul>
 						<p class="note">Reading Systems may choose to use third-party libraries such as MathJax to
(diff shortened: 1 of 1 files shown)


epub33/rs/index.html
```diff
@@ -337,9 +337,10 @@ <h4>Metadata</h4>
 					<dt id="conf-metadata-link">The <code>link</code> element</dt>
 					<dd>
 						<p>Retrieval of Remote Resources is OPTIONAL.</p>
-						<p>The language identified in an <code>hreflang</code> attribute is not authoritative. Language
-							information defined in a linked resource determines its language.</p>
-						<p>Reading System do not have to use or present linked resources, even if they recognize the
+						<p>The language identified in an <code>hreflang</code> is purely advisory.
+							Upon fetching the resource, a Reading System must only use the language information associated with the resource to determine its language, not metadata included in the link to the resource.
+						</p>
+						<p>Reading Systems do not have to use or present linked resources, even if they recognize the
```
dlazin: Should this be MAY, or similar? (Reverse the negative, if so: "Reading Systems MAY use or present linked resources, if they recognize...")

epub33/rs/index.html
```diff
@@ -337,9 +337,10 @@ <h4>Metadata</h4>
 					<dt id="conf-metadata-link">The <code>link</code> element</dt>
 					<dd>
 						<p>Retrieval of Remote Resources is OPTIONAL.</p>
-						<p>The language identified in an <code>hreflang</code> attribute is not authoritative. Language
-							information defined in a linked resource determines its language.</p>
-						<p>Reading System do not have to use or present linked resources, even if they recognize the
+						<p>The language identified in an <code>hreflang</code> is purely advisory.
```
dlazin: Similar to my first comment on https://github.com/w3c/epub-specs/pull/1613#pullrequestreview-629071783, can we give more explicit guidance here about what an RS ought to do with this advisory language setting? That is, if it's not authoritative, should they ignore it entirely? May they use it if they choose? Should they only use it if no language is set at the document level?

epub33/rs/index.html
```diff
@@ -409,9 +410,9 @@ <h3>Spine</h3>
 					specifically activates a hyperlink to such items). Reading Systems MAY also provide the option for
 					users to select whether non-linear content is skipped by default or not.</p>
 
-				<p>When the <code>default</code> value of the <code>page-progression-direction</code> attribute is
-					specified, the Reading System can choose the rendering direction. The <code>default</code> value
-					MUST be assumed when the attribute is not specified. In this case, the reading system SHOULD choose
+				<p>If the <code>page-progression-direction</code> is not specified, the value of <code>default</code>
```
dlazin: Active instead of passive voice, please: "...not specified, the Reading System MUST assume a value of default."

epub33/rs/index.html
```diff
@@ -409,9 +410,9 @@ <h3>Spine</h3>
...
+					MUST be assumed. When the value of the <code>page-progression-direction</code> is set to 
+					<code>default</code> (either explicitly or implicitly) the Reading System SHOULD choose
```
```suggestion
					<code>default</code> (either explicitly or implicitly), the Reading System SHOULD choose
```

epub33/rs/index.html
```diff
@@ -898,9 +899,9 @@ <h3>Fixed-Layout Properties</h3>
 				<section id="layout">
 					<h4>The <code>rendition:layout</code> Property</h4>
 
-					<p>The default global value is <code>reflowable</code> if no <code>meta</code> element carrying this
-						property occurs in the <a href="https://www.w3.org/TR/epub-33/#elemdef-opf-metadata"
-								><code>metadata</code> section</a> [[!EPUB-33]].</p>
+					<p>If no <code>meta</code> element carrying this
+						property occurs in the <a href="https://www.w3.org/TR/epub-33/#elemdef-opf-metadata"><code>metadata</code> 
+						section</a> [[!EPUB-33]], then the Reading System MUST assume the value <code>reflowable</code>.</p>
```
```suggestion
						section</a> [[!EPUB-33]], the Reading System MUST assume the value <code>reflowable</code>.</p>
```

mattgarrish:
> I am not sure. The real normative statement is in the bullet item. I leave this to @mattgarrish, this is a generic styling issue...

We sort of endlessly circle around this issue. We had these as MUST in the past, but it's been noted over the years that this becomes confusing and seemingly contradictory with "should" and "may" bullet items, as you note.

When I was looking at the xml section in the last pr, I nearly changed the list to drop all the "its" at the start of each bullet, which would do away with this confusion. Not specifically thinking about this at the time, but only because it seemed a lot clearer.

For example, each conformance list could be simplified along these lines:

> To support X, Reading Systems:
> - MUST do Y
> - MAY do Z
> - SHOULD do A

Thoughts?

mattgarrish:
>> the RS MUST apply a global default...
>  if reviewers agree this is a better choice, the other instances will have to be changed, too.

Sigh, this reverts the change the group just recently asked to be made in https://github.com/w3c/epub-specs/pull/1562

(Not that I disagree with the way it was, as the reading system has to assume a value in the absence of one before it can move to taking whatever actions are required of it.)

iherman:
> > I am not sure. The real normative statement is in the bullet item. I leave this to @mattgarrish, this is a generic styling issue...
> 
> We sort of endlessly circle around this issue. We had these as MUST in the past, but it's been noted over the years that this becomes confusing and seemingly contradictory with "should" and "may" bullet items, as you note.
> 
> When I was looking at the xml section in the last pr, I nearly changed the list to drop all the "its" at the start of each bullet, which would do away with this confusion. Not specifically thinking about this at the time, but only because it seemed a lot clearer.
> 
> For example, each conformance list could be simplified along these lines:
> 
> > To support X, Reading Systems:
> > 
> > * MUST do Y
> > * MAY do Z
> > * SHOULD do A
> 
> Thoughts?

I as not sure how that would affect §4.1.1.2 (that is what we are talking about, aren't we?). Or are you worried about the series of separate sections and replace it with a list like above?

iherman:
> > > the RS MUST apply a global default...
> > > if reviewers agree this is a better choice, the other instances will have to be changed, too.
> 
> Sigh, this reverts the change the group just recently asked to be made in #1562
> 
> (Not that I disagree with the way it was, as the reading system has to assume a value in the absence of one before it can move to taking whatever actions are required of it.)

I must admit I am a little bit lost. I believe the rewrite, offered by @dlazin, did not change the semantics, it only made it more readable...

mattgarrish:
> I believe the rewrite, offered by @dlazin, did not change the semantics, it only made it more readable...

To me and you and Dan, yes, but this is what we had prior to #1562:

> The default value <code>reflowable</code> MUST be assumed by <a>EPUB Reading Systems</a> as the global value if no <code>meta</code> element carrying this property occurs in the <a href="https://www.w3.org/TR/epub-33/#elemdef-opf-metadata"><code>metadata</code> section</a> [[!EPUB-33]].

PR #1562 changed it only a few weeks ago based on a group decision to:

> The default global value is reflowable if no meta element carrying this property occurs in the metadata section [EPUB-33]. 

This is what you're now proposing:

> If no <code>meta</code> element carrying this property occurs in the <a href="https://www.w3.org/TR/epub-33/#elemdef-opf-metadata"><code>metadata</code> section</a> [[!EPUB-33]], then the Reading System MUST assume the value <code>reflowable</code>.

It's basically back where it was, which is problematic from a procedural point of view. Do we need to go back to the group again given that it was just resolved not to use this language?

mattgarrish:
> Or are you worried about the series of separate sections and replace it with a list like above?

I believe that was what was being asked for, yes.

> If so, there are ~10 instances of "has to" in the spec that should change in sync ... but that suggests this wording may be intentional.

I'd like to find a consistent answer to these lead-in sentences that avoids layering in a normative-like statement on top of them. Dropping the "its" is one possible way of doing that. I don't know if it works everywhere, though.

The other alternative is turning all the bullets into paragraphs, but that seems kind of cumbersome.

epub33/rs/index.html
dlazin: And I do like this change — recommend doing the same for all other occurrences of "default global value."

dlazin (commented):
> For example, each conformance list could be simplified along these lines:
> 
> > To support X, Reading Systems:
> > 
> > * MUST do Y
> > * MAY do Z
> > * SHOULD do A
> 
> Thoughts?

That sounds great!

> > > the RS MUST apply a global default...
> > > if reviewers agree this is a better choice, the other instances will have to be changed, too.
> 
> Sigh, this reverts the change the group just recently asked to be made in #1562
> 
> (Not that I disagree with the way it was, as the reading system has to assume a value in the absence of one before it can move to taking whatever actions are required of it.)

Ah, that's too bad. Still, I think the current text is good, although the problem that Dave raised in https://github.com/w3c/epub-specs/issues/1313 remains. I think this *is* testable. You need two elements, one with explicit `auto` and one with no attribute. If the behavior matches, the RS passes. It's up to the RS to determine what `auto` means; we are only requiring that `auto` is the default (which is very normal in HTML/CSS).

> It's basically back where it was, which is problematic from a procedural point of view. Do we need to go back to the group again given that it was just resolved not to use this language?

I agree that circular changes are a problem, so we probably should take this back to a meeting, but I hope my testing suggestion above helps us establish why this is OK. (Maybe we should reopen #1313 and discuss there.)

mattgarrish:
> I think this _is_ testable. You need two elements, one with explicit `auto` and one with no attribute.

Right, the problem is the resolution more than the proposal. It just makes it a blocker on the pull request. Maybe we can get this quickly resolved on this week's call, though.

epub33/rs/index.html
mattgarrish: Maybe go even simpler with: Support for linked resources is OPTIONAL. The rest is just filler.

Could probably be wrapped up with the statement about retrieval, too, since both are optional.

iherman:
> Ah, that's too bad. Still, I think the current text is good, although the problem that Dave raised in #1313 remains. I think this _is_ testable. You need two elements, one with explicit `auto` and one with no attribute. If the behavior matches, the RS passes. It's up to the RS to determine what `auto` means; we are only requiring that `auto` is the default (which is very normal in HTML/CSS).

Indeed. This sounds like a new element that might influence the [resolution](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-01-07-epub#resolution2) that led to #1562.

@dauwhe ? Would what @dlazin proposes answer your original question in #1313?

>  (Maybe we should reopen #1313 and discuss there.)

Procedurally that is probably the way to go.

iherman:
> I agree that circular changes are a problem, so we probably should take this back to a meeting, but I hope my testing suggestion above helps us establish why this is OK. (Maybe we should reopen #1313 and discuss there.)

@dlazin @mattgarrish I reversed this text for now, but left the new text in comments. This will make it possible to close this mega-issue-pr, and come back to this easily later.

epub33/rs/index.html
iherman: Here is what I did in the latest commit:

- The first sentence now says "Retrieval and support of Remote Resources is OPTIONAL."
- I turned all the MUST-s and MAY-s into must-s and may-s (because the whole section is on an optional feature)

That being said, the proposal of @dlazin would read as:

> Reading Systems may use or present linked resources, if they recognize the relationship defined in the <code>rel</code> attribute.

which is slightly different; it does not say what the RS has to do if they do not recognize the relationship. I opted for simply commenting out (for now) the whole sentence, also as a consequence of making everything optional...

epub33/rs/index.html
iherman: I do not think we can say _anything_ here. The [HTML spec](https://html.spec.whatwg.org/multipage/links.html#attr-hyperlink-hreflang) does not say much more about it either. I would think we should not try to go there...

epub33/rs/index.html
iherman: yep. Resolving it in the next commit.

epub33/rs/index.html
iherman: yep

epub33/rs/index.html
iherman: I would wait for #1619 to be merged and then look at it again... Do not want to do any global change, it would lead to merge mess.

iherman:
A procedural thing...

@dlazin and I agreed (on slack) that using such a mega-issue and mega-pr was a bad idea; it leads to confusion. I would propose that:

- we merge #1619 into this branch
- we keep the original text that came from #1313
- we merge this PR and close #1612

And then @dlazin would look at the issues again. If some of them remain, raise individual items and possible PR-s. 

Otherwise I am really worried that we will get into a big, tangled editorial mess...

@mattgarrish ?

epub33/rs/index.html
mattgarrish: Not sure I agree with this change. Wouldn't this mean that there are no normative requirements in media overlays because it's an optional feature? Supporting the feature and how you have to support it if you do are separate concerns.

mattgarrish:
> using such a mega-issue and mega-pr was a bad idea

Yes, atomize! Grouping issues is always a bad idea unless they're typos or other no-brainers.

epub33/rs/index.html
dlazin: +1 to Matt's point. (I think he's referring to the "I turned all the MUST-s and MAY-s into must-s and may-s (because the whole section is on an optional feature)". I've seen other instances of MUSTs inside optional features.

No opinion on "Retrieval and support of Remote Resources is OPTIONAL." vs "Reading Systems may use or present linked resources, if they recognize the relationship defined in the rel attribute." I don't know enough about the implications.

epub33/rs/index.html
dlazin: Can we link the word "advisory" to that section in the HTML spec? I think that helps say "same as here".

epub33/rs/index.html
dlazin: OK, it's merged now.

epub33/rs/index.html
mattgarrish: > Retrieval and support of Remote Resources is OPTIONAL.

This should be "linked resources". Linked resources are not necessarily remote resources (i.e., a linked resource can be located inside the container).

> Reading Systems may use or present linked resources

This statement is likely incomplete which is why I'm good with plain old "support". The reading system can do whatever it wants with linked resources. If it takes a metadata record and sends it to a downstream process, for example, is that really "using" it?

> if they recognize the relationship defined in the rel attribute

This kind of stands to reason. Also, by only talking about the relationships a reading system recognizes it seems to require another statement about the ones it doesn't. In either case, using the linked resources is optional. We can let devs sort out the ones they want to use.

epub33/rs/index.html
iherman: I would not think so. Formally speaking _this_ `hreflang` is not _that_ `hreflang`, it is only inspired by it... A normative reference to the HTML spec would be misleading imho.

iherman:
Note that #1313 has been reopened.

epub33/rs/index.html
iherman: Something happened in the previous changes and commits, because the sentence on "Reading System may use or present..." sentence did not appear in it... And I think that is all right: the first sentence is, sort of, covering everything, and the reference to `rel` does not seem to add information. As @mattgarrish said, its usage stands to reason.

I have made some changes and, to avoid misunderstandings, this is how the text reads in the last iteration:

> **The link element**
>
> Retrieval and support of linked resources is OPTIONAL.
>
> The language identified in an `hreflang` is purely advisory. Upon fetching the resource, a Reading System must only use the language information associated with the resource to determine its language, not the metadata included in the link to the resource.
>
> In the case of a [linked metadata record](https://www.w3.org/TR/epub-33/#record) [EPUB-33], Reading Systems MUST NOT skip processing the metadata expressed in the Package Document and only use the information expressed in the record. Reading Systems may compile metadata from multiple linked records; they do not have to select only one record.
>
> When it comes to resolving discrepancies and conflicts between metadata expressed in the Package Document and in linked metadata records, Reading Systems MUST use the document order of `link` elements in the Package Document to establish precedence (i.e., metadata in the first linked record encountered has the highest precedence and metadata in the Package Document the lowest, regardless of whether the `link` elements occur before, within or after the package metadata elements).
>
 > Reading Systems MUST ignore any instructions contained in linked resources related to the layout and rendering of the EPUB Publication.

Are we o.k. now?

mattgarrish: > a Reading System must only use the language information associated with the resource

Shouldn't that one also be caps, or is that covered in the other issue?

iherman: I guess it was a CAPS and that started this part of the issue:-)

I actually wonder whether we should not remove that part of the statement altogether. If the PR #1613 is merged, there is a separate section that covers all this language issues. The essence of that PR is that these types of statements can go.

epub33/rs/index.html
```diff
@@ -336,22 +336,31 @@ <h4>Metadata</h4>
...
-								>linked metadata record</a> [[!EPUB-33]], Reading Systems MUST NOT skip processing the
+						<p>Retrieval and support of linked resources is OPTIONAL.</p>
+
+						<!-- <p>Reading System may use support linked resources, even if they recognize the
+							relationship defined in the <code>rel</code> attribute.
+						</p> -->
+
+						<p>
+							The language identified in an <code>hreflang</code> is purely advisory.
+							Upon fetching the resource, a Reading System must only use the language information associated with the resource to determine its language, not the metadata included in the link to the resource.
```
dlazin, line 347: Should this "must" be all-caps? I suspect it is intentionally not (already covered in a MUST elsewhere?), but if so, the MUST NOT on line 352 seems like it should be lowercase as well; I think that is talking about the same thing. Or we could just make them both all-caps so that at least no one has to second-guess whether it's normative, even if it is repetitive.

epub33/rs/index.html
```diff
@@ -569,22 +571,22 @@ <h4>HTML Deviations and Constraints</h4>
...
-								<p id="confreq-mathml-rs-anno">It MAY support rendering of <a
+								<p id="confreq-mathml-rs-anno">MAY support rendering of <a
 										href="https://www.w3.org/TR/MathML3/chapter4.html">Content MathML</a> found in
 										<code>annotation-xml</code> elements.</p>
 							</li>
 							<li>
-								<p id="confreq-mathml-rs-render">If it has a <a>Viewport</a>, it MUST support visual
-									rendering of Presentation MathML.</p>
+								<p id="confreq-mathml-rs-render">MUST support visual rendering of Presentation MathML if
+									it has a <a>Viewport</a>.</p>
```
dlazin, line 589: Suggest "MUST, if it has a Viewport, support..." (not particularly important here, but important below, and this ensures consistency with the next change).

epub33/rs/index.html
```diff
@@ -634,33 +636,33 @@ <h3>SVG Content Documents</h3>
...
-						<p id="confreq-svg-rs-css">If it has a <a>Viewport</a>, it MUST support the visual rendering of
-							SVG using CSS as defined in <a href="https://www.w3.org/TR/SVG/styling.html">Styling</a>
-							[[!SVG]], and it SHOULD support all properties defined in the <a
-								href="https://www.w3.org/TR/SVG/propidx.html">Property Index</a> [[!SVG]]. In the case
-							of embedded SVG, it MUST also conform to the constraints defined in <a
-								href="#sec-xhtml-svg-css"></a>.</p>
+						<p id="confreq-svg-rs-css">MUST support the visual rendering of SVG using CSS as defined in <a
+								href="https://www.w3.org/TR/SVG/styling.html">Styling</a> [[!SVG]] and it SHOULD support
+							all properties defined in the <a href="https://www.w3.org/TR/SVG/propidx.html">Property
+								Index</a> [[!SVG]] if it has a <a>Viewport</a>. In the case of embedded SVG, a Reading
```
dlazin, line 656: Suggest "MUST, if it has a ViewPort, support..."

Otherwise the contextual viewport part comes quite late; easier to understand by moving it up.

dlazin (approved):

epub33/rs/index.html
iherman: My proposal was to remove that sentence altogether, in view of the other PR on the I18N section...

epub33/rs/index.html
iherman: @dlazin was it part of the original Mega issue? If not, can we have this as a separate issue? Otherwise we are making our lives impossible...

dlazin: It was not. This was a comment I wrote on Matt's PR that rephrased these sections (removing the "It" references), but I didn't get that comment in before he merged.

I'll send a separate PR for these edits after you merge. So you can ignore this comment.

epub33/rs/index.html
dlazin: Ignore; I will send a separate PR.

iherman:
@mattgarrish any reasons not to merge this mega-pr, and then handle the remaining things in separate issues if needed? We are creating a huge bowl of spaghetti...

mattgarrish:
Ya, the normative keyword issues looks resolved. The rest is getting hard to parse, so let's merge.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-04-15-epub#section2) on 2021-04-15

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-04-15-epub#resolution1):  Merge PR 1616***

### 2. language around rendition:flow 

_See github issue [#1313](https://github.com/w3c/epub-specs/issues/1313)._




_See github pull request [#1616](https://github.com/w3c/epub-specs/pull/1616)._




> *Dave Cramer:* See [Relevant comment within the PR](https://github.com/w3c/epub-specs/pull/1616#pullrequestreview-629109802)

**Dave Cramer:** I was concerned that we say RS MUST use auto as a default value, but auto means "RS, do whatever you want"  
… not sure that it was testable  
… but Dan thinks it is testable  
… so maybe we should just undo that resolution  
… if it is testable i'm fine with restoring that requirement  

**Brady Duga:** agree. Not worth spending much time on.  
… and agree with dlazin's testing method for it  

**Dave Cramer:** yes, we just need to see that things are consistent between having explicit auto value and "implied" auto  

> **Proposed resolution: Merge PR 116** *(Wendy Reid)*

> *Brady Duga:* +1

> *Dave Cramer:* +1

> *Wendy Reid:* +1

> *Matthew Chan:* +1

> *Dan Lazin:* +1

> *Masakazu Kitahara:* +1

> *Toshiaki Koike:* +1

> *Marisa DeMeglio:* +1

> *Shinya Takami (高見真也):* +1

> ***Resolution #1: Merge PR 1616***

</details>

Issue #1608, opened 2021-04-04
Reading system accessibility


mattgarrish:
The appendix on reading system accessibility was removed from the accessibility specification during the iso process, but the reading system specification still references it:

> It SHOULD meet the accessibility requirements for Reading Systems in [EPUB-A11Y-10].

Should we restore this appendix, or how else can we address this?

GeorgeKerscher:
IMO it is important to keep Reading System accessibility in the appendix. People often confuse document accessibility, software accessibility, and the importance of having both.

murata2makoto:
I prefer a separate specification.   This is what WAI does.

mattgarrish:
> I prefer a separate specification.

That's not in scope or realistic for this revision. We need something we can use now.

Given that [there are only two bullets](http://idpf.org/epub/a11y/#app-ac-rs-conf) for reading systems, could we integrate these directly in the reading system specification rather than farm them out to the accessibility specification?

avneeshsingh:
Separate specifications for reading systems accessibility is difficult to fit in the current schedule. We need to wrap up before November 2021 (CR timeline). 
As per my understanding, the purpose of mentioning reading systems accessibility in the appendix of EPUB accessibility is to make it clear that EPUB accessibility spec is focused on the content, and one need to focus separately on accessibility of reading systems and to point to couple of resources for improving reading systems accessibility. Currently these to resources are accessibility testing protocol on epubtest.org and W3C user agent accessibility guidelines.
I am good with adding such a brief information for reading systems accessibility in appendix of EPUB accessibility 1.1.

murata2makoto:
I then like Matt's suggestion.  Incorporate the two bullets in the reading system specification.   Adding normative requirements that are not in the said scope of the accessibility  specification is not the right way to go forward, IMHO.

mattgarrish:
It actually looks like we only have one requirement: that reading systems should conform to UAAG.

I don't believe we can require reading systems to pass a loosely defined test suite no matter which specification we put it in. I'm pretty sure we'll run afoul of W3C's requirements for normative references if we try to cite them normatively.

An informative note is likely all we could use to point devs to them.

GeorgeKerscher:
[...]

PR #1620, opened 2021-04-07
User interaction may overwrite CSS

### LINKED ISSUE #1605: Letting users override CSS "as desired" is unclear
https://w3c.github.io/epub-specs/epub33/rs/#confreq-css-user-styles says `It SHOULD allow users to override the EPUB Creator's style sheets as desired.`

First, `SHOULD` seems too aggressive here. If an RS doesn't want to, isn't that an OK decision? I recommend dropping to `MAY`.

Secondly, "as desired" suggests letting users upload their own stylesheets, or at least suggests nothing in particular. I think we mean "should let users change font size, dark mode, and maybe a couple other things like line-spacing or paragraph justification."

At most, that handful of options should (maybe) be `SHOULD`, and most other properties `SHOULD NOT` be user-overrideable; it's way too hard for a publisher to ensure their book won't break if users can override arbitrary (or just different) properties in different RSes.

Can we consider dropping to `MAY`, and rephrase "as desired" to be more specific, possibly adding a list of `SHOULD NOT`s?

iherman:
Fix #1605

The only change is using the sentence in https://github.com/w3c/epub-specs/issues/1605#issuecomment-814300267

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -717,8 +717,7 @@
 									><code>style</code> attributes</a>.</p>
 					</li>
 					<li>
-						<p id="confreq-css-user-styles">It SHOULD allow users to override the EPUB Creator's style
-							sheets as desired.</p>
+						<p id="confreq-css-user-styles">It MAY override parts of the EPUB Creator's style sheet as a result of user interaction.</p>
 					</li>
 				</ul>
 


mattgarrish:
I agree with the initial lack of clarity about what this bullet is requiring of reading systems, but we're losing the 
recommendation that reading systems provide user control over styling in the rewrite. From an accessibility perspective, that's a very important thing to have.

But maybe if we have to write an (informative) accessibility section in this document per #1608 it can go there.

I'll take a skim through the html spec and see if they touch on this at all.

dlazin (approved):
Thanks!

mattgarrish:
I couldn't find anything in the HTML or CSS specs. Only UAAG goes into this a bit, but even it's pretty vague. Short of mandating the control a user has to have, the language we had was about as good as you can get.

I'm going to take this up when I get to the accessibility section instead.

PR #1624, opened 2021-04-09
Add dedicated accessibility sections

### LINKED ISSUE #1608: Reading system accessibility

mattgarrish:
As noted in #1608 and discussed on the last accessibility task force call, this pull request adds new accessibility sections to both the core and reading system specifications.

For the core spec, the recommendation to adhere to the accessibility specification remains with only some additional context added.

The reading system section probably needs the most review. I've marked it as informative and included a general list of areas of concern based largely off what I was finding in the test suite. I've also referenced the test suite and UAAG without making normative requirements for them - the section is informative at this point unlike on the authoring side.

I expect this will still need massaging, so please suggest any changes you think would help improve this.

Fixes #1608

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1624.html" title="Last updated on Apr 16, 2021, 5:28 PM UTC (63cb2c3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1624/372f2f5...63cb2c3.html" title="Last updated on Apr 16, 2021, 5:28 PM UTC (63cb2c3)">Diff</a>


mattgarrish:
- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/feature/add-a11y-sections/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/main/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/feature/add-a11y-sections/epub33/rs/index.html)

clapierre (approved):
Looks great Matt!

epub33/core/index.html
```diff
@@ -8071,12 +8067,37 @@ <h3>Navigation Document Overlays</h3>
 				</div>
 			</section>
 		</section>
-		<!-- <section id="sec-privacy">
-			<h2>Privacy</h2>
-			<div class="ednote">
-				<p>The working group will address issues related to privacy of EPUB Publications in a future draft.</p>
+		<section id="sec-accessibility">
+			<h2>Accessibility</h2>
```
iherman, line 8076: Shouldn't this section be also informative, i.e., not normative?

iherman (approved):

epub33/core/index.html
mattgarrish: Not unless we move the recommendation out of it again. It feels kind of strange to recommend accessibility in one place but talk about it another, but we could point to the recommendation where it was.

iherman: You may be right. The only contrast that may come to the fore is that this section very much reads like the, e.g., security and privacy sections, and those are indeed informative...

The only worry I have is how would we enforce this if it is normative. But we may want to leave this open for now and see that later.

mattgarrish: > The only worry I have is how would we enforce this if it is normative.

Enforcing accessibility is better left to legislation, but just because you can't validate every accessibility check doesn't mean we shouldn't recommend accessibility. What havoc would it wreak on scripting if we have to test what scripts are doing?

We also have the Ace validator to show that we do, in fact, have an implementation of a checker for machine-identifiable issues.

But, looking at the RS spec, the scripting considerations are separate from scripting support, so maybe it isn't such a bad thing. Let me try a rewrite or three.

iherman: +1

epub33/core/index.html
```diff
@@ -8071,12 +8071,39 @@ <h3>Navigation Document Overlays</h3>
...
-				<p>The working group will address issues related to privacy of EPUB Publications in a future draft.</p>
+		<section id="sec-accessibility" class="informative">
+			<h2>Accessibility</h2>
+
+			<p>EPUB 3 is built upon the Open Web Platform expressly so that it can leverage the structure, semantics
+				and, by extension, accessibility built into its underlying technologies.</p>
+
+			<p>The requirements and practices for creating accessible web content have already been documented in the
+				W3C's <a href="https://www.w3.org/TR/wcag/">Web Content Accessibility Guidelines (WCAG)</a> [[!WCAG]].
+				These guidelines also form the basis for defining accessibility in EPUB Publications.</p>
```
iherman, line 8083: I presume it should be [[WCAG]] now; I guess that is why respec is a bit unhappy and puts some sort of an underling underneath...

iherman (approved):

epub33/core/index.html
mattgarrish: Ya, I should probably try rendering the document again when slapping an informative label on something... ;)

iherman:
@avneeshsingh @dauwhe @shiestyle @wareid can you comment on this PR? Would be nice to merge this if it gets your approval...

shiestyle (approved):

epub33/core/index.html
```diff
@@ -8071,13 +8071,40 @@ <h3>Navigation Document Overlays</h3>
...
+					Accessibility standard</a>. A benefit of following this recommendation is that it helps ensure that
+				EPUB Publications meet the accessibility requirements legislated in jurisdictions around the world,
+				ensuring EPUB Creators are not locked out of potential markets.</p>
+
+			<p><a>EPUB Creators</a> are encouraged to look beyond legal imperatives and treat accessibility as a
+				requirement for all their content. The more accessible that EPUB Publications are, the greater the
+				potential audience for them.</p>
+
+			<div class="note">
+				<p>Acessibility requirements are not integrated in this specification in order to allow them to adapt
```
wareid: Typo "Accessibility"

wareid (approved):
Found one typo but otherwise this looks great.

epub33/core/index.html
iherman: Thanks! Handled

Issue #1614, opened 2021-04-06
What should a RS do with CSS direction property?


iherman:
[§3.3.2](http://localhost:8001/LocalData/github/EPUB/epub-specs/epub33/core/#sec-css-req) says:

>* It MAY include any CSS properties, with the following exceptions:
>    * It MUST NOT use the direction property [CSS-Writing-Modes-3]. Use the [HTML] dir attribute to set the inline base direction.
>    * It MUST NOT use the unicode-bidi property [CSS-Writing-Modes-3]. Use [HTML] bdo elements and dir attributes to control bidirectionality.

I am not 100% sure why this restriction is present. However, two questions do arise:

1. What happens if I use SVG (which does not have a `dir` attribute, only language, or a `bdo` element)?
2. What should be the behavior of a RS be if it finds one of these two properties (in HTML). Is it a "_MUST_ ignore" ?

iherman:
Whatever the decision for the RSS system is, I believe it is better to add this [§4.3](https://w3c.github.io/epub-specs/epub33/rs/#sec-css) of the RS spec.

dauwhe:
> I am not 100% sure why this restriction is present. 

CSS [advises people not to use `direction`](https://www.w3.org/TR/css-writing-modes-3/#direction) with HTML documents: 

> Because HTML UAs can turn off CSS styling, we recommend HTML authors to use the HTML dir attribute and <bdo> element to ensure correct bidirectional layout in the absence of a style sheet. Authors should not use direction in HTML documents.

Basically, fantasai told us never to use `direction` :)

iherman:
But... do we allow RS-s to turn off CSS styling? Ie, is this relevant for us? (O.k., there are older RS-s that do not do CSS at all I guess...)

In any case, I believe this argument should be in the text because, at this moment, it really looks like a fairly ad-hoc constraint. Also, if I read the CSS text, it is a 'recommend' and not 'RECOMMEND'. In other words, we have made it a MUST NOT, which means RS-s have to check this, epubcheck have to test it, RS-s have to do something if it is violated, etc...

Wouldn't it be possible to change the text so that:

1. we essentially take over (or link to) the CSS text pointing out that it is safer to use the HTML features (note, this is not possible in SVG, so this advise is only relevant for HTML)
2. all this is put into an informative note rather than a normative restriction

dauwhe:
[...]

PR #1627, opened 2021-04-11
Added the SVG reference and some notes for CSS restrictions

iherman:
This is a PR for #1614 (and for the corresponding [WG resolution](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-04-08-epub#resolution2)).

- Added a reference to the SVG `direction` attribute (to be used instead of the CSS `direction` property). I had to rephrase the bullet item texts to make it a bit more readable
- Added a note for the background of these restrictions
- Added a note in the RS document that RS-s can "do what they damn well please" :-)

One discrepancy I encountered is that there seems to be no equivalent to a `bidi` element in SVG2. Ie, the restriction on the  `unicode-bidi` CSS property does not apply to SVG2.

(I may have misunderstood something. Maybe @fantasai can help out on this.)

* For Preview  on the EPUB 3.3 Reading Systems:
    * [Preview](https://raw.githack.com/w3c/epub-specs/css-dir-property-issue-1614/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/css-dir-property-issue-1614/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1627.html" title="Last updated on Apr 16, 2021, 5:29 PM UTC (55a0733)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1627/edb9e7b...55a0733.html" title="Last updated on Apr 16, 2021, 5:29 PM UTC (55a0733)">Diff</a>


mattgarrish:
> One discrepancy I encountered is that there seems to be no equivalent to a `bidi` element in SVG2. Ie, the restriction on the `unicode-bidi` CSS property does not apply to SVG2.

The SVG `unicode-bidi` attribute maps to the CSS property, so why would it be appropriate to use CSS instead of the attribute?

That HTML also has a `bdo` element doesn't appear of consequence.

mattgarrish:
I think we should state more generally that authoring restrictions are not applicable to reading systems, as the CSS properties are only one case of restrictions we impose on HTML, MathML, SVG, etc.

I was thinking after the call we should probably have a statement more like the following somewhere:

> The definition of EPUB Content Documents includes various authoring restrictions to optimize the cross-compatibility of content (e.g., prohibiting CSS for setting language and direction). **Unless stated otherwise in this specification, Reading Systems MAY support these restricted features.**

(Bolding's sort of a polite way of screaming to do what you please.)

Maybe this could float at the start of the content documents section, or maybe it would be better to incorporate into a section on authoring restrictions?

iherman:
> > One discrepancy I encountered is that there seems to be no equivalent to a `bidi` element in SVG2. Ie, the restriction on the `unicode-bidi` CSS property does not apply to SVG2.
> 
> The SVG `unicode-bidi` attribute maps to the CSS property, so why would it be appropriate to use CSS instead of the attribute?
> 

There is something strange happening that I do not understand. I see no _definition_ for the `unicode-bidi` property in SVG. It is used in an example and in the text in [the section on direction](https://www.w3.org/TR/SVG2/text.html#DirectionProperty) but it all refers to the CSS spec only.

Is our requirement for the SVG case something like:

>   It MUST NOT use the unicode-bidi property [CSS-Writing-Modes-3] for [SVG] Content Documents. Use the [SVG] unicode-bidi and direction attributes to control bidirectionality.

Knowing that I _cannot_ put a normative reference to an SVG reference to the `unicode-bidi` attribute but only to the CSS property definition? This is what bothered me...

I may miss something fundamental...

iherman:
> I think we should state more generally that authoring restrictions are not applicable to reading systems, as the CSS properties are only one case of restrictions we impose on HTML, MathML, SVG, etc.
> 
> I was thinking after the call we should probably have a statement more like the following somewhere:
> 
> > The definition of EPUB Content Documents includes various authoring restrictions to optimize the cross-compatibility of content (e.g., prohibiting CSS for setting language and direction). **Unless stated otherwise in this specification, Reading Systems MAY support these restricted features.**
> 
> (Bolding's sort of a polite way of screaming to do what you please.)
> 
> Maybe this could float at the start of the content documents section, or maybe it would be better to incorporate into a section on authoring restrictions?

I actually like that, is it more future proof. It may be put into [§4](http://localhost:8001/LocalData/github/EPUB/epub-specs/epub33/rs/#sec-contentdocs) _before_ §4.1.

iherman:
@mattgarrish, I have adopted your proposed change in https://github.com/w3c/epub-specs/pull/1627/commits/dd58c5037a0194b59b6348939c5b8e5c937a72f3. However, I have not used the boldface; I do not think we use that type of typesetting trick elsewhere in the document, so it looked a bit out of place…

mattgarrish:
> I see no _definition_ for the `unicode-bidi` property in SVG. It is used in an example and in the text in [the section on direction](https://www.w3.org/TR/SVG2/text.html#DirectionProperty) but it all refers to the CSS spec only.

Right, the formal process of using css properties as attributes is defined under "[presentation attributes](https://www.w3.org/TR/SVG2/styling.html#PresentationAttributes)".

For all the listed CSS properties there are equivalent svg attributes. The attribute are exact mappings to the css properties, which is why I don't believe we need to worry about `bdo`. That's for html which doesn't have this equivalency.

That's why they just talk about the "property" but then use it as an attribute. You could always link to the presentation attribute section, I suppose, as someone can find their way from there to the definition.

mattgarrish:
> I have not used the boldface; I do not think we use that type of typesetting trick elsewhere in the document

Oh, come on, take a walk on the wild side... :)

([Never say never](https://www.w3.org/TR/epub-33/#sec-docs-assoc-style)... although I thought we had more.)

iherman:
On the SVG bi-di stuff: here is what I have just committed for that item:

> It MUST NOT use the <a href="https://www.w3.org/TR/css3-writing-modes/#unicode-bidi"><code>unicode-bidi</code> property</a> [[!CSS-Writing-Modes-3]] for [[!SVG]] Content Documents. Use the [[!SVG]] <a href="https://www.w3.org/TR/SVG2/styling.html#PresentationAttributes">presentation attribute alternative</a> for <code>unicode-bidi</code> and the <a href="https://www.w3.org/TR/SVG2/text.html#DirectionProperty"><code>direction</code> attribute</a> to control bidirectionality.

How does that sound?

iherman:
> Oh, come on, take a walk on the wild side... :)

O.k., you win :-)

mattgarrish (approved):

iherman:
@mattgarrish I have finalized the stuff (essentially rolling back my last commit and modifying a little bit the note you provided).

iherman:
@dauwhe @wareid could you look at this? It would be time to merge (also trying to avoid complex merge conflicts...)

wareid (approved):

PR #1634, opened 2021-04-12
Modification in the RS Appendix A.

### LINKED ISSUE #716: Issues with epubReadingSystem object
From Daniel Glazman:

The "layout-changes" feature covers far too much scope. The layout for a given element can be drastically changed modifying a class, any attribute, the style attribute, it can even be modified inserting elements before that element or directly tweaking the stylesheets.

In that light, chances are reading systems will always reply true to this if scripting is allowed. If they always reply true, querying that feature is useless.

Similarly, the dom-manipulation feature will always be true if the spine-scripting feature is true because if you can't manipulate the DOM from script, scripting is pretty useless...

(Repetition of issues #691 and #692 elided.)

All in all, |hasFeatures()| seems to me completely useless. My recommendation is simple: drop it. But wait, in that case, |epubReadingSystem| exists only to give us |name|, |version| and |layoutStyle|. Is it really worth it then? Do we really need the bloat of an extra global JS object for that and don't we already have stuff for that in the |navigator| object anyway?

### LINKED ISSUE #1420: Usage of WebIDL, Appendix A in RS.

iherman:
This is answering the [WG resolution](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-04-09-epub#resolution1) on `epubReadingSystem`, ie, on issues #1610, #1420, and #716.

Remarks:

1. After all, I shied away of turning  WebIDL into Javascript. The formalism stemming from WebIDL permeates the whole section, and it would need a complete rewrite, let alone the fact that the Javascript part would have to be commented extensively to make sense of it. Instead, I have added a separate paragraph trying to make clear that this is _not_ required to be implemented by _every_ Web browser. Maybe we can get away with this...
2. I have changed the title of the section, removing the 'Javascript' from it. WebIDL is supposed to be language dependent...
3. I also realized that the note underneath the WebIDL portion refers to the, now defunct, Worker specification. That specification has been incorporated into the HTML spec, so I updated the pointers (which, incidentally, means that the reference to Workers is now gone from the reference list)

An extra remark on (1). The text I added has the following text:

> Reading System that, conceptually, act as the core rendering engines of Web browsers providing a scripting environment based on the [DOM] specification.

I believe this statement is true, but I also believe it would be better to make this statement to the RS document somewhere else. After all, we all _think_ in these terms, but it is nowhere stated...

Fix #716 
Fix #1420

See:

    * [Preview](https://raw.githack.com/w3c/epub-specs/rs-webidl-section/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/rs-webidl-section/epub33/rs/index.html)


mattgarrish (approved):

epub33/rs/index.html
```diff
@@ -838,6 +839,10 @@ <h3>Scripting</h3>
 					</li>
 				</ul>
 
+				<p class="ednote">Note that the definition of <code>epubReadingSystem</code> is currently markes as "at
```
bduga: s/markes/marked/

epub33/rs/index.html
```diff
@@ -838,6 +839,10 @@ <h3>Scripting</h3>
...
+					risk". If, in the final version of this document, the object becomes non-normative, then each "MUST"
```
bduga, line 844: This is not a comment on this line, but I guess I can't comment on unchanged lines? This interface is weird.

Anyway, line 833 above has JavaScript navigator. Should we drop "JavaScript", since you dropped it below?

epub33/rs/index.html
```diff
@@ -1818,11 +1823,21 @@ <h4>Event Model Considerations</h4>
...
+
+			<p class="issue atrisk" title="Are there enough implementations?">Although this section is
+					<em>normative</em>, it is not clear whether implementations of this interface are widespread enough
+				to warrant this status. If it turns out not to be the case, this section will be marked as
+					<em>non-normative</em> in the final version of the specification.</p>
 
 			<section id="app-ers-idl">
 				<h3>Interface Definition</h3>
 
+				<p> This interface definition uses the [[!WEBIDL]] notation, because it is aimed Reading System that,
```
bduga: s/aimed/aimed at/
s/System/Systems/

epub33/rs/index.html
```diff
@@ -1818,11 +1823,21 @@ <h4>Event Model Considerations</h4>
...
+					conceptually, act as the core rendering engines of Web browsers providing a scripting environment
```
bduga: Oddly phrased. Maybe something more like "act as the core rendering engines of epub content in Web browsers"? Still a little clumsy.

mattgarrish: This also looks like a note. I've made it one in the last commit and tried to rephrase.

epub33/rs/index.html
mattgarrish: I would think we should link/reference the definition in the HTML spec.

mattgarrish: I've removed the references to javascript and the navigator object from the reading system requirement as this is explained in the object definition. I've also added a link on the navigator object reference where it's introduced in the definition.

bduga (approved):
Nice! Looks good to me.

rdeltour (approved):

iherman:
@dauwhe @wareid I think we should have your approval before merging this. 

The main reason is that, after all, I shied away from removing WebIDL (at least for now) because I realized it would make things way more complicated, alas! (Javascript is really poor in this respect). What I hope is that the separate note may help us to overcome potential problems with browser vendors (I may check this with the director in advance. Should I?). In this respect, I did not follow the WG resolution by the letter...

dauwhe:
My only concern right now is with language.

>  It MUST implement the `epubReadingSystem` extension object  defined in...

Is "extension object" the correct phrase here? I know we are extending an object, but the phrase doesn't seem to be commonly used. I don't know if this is any better: 
 
>  It MUST extend the `navigator` [HTML] object with the `epubReadingSystem` interface, defined in...

iherman:
> My only concern right now is with language.
> 
> > It MUST implement the `epubReadingSystem` extension object  defined in...
> 
> Is "extension object" the correct phrase here? I know we are extending an object, but the phrase doesn't seem to be commonly used. I don't know if this is any better:
> 
> > It MUST extend the `navigator` [HTML] object with the `epubReadingSystem` interface, defined in...

You are probably right. I will do the change (not immediately, I am sitting at a virtual conference...).

iherman:
@dauwhe I have made the change. I have also moved the 'disclaimer' on browser implementations one level higher up in the sections, to make it relevant to the full of Appendix A, and not only the strict WebIDL stuff.

mattgarrish:
> It MUST extend the `navigator` [HTML] object with the `epubReadingSystem` interface defined in...

Minor quibble, but are we extending the interface or the object?

You get an object from it, but the HTML spec defines an interface. I found a couple of other specs that extend the "interface":

- https://www.w3.org/TR/gamepad/#navigator-interface-extension
- https://www.w3.org/TR/pointerevents2/#extensions-to-the-navigator-interface

And, of course, one that talks about extending the "object" is https://www.w3.org/TR/battery-status/#the-navigator-interface

Which camp should we join? 😄

dauwhe:
> Minor quibble, but are we extending the interface or the object?

I have no idea! :) 

The section heading in the HTML spec is for the Navigator object. It is then described as an attribute of the window interface. 

I wonder if they even know! I'll see if I can ask on twitter... lots of spec-writing folks follow me.

iherman:
I do not really speak WebIDL, but I do speak languages with interfaces like TypeScript. From that point of view the language is fine. The Navigator object implements a bunch of interfaces; what we say is that the HTML Navigator object must add one more interface to the lot. That sounds pretty fine to me...

iherman:
@dauwhe, in case you contact spec-people, there is another question for which we have no answer: what is the equivalent to the Navigator object for an SVG content file?

I think we may have a hole here in the spec, though we should leave this to another PR.

@danielweck, as an SVG geek, do you know this from the top of your head? How does a script running in SVG gets hold of the top level environment?

iherman:
Actually, maybe the language is not the right one (to answer @mattgarrish's question). What we do is:

```
partial interface Navigator {
```

which means that we extend the Navigator _interface_. Ie, my remark "The Navigator object implements a bunch of interfaces;" is not the way WebIDL looks at things; it gradually extends the interface.

Ie, to day its says 

> It MUST extend the Navigator object [HTML]

and it should probably say

> It MUST extend the Navigator interface [HTML]

---

I note, however, that the HTML spec is a [little bit sloppy](https://html.spec.whatwg.org/multipage/system-state.html#the-navigator-object). The text says:

> The navigator attribute of the Window interface must return an instance of the Navigator interface

Ie, the capitalized version is the interface, and the non-capitalized version is for the object implementing the interface. However, the title of the section says:

> 8.9.1 The Navigator object

So... it is sloppy all around. Nobody will jump at us if we leave it as is, but I can change object -> interface...

wareid (approved):

iherman:
> Ie, to day its says
> 
> > It MUST extend the Navigator object [HTML]
> 
> and it should probably say
> 
> > It MUST extend the Navigator interface [HTML]

alternatively, we can also say

> It MUST extend the navigator object [HTML]

considering the naming convention, whereby 'Navigator' is the interface, and 'navigator' is the object.

iherman:
Just to move things I have made the change 'Navigator' -> 'navigator'. We can roll that back, but maybe we can merge this and move on.

(I believe handling SVG should be treated separately.)

wareid (approved):

PR #1644, opened 2021-04-19
remove custom attributes section from core spec

mattgarrish:
I noticed while opening epubcheck issues that only the custom attributes section from the reading system spec was removed.

This PR also removes their use from authoring.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1644.html" title="Last updated on Apr 19, 2021, 12:53 PM UTC (b9b673e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1644/5bc4f65...b9b673e.html" title="Last updated on Apr 19, 2021, 12:53 PM UTC (b9b673e)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3676,25 +3676,6 @@ Spine:
 						<p>Use of the <code>trigger</code> element is <a href="#deprecated">deprecated</a>. Refer to its
 							definition in [[!EPUBContentDocs-301]] for usage information.</p>
 					</section>
-
-					<section id="sec-xhtml-custom-attributes">
-						<h5>Custom Attributes</h5>
-
-						<p><a>EPUB Creators</a> MAY specify <a
-								href="https://www.w3.org/TR/epub-rs-33/#sec-xhtml-custom-attributes">custom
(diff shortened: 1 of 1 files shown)


iherman (approved):
Shame on me, I should have done it when I removed it from RS.

But: there is also a reference to custom attributes in the Multiple Rendition document. It says

>  If both rendition selection attributes conformant to this specification and custom attributes are defined, the latter SHOULD be ignored.

at the end of [§4.5](https://w3c.github.io/epub-specs/epub33/multi-rend/#rendition-selection-proc-model). Maybe that should also be removed, to keep things in sync...

mattgarrish:
We should probably leave that section as-is. Custom attributes are also allowed in the container.xml file (not the same as in HTML) so if we remove them it could be more problematic.

I believe at least one producer in Japan(?) was making multiple rendition publications using their own attributes (for internal purposes) before we developed that document. That's why we had to be careful not to require only the model we developed.

iherman:
@mattgarrish I believe we can merge this by now. It is the missing part of an earlier resolution...

PR #1647, opened 2021-04-24
Group non characters in character restrictions

mattgarrish:
As I mentioned on the call yesterday, would it make sense to group all the non characters together like in this PR so it's clearer what we're restricting?

There's some overlap with the subsequent character restrictions, but that's hardly consequential.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1647.html" title="Last updated on Apr 25, 2021, 5:58 PM UTC (164b4af)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1647/5b6ea1a...164b4af.html" title="Last updated on Apr 25, 2021, 5:58 PM UTC (164b4af)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5795,14 +5795,23 @@ Spine:
 									<p>Private Use Area (<code>U+E000 … U+F8FF</code>)</p>
 								</li>
 								<li>
-									<p>Non characters in Arabic Presentation Forms-A (<code>U+FDD0 … U+FDEF</code>)</p>
+									<p>All Unicode Non Characters, specifically:</p>
+									<ul>
+										<li>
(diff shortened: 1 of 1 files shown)


iherman (approved):
I am fine with the regrouping. I leave it to others to decide whether the right Unicode intervals are included...

dauwhe (approved):

xfq:
This PR looks good to me.

The i18n WG noticed that restrictions on Tags and Variation Selectors Supplement (U+E0000 … U+E0FFF) may prevent the use of [emoji tag sequence (ETS)](https://unicode.org/reports/tr51/#def_emoji_tag_sequence). If authors are not expected to use them in file and path names, there is no problem, tho.

Issue #1390, opened 2020-10-30
Prefixed CSS properties' tables are unclear


iherman:
I am not sure I get the intention of the tables in [§5.3.3. Prefixed properties](https://w3c.github.io/epub-specs/epub33/core/#sec-css-prefixed). I _suspect_ the message is:

- in earlier versions of EPUB 3 prefixed CSS properties were used because there were no final CSS properties for the specific functionalities
- these functionalities now exists in CSS 3
- new contents should use the standard functionalities, but the old, prefixed properties remain valid (and the RS conformance says that RS-s must continue to accept and implement those)
- here are the tables how to map the old properties, and their allowed values, to the final CSS properties and their values

If that is the case, I do not really understand 

- what "Prior EPUB Mapping" really means
- what does "no change" means under the heading "Mapping to CSS Text 3"
- it is also not really clear from the table what the corresponding property is (although, in most of the cases, this is probably the one that one gets dropping the `-epub-` prefix, but not always)

I would look for a table somewhat of the sort:

| Prefixed property | Value | Standard CSS property | Value |
| :--------------- | :----| :------------------- | :--- |
| `-epub-text-orientaton` | `upright` | `text-orientation` | `upright` |
| etc |
| `-epub-text-combine-horizontal` | `none` | `text-combine-upright` | `none` |
| etc |
| `-epub-hyphens` | `none` \| `manual` \| `auto` | `hyphens` | `none` \| `manual` \| `auto`
| etc |

I realize the table includes some items that are considered to be deprecated; imho these should be part of a separate table.

Does this make sense?

dauwhe:
This is all my fault. I spent an immense amount of time trying to trace the history of these properties and values. The CSS Writing modes spec has changed A LOT over the years. The epub-prefixed properties were taken from an earlier version of writing modes, but by the time 3.1/3.2 came along there had been significant changes. So it seemed important to document this, and provide some guidance to content authors who should move away from the prefixed properties. 

I coordinated quite a bit with Murata-san and others who knew what properties were actually supported by Japanese reading systems. The deprecated values here were not supported, and were not in the EBPAJ documents, and so we could deprecate them without affecting any existing content. 

So, for example, one possible value of `-epub-text-orientation` is `mixed`. EPUB continues to support this value, even though the best practice would be to use the unprefixed `text-orientation: mixed`. But there was also an older value allowed in EPUB: `-epub-text-orientation: vertical-right`. This has been deprecated. 

[...]

Issue #1389, opened 2020-10-30
CSS Writing Modes reference


iherman:
[§5.3.3.1 CSS Writing Modes](https://w3c.github.io/epub-specs/epub33/core/#sec-css-prefixed-writing-modes) refers to [CSS-Writing-Modes-3-20151215](https://w3c.github.io/epub-specs/epub33/core/#bib-css-writing-modes-3-20151215) which is now outdated. I would think we should refer to [css-writing-modes-3](https://www.w3.org/TR/css-writing-modes-3/) to refer to the latest version...

iherman:
Actually, the same comment applies to other CSS modules in the same section: undated references should be used to ensure referring to the latest version.

I realize there that some mapping may go wrong in newer versions of the respective CSS modules, which may require future versions to of the EPUB document to be upated. But the current approach "baking in" outdated CSS document version is probably not a good idea (and I suspect will also be rejected by future document reviews.)

dauwhe:
Given the dramatic changes that CSS Writing Modes went through, we had to point to a very specific version or else nothing would make sense. We can of course re-evaluate that now that CSS Writing Modes is a **W3C RECOMMENDATION**!!!!

murata2makoto:
[§5.3.3.1 CSS Writing Modes](https://w3c.github.io/epub-specs/epub33/core/#sec-css-prefixed-writing-modes) is for the -epub- prefix.  I believe that many EPUB publications in Japan used this prefix while CW was a WD.  These EPUB publications will not go away.  I thus oppose to any changes.

toshiakikoike:
Probably the most influential is,
 "-epub-text-combine"

Especially for those produced around 2011-2013, it is specified as follows.

```
.tcy {
  -epub-text-combine:horizontal;
}
```
This is expected to be treated as 
 "text-combine-upright: all;"

iherman:
> [§5.3.3.1 CSS Writing Modes](https://w3c.github.io/epub-specs/epub33/core/#sec-css-prefixed-writing-modes) is for the -epub- prefix. I believe that many EPUB publications in Japan used this prefix while CW was a WD. These EPUB publications will not go away. I thus oppose to any changes.

@murata2makoto nobody proposed any changes, let alone removal, of those features. But the current text is, at least as far as I am concerned, extremely confusing.

murata2makoto:
@iherman 

[...]

PR #1655, opened 2021-04-29
Complete rewrite of section on prefixed CSS properties

dauwhe:
This attempts to address #1389 and #1390.

***
<a href="https://pr-preview.s3.amazonaws.com/dauwhe/epub-specs/pull/1655.html" title="Last updated on May 3, 2021, 7:42 PM UTC (37d5191)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1655/1a90e58...dauwhe:37d5191.html" title="Last updated on May 3, 2021, 7:42 PM UTC (37d5191)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -4092,11 +4092,12 @@ Spine:
 					<section id="sec-css-prefixed-writing-modes">
 						<h5>CSS Writing Modes</h5>
 
-						<p>The following lists the <code>-epub-</code> prefixed properties for
-							[[CSS-Writing-Modes-3]]. .</p>
+						<p>This section describes the <code>-epub-</code> prefixed properties for
+							[[CSS-Writing-Modes-3]].</p>
 
 						<section id="sec-css-prefixed-writing-modes-text-orientation">
 							<h6><code>-epub-text-orientation</code></h6>
+							<p>This is a prefixed version of the CSS <code>text-orientation</code> property.</p>
 							<table class="def propdef">
 								<tbody>
 									<tr>
@@ -4145,6 +4146,8 @@ Spine:
 						
 						<section id="sec-css-prefixed-writing-modes-writing-mode">
 							<h6><code>-epub-writing-mode</code></h6>
+							<p>This is a prefixed version of the CSS <code>writing-mode</code> property, 
+							with the same syntax and behavior.</p>
 							<table class="def propdef">
 								<tbody>
 									<tr>
@@ -4164,11 +4167,9 @@ Spine:
 						</section>
 						
 						<section id="sec-css-prefixed-writing-modes-text-combine">
-							<h6><code>-epub-text-combine-horizontal</code> and <code>-epub-text-combine</code>(Deprecated)</h6>
+							<h6><code>-epub-text-combine-horizontal</code> and <code>-epub-text-combine</code>(deprecated)</h6>
 							
-							<p>The <code>-epub-text-combine</code> property is deprecated. Authors should use 
-							<code>text-combine-upright</code>, although <code>-epub-text-combine-horizontal</code>
-							is still supported.</p>
+							<p>These properties no longer exist in CSS, having been replaced with <code>text-combine-upright</code>.</p>
 							<table class="def propdef">
 								<tbody>
 									<tr>
@@ -4185,7 +4186,57 @@ Spine:
 									</tr>
 								</tbody>
 							</table>
-		
+							<p>The <code>-epub-text-combine</code> property is deprecated. Authors should use 
+							<code>text-combine-upright</code>, although <code>-epub-text-combine-horizontal</code>
+							is still supported.</p>
+							<table class="def propdef">
+								<tbody>
+									<tr>
+										<th>Name: </th>
+										<td>
+											<dfn>-epub-text-combine</dfn> (deprecated)
+										</td>
+									</tr>
+									<tr class="value">
+										<th>Value:</th>
+										<td>
+											none | horizontal | horizontal &lt;number&gt; 
+										</td>
+									</tr>
+								</tbody>
+							</table>
+							<p>The following table shows how to map from the prefixed versions to the current CSS version.</p>
+							<table class="data">
+								<thead>
+									<tr>
+										<th>Prefixed version</th>
+										<th>Version to be used</th>
+									</tr>
+								</thead>
+								<tbody>
+									<tr>
+										<td><code>-epub-text-combine-horizontal: none</code></td>
+										<td><code>text-combine-upright: none</code></td>
+									</tr>
+									<tr>
+										<td><code>-epub-text-combine-horizontal: all</code></td>
+										<td><code>text-combine-upright: all</code></td>
+									</tr>
+									<tr>
+										<td><code>-epub-text-combine: none</code></td>
+										<td><code>text-combine-upright: none</code></td>
+									</tr>
+									<tr>
+										<td><code>-epub-text-combine: horizontal</code></td>
+										<td><code>text-combine-upright: all</code></td>
+									</tr>
+									<tr>
+										<td><code>-epub-text-combine: horizontal &lt;number&gt;</code></td>
+										<td>Error</td>
+									</tr>
+								</tbody>
+							
+							</table>
 						</section>
 					</section>
 
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -4096,7 +4096,7 @@ Spine:
 							[[CSS-Writing-Modes-3]].</p>
 
 						<section id="sec-css-prefixed-writing-modes-text-orientation">
-							<h6><code>-epub-text-orientation</code></h6>
+							<h6>The <code>-epub-text-orientation</code> Property</h6>
 							<p>This is a prefixed version of the CSS <code>text-orientation</code> property.</p>
 							<table class="def propdef">
 								<tbody>
@@ -4145,7 +4145,7 @@ Spine:
 						</section>
 						
 						<section id="sec-css-prefixed-writing-modes-writing-mode">
-							<h6><code>-epub-writing-mode</code></h6>
+							<h6>The <code>-epub-writing-mode</code> Property</h6>
 							<p>This is a prefixed version of the CSS <code>writing-mode</code> property, 
 							with the same syntax and behavior.</p>
 							<table class="def propdef">
@@ -4167,7 +4167,7 @@ Spine:
 						</section>
 						
 						<section id="sec-css-prefixed-writing-modes-text-combine">
-							<h6><code>-epub-text-combine-horizontal</code> and <code>-epub-text-combine</code>(deprecated)</h6>
+							<h6>The <code>-epub-text-combine-horizontal</code> and <code>-epub-text-combine</code>(deprecated) Properties</h6>
 							
 							<p>These properties no longer exist in CSS, having been replaced with <code>text-combine-upright</code>.</p>
 							<table class="def propdef">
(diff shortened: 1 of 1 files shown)


shiestyle (approved):

epub33/core/index.html
```diff
@@ -4092,348 +4092,376 @@ <h4>Prefixed Properties</h4>
 					<section id="sec-css-prefixed-writing-modes">
 						<h5>CSS Writing Modes</h5>
```
iherman, line 4098: The intro paragraph of §3.3.3 is clear for a reader who is already familiar with EPUB, more exactly with earlier versions of EPUB 3. For a newcomer, who gets into the world of EPUB via EPUB 3.3 this paragraph, the whole section is really cryptic and comes out of the blue...

I think some explanatory words on how these properties came about, that this specification includes this section to ensure backward compatibility, etc, would be necessary. It should also suggest (there are already references to this) that new authors for new content should probably ignore the whole section altogether. Etc.

epub33/core/index.html
```diff
@@ -4092,348 +4092,376 @@ <h4>Prefixed Properties</h4>
...
-								</tr>
-							</tbody>
-						</table>
+						<p>This section describes the <code>-epub-</code> prefixed properties for
+							[[CSS-Writing-Modes-3]].</p>
+
+						<section id="sec-css-prefixed-writing-modes-text-orientation">
+							<h6>The <code>-epub-text-orientation</code> Property</h6>
+							<p>This is a prefixed version of the CSS <code>text-orientation</code> property.</p>
+							<table class="def propdef">
```
iherman: For this table (and the others, or as some sort of an introductory remark) it should be made clear that, unless otherwise noted, the semantics of a property value for the prefixed version is identical to what CSS specifies (is that correct?). This is stated in, say, the `-epub-writing-mode` property, but not everywhere. In the absence of this the spec is unclear what the meaning of, say, the `vertical-rl` value means

epub33/core/index.html
```diff
@@ -4092,348 +4092,376 @@ <h4>Prefixed Properties</h4>
...
+											normal | keep-all | break-all
+										</td>
+									</tr>
+								</tbody>
+							</table>
+						</section>
+						<section id="sec-css-prefixed-text-text-transform">
+						<h6>The <code>text-transform</code> Property</h6>
+						<p>This is a prefixed value for the <code>text-transform</code> property.</p>
+							<table class="def propdef">
```
iherman: This is an exception... it is _not_ a prefixed property! I guess it should be treated differently and how exactly it relates to the CSS equivalent...

epub33/core/index.html
dauwhe: This is unchanged from EPUB 3.2. I will add an introductory sentence or two, but is the existing text good enough for suggesting these shouldn't be used today? 

> EPUB Creators are advised to use CSS-native solutions for the removed properties where and when they are available.

> EPUB Creators currently using these prefixed properties are advised to move to unprefixed versions as soon as support allows...

epub33/core/index.html
dauwhe: Correct. The section heading describes the text-transform property. The text describes it as a prefixed value of that property. The table shows what the CSS equivalent is. I'm not sure what you're asking me to change.

epub33/core/index.html
bduga: I wouldn't be opposed to an explicit SHOULD NOT, though what you have seems pretty clear to me.

epub33/core/index.html
dauwhe: That's pretty much the very definition of a prefixed property. I have tried to be very clear about any differences in behavior from current CSS specs. Should I construct larger tables for each property, to explicitly identify that `-epub-writing-mode: vertical-rl` means exactly the same thing as `writing-mode: vertical-rl`?

dauwhe:
@iherman Aside from the details, do you think this a better approach than what is currently in the spec? Does this address the concerns you raised in the original issue?

epub33/core/index.html
```diff
@@ -4092,348 +4097,376 @@ <h4>Prefixed Properties</h4>
...
+									</tr>
+									<tr class="value">
+										<th>Value:</th>
+										<td>
+											none | all
+										</td>
+									</tr>
+								</tbody>
+							</table>
+							<p>The <code>-epub-text-combine</code> property is deprecated. Authors should use
```
bduga: This paragraph is confusing. First, it seems odd to have it between the two properties, since it refers to both of them, so perhaps moving it up would help. But even then, it is confusing - one property is deprecated, but the other is still supported? But also don't use it, and it doesn't exist in CSS. The table below clarifies things, but the "still supported" seems misleading.

epub33/core/index.html
dauwhe: I've added a sentence to the opening para saying that prefixed properties behave the same as their unprefixed equivalent unless otherwise noted. Does this address your concerns?

epub33/core/index.html
```diff
@@ -4092,348 +4097,376 @@ <h4>Prefixed Properties</h4>
...
+						</section>
+						<section id="sec-css-prefixed-text-epub-line-break">
+						<h6>The <code>-epub-line-break</code> Property</h6>
+						<p>This is a prefixed version of the <code>line-break</code> property.</p>
+							<table class="def propdef">
+								<tbody>
+									<tr>
+										<th>Name: </th>
+										<td>
+											<dfn>-epub-linebreak</dfn>
```
bduga: s/-epub-linebreak/-epub-line-break/

epub33/core/index.html
```diff
@@ -4092,348 +4097,376 @@ <h4>Prefixed Properties</h4>
...
+									</tr>
+									<tr class="value">
+										<th>Value:</th>
+										<td>
+											none | manual | auto | all
+										</td>
+									</tr>
+								</tbody>
+							</table>
+							<p class="note">The <code>auto</code> value is no longer supported in CSS.</p>
```
bduga: s/auto/all/

epub33/core/index.html
```diff
@@ -4092,348 +4097,376 @@ <h4>Prefixed Properties</h4>
...
+									<tr>
+										<th>Name: </th>
+										<td>
+											<dfn>-epub-linebreak</dfn>
+										</td>
+									</tr>
+									<tr class="value">
+										<th>Value:</th>
+										<td>
+											auto | loose | normal | strict
```
bduga: This list isn't exhaustive, which makes sense since 'anywhere' wasn't a value of -epub-line-break. But does that mean it is an error to have -epub-line-break: anywhere? What should reading systems do in that case? Maybe it makes sense to mention that we allow all the values allowed on the unprefixed property?

Also applies to other properties (eg -epub-word-break)

epub33/core/index.html
dauwhe: This is so messy. 3.0.1 deprecated `-epub-text-combine` and added `-epub-text-combine-horizontal`. So the latter is not-yet-deprecated and the former is deprecated and you shouldn't use any of them. So reading systems MAY support `-epub-text-combine` and MUST support `-epub-text-combine-horizontal`?

epub33/core/index.html
iherman: I agree with Brady, but I am not sure what SHOULD NOT means in terms of epubcheck and whether that would be acceptable...

epub33/core/index.html
iherman: > I've added a sentence to the opening para saying that prefixed properties behave the same as their unprefixed equivalent unless otherwise noted. Does this address your concerns?

Yes! Thx

epub33/core/index.html
iherman: Ouch, sorry, my bad. It indeed says 'prefixed _value_' (emphasis mine).

(If I want to be picky: the section heading says 'Prefixed Properties', and this section refers to a prefixed value... But that is something we can let go, I guess...)

iherman:
> @iherman Aside from the details, do you think this a better approach than what is currently in the spec? Does this address the concerns you raised in the original issue?

Absolutely. I should have thanked you in my previous comments... :-)

epub33/core/index.html
dauwhe: Yeah. I think our general principle is that we should allow reading systems to try to do the right thing, even when authors are behaving strangely. So if you see `-epub-line-break: anywhere` the RS can do `line-break: anywhere`.

epub33/core/index.html
bduga: Bleh. I was reading this from an authoring perspective (weird, I know), but this also applies to the RS behavior. I guess the "authors should use" misdirected me. So "authors SHOULD use text-combine-upright, and Reading Systems MUST support content authored with -epub-text-combine-horizontal. If you use -epub-text-combine you are totes hosed, loser." Or maybe remove this paragraph altogether, note the deprecation, and rely on the transformation table (below).

dauwhe: I'm going to delete the middle para and go with this at the start:

```html
<p>These are prefixed versions of <code>text-combine-upright</code>, 
although <code>-epub-text-combine</code> is deprecated. 
See the table below for how values of both properties translate to unprefixed CSS.</p>
```

epub33/core/index.html
dauwhe: SHOULD NOT for prefixed properties in general? That would trigger an EPUBCheck warning, which would break backward compatibility. 

I also think these are relatively easy to continue supporting--some are even built into browsers.

epub33/core/index.html
dauwhe: Oops! Fixed.

epub33/core/index.html
dauwhe: Fixed.

epub33/core/index.html
dauwhe: Added a note to that effect at the beginning. There's also the converse question: values that were used when the prefixed properties were created, but that are no longer part of CSS (like `hyphens: all`).

epub33/core/index.html
iherman: > That would trigger an EPUBCheck warning, which would break backward compatibility.

That is what I was worried about. We should not change it then.

iherman:
I do like the new introduction in §3.3.3, and it addresses my concern. Thanks. 

One minor issue, though: the reference to EPUBContentDocs-301 is currently normative. I wonder whether this is right: we do not (normatively) rely on that document any more, do we? If that is indeed the case, this should be a non-normative reference.

(Editorially, in the respec source and its new syntax, it shuld be `[[?EPUBContentDOcs-301]]`)

mattgarrish:
My only question is whether it might make sense to move all these prefixed properties to an appendix?

mattgarrish:
> the reference to EPUBContentDocs-301 is currently normative

Are cautions normative or informative? I would have expected the reference to be treated informally.

It could also be getting picked up as normative elsewhere in the document (e.g., deprecated elements). I believe respec will always insert a single normative entry unless all the references are informative.

iherman:
> > the reference to EPUBContentDocs-301 is currently normative
> 
> Are cautions normative or informative? I would have expected the reference to be treated informally.
> 
> It could also be getting picked up as normative elsewhere in the document (e.g., deprecated elements). I believe respec will always insert a single normative entry unless all the references are informative.

That is correct. If the same reference appears elsewhere in the document, I would think they should be all explicitly informative...

mattgarrish:
> If the same reference appears elsewhere in the document, I would think they should be all explicitly informative.

That wouldn't work, as it's pretty common to cite normative specs in informative sections. The context determines if the reference is informative or normative. Otherwise we could never reference html in an informative section, for example, as we'd run afoul of the referencing rules.

In this case, while the reference here looks informative, when we deprecate elements/attributes/properties and point authors back to their old definitions, it doesn't make the references informative.

But it does bring into question what the point of overriding references is when informative references in normative sections aren't differentiated in any way. As far as users can tell, the informative reference will appear normative.

mattgarrish:
> Are cautions normative or informative?

Part of the problem here is there is no "caution" class.  It appears we have a couple of other instances of this that aren't formatting as anything, too.

The "warning" class outputs a red warning box, but it's oddly also not documented in the respec manual.

iherman:
> > Are cautions normative or informative?
> 
> Part of the problem here is there is no "caution" class. It appears we have a couple of other instances of this that aren't formatting as anything, too.
> 
> The "warning" class outputs a red warning box, but it's oddly also not documented in the respec manual.

My apologies for having started this issue bound to this PR, @dauwhe. I think the review of informative/normative references should be reviewed at some point, but this is not related to this particular issue...

iherman (approved):

dauwhe:
@mattgarrish @iherman Based on Matt's comment I moved this to an appendix.

mattgarrish (approved):

PR #1521, opened 2021-02-24
add early draft of explainer for TAG, including security and privacy questionnaire

dauwhe:
This is just the very beginning of a long process... I started trying to answer the questions in the security and privacy questionnaire.

***
<a href="https://pr-preview.s3.amazonaws.com/dauwhe/epub-specs/pull/1521.html" title="Last updated on Apr 28, 2021, 2:14 PM UTC (b163f76)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1521/f430175...dauwhe:b163f76.html" title="Last updated on Apr 28, 2021, 2:14 PM UTC (b163f76)">Diff</a>

PR #1656, opened 2021-04-29
Security and Privacy Questionnaire Rough Draft

dauwhe:
This adds two documents: The skeleton of an explainer for the TAG, and an in-progress security and privacy questionnaire. 

See also comments in #1521. I have tried to address many of those, but I think there are still some differences of opinion.


wareid (approved):

shiestyle (approved):

epub33/explainers/EPUB-33-security-privacy.md
```diff
@@ -0,0 +1,103 @@
...
+
+### 14. How does this specification work in the context of a user agent’s Private Browsing or "incognito" mode?
+
+This seems a bit specific to browsers... 
+
+
+### 15. Does this specification have a "Security Considerations" and "Privacy Considerations" section?
+
+Yes, but it feels quite inadequate. 
+
```
iherman, line 96: What does this "feels quite inadequate" means? That it is inadequate for this spec? That it is inadequate because we did not do our job? etc.

I would simply remove that statement. The answer is "yes".

epub33/explainers/EPUB-33-security-privacy.md
```diff
@@ -0,0 +1,103 @@
...
+## Links to specifications
+
+* [EPUB 3.3 Overview](https://w3c.github.io/epub-specs/epub33/overview/), a non-normative overview of EPUB 3.3
+* [EPUB 3.3 Core](https://w3c.github.io/epub-specs/epub33/core/), the specification of the file format
+* [EPUB 3.3 Reading Systems](https://w3c.github.io/epub-specs/epub33/rs/), the specification for user agents, known as EPUB Reading Systems. 
+* [EPUB Accessibility 1.1](https://w3c.github.io/epub-specs/epub33/a11y/)
+* [EPUB Accessibility Techniques 1.1](https://w3c.github.io/epub-specs/epub33/a11y-tech/)
+
+
```
iherman, line 13: Maybe worth indicating which documents are rec-track.

epub33/explainers/EPUB-33-security-privacy.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Introduction
+
+EPUB files are presented to end users by EPUB Reading Systems, which can be constructed as web applications but are more likely to be native apps. The EPUB specifications define the file format in great detail, but generally give user agents wide latitude, subject mostly to the requirements of the embedded technologies, such as HTML, SVG, XML, and CSS. 
+
```
iherman, line 18: I think it is worth mentioning that the majority of current reading systems are based on existing Browser Engines like Blink or WebKit. Which also means that the security and privacy features (or bugs:-) are often inherited from those engines and not inherent to the reading system itself.

epub33/explainers/EPUB-33-security-privacy.md
```diff
@@ -0,0 +1,103 @@
...
+EPUB Reading Systems typically collect information on the reading habits of their users. In some cases this information is necessary to provide a good user experience; in other cases the information is collected for other users—the reading system company or the publisher of the book. 
+
+1. Tracking reading position. You really do want the reading system to remember that you just started chapter 5. 
+2. Remembering highlights, notes, bookmarks. 
+3. Remembering user preferences around fonts, visual themes (including dark/light), text justification, etc. 
+4. Recording progress through a book and whether a book was finished. 
+5. Some reading systems track every user reading session, including the time of day, the duration, how many pages were read, what book, the user’s IP address if a web-based reader, etc. 
+
+
```
iherman, line 35: Worth noting that most of these informations are usually "sand-boxed": it is available for the same user reading the same book from different devices (ie, when I read on an iPad and then I read the same book on my iPhone, I have access to those) but they are _not_ available _outside_ that single book. They are definitely not available to "other parties" (as stated in the question).

epub33/explainers/EPUB-33-security-privacy.md
```diff
@@ -0,0 +1,103 @@
...
+The EPUB specification does not mention reading system features like bookmarks or highlights. There is no requirement to remember reading state. 
+
+### 3. How does this specification deal with personal information or personally-identifiable information or information derived thereof?
+
+The EPUB specifications do not mention personal information.
+
+### 4. How does this specification deal with sensitive information?
+
+Merely the fact that you are reading a particular book can be sensitive information. Again, the specification offers no guidance to reading systems on this matter. 
+
```
iherman, line 48: Worth noting that the EPUB content itself does not contain, per specification, any sensitive information.

epub33/explainers/EPUB-33-security-privacy.md
```diff
@@ -0,0 +1,103 @@
...
+### 5. Does this specification introduce new state for an origin that persists across browsing sessions?
+
+Reading systems, in order to meet user expectations, must retain some information across browsing sessions, for example to remember reading location. The spec has a non-normative section saying that reading systems should treat each EPUB as a unique origin, but does not say anything about most reading system behaviors.
+
```
iherman, line 52: Hm. We may have an issue whether the 'origin' stuff is normative or not. 

The (normative) [EPUB content section](https://w3c.github.io/epub-specs/epub33/core/#sec-scripted-content) says: "Reading Systems are required to behave as though a unique origin [URL] has been assigned to each EPUB Publication", but the reading system specs notes this in the section on security and privacy only. It is not _really_ clear whether this requirement is normative or not. Do I miss something, @mattgarrish 

(In my personal view that statement should be normative)

epub33/explainers/EPUB-33-security-privacy.md
```diff
@@ -0,0 +1,103 @@
...
+### 6. What information from the underlying platform, e.g. configuration data, is exposed by this specification to an origin?
+
+Many EPUB reading systems do not support scripting at all. For those that do, scripts in EPUBs that seek this sort of information might be blocked by the application before they even reach the platform level. If not, the risks are likely similar to that of HTML/CSS. 
+
```
iherman, line 56: See my remark above: the risks are similar to that of HTML/CSS and, in practice, are handled by the underlying Web engine.

epub33/explainers/EPUB-33-security-privacy.md
```diff
@@ -0,0 +1,103 @@
...
+
+### 7. Does this specification allow an origin access to sensors on a user’s device?
+
+EPUB does have requirements for user agents to respond to device orientation (`portrait | landscape`) for fixed-layout EPUBs. 
+
+There has been talk about scripted EPUBs using geolocation information; I don’t know if this has happened in the wild yet, or if any reading systems would allow access to this information. This would happen through existing Web APIs; EPUB does not specify new APIs to interact with sensors. 
+
```
iherman, line 63: I do not think we should go into these details. EPUB does not include any extra access to sensors beyond the Web APIs. That is it.

iherman: It is actually important to stress this elsewhere (maybe in the intro): EPUB does _not_ create _any_ new API, access method, etc. It relies on existing Web API-s. Ie (and that is important to emphasize) _EPUB does not fork the Web!_

epub33/explainers/EPUB-33-security-privacy.md
```diff
@@ -0,0 +1,103 @@
...
+Again, this depends on the architecture of the reading system. 
+
+### 9. Does this specification enable new script execution/loading mechanisms?
+
+No. Scripting is not forbidden, but follows the HTML model.
+
+### 10. Does this specification allow an origin to access other devices?
+
+EPUB reading systems often implement syncing behaviour across multiple devices with the same user account. This is not required or mentioned by the spec. 
+
```
iherman, line 75: Actually... to be very precise, syncing behaviour across multiple devices for the same EPUB publication. I do not think that there is a way for a script runing in one EPUB publication to access information on another EPUB publication, even if it belongs to the same user account.

epub33/explainers/EPUB-33-security-privacy.md
```diff
@@ -0,0 +1,103 @@
...
+### 11. Does this specification allow an origin some measure of control over a user agent’s native UI?
+
+Only to the extent that the reading system or app platform would allow full-screen, etc. 
+
```
iherman, line 79: I think the question is whether I can access the user agent's (ie, the RS) native UI from my EPUB content. I believe the answer is simply no (the fact that the RS my allow full-screen or not is irrelevant imho).

epub33/explainers/EPUB33-explainer.md
```diff
@@ -0,0 +1,134 @@
...
+
+EPUB 3.3 includes the following specifications:
+
+* [EPUB 3.3 Overview](https://w3c.github.io/epub-specs/epub33/overview/), a non-normative overview of EPUB 3.3
+* [EPUB 3.3 Core](https://w3c.github.io/epub-specs/epub33/core/), the specification of the file format
+* [EPUB 3.3 Reading Systems](https://w3c.github.io/epub-specs/epub33/rs/), the specification for user agents, known as EPUB Reading Systems. 
+* [EPUB Accessibility 1.1](https://w3c.github.io/epub-specs/epub33/a11y/)
+* [EPUB Accessibility Techniques 1.1](https://w3c.github.io/epub-specs/epub33/a11y-tech/)
+
```
iherman, line 29: Note that the techniques is non-normative. If we list it here then we should probably add the multiple rendition document, too, or list only the rec-track documents.

epub33/explainers/EPUB33-explainer.md
```diff
@@ -0,0 +1,134 @@
...
+## Goals
+
+## Non-goals
+
+
+## Security and Privacy Questionnaire
+
+EPUB files are presented to end users by EPUB Reading Systems, which can be constructed as web applications but are more likely to be native apps. The EPUB specifications define the file format in great detail, but generally give user agents wide latitude, subject mostly to the requirements of the embedded technologies, such as HTML, SVG, XML, and CSS. 
+
```
iherman: Maybe worth emphasizing what I said above: that reading systems are mostly based on Blink/Webkit, that EPUB does not add new APIs, does not fork the Web, etc. 

(We have to be slightly defensive here...)

epub33/explainers/EPUB33-explainer.md
```diff
@@ -0,0 +1,134 @@
...
+### 1. What information might this feature expose to Web sites or other parties, and for what purposes is that exposure necessary?
+
+EPUB Reading Systems typically collect information on the reading habits of their users. In some cases this information is necessary to provide a good user experience; in other cases the information is collected for other users—the reading system company or the publisher of the book. 
+
+1. Tracking reading position. You really do want the reading system to remember that you just started chapter 5. 
+2. Remembering highlights, notes, bookmarks. 
+3. Remembering user preferences around fonts, visual themes (including dark/light), text justification, etc. 
+4. Recording progress through a book and whether a book was finished. 
+5. Some reading systems track every user reading session, including the time of day, the duration, how many pages were read, what book, the user’s IP address if a web-based reader, etc. 
+
```
iherman: Worth repeating what I said above: there is an exchange of info in some sense, but only for the same publication.

iherman:
It is strange that the security/privacy and the TAG questionnaire sound almost the same... Anyway, my comments on the questionnaire should then be synced with the TAG document as well...

epub33/explainers/EPUB-33-security-privacy.md
dauwhe: I've made that change. But I do think we need a lot of input from people who know more than us before this section can be considered adequate :)

epub33/explainers/EPUB-33-security-privacy.md
dauwhe: I think Jiminy has shown that scripts running in one EPUB can indeed access information in other EPUBs.

epub33/explainers/EPUB-33-security-privacy.md
dauwhe: Do we know that the information is sandboxed?

epub33/explainers/EPUB-33-security-privacy.md
dauwhe: We do have the `epubReadingSystem` object...

epub33/explainers/EPUB-33-security-privacy.md
iherman: Absolutely true, and we hope they will give us some more input!

epub33/explainers/EPUB-33-security-privacy.md
iherman: I have never heard of the opposite but, indeed, somebody who knows the Reading System landscape better than I do should tell us...

epub33/explainers/EPUB-33-security-privacy.md
iherman: ... which we may turn into optional:-) But you are right, it is a minor addition to the existing structures. We can add something like "(except for single entry to get back an identification of the RS)". I wish all Web API-s were as simple:-)

epub33/explainers/EPUB-33-security-privacy.md
iherman: Really? I would consider that a bug (but obviously a privacy/security breach). If the EPUB instance is a single origin, then this could be considered as a CORS violation of some sort...

epub33/explainers/EPUB-33-security-privacy.md
dauwhe: I have made that change. By the way, how would one tell that the A11Y techniques doc is not rev-track from looking at the document itself?

epub33/explainers/EPUB-33-security-privacy.md
dauwhe: I have added a few words mentioning the use of browser engines.

epub33/explainers/EPUB-33-security-privacy.md
dauwhe: Added.

epub33/explainers/EPUB-33-security-privacy.md
dauwhe: One of the reasons to do all these reviews is to hear from people who know more about this than we do :)

epub33/explainers/EPUB-33-security-privacy.md
dauwhe: Since we already mentioned this, I'm comfortable with not repeating it.

epub33/explainers/EPUB-33-security-privacy.md
dauwhe: > I do not think we should go into these details. EPUB does not include any extra access to sensors beyond the Web APIs. That is it.

But EPUB does allow scripting, and scripting might allow access to sensors. Providing some context on how that is actually playing out in the EPUB ecosystem seems useful to me.

epub33/explainers/EPUB-33-security-privacy.md
dauwhe: Given that what we say about origins (especially in 3.2 and earlier), this might not even violate the spec.

epub33/explainers/EPUB-33-security-privacy.md
dauwhe: Updated.

epub33/explainers/EPUB33-explainer.md
dauwhe: This was meant as a convenience for people who are looking at the questionnaire. I've already made a change indicating that a11y techniques is non-normative. If we remove all non-normative documents we lose the EPUB Overview. I am opposed to adding the multiple rendition note, as that is far outside the core of EPUB.

epub33/explainers/EPUB33-explainer.md
dauwhe: Note this is moot, as I forgot to delete this text when I split the explainer and the security questionnaire.

epub33/explainers/EPUB33-explainer.md
dauwhe: Moot as this has been deleted from the explainer (and remains in the security questionnaire).

epub33/explainers/EPUB-33-security-privacy.md
iherman: On the editor's draft? It is well hidden: there is a sentence in the status section which says "The group does not expect this document to become a W3C Recommendation.". It is generated by respec.

epub33/explainers/EPUB-33-security-privacy.md
iherman: ok

epub33/explainers/EPUB-33-security-privacy.md
iherman: > But EPUB does allow scripting, and scripting might allow access to sensors

... via API-s that are defined on the Open Web Platform. The important point is that we do not introduce anything new in this respect. Selfishly: this also means that the related security/privacy problems are not for us to solve...

epub33/explainers/EPUB-33-security-privacy.md
iherman: Why? This means that another EPUB instance has definitely a different origin, and that means scripts are not supposed to get there...

epub33/explainers/EPUB-33-security-privacy.md
dauwhe: I already have 

> This would happen through existing Web APIs; EPUB does not specify new APIs to interact with sensors. 

Is this not enough?

epub33/explainers/EPUB33-explainer.md
iherman: To be clear: I agree to list notes, as they are part of the picture. We just have to be consistent and list them all.

epub33/explainers/EPUB-33-security-privacy.md
iherman: I missed that... sorry.

epub33/explainers/EPUB33-explainer.md
dauwhe: Consistency is not the only virtue. I was using my editorial judgment in deciding what links would be most useful to a reviewer. In my opinion, providing a link to the multiple renditions document would merely serve to muddy the waters. Multiple Renditions are not a thing!

But I do not wish to spend much more time on this issue. And so I will add a link to the multiple renditions document. Should I also add a link to EPUB-CFI? Which, of course, would remind everyone of an EPUB technology that is unrelated to the web...

iherman: You are right about MR being... well, not really a thing:-)

For the time being we do not have a WG resolution to really publish epubcfi. It is "only" on our repo and the decision whether it will become a WG note is way down the line. So no, I would not put it there.

dauwhe:
I'm going to merge this, so at least the explainers will be in our repo and people can comment on them. I expect lots and lots of changes to both of these documents before they are submitted. It might be easier if we can discuss particular points in new issues, rather than continuing to comment on the PR.

PR #1658, opened 2021-05-01
Editorial cleanup

mattgarrish:
I started this PR off with the modest goal of reviewing some of the more awful workarounds to normative keywords we've used in the past now that we reference rfc 8174, but it kind of grew out of hand.

It's still all editorial, but as I was looking at our "strong encouragements" and thing you "need to" or "can" do, I spotted some reading system requirements still in the core spec that were missed because they were written in the passive voice (primarily in the meta-inf files). That led me to also try and weed out as many passive requirements and statements as I could from the core and reading system specs, which will explain the extensive changes.

The requirements are largely all the same; it's only the phrasing that's changed. The one exception was in the identifier and last modification sections. We were using normative keywords to describe best practices. For example, the text was stating that creators SHOULD change the unique identifier as infrequently as possible and explaining when new ones SHOULD NOT be issued, but these are not normative. The last modification date similarly stated that it MUST be updated when changes are made, but this again is advice we could never check. I've lowercased these.

I ended the review at the vocabularies. They're not as critical as there are few normative requirements in them. I'll try to return to them later when there are fewer open PRs again.

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/editorial/rfc8174/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/editorial/rfc8174/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1658.html" title="Last updated on May 1, 2021, 12:55 AM UTC (a1ab9cb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1658/f430175...a1ab9cb.html" title="Last updated on May 1, 2021, 12:55 AM UTC (a1ab9cb)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -95,8 +95,8 @@
 				<p>EPUB 3 has been widely adopted as the format for digital books (ebooks), and this revision continues
 					to increase the format's capabilities to better support a wider range of publication requirements,
 					including complex layouts, rich media and interactivity, and global typography features. The
-					expectation is that the EPUB 3 format will be utilized for a broad range of content, including
-					books, magazines, and educational, professional, and scientific publications.</p>
+					expectation is that publishers will utilize the EPUB 3 format for a broad range of content,
+					including books, magazines, and educational, professional, and scientific publications.</p>
 
 				<p>This specification represents the core of EPUB 3 and includes the conformance requirements for
 						<a>EPUB Publications</a> — the product of the standard. The other specifications that comprise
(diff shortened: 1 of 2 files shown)


iherman (approved):
Wow! The number changed item in the diff files is around 200 :-)

We should merge and actually publish these asap. Otherwise we get into PR hell (I hope that will not be the case for current open PR-s already...)

mattgarrish:
> The number changed item in the diff files is around 200 :-)

Ya, we tend to stray into talking passively about things that must or should be set rather than just saying directly that epub creators do these things. Something to watch for going forward.

> We should merge and actually publish these asap. Otherwise we get into PR hell

Right, that's why I stopped at the vocabularies. I don't see this as a controversial change, so I'll merge it in and fix up the open PRs.

Issue #1087, opened 2018-08-05
Most readers probably don't follow the MUST about the requiredExtensions attribute for SVG


BCWalters:
[SVG Reading System conformance](https://w3c.github.io/publ-epub-revision/epub32/spec/epub-contentdocs.html#confreq-svg-rs-reqdExt)

>It MUST recognize the value "http://www.idpf.org/2007/ops" of the requiredExtensions attribute as representing the occurrence of XHTML Content Document fragments (e.g., when the attribute is included on the foreignObject element or children of the switch element).

Presumably there was a good historical reason for this that's lost on me, but I doubt most EPUB readers do anything with this MUST.

mattgarrish:
OPS 2.0.1 has this to say about it:

> Since OPS is, by definition, supported in EPUB Reading Systems, the requiredExtension attribute on SVG elements with the value http://www.idpf.org/2007/ops must always evaluate to true.

But since we've deprecated out almost all epub-created extensions, I'm not sure if this is still necessary.

@rkwright @iherman @murata2makoto any thoughts?

iherman:
@rkwright should know infinitely more about this than me, but I chime in anyway...

- If I look at the text purely from standard perspective then something like this URL is needed for the `<switch>` element to work to signal whether the SVG implementation does implement `<foreignObject>`. Strangely enough (@rkwright ?) there is no mention in the text on some form of a registry or whatever for what extensions means; I must admit I did not really understood from the SVG text how this thing should work.

- In practice, however, it seems that the traditional example of combining switch and foreign object (with HTML) simply works in browser engines (I tried it in Safari, Chrome and Firefox; I do not have Edge on my mac but I would expect that to work, too). In other words, @BCWalters is right in the sense that the attribute is probably simply ignored by implementations, ie, a MUST in our standard sounds indeed superfluous...

murata2makoto:
I do not know if non-EPUB SVG often uses the svg:switch and svg:foreignObject elements as representing the occurrence of XHTML Content Document fragments.  If this is the case, I think that we can lift the requirement that requiredExtensions shall be "http://www.idpf.org/2007/ops".

dauwhe:
So what happens here? Note that browsers don't have interop on this—Chrome shows "baz" and Firefox/Safari show "bar". 

```
<svg width="400px" height="300px" viewBox="0 0 400 300"
     xmlns="http://www.w3.org/2000/svg">
<switch>
     <foreignObject width="100" height="50"
[...]

PR #1662, opened 2021-05-04
remove requirement to support SVG requiredExtensions attribute per #1087

dauwhe:
***
<a href="https://pr-preview.s3.amazonaws.com/dauwhe/epub-specs/pull/1662.html" title="Last updated on May 4, 2021, 5:01 PM UTC (49b5dab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1662/010a52f...dauwhe:49b5dab.html" title="Last updated on May 4, 2021, 5:01 PM UTC (49b5dab)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3969,11 +3969,6 @@ Spine:
 										that conforms to the XHTML Content Document model defined in <a
 											href="#sec-xhtml-req"></a>.</p>
 								</li>
-								<li>
-									<p id="confreq-svg-foreignObject-reqext">MUST have its
-											<code>requiredExtensions</code> attribute, when set, specify the value
-											"<code>http://www.idpf.org/2007/ops</code>".</p>
-								</li>
 							</ul>
 						</li>
 						<li>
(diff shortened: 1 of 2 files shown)


iherman (approved):

iherman (approved):
I believe this change should be recorded in the changes' section.

mattgarrish:
Fine with this change (except needs a change entry as already noted), but are authors still required to set the attribute or are we dropping that requirement, too?

iherman:
> Fine with this change (except needs a change entry as already noted), but are authors still required to set the attribute or are we dropping that requirement, too?

Looking up the original issues it seems that this attribute does more harm than good; [my own tests](https://github.com/w3c/epub-specs/issues/1087#issuecomment-718880480) resulted in SVG content that was not displayed (eg, on Thorium). So yes, I believe authors should forget about this.

dauwhe:
I removed the corresponding line from the core spec, and added an entry to the change log.

mattgarrish (approved):

PR #1664, opened 2021-05-05
Move registries to notes

mattgarrish:
An attempt to do away with formally referencing the old idpf registries so we don't have to port them forward.

I've noted the obsolescence of the old IDPF registries without fully removing the links to them. (I don't believe we need references if we're only providing a link to where they were defined. @iherman?)

The only registry with values that weren't purely informative/repetitive was for the authority property. I've integrated these into the definition.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1664.html" title="Last updated on May 7, 2021, 12:24 PM UTC (86e46e7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1664/821c764...86e46e7.html" title="Last updated on May 7, 2021, 12:24 PM UTC (86e46e7)">Diff</a>

### DIFF
--- a/epub33/common/js/biblio.js
+++ b/epub33/common/js/biblio.js
@@ -3,14 +3,6 @@ var biblio = {
 		"title": "EPUB Accessibility Frequently Asked Questions",
 		"href": "http://www.idpf.org/epub/guides/a11y-faq"
 	},
-	"AttributeExtensions": {
-		"title": "EPUB Custom Attribute Extensions for Content Documents",
-		"href": "http://www.idpf.org/epub/extensions/attributes"
-	},
-	"AuthorityRegistry": {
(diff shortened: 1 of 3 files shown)


iherman (approved):
Although this is not a major technical change, but the statement whereby we no longer consider the registry is a fairly substantial. Ie, I wonder whether it is not worth adding this reference to the change section, too.

mattgarrish:
> but the statement whereby we no longer consider the registry is a fairly substantial

Ya, but you could argue it's just an admission of reality. ;)

Nothing changes from a validation perspective, but I've added a change log entry explaining the update.

dauwhe (approved):

wareid (approved):

PR #1670, opened 2021-05-09
Change to using URL standard

### LINKED ISSUE #808: OCF Document: RFC3986 or RFC3987?
(I know this is a _huge_ can of worms, and I regularly get it wrong; maybe it is the case this time, too...)

The issue is what the allowed characters are for file names. More specifically, what the allowed characters are for the value of `full-path` in the `container.xml` file.
- In Section 3.3, the reference is to RFC3986 (URI) _and_ RFC3987 (IRI) are both listed. Which sounds about right; this means a file name of the form `/téléphone` is all right, although this would not be acceptable via RFC3986. (Although I wonder whether the reference to RFC3986 is necessary in the first place; in this respect RFC3987 supersedes RFC3986, doesn't it)
- In 3.5.3.1, towards the end, only RFC3986 is reference in the defintion of `full-path`. This seems to be in contradiction with what is in Section 3.3
- Section 3.4 seems to give a specific definition for file names; it is not clear to me whether this is necessary. Isn't it enough to refer to the same RFC3987 (well, the `path` portion thereof) to avoid mixup? Isn't it enough to quote RFC3987 and, if necessary, list the possible restrictions (I have not checked which of the characters listed in the 4th bullet point are excluded from the path segment of an IRI anyway) when it comes to file name? (Yes, of course, we have to refer to the last portion of an IRI path as the file name.)
- And here is the rub: I think there _is_ a discrepancy at this moment. If I correctly interpret either RFC3986 _or_ RFC3987, the space (i.e., U+0020) character is _excluded_ from the path, whereas Section 3.4 does not exclude it. This means that the examples are also wrong, because `full-path="EPUB/Great Expectation.opf"` is _not_ a valid value for `@full-path`, although it is indeed a valid file name per Section 3.4:-(

As i said, it _is_ a can of worm, and one of you guys may prove me wrong in my interpretation... But if I am right, my proposal would be:
1. Use RFC3987 only in the references (with a possible note relating it to RFC3986)
2. Reduce section 3.4 by saying that a file name should be conform to the path of RFC3987, minus a number of characters (if necessary, ie, check whether all those entries are necessary)
3. Change the examples...

### LINKED ISSUE #1303: Are fragment identifiers allowed in manifest?
See https://github.com/w3c/epubcheck/issues/1085

It appears that `href` attributes anywhere in the package file must be IRIs. 

> Each item element in the manifest identifies a Publication Resource by the IRI [RFC3987] provided in its href attribute. The IRI MAY be absolute or relative. In the case of relative IRIs, Reading Systems MUST use the IRI of the Package Document as the base when resolving these to absolute IRIs. The resulting absolute IRI MUST be unique within the manifest scope.

But we don't appear to forbid fragment identifiers. All we say about `href` is

> An absolute or relative IRI reference [RFC3987] to a resource.

EPUBCheck complains loudly about fragment identifiers, I think because it misinterprets them as incorrect file extensions.

I'm not sure if we need to do anything here.

### LINKED ISSUE #1673: Handling of invalid properties
For the property data type, we say Reading System concatenation of the base URL and value must result in a valid url string.

So what happens if it doesn't? Must Reading Systems ignore the value? Should they ignore it? Or do we not care?

mattgarrish:
I've updated the authoring and reading system specifications to use the URL standard terminology.

There are a number of changes in this PR, but they basically all break down along these lines:

- relative IRI has become "relative-URL string" where it doesn't seem to make sense to allow fragment identifiers (manifest item and media overlays audio src)
- relative IRI has become "relative-URL-with-fragment string" for all other cases
- same changes as above for absolute IRIs
- where we hadn't broken our relative from absolute, I went with "valid URL string"
- base IRI has become base URL

The one tough case was in ocf, where the rootfile and link elements required the IRI be "a path component [RFC3986] which MUST take the form of a path-rootless [RFC3986] only". My best interpretation of this is that it is the equivalent of the URL standard's "path-relative-scheme-less-URL string".

I've also used this definition for the property datatype definition to replace irelative-ref in the ebnf.

Please let me know if you think I've made a mistake with any of these.

Otherwise, I've moved a few more passively expressed reading system requirements over to the proper document. These were all related to creating URLs from the default vocabulary base URLs. Reading systems are the ones that create these expanded URLs.

Fixes #808 

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-808/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-808/epub33/rs/index.html)

EDIT: Also fixes #1673 by adding requirement for reading systems to ignore properties that expand to invalid URL strings and also prefixes without a referent.

EDIT 2: Also fixes #1303 by restricting manifest items to fragment-less URLs.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1670.html" title="Last updated on May 14, 2021, 12:12 PM UTC (039d181)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1670/0f951c6...039d181.html" title="Last updated on May 14, 2021, 12:12 PM UTC (039d181)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -8197,8 +8197,9 @@ html.my-document-playing * {
 							href="#sec-property-datatype"><var>property</var> value</a> is expected. EPUB Creators MUST
 						NOT add a prefix to terms and properties from a default vocabulary.</p>
 
-					<p>EPUB Creators MUST NOT assign a prefix to the URLs associated with these vocabularies using the
-							<a href="#sec-prefix-attr"><code>prefix</code></a> attribute.</p>
+					<p>EPUB Creators MUST NOT assign a prefix to the <a
+							href="https://url.spec.whatwg.org/#concept-base-url">base URLs</a> [[URL]] associated with
+						these vocabularies using the <a href="#sec-prefix-attr"><code>prefix</code></a> attribute.</p>
 
 					<div class="note">
 						<p>Refer to the definition of each attribute that takes a <a href="#sec-property-datatype"
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -2733,7 +2733,7 @@
 									href="https://url.spec.whatwg.org/#absolute-url-string">absolute-</a> or <a
 									href="https://url.spec.whatwg.org/#relative-url-string">relative-URL string</a>
 								[[URL]], but they MUST ensure each URL is unique within the <code>manifest</code> scope
-								after <a href="https://www.w3.org/TR/epub-rs-33/#sec-pkg-doc-relative-iris">resolution
+								after <a href="https://www.w3.org/TR/epub-rs-33/#sec-pkg-doc-relative-urls">resolution
 									to an absolute URL</a> [[EPUB-RS-33]].</p>
 
 							<p id="attrdef-item-media-type">The Publication Resource identified by an <code>item</code>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5402,10 +5402,9 @@ Spine:
 						OCF Abstract Container and not relative to the <code>META-INF</code> directory.</p>
 
 					<p>All <a href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
-							>relative-URL-with-fragment strings</a> [[URL]] MUST identify resources within the OCF
-						Abstract Container (i.e., at or below the Root Directory) after <a
-							href="https://url.spec.whatwg.org/#concept-url-parser">parsing to URL records</a>
-						[[URL]].</p>
+							>relative-URL-with-fragment strings</a> [[URL]] MUST, after <a
+							href="https://url.spec.whatwg.org/#concept-url-parser">parsing to URL records</a> [[URL]],
+						identify resources within the OCF Abstract Container (i.e., at or below the Root Directory).</p>
 				</section>
 
 				<section id="sec-container-filenames">
@@ -9436,6 +9435,8 @@ EPUB/images/cover.png</pre>
 				-->
 
 				<ul>
+					<li>12-May-2021: Changed all references to URIs/IRIs to URLs and references to RFCs 3986 and 3987 to
+						the URL standard. See <a href="https://github.com/w3c/epub-specs/issues/808">issue 808</a>.</li>
 					<li>06-May-2021: Noted that the working group will no longer maintain the publication type and
 						collection role registries and removed dependence on the latter for validating collection types
 						(use of NMToken values remains restricted to extension specifications). The registry of
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -6756,7 +6756,7 @@ store destination as source in ocf
 										<p>Identifies an associated fragment of an EPUB Content Document.</p>
 										<p>The value MUST be a <a
 												href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
-												>relative-URL-with-fragment string</a> [[URL]] with a non-null <a
+												>relative-URL-with-fragment string</a> [[URL]] with a <a
 												href="#sec-media-overlays-fragids">fragment identifier</a>.</p>
 									</dd>
 								</dl>
@@ -6835,7 +6835,7 @@ store destination as source in ocf
 										<p>Identifies an associated fragment of an EPUB Content Document.</p>
 										<p>The value MUST be a <a
 												href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
-												>relative-URL-with-fragment string</a> [[URL]] with a non-null <a
+												>relative-URL-with-fragment string</a> [[URL]] with a <a
 												href="#sec-media-overlays-fragids">fragment identifier</a>.</p>
 										<p>Refer to <a href="#sec-media-overlays-structure"></a> for more
 											information.</p>
@@ -6969,7 +6969,7 @@ store destination as source in ocf
 										<p>Identifies the associated fragment of an EPUB Content Document.</p>
 										<p>The value MUST be a <a
 												href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
-												>relative-URL-with-fragment string</a> [[URL]] with a non-null <a
+												>relative-URL-with-fragment string</a> [[URL]] with a <a
 												href="#sec-media-overlays-fragids">fragment identifier</a>.</p>
 									</dd>
 									<dt>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5369,11 +5369,11 @@ Spine:
                 </pre>
 					</aside>
 
-					<p>For <a href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
-							>relative-URL-with-fragment strings</a> [[URL]], the <a
-							href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> [[URL]] is determined by
-						the relevant language specifications for the given file formats. For example, CSS defines how
-						relative URL references work in the context of CSS style sheets and property declarations
+					<p>The relevant language specification for a given file format determines the <a
+							href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> [[URL]] used to parse <a
+							href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
+							>relative-URL-with-fragment strings</a> [[URL]]. For example, CSS defines how relative URL
+						references work in the context of CSS style sheets and property declarations
 						[[CSSSnapshot]].</p>
 
 					<div class="note">
@@ -5402,8 +5402,10 @@ Spine:
 						OCF Abstract Container and not relative to the <code>META-INF</code> directory.</p>
 
 					<p>All <a href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
-							>relative-URL-with-fragment strings</a> [[URL]] references MUST resolve to resources within
-						the OCF Abstract Container (i.e., at or below the Root Directory).</p>
+							>relative-URL-with-fragment strings</a> [[URL]] MUST identify resources within the OCF
+						Abstract Container (i.e., at or below the Root Directory) after <a
+							href="https://url.spec.whatwg.org/#concept-url-parser">parsing to URL records</a>
+						[[URL]].</p>
 				</section>
 
 				<section id="sec-container-filenames">
@@ -6752,10 +6754,11 @@ store destination as source in ocf
 										<code>[optional]</code>
 									</dt>
 									<dd>
-										<p>The <a href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
-												>relative-URL-with-fragment string</a> [[URL]] of the corresponding EPUB
-											Content Document, which MUST have a <a href="#sec-media-overlays-fragids"
-												>fragment identifier</a> that references the specific content.</p>
+										<p>Identifies an associated fragment of an EPUB Content Document.</p>
+										<p>The value MUST be a <a
+												href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
+												>relative-URL-with-fragment string</a> [[URL]] with a non-null <a
+												href="#sec-media-overlays-fragids">fragment identifier</a>.</p>
 									</dd>
 								</dl>
 							</dd>
@@ -6830,10 +6833,11 @@ store destination as source in ocf
 										<code>[required]</code>
 									</dt>
 									<dd>
-										<p>The <a href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
-												>relative-URL-with-fragment string</a> [[URL]] of the corresponding EPUB
-											Content Document, which MUST have a <a href="#sec-media-overlays-fragids"
-												>fragment identifier</a> that references the specific content.</p>
+										<p>Identifies an associated fragment of an EPUB Content Document.</p>
+										<p>The value MUST be a <a
+												href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
+												>relative-URL-with-fragment string</a> [[URL]] with a non-null <a
+												href="#sec-media-overlays-fragids">fragment identifier</a>.</p>
 										<p>Refer to <a href="#sec-media-overlays-structure"></a> for more
 											information.</p>
 									</dd>
@@ -6963,10 +6967,11 @@ store destination as source in ocf
 										<code>[required]</code>
 									</dt>
 									<dd>
-										<p>The <a href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
-												>relative-URL-with-fragment string</a> [[URL]] of the corresponding EPUB
-											Content Document, which MUST have a <a href="#sec-media-overlays-fragids"
-												>fragment identifier</a> that references the specific content.</p>
+										<p>Identifies the associated fragment of an EPUB Content Document.</p>
+										<p>The value MUST be a <a
+												href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
+												>relative-URL-with-fragment string</a> [[URL]] with a non-null <a
+												href="#sec-media-overlays-fragids">fragment identifier</a>.</p>
 									</dd>
 									<dt>
 										<code>id</code>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1303,7 +1303,7 @@
 							</dt>
 							<dd>
 								<p>Establishes an association between the current expression and the element or resource
-									identified by its value. EPUB Creators MUST use, as the value, a <a
+									identified by its value. EPUB Creators MUST use as the value a <a
 										href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
 										>relative-URL-with-fragment string</a> [[URL]] that references the resource or
 									element they are describing.</p>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1140,8 +1140,8 @@
 								metadata.</p>
 						</li>
 						<li>
-							<p>A <a href="#sec-manifest-elem">manifest</a> — identifies via IRI [[RFC3987]], and
-								describes via MIME media type [[RFC4839]], the set of <a>Publication Resources</a>.</p>
+							<p>A <a href="#sec-manifest-elem">manifest</a> — identifies via URL [[URL]], and describes
+								via MIME media type [[RFC4839]], the set of <a>Publication Resources</a>.</p>
 						</li>
 						<li>
 							<p>A <a href="#sec-spine-elem">spine</a> — an ordered sequence of ID references to top-level
@@ -1226,7 +1226,8 @@
 								<code>href</code>
 							</dt>
 							<dd>
-								<p>An absolute or relative IRI reference [[RFC3987]] to a resource.</p>
+								<p>A <a href="https://url.spec.whatwg.org/#valid-url-string">valid URL string</a>
+									[[URL]] that references a resource.</p>
 								<div class="example">
 									<pre>&lt;link rel="record"
       href="meta/9780000000001.xml" 
@@ -1302,8 +1303,10 @@
 							</dt>
 							<dd>
 								<p>Establishes an association between the current expression and the element or resource
-									identified by its value. The value of the attribute must be a relative IRI
-									[[RFC3987]] that references the resource or element being described.</p>
+									identified by its value. EPUB Creators MUST use, as the value, a <a
+										href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
+										>relative-URL-with-fragment string</a> [[URL]] that references the resource or
+									element they are describing.</p>
 								<aside class="example">
 									<p>The following example shows the <code>refines</code> element used to indicate a
 										creator is the illustrator.</p>
@@ -2151,7 +2154,7 @@
 								</aside>
 
 								<aside class="example">
-									<p>The following example shows the use of an IRI for the scheme.</p>
+									<p>The following example shows the use of a URL for the scheme.</p>
 									<pre>&lt;dc:subject id="sbj01"&gt;Number Theory&lt;/dc:subject&gt;
 &lt;meta refines="#sbj01" property="authority">http://www.ams.org/msc/msc2010.html&lt;/meta>
 &lt;meta refines="#sbj01" property="term">11&lt;/meta></pre>
@@ -2290,9 +2293,7 @@
 
 							<p>The <a href="#app-meta-property-vocab">Meta Properties Vocabulary</a> is the <a
 									href="#sec-default-vocab">default vocabulary</a> for use with the
-									<code>property</code> attribute. If the attribute's value does not include a prefix,
-								the following IRI [[RFC3987]] stem MUST be used to generate the resulting IRI:
-									<code>http://idpf.org/epub/vocab/package/meta/#</code></p>
+									<code>property</code> attribute.</p>
 
 							<p>EPUB Creators MAY add terms from other vocabularies as defined in <a
 									href="#sec-vocab-assoc"></a>.</p>
@@ -2492,8 +2493,8 @@
 
 							<p id="attrdef-link-media-type">The <a href="#attrdef-media-type"><code>media-type</code>
 									attribute</a> is OPTIONAL when a linked resource is located outside the EPUB
-								Container, as more than one media type could be served from the same IRI. EPUB Creators
-								MUST specify the attribute for all <a>Local Resources</a>.</p>
+								Container, as more than one media type could be served from the same URL [[URL]]. EPUB
+								Creators MUST specify the attribute for all <a>Local Resources</a>.</p>
 
 							<p id="attrdef-hreflang">The OPTIONAL <code>hreflang</code> attribute identifies the
 								language of the linked resource. The value MUST be a <a
@@ -2529,9 +2530,7 @@
 
 							<p>The <a href="#app-link-vocab">Metadata Link Vocabulary</a> is the <a
 									href="#sec-default-vocab">default vocabulary</a> for the <code>rel</code> and
-									<code>properties</code> attributes. If any of these attributes' values do not
-								include a prefix, the following IRI [[RFC3987]] stem MUST be used to generate the
-								resulting IRI for them: <code>http://idpf.org/epub/vocab/package/link/#</code></p>
+									<code>properties</code> attributes.</p>
 
 							<p><a>EPUB Creators</a> MAY add relationships and properties from other vocabularies as
 								defined in <a href="#sec-vocab-assoc"></a>.</p>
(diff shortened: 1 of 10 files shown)


epub33/core/vocab/item-properties.html
```diff
@@ -1,18 +1,15 @@
...
-	
 	<p>The <strong>Applies to</strong> field indicates which Publication Resource type(s) the given property MAY
 		be specified on, the <strong>Cardinality</strong> field indicates the number of times the property MUST
 		appear within the Package Document scope, and the <strong>Usage</strong> field indicates usage
 		conditions.</p>
-	
-	<p>Properties without a prefix are referenceable using the base IRI
+	<p>Properties without a prefix are referenceable using the <a
+			href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> [[URL]]
 			<code>http://idpf.org/epub/vocab/package/item/#</code>.</p>
```
iherman: Wouldn't it better to rephrase this as:

> The base URL <a href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> [[URL]] is <code>http://idpf.org/epub/vocab/package/item/#</code>.</p>

To be identical to a similar statement in §D.4 (albeit for rendition properties). I find it also cleaner

epub33/core/vocab/itemref-properties.html
```diff
@@ -6,7 +6,8 @@ <h3>Spine Properties Vocabulary</h3>
 				><code>itemref</code> element's</a>
 		<a href="#attrdef-itemref-properties"><code>properties</code> attribute</a>.</p>
 
-	<p>Properties are referenceable using the base IRI
+	<p>Properties are referenceable using the <a
+		href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> [[URL]]
 			<code>http://idpf.org/epub/vocab/package/itemref/#</code>.</p>
```
iherman, line 11: Same comment as above for the Manifest property vocabulary

epub33/core/vocab/meta-property.html
```diff
@@ -12,7 +12,8 @@ <h3>Meta Properties Vocabulary</h3>
 		value (using [[!XMLSCHEMA-2]] datatypes), the <strong>Cardinality</strong> field indicates the number of
 		times the property MAY be attached to another property, and the <strong>Extends</strong> field
 		identifies the properties it MAY be attached to.</p>
-	<p>Properties without a prefix are referenceable using the base IRI
+	<p>Properties without a prefix are referenceable using the <a
+		href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> [[URL]]
 			<code>http://idpf.org/epub/vocab/package/meta/#</code>.</p>
```
iherman: Same comment as above for the Manifest properties

epub33/core/vocab/rendering.html
```diff
@@ -10,8 +10,8 @@ <h3>Package Rendering Vocabulary</h3>
 			System</a>). If a Reading System supports the desired rendering, these properties enable the
 		user to be presented the content as the EPUB Creator optimally designed it.</p>
 	
-	<p>The base IRI for referencing these properties is
-		<code>http://www.idpf.org/vocab/rendition/#</code>.</p>
+	<p>The <a href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> [[URL]] for referencing
+		these properties is <code>http://www.idpf.org/vocab/rendition/#</code>.</p>
```
iherman: And again... :-)

epub33/rs/index.html
```diff
@@ -1591,22 +1618,23 @@ <h2>Vocabulary Association Mechanisms</h2>
...
+							<p>If the property consists of a prefix and reference, obtain the URL by concatenating the
+									<a href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> [[URL]]
+								associated with the prefix to the reference. If the EPUB Creator has not defined a
+								matching prefix, the property is invalid and the Reading System MUST ignore it.</p>
 						</li>
 					</ul>
-					<p>The resulting IRI MUST be valid to [[RFC3987]]. Reading Systems do not have to resolve this IRI,
-						however.</p>
+					<p>The result MUST be valid a <a href="https://url.spec.whatwg.org/#valid-url-string">valid URL
+							string</a> [[URL]]. Reading Systems do not have to resolve this URL, however.</p>
```
iherman: Remove the 'valid' from 'MUST be valid'

iherman (approved):
I did not find any problems with the mapping from the RFC concepts to the URL spec... All my comments are minor editorial one.

Thanks!

epub33/rs/index.html
marisademeglio: Not sure if this is what you mean, @iherman , but I would remove both 'valid's

marisademeglio (approved):

epub33/rs/index.html
iherman: That is not what I meant; the current text is "MUST be valid a valid URL".

Why do you think of removing valid altogether?

llemeurfr (approved):

epub33/rs/index.html
mattgarrish: The typo definitely needs fixing, but "valid URL string" is a formal term in the URL spec. We should keep that as it ensures validation.

epub33/core/vocab/item-properties.html
mattgarrish: I'm going to change these to match the rendition vocab's: "The base URL for referencing these properties is ...".

I'm also going to link them up to the default vocabulary section and modify this sentence as follows so it's clearer why we have these base URLs defined:

> EPUB Creators MUST NOT assign a prefix to the ___base___ URLs ___[[URL]]___ associated with these vocabularies using the prefix attribute.

Otherwise, I expect we'll just confuse people more why we have URLs for properties that don't need or have prefixes.

epub33/rs/index.html
marisademeglio: I was thinking that requiring something to be a URL would automatically mean a "valid URL" because is an invalid URL really a URL?  But I would never tamper with formal terms, so, ok :)

epub33/core/index.html
```diff
@@ -5352,32 +5354,37 @@ <h4>File and Directory Structure</h4>
...
 
-					<p>For relative IRI references, the Base IRI [[RFC3986]] is determined by the relevant language
-						specifications for the given file formats. For example, CSS defines how relative IRI references
-						work in the context of CSS style sheets and property declarations [[CSSSnapshot]].</p>
+					<p>For <a href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
+							>relative-URL-with-fragment strings</a> [[URL]], the <a
+							href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> [[URL]] is determined by
+						the relevant language specifications for the given file formats. For example, CSS defines how
+						relative URL references work in the context of CSS style sheets and property declarations
+						[[CSSSnapshot]].</p>
```
rdeltour: The current wording,
> For _relative-URL-with-fragment strings_, the _base URL_ is determined by the relevant language specifications for the given file formats.

may be a little imprecise: a _relative-URL-with-fragment strings_ doesn't **have** a base URL. A _base URL_ is used by a _URL parser_ to parse a relative string into a URL.

Suggested rewording:

> The _base URL_ used to parse  _relative-URL-with-fragment strings_ is determined by the relevant language specifications for the given file formats.

epub33/core/index.html
```diff
@@ -5394,8 +5401,9 @@ <h4>Relative IRIs for Referencing Other Components</h4>
 					<p>then the path <code>EPUB/Great_Expectations.opf</code> is relative to the root directory for the
 						OCF Abstract Container and not relative to the <code>META-INF</code> directory.</p>
 
-					<p>All relative IRI references MUST resolve to resources within the OCF Abstract Container (i.e., at
-						or below the Root Directory).</p>
+					<p>All <a href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
+							>relative-URL-with-fragment strings</a> [[URL]] references MUST resolve to resources within
+						the OCF Abstract Container (i.e., at or below the Root Directory).</p>
```
rdeltour: - this statement is more about **URLs** (aka **URL records**) than URL strings
- "resolve" isn't defined in the URL standard. 

Since the part about parsing relative URLs is already specified a few paragraphs above, we could say:

> The URLs returned when parsing _relative-URL-with-fragment strings_ MUST identify resources with the OCF Abstract Container.

epub33/core/index.html
```diff
@@ -6743,9 +6752,10 @@ <h5>The <code>body</code> Element</h5>
 										<code>[optional]</code>
 									</dt>
 									<dd>
-										<p>The relative IRI reference [[RFC3987]] of the corresponding EPUB Content
-											Document, including a <a href="#sec-media-overlays-fragids">fragment
-												identifier</a> that references the specific content.</p>
+										<p>The <a href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
+												>relative-URL-with-fragment string</a> [[URL]] of the corresponding EPUB
+											Content Document, which MUST have a <a href="#sec-media-overlays-fragids"
+												>fragment identifier</a> that references the specific content.</p>
```
rdeltour: This somewhat mixes the serialization terminology (strings) and the model (URL, fragment).

Possible rewording:
> A _relative-URL-with-fragment string_ representing the _URL_ of the corresponding EPUB Content Document, which MUST have a non-null _fragment_.

epub33/core/index.html
```diff
@@ -6820,9 +6830,10 @@ <h5>The <code>seq</code> Element</h5>
 										<code>[required]</code>
 									</dt>
 									<dd>
-										<p>The relative IRI reference [[RFC3987]] of the corresponding EPUB Content
-											Document, including a <a href="#sec-media-overlays-fragids">fragment
-												identifier</a> that references the specific content.</p>
+										<p>The <a href="https://url.spec.whatwg.org/#relative-url-with-fragment-string"
+												>relative-URL-with-fragment string</a> [[URL]] of the corresponding EPUB
+											Content Document, which MUST have a <a href="#sec-media-overlays-fragids"
+												>fragment identifier</a> that references the specific content.</p>
```
rdeltour: Same comment as above.

epub33/core/index.html
```diff
@@ -6952,9 +6963,10 @@ <h5>The <code>text</code> Element</h5>
...
```
rdeltour: Same comment as above.

epub33/rs/index.html
```diff
@@ -275,16 +275,20 @@ <h2>Package Document Processing</h2>
 					href="https://www.w3.org/TR/epub-33/#sec-package-doc">Package Document</a> [[EPUB-33]].</p>
 
 			<section id="sec-pkg-doc-relative-iris">
-				<h4>Resolving Relative IRIs</h4>
+				<h4>Resolving Relative URLs</h4>
```
rdeltour: Why not adopt the URL standard’s  "parsing" terminology here, instead of "resolving"?

epub33/rs/index.html
```diff
@@ -275,16 +275,20 @@ <h2>Package Document Processing</h2>
...
 
-				<p>To obtain an absolute IRI reference from a relative IRI reference [[RFC3987]] in the Package
-					Document, Reading Systems MUST use the Base IRI [[RFC3986]] of the Package Document.</p>
+				<p>To obtain an <a href="https://url.spec.whatwg.org/#absolute-url-with-fragment-string"
+						>absolute-URL-with-fragment string</a> [[URL]] from a <a
+						href="https://url.spec.whatwg.org/#relative-url-with-fragment-string">relative-URL-with-fragment
+						string</a> [[URL]] in the Package Document, Reading Systems MUST use the <a
+						href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> [[URL]] of the Package
+					Document.</p>
```
rdeltour: This mixes strings and model terminology.

In the URL standard:
- a base URL is a URL record, it is used to parse a string into another URL record.
- a relative URL string needs a base to be parsed into a URL.
- an absolute URL string doesn't need a base URL to be parsed. It is not something typically "obtained" from a relative string (except indirectly by serializing the URL returned from parsing the relative string: "relative URL string + base URL" → URL record → "absolute URL string").

Possible rewording:
> To _parse_ a _relative-URL-with-fragment string_ in the Package Document, Reading Systems MUST use the URL of the Package Document as the _base URL_.

epub33/rs/index.html
```diff
@@ -275,16 +275,20 @@ <h2>Package Document Processing</h2>
...
 
-				<p>The Base IRI of the Package Document is the combination of the Base IRI of the <a>EPUB Container</a>
+				<p>The base URL of the Package Document is the combination of the base URL of the <a>EPUB Container</a>
 					together with the path to Package Document (relative to the <a>Root Directory</a>) when an EPUB
 					Publication is zipped.</p>
```
rdeltour: "combination" is imprecise here. We could say something along these lines:
> the URL of the Package Document is the URL returned by _parsing_  [URL] the path to the Package Document (relative to the Root Directory), with the URL of the EPUB Container used as the base URL.

epub33/rs/index.html
```diff
@@ -275,16 +275,20 @@ <h2>Package Document Processing</h2>
...
 
-				<p>This specification does not require a specific IRI scheme for referencing the path to the Package
+				<p>This specification does not require a specific URL scheme for referencing the path to the Package
 					Document within the EPUB Container.</p>
```
rdeltour: I'm not sure what this means, but this predates this specific PR 😊

epub33/rs/index.html
```diff
@@ -409,8 +421,11 @@ <h4>Metadata</h4>
 			<section id="sec-pkg-doc-manifest">
 				<h3>Manifest</h3>
 
-				<p>When an <code>href</code> attribute contains a relative IRI, Reading Systems MUST use the IRI of the
-					Package Document as the base when resolving to an absolute IRI.</p>
+				<p>When an <code>href</code> attribute contains a <a
+						href="https://url.spec.whatwg.org/#relative-url-string">relative-URL string</a> [[URL]], Reading
+					Systems MUST use the URL of the Package Document as the <a
+						href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> [[URL]] when resolving to an
+						<a href="https://url.spec.whatwg.org/#absolute-url-string">absolute-URL string</a> [[URL]].</p>
```
rdeltour: same as for previous comments: "resolve to an absolute URL string" isn't well defined. I suggest we use "parsing" instead of "resolve" (with ref to the URL standard parsing algorithm), and the conceptual  "URL record" instead of "absolute-URL string".

epub33/rs/index.html
```diff
@@ -372,6 +376,10 @@ <h4>Metadata</h4>
 						<p>Reading Systems SHOULD ignore all <code>meta</code> elements whose <code>property</code>
 							attributes define expressions they do not recognize. A Reading System MUST NOT fail when
 							encountering unknown expressions.</p>
+						<p>If the <code>property</code> attribute's value does not include a prefix, Reading Systems
+							MUST use the following <a href="https://url.spec.whatwg.org/#concept-base-url">base URL</a>
+							[[URL]] to generate the resulting URL:
```
rdeltour: I suggest rewording 
> Reading Systems MUST use the following base URL to generate the resulting URL

into
> Reading Systems MUST parse it into a URL by using `https://example.org` as the base URL.

rdeltour: Applies also severals times below.

epub33/rs/index.html
```diff
@@ -1591,22 +1618,23 @@ <h2>Vocabulary Association Mechanisms</h2>
...
+					<p>A Reading System MUST use the following rules to create a URL [[URL]] from a property:</p>
 					<ul>
 						<li>
-							<p>If the property consists only of a reference, obtain the IRI by concatenating the IRI
-								stem associated with the <a href="https://www.w3.org/TR/epub-33/#sec-default-vocab"
-									>default vocabulary</a> [[EPUB-33]] to the reference.</p>
+							<p>If the property consists only of a reference, obtain the URL by concatenating the <a
+									href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> [[URL]] associated
+								with the <a href="https://www.w3.org/TR/epub-33/#sec-default-vocab">default
+									vocabulary</a> [[EPUB-33]] to the reference.</p>
```
rdeltour: base URL is not a string, it cannot be concatenated. The result of serializing the base URL can be concatenated.

epub33/rs/index.html
```diff
@@ -1591,22 +1618,23 @@ <h2>Vocabulary Association Mechanisms</h2>
...
 						</li>
 						<li>
-							<p>If the property consists of a prefix and reference, obtain the IRI by concatenating the
-								IRI stem associated with the prefix to the reference. If the EPUB Creator has not
-								defined a matching prefix, the property is invalid and the Reading System MUST ignore
-								it.</p>
+							<p>If the property consists of a prefix and reference, obtain the URL by concatenating the
```
rdeltour: same as above

epub33/rs/index.html
```diff
@@ -1591,22 +1618,23 @@ <h2>Vocabulary Association Mechanisms</h2>
...
+									<a href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> [[URL]]
+								associated with the prefix to the reference. If the EPUB Creator has not defined a
+								matching prefix, the property is invalid and the Reading System MUST ignore it.</p>
 						</li>
 					</ul>
-					<p>The resulting IRI MUST be valid to [[RFC3987]]. Reading Systems do not have to resolve this IRI,
-						however.</p>
+					<p>The result MUST be a <a href="https://url.spec.whatwg.org/#valid-url-string">valid URL string</a>
+						[[URL]]. Reading Systems do not have to resolve this URL, however.</p>
```
rdeltour: s/resolve/parse/ ?

rdeltour (changes requested):
Relying on the well-defined URL standard is a good move, well done! 👍

I find however that the PR sometimes mixes the serialization terminology (strings) and the concepts (URL record). For instance it often says "resolve relative URL string to absolute URL string" when the URL standard would rather say "parse relative URL string into URL". See the inline comments for details. 😊

epub33/core/index.html
mattgarrish: I like where you're going with this, but the formulation seems to put the requirement on unspecified software doing the parsing. You can also read it as conditional - is it okay for relative URLs not to resolve to resources in the container if they aren't parsed, for example?

Maybe we can switch it around so there's still a hard requirement for all relative urls:

"All [relative-URL-with-fragment strings](https://url.spec.whatwg.org/#relative-url-with-fragment-string) [URL] MUST identify resources within the OCF Abstract Container (i.e., at or below the Root Directory) after [parsing to URL records](https://url.spec.whatwg.org/#concept-url-parser) [URL]."

epub33/core/index.html
mattgarrish: I wonder in this case if we're squeezing too much information into a single sentence.

Would it make sense to separate what textref is for from what its value must be, so reformulate more like the epub:type attribute definition along these lines:

"Identifies an associated fragment of an EPUB Content Document.

The value MUST be a [relative-URL-with-fragment string](https://url.spec.whatwg.org/#relative-url-with-fragment-string) [URL] with a non-null fragment identifier."

Any thoughts on this @marisademeglio ?

epub33/rs/index.html
mattgarrish: Looking again, these probably shouldn't be references to the URL term. The vocabularies have "base URLs" defined for them, but because of the mashup of vocabularies in the authoring spec they're only base URLs in concept (i.e., each vocabulary doesn't actually have its own URL, only string values that reflect what they once were).

epub33/rs/index.html
mattgarrish: "resolve" is typically used in our specs to mean you don't have to go out and get the resource. In this case, both are probably true:

"Reading Systems do not have to parse this URL string [[URL]] or attempt to retrieve the referenced resource, however."

mattgarrish: It does sort of beg the question what happens if the URL string is not valid? Does the property have to be ignored?

But I'll open a separate issue for this.

epub33/rs/index.html
mattgarrish: I've rewritten the sections relating to property data types to not rely on the URL standard. These are effectively CURIEs and even when expanded they don't have to be parsed to URLs.

I've changed the RS section to "Expanding property Data Types" and tried to use this "expanding" phrasing along with "prefix URL" wherever these come up.

@iherman, you probably have the best grasp on whether these changes are in keeping with how curies are used, for example, in RDFa. Do these changes make sense to you?

epub33/rs/index.html
mattgarrish: The URL standard doesn't apply within zipped EPUBs, so I don't think this wording would work. You need some kind of fragment identifier to identify the path within the container. That's why it isn't a precise statement, as the next paragraph goes on to spell out.

Maybe we could make it a bit clearer by rearranging a bit and noting a fragment identifier is involved:

"When an EPUB Publication is zipped, the base URL of the Package Document is obtained from the URL of the EPUB Container together with a fragment identifier that specifies the path to the Package Document (relative to the Root Directory)."

epub33/rs/index.html
mattgarrish: There isn't a universal scheme for referencing into zip files. You could use `jar:`, I believe, or `http(s)` with the path after a hash. Probably other ways not immediately coming to mind.

epub33/rs/index.html
iherman: @mattgarrish looking at [RDFa CURIE Definition](https://www.w3.org/TR/rdfa-core/#s_curies) what you write looks correct to me.

However: the RDFa spec goes into great length specifying what happens if, say the prefix is an empty string, or the referent is empty, etc. Are we sure that those corner cases are covered? I mean we can say that a prefixes MUST be non-empty, and referents MUST be non-empty (it probably makes sense for our cases, which is much simpler than the vocabularies used in RDFa). Are these clearly specified somewhere?

mattgarrish: > the RDFa spec goes into great length specifying what happens if, say the prefix is an empty string

We're much more minimal. If a prefix URL isn't specified or it's not reserved then reading systems must ignore the value. That's why I found it odd that we didn't say anything if the resulting URL is invalid, and why we should correct that to match the rejection when a prefix doesn't have a URL.

> the referent is empty

That's another useful case to note. We should include another bullet to reject this case.

epub33/rs/index.html
rdeltour: ah right. thanks!

epub33/core/index.html
rdeltour: _fragment identifier_ isn't defined in the URL standard. It defines _fragment_ (the component of a URL record) or _URL-fragment string_ (used in the parsing algorithm). If we stay at the string level, the latter should be used. But I'm not sure the term intended to be used outside of the parser algorithm? My gut feeling is that a statements on the URL record would be preferable.

epub33/core/index.html
rdeltour: Strictly speaking it's the URL record that identifies the resource, not the string. But you make good points. Works for me!

epub33/core/index.html
marisademeglio: > "Identifies an associated fragment of an EPUB Content Document.
> 
> The value MUST be a [relative-URL-with-fragment string](https://url.spec.whatwg.org/#relative-url-with-fragment-string) [URL] with a non-null fragment identifier."
> 
> Any thoughts on this @marisademeglio ?

I think this is very clear. One point though, along the lines of what @rdeltour has been mentioning, "non-null" only appears in the parser algorithm section of the URL spec. I think to express this requirement in URL-speak, we would have to say "a fragment identifier with one or more URL Units". Or could we say "non-empty"? 

Alternately, we could revisit the requirement of having fragment identifiers on MO text references. They're useful, sure, but are they necessary to require all the time? Pointing to the document is enough, if the document is very simple and would only have one sync point anyway. And if you need more, you can use fragment IDs, but you don't have to.

epub33/core/index.html
mattgarrish: Would that be clearer if we moved the parsing statement forward so it reads:

All relative-URL-with-fragment strings [URL] MUST, after parsing to URL records [URL], identify resources within the OCF Abstract Container (i.e., at or below the Root Directory).

epub33/core/index.html
mattgarrish: I was intentionally trying to avoid the URL standard in this case. 😄 

We have a section on [fragment identifiers for media overlays](https://w3c.github.io/epub-specs/epub33/core/#sec-media-overlays-fragids), so "fragment identifier" is linked to that section for further elaboration.

Does that work?

mattgarrish: > Alternately, we could revisit the requirement of having fragment identifiers on MO text references.

Do you want to open a new issue for this? Probably not a good idea to get into it in this pull request.

marisademeglio: > > Alternately, we could revisit the requirement of having fragment identifiers on MO text references.
> 
> Do you want to open a new issue for this? Probably not a good idea to get into it in this pull request.

Just leaving a note here for my future self: after sketching out a summary of this to open an issue, I've talked myself out of it -- pretty sure the only people whose lives would be made easier would be spec authors ! It introduces a bit of ambiguity that I don't think is worth it.

marisademeglio: > I was intentionally trying to avoid the URL standard in this case. 😄
> 
> We have a section on [fragment identifiers for media overlays](https://w3c.github.io/epub-specs/epub33/core/#sec-media-overlays-fragids), so "fragment identifier" is linked to that section for further elaboration.
> 
> Does that work?

Yes, so can we omit "non-null", since it seems like all requirements on fragment identifiers then belong in that section 7.3.2.2 ?

As in

"The value MUST be a relative-URL-with-fragment string [URL] with a [fragment identifier](https://w3c.github.io/epub-specs/epub33/core/#sec-media-overlays-fragids)."

I also wonder if the first paragraph of the [fragment identifier section](https://w3c.github.io/epub-specs/epub33/core/#sec-media-overlays-fragids) is necessary. It gives context but it seems to be purely informative.

mattgarrish: > Yes, so can we omit "non-null"

Works for me!

> I also wonder if the first paragraph of the [fragment identifier section](https://w3c.github.io/epub-specs/epub33/core/#sec-media-overlays-fragids) is necessary.

I'd err on the side of leaving it for context. I think it would be confusing for anyone entering the section without following one of the links.

wareid (approved):

shiestyle (approved):

mattgarrish:
I believe everything has been addressed now, so I'm going to go ahead and merge this. New issues are welcome if anyone spots anything amiss.

PR #1671, opened 2021-05-10
Added the text on requiring common origin

### LINKED ISSUE #1659: Is our requirement on (scripting) origin normative or not?
At the moment, the RS document says, in [$15.1](https://w3c.github.io/epub-specs/epub33/rs/#sec-scripted-content-security):

> Reading Systems must behave as if a unique origin [[URL](https://w3c.github.io/epub-specs/epub33/rs/#bib-url)] were allocated to each EPUB Publication. Adopting this approach will isolate publications from each other, thereby limiting access to cookies, DOM storage, etc.

This statement is in a non-normative section and "must" is used as opposed to "_MUST_". As this is the only place this restriction is mentioned, the conclusion is that the requirement is not normative for Reading Systems.

The (normative) [EPUB Content section §3.4](https://w3c.github.io/epub-specs/epub33/core/#sec-scripted-support) says:

> Reading Systems are required to behave as though a unique [origin](https://url.spec.whatwg.org/#origin) [[URL](https://w3c.github.io/epub-specs/epub33/core/#bib-url)] has been assigned to each EPUB Publication.

It says "required" and not "_REQUIRED_" but that is fine, even if the intention is that this feature would be non-normative; after all, the requirement should indeed be in the RS spec and not in the content. However, the reader of the Core spec may believe that this is the only reason that the term is as is, but that the Reading System is indeed _REQUIRED_, ie, that it is a normative requirement. As most of the authors will not read the RS document, this may lead to false expectations. 

Two fundamental question: is the requirement, whereby RS-s must ensure a unique [origin](https://url.spec.whatwg.org/#origin) for all scripts in content documents:

1. **normative**
    - in this case the statements should be added into a normative section of the RS document (because we forgot to do that)
2. **informative**
    - in this case the statement in the Content document should be changed to something that conveys the information that it is by means sure that this restriction is applied. For example:
    > Reading systems may behave as though a unique [origin](https://url.spec.whatwg.org/#origin)"

iherman:
Have added the text to §4.4, per latest [WG resolution](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-05-07-epub#resolution1).

I have added a clarification note to remove ambiguities that may remain; I am not sure whether it is possible to express that in proper normative text...

The restriction is currently on spine level scripts; I am not sure this is also necessary for container constrained scripts (which may have their own security measures). Advice is welcome whether the same restriction should also apply.

Fix #1659

See:

    * [Preview](https://raw.githack.com/w3c/epub-specs/script-origin/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/script-origin/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -777,6 +777,16 @@
 						<p id="confreq-rs-scripted-ua">It MAY render <a>Scripted Content Documents</a> as an
 							interactive, scripted user agent according to [[HTML]].</p>
 					</li>
+
+					<li>
+						<p id="confreq-rs-scripted-origin">
+							It MUST provide a common <a href="https://url.spec.whatwg.org/#origin" >origin</a> [[URL]], shared by all <a href="https://www.w3.org/TR/epub-33/#sec-scripted-spine">spine-level scripts</a> of the EPUB Publication. That <a href="https://url.spec.whatwg.org/#origin" >origin</a> [[URL]] MUST be <em>unique</em> for each EPUB Publication instance. 
+						</p>
+						<p class=note>
+							The unicity of the origin per EPUB Publication instance means that if two different users acquire a copy of the same EPUB Publication, the origins will be different for the two users on those copies even if the same Reading System is used. 
+						</p>
+					</li>
+
 					<li>
 						<p id="confreq-rs-scripted-container">It MUST NOT allow a container-constrained script to modify
 							the DOM of the parent Content Document or other contents in the EPUB Publication and MUST
@@ -2116,6 +2126,9 @@ alert("Feature " + feature + " supported?: " + conformTest);</pre>
 				-->
 
 				<ul>
+					<li>
+						12-May-2021: Providing a shared and unique origin for scripts is a requirements for Reading Systems that support scripting. See <a href="https://github.com/w3c/epub-specs/issues/1659">issue 1659</a>.
+					</li>
 					<li>06-May-2021: Recommend that Reading Systems support references to HTML target elements and SVG
 						fragment identifiers in <code>textref</code> attributes and the <code>text</code> element's
 							<code>src</code> attribute. Support for other fragment identifier schemes is optional. See
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1871,7 +1871,8 @@
 					In handling untrusted scripts this specification recommends that Reading Systems establish 
 					a unique <a href="https://url.spec.whatwg.org/#origin">origin</a> [[URL]] allocated to each 
 					<a>EPUB Publication</a> (see <a href="#sec-scripted-content"></a>). Adopting this
-					approach isolates publications from each other, thereby limiting access to cookies, DOM storage, etc. Examples of Web APIs that are tied to the concept of “origin” include Web Storage [[?WEBSTORAGE]] and 
+					approach isolates publications from each other, thereby limiting access to cookies, DOM storage, etc. 
+					Examples of Web APIs that are tied to the concept of “origin” include Web Storage [[?WEBSTORAGE]] and 
 					IndexedDB [[?INDEXEDDB]], which EPUB content documents can interact with via scripting.
 					This specification recommends the following practices:
 				</p>
@@ -1897,14 +1898,15 @@
 					<li>
 						<p>
 							The unique origin for a given publication (and by extension, for all its content documents)
-							must be preserved across "renders", i.e. when closing/re-opening publications, or when navigating back-and-forth between documents within an opened publication.
+							must be preserved across "renders", i.e. when closing/re-opening publications, or when navigating 
+							back-and-forth between documents within an opened publication.
 						</p>
 					</li>
 
 					<li>
 						<p>
 							Reading Systems must ensure that the data tied to specific <a href="#confreq-rs-scripted-origin">origins</a>
-							(e.g., Web Storage [[?WEBSTORAGE]] or IndexedDB [[?INDEXEDDB]]) is preserved between "reading".
+							(e.g., Web Storage [[?WEBSTORAGE]] or IndexedDB [[?INDEXEDDB]]) is preserved between reading sessions.
 							Reading System implementations that integrate the capabilities of web browser engines (e.g., embedded WebView component) would typically use the provided cache/persistence APIs in order to ensure that origin-bound data is preserved. This specification does not mandate particular mitigation strategies for dealing with size limitations (e.g., cache invalidation, storage reset), but merely recommends to follow best practices as used on the Open Web Platform.
 						</p>
 					</li>
@@ -1930,9 +1932,7 @@
 					<li>
 						<p>
 							Reading Systems that allow local storage should also provide methods for users to inspect,
-							disable, and delete that data. The Reading System must destroy the data if the corresponding
-							EPUB Publication is deleted from its "bookshelf" if it implements one; otherwise it must destroy the
-							data before the Reading System closes.
+							or delete that data. 
 						</p>
 					</li>
 				</ul>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1805,7 +1805,7 @@
 			<section id="sec-scripted-content-security">
 				<h4>Scripting Considerations</h4>
 
-				<p><a>EPUB Creators</a> and Reading System developers who also support scripting  must be aware of the 
+				<p>Reading System developers who also support scripting  must be aware of the 
 					security issues that arise when Reading Systems execute scripted content. As the underlying 
 					scripting model employed by Reading Systems and browsers is the same, developers must take into consideration the same kinds of issues encountered in Web contexts.</p>
 
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -782,9 +782,6 @@
 						<p id="confreq-rs-scripted-origin">
 							It MUST provide a common <a href="https://url.spec.whatwg.org/#origin" >origin</a> [[URL]], shared by all <a href="https://www.w3.org/TR/epub-33/#sec-scripted-spine">spine-level scripts</a> of the EPUB Publication. That <a href="https://url.spec.whatwg.org/#origin" >origin</a> [[URL]] MUST be <em>unique</em> for each EPUB Publication instance. 
 						</p>
-						<p class=note>
-							The unicity of the origin per EPUB Publication instance means that if two different users acquire a copy of the same EPUB Publication, the origins will be different for the two users on those copies even if the same Reading System is used. 
-						</p>
 					</li>
 
 					<li>
@@ -808,6 +805,15 @@
 							contexts.</p>
 					</li>
 				</ul>
+
+				<div class=note>
+					<p>
+						This specification does not mandate any particular implementation technique for the creation of a unique <a href="https://url.spec.whatwg.org/#origin">origin</a> in the absence of a reliable Web origin (e.g., HTTP URL scheme + host + port). The necessary heuristics may include the combination of the Publication's <a href="https://www.w3.org/TR/epub-33/#dfn-unique-identifier">unique identifier</a> (although, in practice, these identifiers may in fact not guarantee globally-unique identification, that is why it is recommended to combine multiple techniques), filesystem path, <a href="https://www.w3.org/TR/epub-33/##dfn-zip-container">OCF Zip Container</a> checksum, etc.
+					</p>
+					<p>
+						The unicity of the <a href="https://url.spec.whatwg.org/#origin" >origin</a> per EPUB Publication instance means that if two different users acquire a copy of the same EPUB Publication, the origins will be different for the two users on those copies even if the same Reading System is used. 
+					</p>
+				</div>
 
 				<p class="ednote">Note that the definition of <code>epubReadingSystem</code> is currently marked as "at
 					risk". If, in the final version of this document, the object becomes non-normative, then each "MUST"
@@ -1771,7 +1777,7 @@
 			<p>The particularity of an <a>EPUB Publication</a> is its structure. The EPUB format provides a means of
 				representing, packaging, and encoding structured and semantically enhanced Web content — including HTML,
 				CSS, SVG, and other resources — for distribution in a single-file container. This specification does not
-					<em>add</em> any new features, APIs, etc. This also means that, essentially, the security and
+					<em>add</em> any new features, complex APIs, etc. This also means that, essentially, the security and
 				privacy issues are reliant on the features of those formats. In particular:</p>
 
 			<ul>
@@ -1799,10 +1805,9 @@
 			<section id="sec-scripted-content-security">
 				<h4>Scripting Considerations</h4>
 
-				<p><a>EPUB Creators</a> and Reading System developers must be aware of the security issues that arise
-					when Reading Systems execute scripted content. As the underlying scripting model employed by Reading
-					Systems and browsers is the same, developers must take into consideration the same kinds of issues
-					encountered in Web contexts.</p>
+				<p><a>EPUB Creators</a> and Reading System developers who also support scripting  must be aware of the 
+					security issues that arise when Reading Systems execute scripted content. As the underlying 
+					scripting model employed by Reading Systems and browsers is the same, developers must take into consideration the same kinds of issues encountered in Web contexts.</p>
 
 				<p>Each Reading System must establish if it can trust the scripts in a particular document. Reading
 					Systems should treat all scripts as untrusted (and potentially malicious), and developers should
@@ -1832,33 +1837,72 @@
 					</li>
 				</ul>
 
-				<p>This specification recommends the following practices for handling untrusted scripts:</p>
-
+				<p>
+					In handling untrusted scripts this specification recommends that Reading Systems establish 
+					a unique <a href="https://url.spec.whatwg.org/#origin">origin</a> [[URL]] allocated to each 
+					<a>EPUB Publication</a> (see <a href="#sec-scripted-content"></a>). Adopting this
+					approach isolates publications from each other, thereby limiting access to cookies, DOM storage, etc. Examples of Web APIs that are tied to the concept of “origin” include Web Storage [[?WEBSTORAGE]] and 
+					IndexedDB [[?INDEXEDDB]], which EPUB content documents can interact with via scripting.
+					This specification recommends the following practices:
+				</p>
 				<ul>
+
 					<li>
-						<p>Reading Systems must behave as if a unique <a href="https://url.spec.whatwg.org/#origin"
-								>origin</a> [[URL]] were allocated to each <a>EPUB Publication</a>. Adopting this
-							approach will isolate publications from each other, thereby limiting access to cookies, DOM
-							storage, etc.</p>
-						<div class="note">
-							<p>Although domain isolation is the most effective way to secure EPUB Publications in
-								Web-based Reading Systems, it is not always a realistic option. Web-based Reading
-								Systems must still maintain isolation in such cases, which may require limiting access
-								to scripting APIs that allow client-side data storage and access.</p>
-						</div>
-						<p>Reading Systems that enable scripting and network access should also include methods to
-							notify the user that network activity is occurring and/or that allow them to disable it.</p>
+						<p>
+							Although domain isolation is the most effective way to secure EPUB Publications in
+							Web-based Reading Systems, it is not always a realistic option. Web-based Reading
+							Systems must still maintain isolation in such cases, which may require limiting access
+							to scripting APIs that allow client-side data storage and access.
+						</p>
 					</li>
 					<li>
-						<p>If a Reading System allows users to store persistent data, it must treat that data as
+						<p>
+							Reading Systems that allow users to add/remove publications from a managed library 
+							(a.k.a “bookshelf”) may maintain the publication’s unique origin when the publication is
+							removed and subsequently re-imported into the content library. Conversely, Reading
+							Systems may create a new unique origin for every newly-added publication.
+						</p>
+					</li>
+
+					<li>
+						<p>
+							The unique origin for a given publication (and by extension, for all its content documents)
+							must be preserved across "renders", i.e. when closing/re-opening publications, or when navigating back-and-forth between documents within an opened publication.
+						</p>
+					</li>
+
+					<li>
+						<p>
+							Reading Systems must ensure that the data tied to specific <a href="#confreq-rs-scripted-origin">origins</a>
+							(e.g., Web Storage [[?WEBSTORAGE]] or IndexedDB [[?INDEXEDDB]]) is preserved between "reading".
+							Reading System implementations that integrate the capabilities of web browser engines (e.g., embedded WebView component) would typically use the provided cache/persistence APIs in order to ensure that origin-bound data is preserved. This specification does not mandate particular mitigation strategies for dealing with size limitations (e.g., cache invalidation, storage reset), but merely recommends to follow best practices as used on the Open Web Platform.
+						</p>
+					</li>
+
+					<li>
+						<p>
+							Reading Systems that support scripting and network access should also include
+							methods to notify the user that network activity is occurring and/or that allow them to disable it.
+						</p>
+					</li>
+
+					<li>
+						<p>
+							If a Reading System allows users/scripts to store persistent data, it must treat that data as
 							sensitive. Scripts may save persistent data through cookies and DOM storage but Reading
 							Systems may block such attempts. Reading Systems that allow users to store data must ensure
 							they do not make that data available to other unrelated documents (e.g., ones that could be
 							spoofed). In particular, checking for a matching document identifier (or similar metadata)
-							is not a valid method to control access to persistent data.</p>
-						<p>Reading Systems that allow local storage should also provide methods for users to inspect,
+							is not a valid method to control access to persistent data.
+						</p>
+					</li>
+
+					<li>
+						<p>
+							Reading Systems that allow local storage should also provide methods for users to inspect,
 							disable, and delete that data. The Reading System must destroy the data if the corresponding
-							EPUB Publication is deleted.</p>
+							EPUB Publication is deleted from its "bookshelf".
+						</p>
 					</li>
 				</ul>
 
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1898,8 +1898,7 @@
 					<li>
 						<p>
 							The unique origin for a given publication (and by extension, for all its content documents)
-							should, when appropriate, be preserved across reading sessions, i.e. when closing/re-opening publications, or when navigating 
-							back-and-forth between documents within an opened publication.
+							should, when appropriate, be preserved across reading sessions. This may include when publications are closed and re-opened, or when navigating back-and-forth between documents within an opened publication.
 						</p>
 					</li>
 
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1901,7 +1901,8 @@
 						<p>
 							Reading Systems that allow local storage should also provide methods for users to inspect,
 							disable, and delete that data. The Reading System must destroy the data if the corresponding
-							EPUB Publication is deleted from its "bookshelf".
+							EPUB Publication is deleted from its "bookshelf" if it implements one; otherwise it must destroy the
+							data before the Reading System closes.
 						</p>
 					</li>
 				</ul>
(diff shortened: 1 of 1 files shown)


iherman:
I have also updated the security section (and added a note to the normative part) based on the [comments](https://github.com/w3c/epub-specs/issues/1659#issuecomment-836563415) of @danielweck

mattgarrish:
Didn't we discuss earlier that unique origins aren't realistic for all web-based reading systems?

I recall the example @lrosenthol gave was dropbox, which can't assign a unique origin to each publication a user loads into their space.

Won't this proposal invalidate such reading systems?

epub33/rs/index.html
```diff
@@ -1761,7 +1777,7 @@ <h2>Security and Privacy</h2>
 			<p>The particularity of an <a>EPUB Publication</a> is its structure. The EPUB format provides a means of
 				representing, packaging, and encoding structured and semantically enhanced Web content — including HTML,
 				CSS, SVG, and other resources — for distribution in a single-file container. This specification does not
-					<em>add</em> any new features, APIs, etc. This also means that, essentially, the security and
+					<em>add</em> any new features, complex APIs, etc. This also means that, essentially, the security and
```
mattgarrish, line 1818: Aren't most APIs pretty complex to the average author? ;)

It's not clear what this addition is getting at, though? (I'm assuming epubReadingSystem object?)

The sentence seems inaccurate taken more broadly, though. We've added the SSML attributes, the (now deprecated but still available) epub:* elements, eRS object, etc.

epub33/rs/index.html
```diff
@@ -1789,10 +1805,9 @@ <h2>Security and Privacy</h2>
 			<section id="sec-scripted-content-security">
 				<h4>Scripting Considerations</h4>
 
-				<p><a>EPUB Creators</a> and Reading System developers must be aware of the security issues that arise
-					when Reading Systems execute scripted content. As the underlying scripting model employed by Reading
-					Systems and browsers is the same, developers must take into consideration the same kinds of issues
-					encountered in Web contexts.</p>
+				<p><a>EPUB Creators</a> and Reading System developers who also support scripting  must be aware of the
```
mattgarrish: We don't really need "epub creators" in this context anymore. It's the RS spec.

epub33/rs/index.html
```diff
@@ -1822,33 +1837,72 @@ <h4>Scripting Considerations</h4>
...
-							<p>Although domain isolation is the most effective way to secure EPUB Publications in
-								Web-based Reading Systems, it is not always a realistic option. Web-based Reading
-								Systems must still maintain isolation in such cases, which may require limiting access
-								to scripting APIs that allow client-side data storage and access.</p>
-						</div>
-						<p>Reading Systems that enable scripting and network access should also include methods to
-							notify the user that network activity is occurring and/or that allow them to disable it.</p>
+						<p>
+							Reading Systems that allow users to add/remove publications from a managed library 
+							(a.k.a “bookshelf”) may maintain the publication’s unique origin when the publication is
```
mattgarrish: If we're going to start using "bookshelf" (which I see also used later) we should fully define what we mean by it in the terminology.

epub33/rs/index.html
iherman: I is indeed referring to `epubReadingSystem`, meaning that we do define a very simple API. Ie, the statement not _adding_ an API is wrong. But the API is dead simple.

I am not sure the epub:* elements are relevant. I believe that statement refers to, essentially, Javascript API-s (those are the ones that are relevant for the origin issue).

epub33/rs/index.html
iherman: Well, it is in a non-normative section (and in quotes), I believe we can get by without a formal definition... I do not we use it normatively.

epub33/rs/index.html
```diff
@@ -1822,33 +1837,72 @@ <h4>Scripting Considerations</h4>
...
+							is not a valid method to control access to persistent data.
+						</p>
+					</li>
+
+					<li>
+						<p>
+							Reading Systems that allow local storage should also provide methods for users to inspect,
 							disable, and delete that data. The Reading System must destroy the data if the corresponding
-							EPUB Publication is deleted.</p>
+							EPUB Publication is deleted from its "bookshelf".
```
mattgarrish: What if the reading system doesn't use a "bookshelf"? Shouldn't the requirement be the same if the user closes off such an application?

epub33/rs/index.html
iherman: Correct. Removed it in https://github.com/w3c/epub-specs/pull/1671/commits/86beb17f4004082c39aeaf05537832ad23fad694

iherman:
> Didn't we discuss earlier that unique origins aren't realistic for all web-based reading systems?
> 
> I recall the example @lrosenthol gave was dropbox, which can't assign a unique origin to each publication a user loads into their space.
> 
> Won't this proposal invalidate such reading systems?

There was a remark referring to this, and that remark was kept in the security consideration section (first bullet item).

I would not be shy to say (I believe @dauwhe referred to something like that on our call) that those Reading Systems may not be 100% conformant, if that is indeed the case. I am not sure that it would be a generic impossibility, though; afaik, Readium does not have a problem with this requirement and the tool is running in a browser (@llemeurfr, is that correct?). I would also like the feedback of @fchasen on this...

epub33/rs/index.html
iherman: Yep, true.

Maybe say:

> The Reading System must destroy the data if the corresponding EPUB Publication is deleted from its "bookshelf" if it implements one; otherwise it must destroy the data before the Reading System closes.

WDYT?

llemeurfr:
> Didn't we discuss earlier that unique origins aren't realistic for all web-based reading systems?

In its comments @danielweck was careful to state that "reading systems must assign publications a unique "origin"". Meaning that the RS creates a "closure" it can manage correctly in an unsafe "region" (e.g the user storage). There is no real security there, afaik, only a responsible use of a shared storage resource.

mattgarrish:
> was careful to state that "reading systems must assign publications a unique "origin"". Meaning that the RS creates a "closure" it can manage correctly in an unsafe "region" (e.g the user storage).

Right, and this is why we added the clarification that a reading system that can't actually assign a unique origin still has to restrict scripting as though one had been assigned (and this was all informative).

But normatively we're saying:

> It MUST provide a common <a href="https://url.spec.whatwg.org/#origin" >origin</a> [[URL]], shared by all <a href="https://www.w3.org/TR/epub-33/#sec-scripted-spine">spine-level scripts</a> of the EPUB Publication.

I wouldn't interpret using closure as being in conformance with this requirement.

But if I'm over-reading the implications of this normative requirement I'll just shut up... 😄

epub33/rs/index.html
mattgarrish: SGTM

epub33/rs/index.html
mattgarrish: > I believe that statement refers to, essentially, Javascript API-s

It's not in a scripting-specific section but the very first paragraph of the privacy and security section describing EPUB 3 as a format.

We have added various features (and that api) that don't exist on the OWP, even if we've been slowly trying to walk them back. So we own any security and privacy issues that might arise from those extensions.

It would probably be better to enumerate the things we've added and state why we don't believe they add any new issues.

mattgarrish: But this is probably better done as a separate issue.

epub33/rs/index.html
iherman: Done in https://github.com/w3c/epub-specs/pull/1671/commits/a3cee4e2ae3e9b54e53d4866d7863604667cae05

iherman:
> > was careful to state that "reading systems must assign publications a unique "origin"". Meaning that the RS creates a "closure" it can manage correctly in an unsafe "region" (e.g the user storage).
> 

I have changed the requirement to

> It MUST assign a unique origin [[URL]], shared by all spine-level scripts of the EPUB Publication.

to be in line with the text of @danielweck that @llemeurfr quoted, but I am not sure that answers the problem of @mattgarrish.

But, looking at the text quited by @mattgarrish 

> we added the clarification that a reading system that can't actually assign a unique origin still has to restrict scripting as though one had been assigned (and this was all informative).

As far as I am concerned, what really counts is that, for a script, the _observable behaviour_ (e.g., when doing a fetch) of the environment is as if a unique origin was present. How exactly this is done by the RS is an internal feature of the system whose details are out of scope of the spec. I.e., if  "restrict scripting as though one had been assigned" is achieved then, in my view, that makes the RS conformant. What am I missing?

@bduga @fchasen @llemeurfr ?

mattgarrish:
> if "restrict scripting as though one had been assigned" is achieved then, in my view, that makes the RS conformant.

The problem I have is that the only accepted implementation is an origin as defined by the URL standard, which requires a unique scheme, host, port and domain. Anything else is invalid, no matter whether it approximates the behaviour of a unique origin or not.

I'd expect a normative rephrasing of the original text to be more along the lines of:

"A Reading System SHOULD assign a unique origin [[URL]] shared by all spine-level scripts of the EPUB Publication. If a unique origin cannot be assigned, a Reading System MUST implement a similar means of ensuring that spine-level scripts cannot interact with other EPUB Publications."

iherman:
> "A Reading System SHOULD assign a unique origin [[URL]] shared by all spine-level scripts of the EPUB Publication. If a unique origin cannot be assigned, a Reading System MUST implement a similar means of ensuring that spine-level scripts cannot interact with other EPUB Publications (e.g., through JavaScript closures)."

I am sympathetic to this text, but what exactly do you mean by "through JavaScript closures"? Also, if this approach is used, does it change the perceivable behaviour of scripts? Or will, say, a fetch or a Web Storage API behave the same way?

mattgarrish:
> what exactly do you mean by "through JavaScript closures"

I took that from @llemeurfr's comment above that closures are used to isolate the scripting. I assumed that was a reference to JS closures (scoping the context of what the publication's scripts can access), but maybe the quotes around it are suggesting something else.

mattgarrish:
> Or will, say, a fetch or a Web Storage API behave the same way?

Reading systems with scripting support don't necessarily allow authors to use APIs. In such cases, why would it matter if a unique origin has been assigned to the epub or not?

That's why I wonder if unique origins are the sole answer to this problem.

iherman:
> > Or will, say, a fetch or a Web Storage API behave the same way?
> 
> Reading systems with scripting support don't necessarily allow authors to use APIs. In such cases, why would it matter if a unique origin has been assigned to the epub or not?

Do not _necessarily_. But we are relying on the fact that scripting means using the Open Web Platform, meaning also the APIs. This does not mean that the RS must implement all the available APIs; very few browsers do that, I presume. But if they do implement a specific API, then it should behave like on the OWP. Ie, scripts should also behave v.a.v. the concept of origin as they do in a browser with the particularity of a common origin per EPUB instance...

> 
> That's why I wonder if unique origins are the sole answer to this problem.

There may be other security issues that are unrelated to origin. But the origin concept is certainly at the heart of the OWP security model...

(We would really need the input of RS implementers here...)

llemeurfr:
>  I took that from @llemeurfr's comment above that closures are used to isolate the scripting. I assumed that was a reference to JS closures 

Sorry, I wasn’t clear enough; I used the term closure in a very broad sense, not the JS technique. Tried to express that a RS makes its best to isolate the info associated with each publication, without being able to rely on pure web mechanisms to do so.

iherman:
> > I took that from @llemeurfr's comment above that closures are used to isolate the scripting. I assumed that was a reference to JS closures
> 
> Sorry, I wasn’t clear enough; I used the term closure in a very broad sense, not the JS technique. Tried to express that a RS makes its best to isolate the info associated with each publication, without being able to rely on pure web mechanisms to do so.

Thanks, @llemeurfr. But does that mean that if the RS uses this approach, then, a script running in this RS cannot rely on the standard origin-related mechanism for security? Ie, does that mean that it becomes impossible for a script to be truly interoperable (if it relies on the concept of origin, that is)?

epub33/rs/index.html
```diff
@@ -1852,33 +1867,73 @@ <h4>Scripting Considerations</h4>
 					</li>
 				</ul>
 
-				<p>This specification recommends the following practices for handling untrusted scripts:</p>
-
+				<p>
+					In handling untrusted scripts this specification recommends that Reading Systems establish
```
bduga: This whole section confuses me, since I thought line 802 was requiring unique origins. But here we say it is recommended and explain alternatives. What am I missing?

epub33/rs/index.html
```diff
@@ -1852,33 +1867,73 @@ <h4>Scripting Considerations</h4>
...
+							(a.k.a “bookshelf”) may maintain the publication’s unique origin when the publication is
+							removed and subsequently re-imported into the content library. Conversely, Reading
+							Systems may create a new unique origin for every newly-added publication.
+						</p>
+					</li>
+
+					<li>
+						<p>
+							The unique origin for a given publication (and by extension, for all its content documents)
+							must be preserved across "renders", i.e. when closing/re-opening publications, or when navigating back-and-forth between documents within an opened publication.
```
bduga: Not sure we need this constraint, and it may not really be true. For instance, in a shared use case it might make sense to assign different origins for each reader of the same local content. I am not saying that is a great design, but I also haven't given enough thought to possible designs to decide if this makes sense. This kind of seems like a recommendation to not break things.

epub33/rs/index.html
```diff
@@ -1852,33 +1867,73 @@ <h4>Scripting Considerations</h4>
...
-							<p>Although domain isolation is the most effective way to secure EPUB Publications in
-								Web-based Reading Systems, it is not always a realistic option. Web-based Reading
-								Systems must still maintain isolation in such cases, which may require limiting access
-								to scripting APIs that allow client-side data storage and access.</p>
-						</div>
-						<p>Reading Systems that enable scripting and network access should also include methods to
-							notify the user that network activity is occurring and/or that allow them to disable it.</p>
+						<p>
+							Reading Systems must ensure that the data tied to specific <a href="#confreq-rs-scripted-origin">origins</a>
+							(e.g., Web Storage [[?WEBSTORAGE]] or IndexedDB [[?INDEXEDDB]]) is preserved between "reading".
```
bduga: Not sure what 'preserved between "reading"' means. Between reading sessions? Though also same comment as above - should we really be recommending this?

epub33/rs/index.html
```diff
@@ -1852,33 +1867,73 @@ <h4>Scripting Considerations</h4>
...
-							is not a valid method to control access to persistent data.</p>
-						<p>Reading Systems that allow local storage should also provide methods for users to inspect,
+							is not a valid method to control access to persistent data.
+						</p>
+					</li>
+
+					<li>
+						<p>
+							Reading Systems that allow local storage should also provide methods for users to inspect,
 							disable, and delete that data. The Reading System must destroy the data if the corresponding
```
bduga: Preexisting, but I am not entirely sure how data is disabled. I guess not deleted but somehow not accessible? I am not at all sure how to implement that.

epub33/rs/index.html
```diff
@@ -1852,33 +1867,73 @@ <h4>Scripting Considerations</h4>
...
-							EPUB Publication is deleted.</p>
+							EPUB Publication is deleted from its "bookshelf" if it implements one; otherwise it must destroy the
```
bduga: This should probably be deleted. Unless we can specify what it means to "delete" content from a "bookshelf", this is about as meaningful as saying RSes should delete data when it should delete data.

epub33/rs/index.html
iherman: Hm. The way I read that section it does not offer alternatives but some additional considerations that RS should take into account. Which part do you feel explain alternatives?

cc @danielweck

epub33/rs/index.html
iherman: @danielweck I am happy to remove this bullet item (@bduga's arguments are compelling); what say you?

epub33/rs/index.html
iherman: I think "reading session" is a much better term. I will change this.

epub33/rs/index.html
iherman: +1

epub33/rs/index.html
iherman: I agree, I am not sure what "disabling" means. I will remove that.

epub33/rs/index.html
iherman: Ah, I think I know what you refer to: the first bullet item right after the quoted paragraph.

This is a leftover of the previous version that I did not remove. If you look at the discussion thread above, the main question is whether that paragraph is needed in the first place, ie, whether the normative requirement of how to handle origins needs some informative text for RSs that, for some reasons, cannot implement what is required.

Re-reading the whole set, even if that type of text _is_ needed, it may be a better option to add it to section 4.4 (a.k.a. line 802) as a (non normative) note. That would make it clearer that it is an advice for the RSs that cannot be fully conformant, but it would remove it from the security section where, I agree, it muddy the waters.

WDYT?

epub33/rs/index.html
danielweck: Well, there are real-world use-cases behind this Reading System conformance requirement, such as:
* EPUBs scripted documents that prompt users to enter their name into an text input field, at an early stage in the reading experience (typically, in the first interactive page after the static cover / intro material), in order to personalise ebook content further into the story narrative. The entered value is preserved in origin-bound LocalStorage.
* EPUBs that store "state" information such as quiz answers, challenge progress, etc. The data structure is saved in origin-bound IndexedDB.
* EPUBs that cache some analytics data, such as "first time" publication or document read event.
* etc.

That being said, there must clearly also be cases where the origin model doesn't matter or is not desirable (due to legacy technical reasons, authoring choices, reading system designs ... or combinations of). So I am not sure how to best reconciliate this, and what kind of testable specification prose can usefully be added / clarified to EPUB v3.x

iherman: First of all, let us not forget that all these points are part of [§15 Security and Privacy](https://w3c.github.io/epub-specs/epub33/rs/#sec-security-privacy), ie, a _non-normative_ section. In other words, none of these remarks are formal RS conformance requirements, but more a matter of guidance, good practices, that implementers should consider in relations to security and privacy. Hence the usage of “must” in that item as opposed to _“MUST”_.

With that in mind, would it be agreeable to all if we keep that item but replace “must” by “should”?

iherman:
An idea... one of the issue we have is whether all RSs are able to implement the strict origin behaviour as required in the current branch. The fact seems to be (see also my question in https://github.com/w3c/epub-specs/pull/1671#issuecomment-842173108) whether the alternative referred to in the current bullet item (see also @bduga's comment https://github.com/w3c/epub-specs/pull/1671#pullrequestreview-661301042) creates a different environment for a script.

If the answer to this question is 'yes', would it make sense to add yet another "feature" in the EpubReadingSystem definition on whether the RS implements the common origin requirement or not? Scripts may then be able to adapt themselves...

epub33/rs/index.html
bduga: I am fine with replacing must with should, or even "when appropriate". Or even list the specific cases @danielweck mentioned (but that seems harder and longer). Though, looking at that list,  *none* of those reasons seem to be related to security and privacy, except perhaps making things less private. Looking more closely, I don't understand the security and privacy implications of the 3rd and fourth bullets at all. These seem like general tips and tricks for scripting support.

epub33/rs/index.html
bduga: I think it is strange to requires something with a MUST, but then acknowledge there are good reasons to not implement the MUST and provide alternative. If there exist good reasons why something should not be implemented, it seems the proper conformance statement is SHOULD, right? I think we should either require it, full stop, or strongly recommend it and give advice on what to do if you can't.

epub33/rs/index.html
danielweck: > security and privacy

Well, content creators (HTML/JS/CSS developers) expect LocalStorage / IndexedDB to be isolated on a same-origin basis. When a reading system does not ensure distinct origins for different publications, data leaks in such a way that a publication script can (intentionally or inadvertantly) read and write data created in the context of another publication.

This is both a privacy and security concern, wouldn't you agree?

bduga: I do agree that data leaks are related to privacy and security, but the bullet items I mention don't talk about leaks. The third bullet discusses preserving origins across reading sessions, and the fourth discusses preserving data across reading sessions. While both of those might be important for a general scripting guidance chapter, they don't seem appropriate to a security and privacy one.

epub33/rs/index.html
iherman: +1 for requiring it, full stop.

iherman:
> An idea... one of the issue we have is whether all RSs are able to implement the strict origin behaviour as required in the current branch. The fact seems to be (see also my question in [#1671 (comment)](https://github.com/w3c/epub-specs/pull/1671#issuecomment-842173108)) whether the alternative referred to in the current bullet item (see also @bduga's comment [#1671 (review)](https://github.com/w3c/epub-specs/pull/1671#pullrequestreview-661301042)) creates a different environment for a script.
> 
> If the answer to this question is 'yes', would it make sense to add yet another "feature" in the EpubReadingSystem definition on whether the RS implements the common origin requirement or not? Scripts may then be able to adapt themselves...

Having talked about it with @dauwhe he made me realize this would be a bad idea. What it says is: "can I inquire whether the reading system is secure?".

I hereby withdraw my idea :-)

epub33/rs/index.html
iherman: @bduga @danielweck 

Looking at the list in §15.1:

- The first item may be removed (or moved elsewhere), see https://github.com/w3c/epub-specs/pull/1671#discussion_r633783893

- I tend to agree with @bduga that item 3 ("The unique origin for a given publication...") and item 4 ("Reading Systems must ensure that the data tied to") are not really privacy/security issue. The question is whether they are indeed valuable notes, however, that we want to keep. Some sort of a "usability" notes for RS implementers. Not being a RS implementer, I cannot judge whether those are useful or whether RS implementers would just shrug seeing those items. But indeed, it should probably be moved... elsewhere, but I do not know where.

Do we, should we plan some sort of RS implementers' guide? @dauwhe @wareid @mattgarrish

dauwhe: > Do we, should we plan some sort of RS implementers' guide?

I don't think so. We already have plenty of work. We do not have demand for such a thing from potential implementers.

iherman: @dauwhe I thought so, but I wanted to be sure :-)

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-05-21-epub#section3) on 2021-05-21

List of resolutions:

- ***[Resolution No. 3](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-05-21-epub#resolution3):  get rid of the hand-waving in PR 1671 on the origin issue***

### 3. Is our scripting requirement normative? 

_See github pull request [#1671](https://github.com/w3c/epub-specs/pull/1671)._




_See github issue [#1659](https://github.com/w3c/epub-specs/issues/1659)._




**Dave Cramer:** Security and Privacy - trying to bring more rigor  

**Ivan Herman:** Unique origin (per epub) is something we have already decided  
… Implementing that resolution I hit two things  
… one editorial, one serious  
… There was a statement which said roughly "you should have a unique origin, but if you can't then do some hacks" (non-normative)  

> *Dave Cramer:* [https://w3c.github.io/epub-specs/epub33/rs/#sec-scripted-content-security](https://w3c.github.io/epub-specs/epub33/rs/#sec-scripted-content-security)

> *:* domain isolation is the most effective way to secure EPUB Publications in Web-based Reading Systems, it is not always a realistic option. Web-based Reading Systems must still maintain isolation in such cases, which may require limiting access to scripting APIs that allow client-side data storage and access.”

**Ivan Herman:** Didn't know what to do with that statement  
… I kept it to security and privacy, but there was a long thread of discussion on the request about what to do with it  
… General hand waving over what to do if you can't handle origins  
… There are several options about how to write it  
… but is tending towards the recommendation to dump it entirely  
… since it is now required to support origin  

**Brady Duga:** to add context, we did not have requirement around origin in the past  
… when we added that requirement, we needed to take out the lines that said it wasn't required. We have to revisit whether this isn't required if we take this out  

**Dave Cramer:** My instinct is to keep the normative requirement and get rid of the hand-waving  
… acknowledging that the architecture of some RSes might have issues with this  
… Don't see the need for the spec to recognize that  
… I would hope there are some tests that check the spirit of this  
… Really don't want cross-script data sharing  
… The web way to do that is to use distinct origins  
… If a RS can pass those tests then I am happy with it, regardless of how it is actually implemented  

> **Proposed resolution: get rid of the hand-waving in PR 1671 on the origin issue** *(Ivan Herman)*

> *Brady Duga:* +1

> *Ivan Herman:* +1

> *Tzviya Siegman:* +1

> *Dave Cramer:* +1

> *Toshiaki Koike:* +1

> *Matt Garrish:* 0

> *Gregorio Pellegrino:* +1

> *Ben Schroeter:* +1

> *Masakazu Kitahara:* +1

> ***Resolution #3: get rid of the hand-waving in PR 1671 on the origin issue***

**Ivan Herman:** Then there was also a more editorial question  
… the entire section has been enriched with more comments  
… but those items, while interesting, are not related to security and privacy  
… comments are around keeping data, which may be correct but is not S&P related  
… Is there a place to move this to? eg general best practices for scripting  
… if yes, then we should keep and enhance, otherwise we need to drop them  

**Dave Cramer:** I don't think we should do something like that (best practices)  
… These seem out of scope for the spec  
… Presumably people will figure it out  

**Brady Duga:** All of these comments are around preserving data, but the spec does not require that  
… you don't have to use web storage, etc  
… It would be tricky to write tips and tricks for specs that are not required  

> *Tzviya Siegman:* This is very niche - tips for supporting optional specs on top of optional features.

**Dave Cramer:** Don't feel qualified to recommend implementation details to devs  
… Seem to be drifting away from core responsibilities of the spec, would prefer fewer words here  

**Ivan Herman:** Essentially in agreement  
… Don't need a formal resolution (this isn't a normative statement), but will go ahead and remove those items  

**Hadrien Gardeur:** Question about previous item: have we discussed what APIs can be used?  
… Is there a policy or rough idea about our position?  

**Brady Duga:** historically, we've viewed scripting as experimental. We've left it up to authors, publishers, systems to decide what to support. It's a strange place to allow third party to decide what to do with user data. It's a scary place to be, and I don't think we should put into spec.  

**Ivan Herman:** An analogy - no one specifies which APIs must be implemented by a browser  
… It is rare (never?) to require specific API implementations  
… Just like a browser, it is up to RS authors to decide what they will implement  

</details>

iherman:
@mattgarrish I have implemented the resolutions of the call; can you make a last minute check and, if things are fine, I believe this can be merged.

epub33/rs/index.html
mattgarrish: I still find this text confusing. Why would we only recommend unique origins to handle untrusted scripts but require them for all scripting?

Shouldn't this say something more like: "To limit the damage of untrusted scripts, this specification requires Reading Systems assign a unique origin ..."

epub33/rs/index.html
```diff
@@ -1849,39 +1875,56 @@ <h4>Scripting Considerations</h4>
...
-				<p>This specification recommends the following practices for handling untrusted scripts:</p>
-
+				<p>
+					In handling untrusted scripts this specification recommends that Reading Systems establish 
+					a unique <a href="https://url.spec.whatwg.org/#origin">origin</a> [[URL]] allocated to each 
+					<a>EPUB Publication</a> (see <a href="#sec-scripted-content"></a>). Adopting this
+					approach isolates publications from each other, thereby limiting access to cookies, DOM storage, etc. 
+					Examples of Web APIs that are tied to the concept of “origin” include Web Storage [[?WEBSTORAGE]] and 
+					IndexedDB [[?INDEXEDDB]], which EPUB content documents can interact with via scripting.
+					This specification recommends the following practices:
```
mattgarrish: This sentence kind of comes out of the blue and isn't clear why the practices are recommended.

Seems like it's missing an "in addition" and/or some connection to untrusted scripts.

iherman: I have adopted your first proposal.

For the second, I actually did some more, separating the security/'origin' problems from the rest of the bullet items which were not really origin related but more privacy one. Can you have a look?

(If you are o.k. with the text, go ahead and merge it...)

PR #1682, opened 2021-05-25
Update guidance around roles and epub:type

mattgarrish:
This PR tries to better match the SEM-001 technique to what we now recommend and validate - namely that role and epub:type do not have to appear together.

I've made the section focus on the use of ARIA role and noted that roles are only a requirement for user interface components.

The epub:type explanation is now in a separate subsection.

Suggestions on how to improve the text welcome.

- Techniques [preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/a11y-epub-type/epub33/a11y-tech/index.html)
- Techniques [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/main/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/a11y-epub-type/epub33/a11y-tech/index.html)

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -756,7 +756,7 @@
 						these requirements. For complete details, refer to <a href="https://www.w3.org/TR/html-aria/"
 							>ARIA in HTML</a> [[HTML-ARIA]].</p>
 
-					<p>EPUB Creators only have to use ARIA roles for interactive controls per <a
+					<p>EPUB Creators only have to use ARIA roles for user interface components per <a
 							href="https://www.w3.org/TR/WCAG/#name-role-value">success criterion 4.1.2</a> (i.e., they
 						are not required to add roles to landmarks and other document structures to conform).</p>
 
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -1549,7 +1549,11 @@
 						- move all changes down to the next section
 				-->
 
-				<ul> </ul>
+				<ul>
+					<li>25-May-2021: Updated the section on ARIA roles to make clear that the <code>epub:type</code>
+						attribute does not have to be used in coordination, nor that roles are required where
+							<code>epub:type</code> is used.</li>
+				</ul>
 			</section>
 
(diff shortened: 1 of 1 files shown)


clapierre (commented):
>&lt;p&gt;EPUB Creators only have to use ARIA roles for interactive controls per &lt;a							href="https://www.w3.org/TR/WCAG/#name-role-value"&gt;success criterion 4.1.2&lt;/a&gt; (i.e., they						are not required to add roles to landmarks and other document structures to conform).&lt;/p&gt;

Doesn't this imply the author shouldn't use role="doc-chapter" , role="doc-glossary", etc. since they are not required to add roles to landmarks and other document structures? 

Not sure I like this.  We want to encourage authors to use these, as this goes beyond the traditional ARIA role and adds semantic structure to the documents which will be leveraged as AT and reading systems evolve.

We are currently working on this with DAISY and the TIES group to determine the current DPUB ARIA roles exposed to AT and how this can be improved for a better overall experience.

mattgarrish:
> Doesn't this imply the author shouldn't use role="doc-chapter" , role="doc-glossary", etc. since they are not required to add roles to landmarks and other document structures?

The techniques are supposed to explain how to meet WCAG for EPUB-specific peculiarities, so it's already a bit unclear why we're discussing ARIA roles when there's nothing different about them.

I'm tempted to throw out any discussion of them and simply leave the section clarifying epub:type.

Would that be less concerning to you?

mattgarrish:
Have a look at the revised text. I've only kept the epub:type differentiation prose and noted that requirements for roles are not satisfied by using epub:type.

clapierre (approved):
I like this a lot better Matt.  We can then come out with some Best practices around the use of DPUB ARIA roles which won't look to be in conflict with this.

Thanks!

mattgarrish:
I'm going to merge this change now as it's been a week since we discussed. Please open a new issue if you find anything you want changed.

PR #1683, opened 2021-05-26
Simplify MathML language

dlazin:
Simplifies MathML language to avoid confusion over what "input-compliant" means, or "render" in the context of a non-visual reading system. This makes the MathML section more consistent with the SVG and CSS sections, which say:
* MUST process SVG embedded in XHTML Content Documents as defined in § 4.2 SVG Content Documents.
* MUST support the official definition of CSS as described in the [CSSSnapshot].

This also moves Presentation MathML rendering up one bullet to keep both Presentation MathML requirements together and help clarify that rendering only applies to visual RSes.

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -599,18 +599,18 @@
 
 						<ul class="conformance-list">
 							<li>
-								<p id="confreq-mathml-rs-behavior">MUST be an input-compliant renderer for <a
+								<p id="confreq-mathml-rs-behavior">MUST process <a
 										href="https://www.w3.org/TR/MathML3/chapter3.html">Presentation MathML</a>, as
 									defined in the [[MATHML3]] specification.</p>
+							</li>
+							<li>
+								<p id="confreq-mathml-rs-render">MUST, if it has a <a>Viewport</a>, support visual
+									rendering of Presentation MathML.</p>
 							</li>
 							<li>
 								<p id="confreq-mathml-rs-anno">MAY support rendering of <a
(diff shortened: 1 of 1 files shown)


epub33/rs/index.html
```diff
@@ -599,19 +599,19 @@ <h3>MathML</h3>
 
 						<ul class="conformance-list">
 							<li>
-								<p id="confreq-mathml-rs-behavior">MUST be an input-compliant renderer for <a
+								<p id="confreq-mathml-rs-behavior">MUST process <a
```
mattgarrish: There are three types of compliant processors per chapter 7 of that specification: input-compliant, output-compliant, and roundtrip-compliant. See https://www.w3.org/TR/REC-MathML/chapter7.html#sec7.2.1

It somewhat stands to reason that a reading system only has to be input-compliant, but unless we can inherit processing from HTML instead of directly from MathML (and not need to say anything here) this change adds ambiguity to what we mean. Unfortunately, HTML still seems a bit open-ended about support, even if input-compliant presentational math processing is all we'll probably ever see supported.

Adding a link to the original wording would certainly help, though.

dlazin: OK, how about "input-compliant processor" (language used there) along with a link? Then "render" only applies to visual RSes. (I recognize that not all rendering is visual; for example, we refer to audio rendering elsewhere. Nonetheless, I think "renderer" is unnecessary here.)

iherman (approved):

epub33/rs/index.html
mattgarrish: LGTM!

PR #1685, opened 2021-05-26
Add data-tests attributes for some existing tests as well as ReSpec test config

dlazin:
Adds data-tests attributes, corresponding to tests also tracked in tinyurl.com/epub-tests, for tests I have written, as well as adding one ID to a normative statement that didn't have an ID. See https://respec.org/docs/#html-attributes.

If this is acceptable, I will rename other existing tests (not written by me) to match, and add data-tests attributes for them too.

In https://github.com/w3c/epub-tests/pull/20, I will rename some tests to match the names in this PR.

***
<a href="https://pr-preview.s3.amazonaws.com/dlazin/epub-specs/pull/1685.html" title="Last updated on May 27, 2021, 6:15 PM UTC (0b6c8a0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1685/81acc27...dlazin:0b6c8a0.html" title="Last updated on May 27, 2021, 6:15 PM UTC (0b6c8a0)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -73,7 +73,11 @@
                 },
                 localBiblio: biblio,
                 preProcess:[inlineCustomCSS],
-                postProcess:[addConformanceLinks, addCautionHeaders]
+                postProcess:[addConformanceLinks, addCautionHeaders],
+				testSuiteURI: "https://github.com/w3c/epub-tests/tree/HEAD/tests/EPUBs/",
+				lint: {
+					"wpt-tests-exist": true,
+				}
             };//]]>
       </script>
 	</head>
@@ -719,28 +723,28 @@
 									<th colspan="3" id="cmt-grp-image" class="tbl-group">Images</th>
 								</tr>
 								<tr>
-									<td id="cmt-gif" data-test="cmt-gif">
+									<td id="cmt-gif" data-tests="cmt-gif.epub">
 										<code>image/gif</code>
 									</td>
 									<td> [[GIF]] </td>
 									<td>GIF Images</td>
 								</tr>
 								<tr>
-									<td id="cmt-jpeg" data-test="cmt-jpeg">
+									<td id="cmt-jpeg" data-tests="cmt-jpeg.epub">
 										<code>image/jpeg</code>
 									</td>
 									<td> [[JPEG]] </td>
 									<td>JPEG Images</td>
 								</tr>
 								<tr>
-									<td id="cmt-png" data-test="cmt-png">
+									<td id="cmt-png" data-tests="cmt-png.epub">
 										<code>image/png</code>
 									</td>
 									<td> [[PNG]] </td>
 									<td>PNG Images</td>
 								</tr>
 								<tr>
-									<td id="cmt-svg" data-test="cmt-svg,confreq-rs-epub3-svg">
+									<td id="cmt-svg" data-tests="cmt-svg,confreq-rs-epub3-svg.epub">
 										<code>image/svg+xml</code>
 									</td>
 									<td>
@@ -749,7 +753,7 @@
 									<td>SVG documents</td>
 								</tr>
 								<tr>
-									<td id="cmt-webp" data-test="cmt-webp">
+									<td id="cmt-webp" data-tests="cmt-webp.epub">
 										<code>image/webp</code>
 									</td>
 									<td> [[WebP-Container]], [[WebP-LB]] </td>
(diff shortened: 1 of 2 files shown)


iherman:
@dlazin I believe testing is on the agenda for our vF2F. What I think we should (also) discuss is the overall architecture of testing in view of the necessity to produce, at the end of the day, and implementation report. I presume these id-s and data are all part of that picture.

I guess we will have to develop some scripts/automatic procedure to extract the (meta)data from all over the place: links in the document, links to the tests, short description metadata of the tests, data about implementation reports, etc. To make it clear, I am happy to help on creating the reporting. It may be a significant amount of work, and we will have to deal with it early. Like now:-)

Actually: there is something in respec already: https://respec.org/docs/#data-tests. We may want to use respec as much as we can. We should try out what it does for us automatically.

dlazin:
> Actually: there is something in respec already: https://respec.org/docs/#data-tests. We may want to use respec as much as we can. We should try out what it does for us automatically.

Ah, very good. I have updated this PR to use data-tests, and I have added the respec config for the test suite. Note that the linter will fail until we also merge https://github.com/w3c/epub-tests/pull/20, since this repo refers to the renaming in that PR.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-05-28-epub#section4) on 2021-05-28

- no resolutions were taken

### 4. testing

_See github pull request [epub-tests#20](https://github.com/w3c/epub-tests/pull/20), [#1685](https://github.com/w3c/epub-specs/pull/1685)._




> *Dan Lazin:* See [Spreadsheet on statements to test](https://docs.google.com/spreadsheets/d/13wgptApSVbQbYzi2L5VXe3D_1QLzJOTGySUAtOxn4sA/edit#gid=0)

**Dan Lazin:** state of testing: some are easy, many are tricky. I put together a [spreadsheet](https://docs.google.com/spreadsheets/d/13wgptApSVbQbYzi2L5VXe3D_1QLzJOTGySUAtOxn4sA/edit#gid=0) that breaks down the rs spec at the time into testable normative statement.  
… roughly 200 tests that need to be written; we are through about 40 - many of the MUSTS  
… what we need is a 'testing guide' that organizes the tests so that people can jump in and contribute  
… how to name, organize, track the tests  
… we should use data tests attribute for test repo  
… here are pull requests  
… we could discuss whether to approve and merge the second PR here  
… then I can write up a document explaining how to participate in testing  
… having trouble in some cases finding reading systems viable to support tests  
… I feel behind on testing; we need to chip away at these  

**Ivan Herman:** need to think about how to report results in a consistent format - some sort of data file that includes metadata about the tests  
… and we need advanced structure on how to report the testing result  
… for audiobooks we had metadata that we could use to produce readable reports  

> *Ivan Herman:* See [Audiobook test results](https://www.w3.org/publishing/groups/publ-wg/implementation/results.html)

**Dan Lazin:** we are going to have a lot more failures than audiobooks testing  

**Ivan Herman:** CR has to prove that spec is implementable through >2 cited implementations  
… not worried about lots of failures, conceptually  
… sometimes we get requests to retest after the publication as implementations catch up  

**dawhe:** one example is manifest fallbacks, which is not widely supported across reading systems  

**Dan Lazin:** maybe we need a regular testing meeting to help answer questions as they come up  
… for example, SVG requirement is hard to test thoroughly  

**dawhe:** I would attend such a meeting  

**Tzviya Siegman:** we need help putting EPUB 3.3 into epubcheck  

**Matt Garrish:** I've been doing this as we go along  

**Ivan Herman:** what is the role of EPUBcheck in our testing strategy? In some sense it's an implementation  
… epubcheck checks content, not RS  
… what does the CR mean for the content document?  
… with a declarative spec, the approach to testing is to see whether those terms and features are used in the community  
… are fallbacks implemented or not? do publishers use those features effectively?  
… we should have evidence for each feature in the package document to justify it to be included in the spec  

**Matt Garrish:** we have a lot of features that are rarely used, or only used by a single feature. Does that mean we should deprecade stuff that isn't in wide use?  

**Ivan Herman:** maybe we have to have a list of those features where we have doubts that they are widely used  
… and ask the community in advance before we deprecate  

**Matt Garrish:** we did a survey of publishers for 3.1  

**Ivan Herman:** we are obliged to honor backwards compatibility, but don't have to keep features that nobody uses  

**George Kerscher:** pronunciation task force is putting forth a new proposal; we should harmonize our requirements with theirs  

**Wendy Reid:** we don't know if manifest fallbacks are used by anyone - should we do a survey across publishers and reading systems for this type of questionable feature?  

**Charles LaPierre:** if we can't find implementations, we could deprecate those because we can't find 2 implementations  
… for Dan's tests, epubtest.org might already have tests or you could use those as a blueprint  

**Tzviya Siegman:** we did an investigation for epub cfi to see if they were in use in Google books  
… so it is possible to this sort of survey  

**Matt Garrish:** it gets complicated when we drop specs from authoring but not reading systems or vice-versa so it would have to have zero support anywhere to have no impact  

> *Tzviya Siegman:* +1 mgarrish

**dawhe:** it's a disservice to authors if they have stuff in their content that we know doesn't work  
… would rather push a little to far and get blowback than not do anything  

**Charles LaPierre:** new developers have to support legacy code; we shouldn't keep supporting if not necessary  

**Matt Garrish:** if we drop fallbacks, are images in the spine valid? there is a ripple effect to getting rid of stuff  

**George Kerscher:** mathml continues to be poorly implemented, but that's a core component of something we really need and support will slowly improve over time  

**Ivan Herman:** there is no risk for mathml even though implementation needs to be improved  

**Dan Lazin:** if we have a weekly meeting we can go through these topics. epubtest repo had a nested hierarchy - does this organization make sense or should it just be flat?  

**Dave Cramer:** fine with me  

**Wendy Reid:** sounds like we have agreed to a regular testing call. We can discuss organization of tests and how to formulate results.  

**Ivan Herman:** Dan, you own the repository and can organize it the way you want.  
… re: the SVG case, this applies to testing html and CSS as well. we need to clearly position ourselves that the reading systems must be compliant with html and css and we are not in a position to test that.  
… we can only rely on the fact that reading systems render html  

**Brady Duga:** there is a large provider that uses a bunch of different ways to render  

**Matt Garrish:** do you want us to review the PR?  

**Dan Lazin:** I would like people to review the individual tests to validate my interpretation of the spec  
… for the PR, is this the right direction? are we polluting the spec?  

**Matt Garrish:** so the specs will be coupled with the tests  

**Dan Lazin:** yes  
… will set of Doodle poll for testing meeting times  
… need to be an owner of the spec to approve pr 1685  
… would like to do a branch in the test repo  

</details>

PR #1691, opened 2021-05-30
Add anchors and data-tests to match w3c/epub-tests#22

dlazin:
Add anchors and data-tests to match https://github.com/w3c/epub-tests/pull/22.

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -294,15 +294,15 @@
 			<section id="sec-pkg-doc-base-dir">
 				<h4>Base Direction</h4>
 
-				<p>If the <code>dir</code> attribute is set and indicates a base direction of <code>ltr</code> or
-						<code>rtl</code>, Reading Systems MUST override the bidi algorithm per <a
-						data-cite="bidi#Higher-Level_Protocols">the higher-level protocols</a> defined in [[BIDI]],
-					setting the paragraph embedding level to 0 if the base direction is <code>ltr</code>, or 1 if the
-					base direction is <code>rtl</code>.</p>
+				<p id="confreq-rs-pkg-dir" data-tests="confreq-rs-pkg-dir.epub">If the <code>dir</code> attribute is set and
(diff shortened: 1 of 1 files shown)


iherman (approved):

iherman:
I think having these test references is great, so we should go ahead with the respec approach. But I wonder whether we can come up with some extra CSS tricks to make those references a bit less emphasized. At the moment the reference is in bold, put on a separate line; a bit obtrusive for my taste. Can we find something a bit better (e.g., put it at the end of the previous paragraph in brackets and not in bold, something like that)? 

As an alternative we may come up with a small piece of javascript and a corresponding button at the start of the file which makes these references' visibility on and off. There are other specs that do such tricks...

(My concern is that these references break the reading flow of the text.)

iherman:
Hm. Actually... it looks like the generated `<details>` element for the tests only appear when the document runs respec. In other words, if I generate a 'final' HTML file (which also means when we publish an 'official' document on /TR) then these elements do not appear in the generated output.

@marcoscaceres is that correct and is this the intention? If that is the case then my comment above becomes moot; as long as these entries help us to keep track of the relevant tests, but do not bother the reader of the final spec, we are probably fine.

marcoscaceres:
> @marcoscaceres is that correct and is this the intention? If that is the case then my comment above becomes moot; as long as these entries help us to keep track of the relevant tests, but do not bother the reader of the final spec, we are probably fine.

Yes, that is correct - they only show up in the "dynamic" version of the document. We remove them when the document is exported.

iherman:
> > @marcoscaceres is that correct and is this the intention? If that is the case then my comment above becomes moot; as long as these entries help us to keep track of the relevant tests, but do not bother the reader of the final spec, we are probably fine.
> 
> Yes, that is correct - they only show up in the "dynamic" version of the document. We remove them when the document is exported.

That makes a lot of sense.

@dlazin then, as far as I am concerned, it is a go... :-)

PR #1697, opened 2021-06-09
Clarify that digital publications must not identify themselves as their pagination source

### LINKED ISSUE #1599: Page source when no source
When we created the requirement to identify the source of page break markers for the accessibility spec, we weren't really considering virtual page numbers.

To workaround the problem of requiring a source, I know we've told people to use an identifier for the current publication, but this doesn't make sense from a metadata perspective. It's an abuse of the `dc:source` element. I'm not sure it makes any sense to users, either.

The solution to this seems fairly simple: `dc:source` is only required if the pages are from a static source.

Any reasons not to make this change?

mattgarrish:
This fixes #1599 as far as the specification goes.

Figuring out an accessibilityFeature value is a change for the schema.org vocabulary.

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -540,13 +540,18 @@
 									<p>Including a recognizable identifier for the statically paginated source, such as
 										its ISBN or ISSN, ensures that users can determine which version the pagination
 										corresponds to.</p>
+									<p>If EPUB Creators insert pagination as a navigation aid for digital-only
+										publications, they must not specify a source (i.e., do not identify the current
+										publication as the source of its own pagination).</p>
 								</dd>
 
 								<dt id="sec-page-src-conf">Meeting this Objective</dt>
 
 								<dd>
-									<p>EPUB Creators MUST identify the source of the pagination in the Package Document
-										metadata.</p>
+									<p>When an EPUB Publication includes <a href="#sec-page-breaks">page break
+											markers</a> and/or a <a href="#sec-page-list">page list</a> that correspond
+										to a statically-paginated version of the publication, EPUB Creators MUST
+										identify that source in the Package Document metadata.</p>
 								</dd>
 							</dl>
 						</section>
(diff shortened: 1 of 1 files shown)


clapierre (changes requested):
I think we should also include something about the accessibilityFeature "PrintPageNumbers" that should only be used when there is also a "Source" and a print equivalent book.  

But in the case of a digital only EPUB where there is no Source as we have now, that the accessibilityFeature they add a new "DigitalPageNumbers" or "EmbeddedPageNumbers" which we can easily add to our list of values.  

This will help in with the discovery of a Print vs. Digital First page numbers, and for determining conformance to the specification since a 3rd party will know there are page #'s / page list inside the EPUB but may not know where the source of the pagebreaks are from.

EG:
Source of Page Breaks: Print Book
```
<meta property="schema:accessibilityFeature">printPageNumbers</meta>
<dc:source id="srcisbn">PrintBookISBN</dc:source>
<meta property="source-of" refines="#srcisbn">pagination</meta>

```
Source of Page Breaks: Digital Only EPUB itself is the source of its page breaks
`<meta property="schema:accessibilityFeature">digitalPageNumbers</meta>`

mattgarrish:
Remember, we can't cite schema.org in the specification, Charles. 🙃 

That's why we had to generalize the whole section on discovery metadata not to be specific to schema.org, and why we don't get into specific values for properties. Making a change like you propose leads us back into this problem.

We can maybe look at this in the techniques when we figure out what the answer is to virtual page breaks. It probably depends on what the task force recommends, too, and if they even expect to use authored tags.

I still think adding another feature is a bad idea, as people don't read specifications -- we're asking for too much nuance. We already know if there's a source that there are print page numbers (the schema.org value was developed in a vacuum from other metadata, so couldn't account for this). I'd personally rather deprecate printPageNumbers and find a more universal term moving forward.

But this is outside the scope of this one change, which is why it's better to resolve the feature value separately.

clapierre:
Understood, don't like it but I guess we live with this for now and provide techniques on how to handle this where we can go look at adding/updating the schema.org values.

PR #1698, opened 2021-06-10
Clarifies that meaningful sequence applies to spreads

### LINKED ISSUE #1695: Review guidance on success criterion 1.3.2
Going back over the WCAG documents, I believe I understand the hesitance they had last time we discussed our focus on spine sequence.

Looking at [our technique](https://w3c.github.io/epub-specs/epub33/a11y-tech/#access-001), the first part seems fine. The order of the markup within a page needs to make sense ___and___ it also needs to make sense across pages because we have to deal with two-page spreads in fixed layouts.

But that's different from ordering documents in the spine and marking non-linear items. This part of the guidance really isn't related to the sequence of markup within/across pages.

If documents aren't sequenced properly, the publication is broken. Period. It's a general usability problem for everyone, not specifically an accessibility issue.

I don't see that we need to spell this out for people, so I'd suggest we remove the text and examples from the technique.

mattgarrish:
I've rewritten the [access-001 technique](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1695/epub33/a11y-tech/index.html#access-001) to make it more specific to synthetic spreads, as this is where I see our issue lying. The order of content in the separate documents still has to make sense as a unit, or there has to be a way to follow the text.

Fixes #1695 

- Techniques [preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1695/epub33/a11y-tech/index.html)
- Techniques [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/main/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1695/epub33/a11y-tech/index.html)

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -559,55 +559,25 @@
 				<h3>Content Access</h3>
 
 				<section class="suppress-numbering" id="access-001">
-					<h4>ACCESS-001: Ensure linear reading order of the publication</h4>
+					<h4>ACCESS-001: Ensure meaningful order of content across spreads</h4>
 
-					<p>[[WCAG2]] <a href="https://www.w3.org/TR/WCAG2/#meaningful-sequence">Success Criterion
-						1.3.2</a> specifies that each Web page have a meaningful order, but the content of an EPUB
-						Publication typically spans multiple documents. It is consequently essential not only that each
-						EPUB Content Document has a meaningful order, but that the order is meaningful from document to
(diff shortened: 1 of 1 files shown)


clapierre (approved):
Great as always Matt.

PR #1703, opened 2021-06-16
file URLs Should Not be used in the manifest

### LINKED ISSUE #1688: `file` URLs in manifest items
Manifest items can identify resources with absolute URLs. So an EPUB can theoretically use local file system resource with `file` URLs.

Using `file` URLs in the manifest is not a good practice, but has never been strictly forbidden (as far as I know). I think it's probably often used by mistake rather than intentionally. But there might be legit use cases, like an internal documentation system (as @mattgarrish pointed out in #1374).

Would it be reasonable to say that absolute URLs SHOULD have a [special scheme](https://url.spec.whatwg.org/#special-scheme) that is not `file`?
In parallel we can also make it explicit what Reading Systems can/should do with absolute URLs as manifest entries.

iherman:
The [WG resolution](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-06-10-epub#resolution2) says "SHOULD NOT" and not "MUST NOT", so this is what the PR uses. There may be a question whether "MUST NOT" is not more appropriate here.

Fixes #1688.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1703.html" title="Last updated on Jun 17, 2021, 9:29 AM UTC (6d49760)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1703/5ce5b14...6d49760.html" title="Last updated on Jun 17, 2021, 9:29 AM UTC (6d49760)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1231,8 +1231,7 @@
 								<code>href</code>
 							</dt>
 							<dd>
-								<p>A <a href="https://url.spec.whatwg.org/#valid-url-string">valid URL string</a>
-									[[URL]] that references a resource. The <a href="https://url.spec.whatwg.org/#url-scheme-string">URL-scheme string</a> [[URL]], if present, SHOULD NOT match for <code>file</code>.</p>
+								<p>A <a href="https://url.spec.whatwg.org/#valid-url-string">valid URL string</a> [[URL]] that references a resource. If the value is an <a href="https://url.spec.whatwg.org/#absolute-url-string">absolute URL</a>, it SHOULD NOT use the <code>file</code> URI scheme [[rfc8089]].</p>
 								<div class="example">
 									<pre>&lt;link rel="record"
       href="meta/9780000000001.xml" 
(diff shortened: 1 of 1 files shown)


rdeltour (approved):
Looks OK to me.

I'm only wondering if using string-level terminology is the right approach for referencing the URL standard. Or if statements like this (on the scheme) are better done on URL records (as opposed to URL strings). Maybe we could ask someone from the URL standard to review a couple key usages in our spec to see if we're good citizens?

iherman:
@rdeltour you are right; the text was fairly convoluted. 

I actually changed it to be consistent with the way we refer to data URLs, i.e., through the relevant RFC. Here is the way it reads now:

> **href**  
> A valid URL string [URL] that references a resource. If the value is an [absolute URL](https://url.spec.whatwg.org/#absolute-url-string), it SHOULD NOT use the "file" URI scheme [rfc8089].
> ...
> **[rfc8089]**  
> The "file" URI Scheme. M. Kerwin. IETF. February 2017. Proposed Standard. URL: https://datatracker.ietf.org/doc/html/rfc8089

mattgarrish (approved):

PR #1714, opened 2021-06-22
Allow use of ssml and pls in svg content documents

### LINKED ISSUE #1710: Are SSML attributes allowed on SVG?
Is there a reason we only define them for XHTML given that they could be put on text/tspan elements in SVG?

The one awkward part is that you can't define lexicons in SVG since it lacks a link element, but there's no reason you can't use SSML or CSS Speech in SVG.

mattgarrish:
The changes are generally straightforward with the exception of what elements the SSML attributes are allowed on. The original wording for ssml:ph, for example, made it global and then limited it to only where it represents a phonetic pronunciation.

This doesn't make as much sense for SVG as it looks like only text, textPath, tspan, title and desc can contain text.

I know we have this for HTML, though, because it would otherwise require an evolving list of where it's allowed. I'm tempted to say we should at least limit the attribute to "palpable content" in HTML, but then it doesn't even make sense on all of those elements (audio, video, img, form elements, iframe, etc.). It also misses the title element, but that's an easy add.

I've largely kept the original wording, as I suppose we can't try to solve every possible authoring sin. Maybe we can address the authoring pitfalls of using ssml:ph on root elements, or the body, or elements without text in another PR without normatively disallowing.

But I'd be open to a format-specific restriction list like:

> EPUB Creators MAY specify on:
> 
> -  In XHTML Content Documents, the `title` element and any palpable content [[HTML]] with which they can logically associate a phonetic equivalent (i.e., that has descendant text content that a Text-to-Speech engine would otherwise render).
> - In SVG Content Docments, the `title`, `desc`, 'text`, `textPath`, and `tspan` elements [[SVG2]].

(We should probably add some reading system restrictions, like ignoring ssml:ph if the element it is on doesn't have any text that would be voiced, but for another PR.)

Fixes #1710 

- TTS [preview](https://cdn.statically.io/gh/w3c/epub-specs/tts/issue-1710/epub33/tts/index.html)
- TTS [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/main/epub33/tts/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/tts/issue-1710/epub33/tts/index.html)

### DIFF
--- a/epub33/tts/index.html
+++ b/epub33/tts/index.html
@@ -183,7 +183,7 @@
 
 				<aside class="example">
 					<p>The following example shows the pronunciation for EPUB added to SVG markup.</p>
-					<pre>&lt;text>&lt;tspan ssml:ph="ipʌb">EPUB&lt;/span> 3&lt;/text></pre>
+					<pre>&lt;text>&lt;tspan ssml:ph="ipʌb">EPUB&lt;/tspan> 3&lt;/text></pre>
 				</aside>
 			</section>
 
--- a/epub33/tts/index.html
+++ b/epub33/tts/index.html
@@ -122,7 +122,7 @@
 				<p>This specification recasts the [[SSML]] <a
 						href="https://www.w3.org/TR/2010/REC-speech-synthesis11-20100907/#S3.1.10"><code>phoneme</code>
 						element</a> as two attributes — <code>ssml:ph</code> and <code>ssml:alphabet</code> — and makes
-					them available within XHTML Content Documents.</p>
+					them available within <a>EPUB Content Documents</a>.</p>
 
 				<p>The attributes allow EPUB Creators to specify the proper phonetic pronunciation for uncommon terms
 					that a TTS engine is likely to mispronounce, as well as to disambiguate heteronyms.</p>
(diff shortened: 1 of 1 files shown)


iherman:
As a cursory look, I have already spotted a missing element in SVG, namely `<g>`: it should be perfectly o.k. to use it by virtue of the statement of the like:

> When the `ssml:ph` attribute appears on an element that has text node descendants, the corresponding document text to which the pronunciation applies is the string that results from concatenating the descendant text nodes, in document order. 

The `<metadata>` element is another one, and I am sure there are more: we get ourselves into trouble if we try to be more precise. The statement above is perfectly fine and, imho, clear.

Note, b.t.w., that the [HTML "palpable content"](https://html.spec.whatwg.org/multipage/dom.html#palpable-content) list misses, for example, the `<title>` element. Which may be fine for what the list is defined for in HTML, but I am not sure it would be acceptable to remove it from the list of elements for tts in epub (e.g., for a11y reasons).

Bottom line: I think the text is good, in this respect, as is...

epub33/tts/index.html
```diff
@@ -234,12 +239,21 @@ <h3>The <code>ssml:alphabet</code> attribute</h3>
...
 
+				<aside class="example">
+					<p>The following example shows a global declaration for the x-SAMPA alphabet on the root
+							<code>svg</code> element.</p>
+					<pre>&lt;svg &#8230; ssml:alphabet="x-sampa">
+   &lt;title>&lt;span ssml:ph="ipVb">EPUB&lt;/span> Adoption Chart&lt;/title>
+   &#8230;
+&lt;/svg></pre>
+				</aside>
+
```
iherman, line 252: I think this should be `<tspan>` and not `<span>`

iherman (approved):

epub33/tts/index.html
iherman: Oops, sorry, I clicked the wrong line for the comment; I am talking about Example 4.

mattgarrish: Ya, obvious cut/paste mistake!

mattgarrish:
> Bottom line: I think the text is good, in this respect, as is...

Ya, this is where I ended up. I think we're better off just adding some authoring guidance about the effect of using ssml:ph -- notably that it replaces all the text content on the element, so put it on something like the body and that's all someone will hear for the document. Plus adding some reading system requirements like ignoring the attribute on elements without text content that would normally be voiced and when the attribute is empty.

PR #1723, opened 2021-06-25
SSML and PLS clarifications

### LINKED ISSUE #1705: PLS processing inconsistencies
Recording this from #1690 to look at again after splitting PLS out to a note.

The PLS reading system requirements:

- confuse the reading system as the text-to-speech engine, and are too simplistic as a definition of how to write a tts engine that supports pls
- are contradictory - pls defines how to select lexemes (first preferred otherwise first in document order) that a conforming tts engine has to follow, not last in document order regardless of preference as we define
- stray into areas the reading system can't control - precedence of lexicon/ssml will be controlled by the tts engine (this is where our ssml definition is similarly underspecified)
- don't even clearly define that you have to obtain and use the linked lexicons

We would have been better off saying only that a reading system:

- must use a PLS-compliant text-to-speech engine to render xhtml content documents
- must initiate the text-to-speech engine using all PLS lexicons linked to from the xhtml content document

### LINKED ISSUE #1706: What happens when SSML alphabet not specified
Another of the holes in our TTS definitions is we don't define what to make of an ssml:ph attribute that doesn't have an ssml:alphabet defined for it. Do we default to... IPA? To whatever the actual TTS engine supports by default?

It's not even required that there be an in-scope alphabet definition.

Also, what happens if the alphabet is defined but not supported? Default to IPA again?

We need to figure out what are really reading system requirements and what information is just passed to a tts engine.

mattgarrish:
This PR adds some clarifications for using the SSML attributes and attaching pronunciation lexicons. In particular, that ssml:ph not be used where there isn't text and be ignored if it's empty as it's not intended to suppress voicing. It also points out the voicings are not replacements for attributes that carry textual information.

I've generally only made the requirements "should not" as it's hard to anticipate if there ever would be a legitimate outside case.

I still find the complexities of HTML and SVG markup vis-a-vis raw SSML leave us only with a basic framework, but I guess small steps are good enough until more developers try to implement these technologies. In other words, I don't anticipate doing anything more to this note myself.

Feedback welcome on the additions, of course.

Fixes #1705 
Fixes #1706 

- TTS [preview](https://cdn.statically.io/gh/w3c/epub-specs/tts/issue-1706/epub33/tts/index.html)
- TTS [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/main/epub33/tts/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/tts/issue-1706/epub33/tts/index.html)

### DIFF
--- a/epub33/tts/index.html
+++ b/epub33/tts/index.html
@@ -435,7 +435,8 @@
 				<h3>Conformance</h3>
 
 				<p><a>Reading Systems</a> with <a>Text-to-Speech</a> (TTS) capabilities SHOULD support <a href="#ssml"
-						>SSML attributes</a> and <a href="#pls">pronunciation lexicons</a> as follows:</p>
+						>SSML attributes</a>, <a href="#pls">pronunciation lexicons</a> and <a href="#css-speech">CSS
+						Speech</a> as follows:</p>
 
 				<dl>
 					<dt>SSML</dt>
@@ -512,6 +513,12 @@
 									carries the <code>ssml:ph</code> attribute.</p>
 							</li>
 						</ul>
+					</dd>
+
+					<dt>CSS Speech</dt>
+					<dd>
+						<p>This document adds no additional requirements for Reading System support to those defined in
+							[[CSS-SPEECH-1]].</p>
 					</dd>
 				</dl>
 			</section>
--- a/epub33/tts/index.html
+++ b/epub33/tts/index.html
@@ -172,9 +172,9 @@
 
 				<p id="ssml-ph-text">EPUB Creators SHOULD NOT use the <code>ssml:ph</code> attribute on elements without
 					text content that a Text-to-Speech engine would normally render (e.g., on empty <code>div</code> or
-						<code>span</code> elements). The attribute is not intended for adding additional voicing only
-					for users listening to TTS playback and Reading Systems are expected to ignore the attribute if it
-					does not replace text they would normally render.</p>
+						<code>span</code> elements). The attribute is not intended to add additional voicing only for
+					TTS playback, and Reading Systems are expected to ignore the attribute if it does not replace text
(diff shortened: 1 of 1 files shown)


epub33/tts/index.html
```diff
@@ -159,19 +159,33 @@ <h3>The <code>ssml:ph</code> attribute</h3>
...
+					results from concatenating the descendant text nodes, in document order. The specified phonetic
+					pronunciation must therefore logically match the element's textual data in its entirety (i.e., not
+					just an isolated part of its content).</p>
+
+				<p id="ssml-ph-text">EPUB Creators SHOULD NOT use the <code>ssml:ph</code> attribute on elements without
+					text content that a Text-to-Speech engine would normally render (e.g., on empty <code>div</code> or
+						<code>span</code> elements). The attribute is not intended for adding additional voicing only
+					for users listening to TTS playback and Reading Systems are expected to ignore the attribute if it
+					does not replace text they would normally render.</p>
+
```
iherman, line 178: I had some difficulties to parse the English sentence (as someone not an expert in TTS). Maybe a comma would help as in

> [...] TTS playback, and Reading Systems are expected [...]

But maybe also

> The attribute is only intended to add voicing for users listening to TTS playback, and [...]

But I may completely misunderstand the sentence which may be a sign of parsing difficulties...

iherman (approved):

epub33/tts/index.html
mattgarrish: Ya, the first part of the sentence if a bit wordy. I've tried to shorten it and also added a comma to separate from the RS explanation.

PR #1724, opened 2021-06-28
Add note about the problem accessing resources

### LINKED ISSUE #1687: Refine the requirements on how RS must process the container directory structure
The EPUB core spec says in the [File and Directory Structure](https://w3c.github.io/epub-specs/epub33/core/#sec-container-file-and-dir-structure) section:

> EPUB Creators MAY locate all other files within the OCF Abstract Container in any location descendant from the Root Directory, provided they are not within the META-INF directory.

But @mattgarrish reports in #1374 that some reading systems do not handle that correctly. They don't allow content that is not in the directory of the Package Document (or any descendant directory).

There are at least two (exclusive) options:
- add language in the RS spec to say more explicitly that RS must be able to handle all container content.
- add restrictions to the Core spec that say authors should not rely on content that is not in the Package Document directory and descendants.

mattgarrish:
I've added a short note about the problem to the OCF section per the resolution in #1687 and also expanded on the note in the multiple renditions spec since that's really where this issue would come into play.

Fixes #1687 

- Multiple Renditions [preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1687/epub33/multi-rend/index.html)
- Multiple Renditions [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/main/epub33/multi-rend/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1687/epub33/multi-rend/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1724.html" title="Last updated on Jun 30, 2021, 9:41 PM UTC (65de0c7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1724/c273d8e...65de0c7.html" title="Last updated on Jun 30, 2021, 9:41 PM UTC (65de0c7)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -9302,6 +9302,9 @@ EPUB/images/cover.png</pre>
 				-->
 
 				<ul>
+					<li>28-June-2021: Added a note discouraging EPUB Creators from referencing resources outside the
+						directory containing the Package Document to avoid interoperability issues. See <a
+							href="https://github.com/w3c/epub-specs/issues/1687">issue 1687</a></li>
 					<li>18-June-2021: Moved requirements for authoring SSML, PLS lexicons and CSS 3 Speech to the <a
 							href="https://www.w3.org/TR/epub-3-tts">EPUB 3 Text-to-Speech Enhancements</a> note. The
 						ability to use these technologies in EPUB 3 Publications remains unchanged. See <a
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1216,7 +1216,10 @@
 								<code>href</code>
 							</dt>
 							<dd>
-								<p>A <a href="https://url.spec.whatwg.org/#valid-url-string">valid URL string</a> [[URL]] that references a resource. If the value is an <a href="https://url.spec.whatwg.org/#absolute-url-string">absolute URL</a>, it SHOULD NOT use the "file" URI scheme [[rfc8089]].</p>
+								<p>A <a href="https://url.spec.whatwg.org/#valid-url-string">valid URL
+									string</a> [[URL]] that references a resource. If the value is an <a
+										href="https://url.spec.whatwg.org/#absolute-url-string">absolute URL</a>, it
+									SHOULD NOT use the "file" URI scheme [[rfc8089]].</p>
 								<div class="example">
 									<pre>&lt;link rel="record"
       href="meta/9780000000001.xml" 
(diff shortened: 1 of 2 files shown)


rdeltour (commented):
The proposed note looks OK to me.

But I'm also wondering why we do not put this as a SHOULD?
After all, the core spec is about defining authoring recommendations. Here we have a know source of interoperability issues, so a normative SHOULD seems appropriate. To adapt RFC2119 language: there may exist valid reasons to put resources in a sibling directory, but the full implications for interoperability must be understood and carefully weighed before doing so.

Unless we're afraid that EPUBCheck-warning-means-rejection will cause havoc, of course.

iherman:
In my view, we are documenting here a misbehavior of some Reading Systems. It is still pending whether we would normatively restrict (even with a SHOULD) the folder structure in the core document. Until that does happen, I would feel uncomfortable putting this as a SHOULD in this spec.

(We can always update that later)

mattgarrish:
> Unless we're afraid that EPUBCheck-warning-means-rejection will cause havoc, of course.

It probably won't cause any actual havoc, but there is concern about making this a warning if we don't need to. We're not sure how far we want to get into normatively hammering authors or developers with this problem when, outside of developing multiple renditions, no one has reported it as a real-world issue they've encountered so far.

There's also that the fix, as @iherman notes, should really be on the reading system side, but there's concern there, too, that even if we make a change now it won't lead to reading systems being changed. It looks like an optimization based on the assumption that the vast majority of content is already self-contained in an /OEBPS or /OPS or /EPUB directory.

So, in effect, we're just documenting what we know about the problem for now (with the option to warn authors as an info message in epubcheck, I suppose). The idea was that we could return to this in the future if it becomes more of an issue.

rdeltour:
OK fine by me. I was thinking we could pave that cowpath in the spec, but if we think it's not ready to be paved or that there's hope that the RS situation will make that useless in the future, then a note is certainly enough 😊.

iherman:
> OK fine by me. I was thinking we could pave that cowpath in the spec, but if we think it's not ready to be paved or that there's hope that the RS situation will make that useless in the future, then a note is certainly enough 😊.

I would hope for the latter...

mattgarrish:
Are we waiting on anything specific before merging this? (i.e., nothing in last week's call negated this PR, did it?)

dauwhe (approved):

dauwhe:
> Are we waiting on anything specific before merging this? (i.e., nothing in last week's call negated this PR, did it?)

I think we should merge. It's a step in the right direction; we may take more steps.

iherman (approved):

iherman:
My apologies, I forgot to add my 👍 in the approvals. Done now, I agree to merge.

PR #1744, opened 2021-06-30
Clarify TTS is not about AT

### LINKED ISSUE #1727: [EPUB 3.3] [question] Difference between Text-To-Speech and Assistive Technologies
When the specification talks about Text-to-Speech (TTS), do we mean the one in assistive technologies (screenreaders) or the one in reading applications? Or both?

Normally the screenreader returns to the user semantic constructs, alternate text, etc. to the user, while the TTS in reading applications reads the text only.

I think it might be helpful to make this more explicit in https://www.w3.org/TR/epub-33/#dfn-text-to-speech

mattgarrish:
This PR fixes #1727 by merging the TTS definition into the media overlays section and pointing the two links to it. It also adds a link to the tts note to the authoring document and adds a paragraph that AT support is not specifically in scope in the tts note's overview.

Reviewing the media overlays section, I got a bit confused by the similarity between the embedded media and tts section -- that both talk about their respective behaviour initiating when an `audio` element isn't present. I assume for TTS we're talking about references to non-media elements, so tweaked the paragraph to that effect, but if you could have a look @marisademeglio and see if it reads properly, I'd appreciate it.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1744.html" title="Last updated on Jul 1, 2021, 11:01 AM UTC (5a9a25a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1744/c273d8e...5a9a25a.html" title="Last updated on Jul 1, 2021, 11:01 AM UTC (5a9a25a)">Diff</a>

### DIFF
--- a/epub33/common/js/biblio.js
+++ b/epub33/common/js/biblio.js
@@ -395,5 +395,17 @@ var biblio = {
 		"title": "XPointer Shorthand Notation",
 		"href": "https://www.w3.org/TR/2003/REC-xptr-framework-20030325/",
 		"date": "25 March 2003"
+	},
+	
+	
+	
+	// TEMP ENTRIES
+	"EPUB-TTS-10": {
+		"authors":[
+			"Matt Garrish"
+		],
+		"title": "EPUB 3 Text-to-Speech Enhancements 1.0",
+		"href": "https://www.w3.org/TR/epub-tts-10",
+		"publisher": "W3C"
 	}
 }
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -518,11 +518,6 @@
 					<dd>
 						<p>The rendering of two adjacent pages simultaneously on a device screen.</p>
 					</dd>
-					<dt><dfn id="dfn-text-to-speech">Text-to-Speech</dfn> (TTS)</dt>
-					<dd>
-						<p>The rendering of the textual content of an <a>EPUB Publication</a> as artificial human speech
-							using a synthesized voice.</p>
-					</dd>
 					<dt>
 						<dfn id="dfn-top-level-content-document" data-lt="Top-level Content Documents">Top-level Content
 							Document</dfn>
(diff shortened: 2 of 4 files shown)


iherman (approved):
- I guess the TTS note reference should be added to our reference list; at present, respec/specref does not find it, and generates an error
- Reading all this it is really a pity the the `text` element is, in fact, a misnomer, because (as the spec says itself) it can refer to resources in general. I guess it is too late, I wonder whether a note should emphasize this point to make the casual reader aware of this...

gregoriopellegrino (approved):

mattgarrish:
> * I wonder whether a note should emphasize this point to make the casual reader aware of this...

I think we can improve this as part of solving #1745.

mattgarrish:
> I guess the TTS note reference should be added to our reference list; at present, respec/specref does not find it, and generates an error

I guess this depends on how quickly we republish. We can't push this in a group with the existing documents as it needs a transition request to FPWG-NOTE, right? Publish it sooner than a refresh and the problem solves itself.

But I guess the nuisance of making/removing a temp entry in the biblio file is offset by not having to worry about the publishing schedule...

gregoriopellegrino (approved):

iherman:
> I guess this depends on how quickly we republish. We can't push this in a group with the existing documents as it needs a transition request to FPWG-NOTE, right? Publish it sooner than a refresh and the problem solves itself.

Indeed. And we may run into timing problems: I will be on vacation in about ten days (Monday a week) for two weeks, so I cannot properly shepherd the publication even if we vote for it tomorrow. (Nevertheless, we should vote for it to get that done, @dauwhe ?) But maybe somebody can keep an eye on the last mile of publication on the Monday and Tuesday of that week...

mattgarrish:
I'm going to merge this now even though Marisa hasn't had a chance to review it as she's off right now. We're going to have to return to the section, and possibly deprecate it, to figure out #1745.

marisademeglio:
The only requirement I see here on EPUB creators is to make the text content render-able by TTS, should the need arise. That's a general accessibility requirement though, not an authoring practrice that's specific to MO. My inclination is to remove this TTS section and put it in the Reading Systems document instead.

I would also remove this from the `par` element definition
```
The audio element is OPTIONAL only if its sibling text element refers to audio or video media 
(see § 7.3.2.4 Embedded Media), or to textual content intended for rendering via text-to-speech.
```

Because it basically says that it's optional all the time (because when would we have text content that cannot be rendered by TTS?). 

Instead we could say that if there is no `audio` element, the timing of the `text` element is determined by the reading system and its abilities to render what the `text` element refers to, be it embedded media or text (or alt-text) content.

PR #1748, opened 2021-07-01
Add additional linking to the cover-image definition

### LINKED ISSUE #1731: [EPUB 3.3] [suggestion] Cover info in metadata section
I know that the cover image is not defined in the `metadata` section of the `opf` file, but in the `manifest` with the property `cover-image`.

I think some people not familiar with the EPUB format and specs might look for how to express this information in the `metadata` section of the specification (for all intents and purposes he cover is a «metadata» of the publication).

Maybe add a note about it that links to the explanation of the `cover-image` property?

mattgarrish:
Per #1731, this PR adds some examples of what the manifest item properties identify, including the cover image.

It also adds a note to the opf2 meta element section explaining that it is retained for identifying covers in epub 2 but that the cover-image property must be used in epub 3.

Hopefully, that should be plenty of guidance in the right direction.

Fixes #1731

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1748.html" title="Last updated on Jul 1, 2021, 2:28 PM UTC (2c45d42)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1748/b355364...2c45d42.html" title="Last updated on Jul 1, 2021, 2:28 PM UTC (2c45d42)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -2785,6 +2785,12 @@
 							<p>EPUB Creators MUST declare all applicable descriptive metadata properties for each
 								Publication Resource in this attribute.</p>
 
+							<p>EPUB Creators MUST declare all applicable descriptive metadata properties for each
+								Publication Resource in this attribute (e.g., if the resource is the <a
+									href="#sec-cover-image">cover image</a>, contains <a href="#sec-scripted"
+									>scripting</a>, or references <a href="#sec-remote-resources">remote
+								resources</a>).</p>
+
 							<p>EPUB Creators MUST declare exactly one <code>item</code> as the <a>EPUB Navigation
 									Document</a> using the <code>nav</code> property.</p>
 
@@ -3441,6 +3447,14 @@ Spine:
 
 							<p>For more information about the <code>meta</code> element, refer to its definition in
 								[[OPF-201]].</p>
+
+							<div class="note">
+								<p>The [[OPF-201]] <code>meta</code> element is retained in EPUB 3 primarily so that
+									EPUB Creators can identify the cover image for compatibility with EPUB 2 Reading
+									Systems. In EPUB 3, the cover image must be identified using the <a
+										href="#sec-cover-image"><code>cover-image</code> property</a> on the <a
+										href="#sec-item-elem">manifest <code>item</code></a> for the image.</p>
+							</div>
 						</section>
 
 						<section id="sec-opf2-guide">
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -2783,9 +2783,6 @@
 									href="#sec-vocab-assoc"></a>.</p>
 
 							<p>EPUB Creators MUST declare all applicable descriptive metadata properties for each
-								Publication Resource in this attribute.</p>
-
-							<p>EPUB Creators MUST declare all applicable descriptive metadata properties for each
 								Publication Resource in this attribute (e.g., if the resource is the <a
 									href="#sec-cover-image">cover image</a>, contains <a href="#sec-scripted"
 									>scripting</a>, or references <a href="#sec-remote-resources">remote


gregoriopellegrino (approved):

epub33/core/index.html
```diff
@@ -3441,6 +3444,14 @@ <h6>The <code>meta</code> Element</h6>
...
 							<p>For more information about the <code>meta</code> element, refer to its definition in
 								[[OPF-201]].</p>
+
+							<div class="note">
+								<p>The [[OPF-201]] <code>meta</code> element is retained in EPUB 3 primarily so that
+									EPUB Creators can identify the cover image for compatibility with EPUB 2 Reading
+									Systems. In EPUB 3, the cover image must be identified using the <a
+										href="#sec-cover-image"><code>cover-image</code> property</a> on the <a
+										href="#sec-item-elem">manifest <code>item</code></a> for the image.</p>
+							</div>
```
iherman, line 3454: I must admit that in my earlier reviews I never thoroughly looked at the [section on legacy meta](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1748.html#sec-opf2-meta), nor have I ever looked at the OPF 2.0 spec. 

After some search I found the definition of the OPF 2.0 version of `meta` in the [Schema](http://idpf.org/epub/20/spec/OPF_2.0.1_draft.htm#AppendixA): it is defined as a possible child element of the `<metadata>` section with possible attributes of `content`, `name`, and `scheme`,  and with an empty content.

In my view, the title of [§2.3.2.3.5](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1748.html#sec-meta-elem) (i.e., the `meta` element as a legacy feature) is then a bit misleading. The `meta` element is not a legacy feature; rather the `meta` element in EPUB 3 has two legacy _attributes_, namely `content` and `name` (and it opens up its content model to text, but that is hardly relevant in terms of "legacy"). I think it would be better to make that clearer in the text. Something like:

> ### Legacy attributes on the `meta` element
>
> The EPUB 3 `meta` element has two legacy attributes (`content` and `name`) that previously provided a means of including generic metadata. The content model of the `meta` element in EPUB 3, as well as a different set of attributes, replaces those attributes.

If this is all correct, then the new Note should also be reformulated:

> The legacy attributes of the [OPF-201] `meta` element have been retained in EPUB 3  primarily so that EPUB Creators can identify the cover image for compatibility with EPUB 2 Reading Systems. In EPUB 3, the cover image must be identified using the cover-image property on the manifest item for the image.

(Note that I have been led down an erroneous path by the current text. While the `guide` element, which is the other legacy feature, is properly defined by the OCF 2.01 spec, the `meta` is not, it just refers to XHTML and that is a bit difficult to find unless one reads through the full text. It took me a certain time to understand what is indeed a "legacy". Maybe by making clear that the attributes have changed only, it becomes clearer.)

mattgarrish: > Maybe by making clear that the attributes have changed only, it becomes clearer.

Well, it's a little more complex than that as you have two very different content models.

Yes, the `name` attribute was technically replaced by `property`, but they're both required in their respective definitions. They can't coexist and you pick the one you want.

The `content` attribute is the more problematic, though, as it was replaced by text content of the element which must be one or more characters in length in EPUB 3 (whereas the OPF version had to be empty). It takes a lot of contortion and "unexplaining" the requirements of EPUB 3 to get back to an understanding of what the meta tag was in EPUB 2.

I don't think we gain anything by trying to walk people through a remapping back to the old meta element, though. It's just as easy to consider it an entirely different kind of meta tag to EPUB 3's.

iherman: > I don't think we gain anything by trying to walk people through a remapping back to the old meta element, though. It's just as easy to consider it an entirely different kind of meta tag to EPUB 3's.

I see the point. However, what led me to this line of thought is the (unfortunate) fact that I did not find any formal definition, in the OCF 2.0 spec, for the `<meta>` element. Ie, we have this strange situation that we have a completely new element with the same name, and if someone tries to understand what the differences are then... the only source we have is the Schema document in the OPF 2.0 spec (which is not an easy read for an average author). That is, actually, the problem I am trying to solve, although I agree with you about the fact that the two elements are fundamentally different.

Maybe (but only maybe) one solution is that, in our spec, we make a direct reference to the XHTML meta element definition, by making it clear that what OCF did was to blindly take over the corresponding XHTML element. Something like  

> The meta element [OPF-201], itself modeled after the HTML meta element, is a legacy feature that previously provided...

And provide a pointer to the relevant portion of the HTML spec.

As for what the pointer would be:  we can point to the  [XHTML 1.1 definition](https://www.w3.org/TR/xhtml-modularization/abstract_modules.html#s_metamodule), although that document does not give any definition as for the use of the term and is only an English version of the DTD 😒. Not helpful. We can go back to the [HTML 4.01 definition](https://www.w3.org/TR/html401/struct/global.html#edef-META) which at least says some words. We could of course simply use the latest official reference [in HTML5](https://html.spec.whatwg.org/multipage/semantics.html#the-meta-element), but it includes additional attributes that are irrelevant for the issue at hand... 

My vote would be to put in a reference to HTML 4.01; we are talking about legacy after all...

mattgarrish: > I did not find any formal definition, in the OCF 2.0 spec, for the `<meta>` element

Ya, EPUB 2 sometimes has a way of letting you down on specifics... 😉 

But I don't think we want to open this can of worms, as it's not actually the HTML element. It's "analogous", which in OPF 2 speak appears to mean it has the same name and only the `name` and `content` attributes. If we link to the HTML definition we're going to add to the confusion, as `dir`, `lang`, `scheme` and `http-equiv` are not allowed.

iherman: Right. That is why I raised the possibility to make a reference to the (outdated) HTML 4.01 spec.The only extra attribute there is `http-equiv`, but it is easy to gloss over that one, because it is indeed very HTML specific.

shiestyle (approved):

epub33/core/index.html
mattgarrish: > The only extra attribute there is `http-equiv`,

It has the others, too: https://www.w3.org/TR/html401/struct/global.html#h-7.4.4.2

It would also be an odd choice to cite for an xml grammar.

Since you can't even find how to specify a cover in the epub 2 spec., I think we're overemphasizing the problem. Short of a new section on covers, which could perhaps at least show an example of epub 2, we're not going to right historical wrongs solely by trying to re-specify the meta tag itself.

iherman: > we're not going to right historical wrongs solely by trying to re-specify the meta tag itself.

I agree. I do feel fairly uncomfortable, but maybe you are right and I am overemphasizing the problem. Let us merge this and forget about EPUB 2 :-)

iherman (approved):

wareid (approved):

PR #1749, opened 2021-07-02
add dedication to Garth Conboy in acknowledgments

dauwhe:


wareid (approved):

rdeltour (approved):

iherman (approved):
Should this text be added to the Reading System spec, too, and not only the EPUB 3.3 core tex?

dauwhe:
> Should this text be added to the Reading System spec, too, and not only the EPUB 3.3 core tex?

Done.

iherman:
@dauwhe: you should be the one merging this when you feel it is right ❤️

PR #1750, opened 2021-07-05
Reorganize acknowledgements

mattgarrish:
Replaces the three acknowledgement files with specification names with one with the dedication and a more generic one without.

I tried to include the list of names so we wouldn't have to duplicate it, but respec doesn't handle recursive inclusion so it would be a pain to add another script to every spec to handle it manually.

Also added an acknowledgements section to the multiple renditions document, even if it is just a note.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1750.html" title="Last updated on Jul 5, 2021, 1:58 PM UTC (7a7b077)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1750/45935cc...7a7b077.html" title="Last updated on Jul 5, 2021, 1:58 PM UTC (7a7b077)">Diff</a>


iherman:
The reason we had three, specification-specific files is because the editors are different, and the text says "the text acknowledge...". Ie, the script removes the editors from the ack list.

I will look into this tomorrow (today is my day off).

mattgarrish:
> The reason we had three, specification-specific files is because the editors are different, and the text says "the text acknowledge...". Ie, the script removes the editors from the ack list.

Ya, but that's easily done away with. All we need to say is something like:

"The following members of the EPUB 3 Working Group contributed to the development of this specification:"

And then it doesn't matter if the editors are listed twice. Otherwise we have to copy the acknowledgements separately into each different specification just so we can avoid a small redundancy of the editors thanking themselves. Doesn't seem worth it with this many documents.

iherman (approved):
I propose to merge this PR, and I will then change the underlying script setups so as to regenerate the final file with an updated list of participants.

Issue #1745, opened 2021-06-30
Can you tts an image in media overlays?


mattgarrish:
According to the [embedded media section](https://w3c.github.io/epub-specs/epub33/core/#sec-audio-video), you can reference images but you must have an audio clip for them because they contain no audio.

But what about images in sections that are TTSed? If we expect the reading system will use the alt text, then why can't you omit an `audio` element so images fall over into that category?

Or maybe what I'm wondering is why are images included in the embedded media section at all?

As far as I remember it used to be called "Embedded Audio and Video" and we only changed it because images were also listed as an example of embedded media in the prose of the section. Was that the wrong fix and we should have just deleted images from the section?

/cc @marisademeglio

OriKuttner:
How an image is TTS'd?
I don't think I understand.

-- 
Ori Idan CEO Helicon Books
http://www.heliconbooks.com

danielweck:
We recently had a discussion in the Thorium issue tracker about an EPUB Media Overlays feature which is not commonly used / authored, and as far as I know rarely implemented: i.e. "filling the gaps" with TTS synthetic voice when pre-recorded human narration is not available (SMIL text references without audio clips).
We implemented support for this feature in the original Readium SDK, but so far in Readium "2" we haven't tackled this, due to lack of real-world adoption / usage.
The use-case seems obvious at a UX surface level (which is why this feature was added in the specification), but looking into practical details I think there are some concerning downsides, such as the inevitable inconsistencies between a professionally-prerecorded audio track and an arbitrary system-dependent synthetic voice (e.g. volume levels, speech rate, voice selection, etc.). These aspects are largely underspecified and it leaves both reading system implementers and end-users with vague expectations.
Rather than fixing this part of the spec., I would be in favour of deprecating the feature, which to my knowledge is poorly supported in the real world. I don't think it would be a big loss, if any at all.
cc @llemeurfr

danielweck:
Spec links:

https://www.w3.org/publishing/epub/epub-mediaoverlays.html#sec-text-to-speech

https://w3c.github.io/epub-specs/archive/epub32/spec/epub-mediaoverlays.html#sec-text-to-speech

danielweck:
Thorium issue tracker discussion: https://github.com/edrlab/thorium-reader/issues/1515#issuecomment-864137485

mattgarrish:
[...]

PR #1753, opened 2021-07-08
Separate handling of audio/video from images in media overlays tts

mattgarrish:
This pull request makes the minimal changes I mentioned in #1745 to clean up the embedded media and TTS sections. I've split the embedded media section to define the handling of audio and video separately from images, and noted that reference to an image without audio is expected to voiced by TTS (since this matches the already-existing prose that there should be text or a text fallback for whatever is destined for TTS).

There may be other issues we need to solve to make the TTS section more robust, but let's handle those separately (if we want to go deeper into this).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1753.html" title="Last updated on Jul 22, 2021, 1:11 PM UTC (9e5ed21)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1753/e674219...9e5ed21.html" title="Last updated on Jul 22, 2021, 1:11 PM UTC (9e5ed21)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7169,16 +7169,19 @@ store destination as source in ocf
 
 					<section id="sec-media-overlays-fragids">
 						<h5>Referencing Document Fragments</h5>
+
 						<p>The <code>epub:textref</code> attribute and the <a href="#elemdef-smil-text"
 									><code>text</code> element's</a>
 							<code>src</code> attribute both require a fragment identifier that references a specific
 							fragment (e.g., an element or text string) of the associated <a>EPUB Content
 							Document</a>.</p>
+
 						<p>For XHTML and SVG Content Documents, the fragment SHOULD be a reference to a <a
 								href="https://html.spec.whatwg.org/multipage/browsing-the-web.html#target-element"
 								>target element</a> [[HTML]] or an <a
 								href="https://www.w3.org/TR/SVG2/linking.html#SVGFragmentIdentifiers">SVG Fragment
 								Identifier</a> [[SVG]], respectively.</p>
+
 						<p>EPUB Creators MAY use other fragment identifier schemes, but Reading Systems may not support
 							such identifiers.</p>
 					</section>
@@ -7203,7 +7206,7 @@ store destination as source in ocf
 							supported.</p>
 					</section>
 
-					<section id="sec-audio-video">
+					<section id="sec-embedded-media">
 						<h5>Embedded Media</h5>
 
 						<p>Any <a>EPUB Content Document</a> associated with a Media Overlay MAY contain embedded media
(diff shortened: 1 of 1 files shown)


iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-07-08-epub#section4) on 2021-07-08

- no resolutions were taken

### 4. Media Overlays and Images 

_See github issue [#1745](https://github.com/w3c/epub-specs/issues/1745)._




_See github pull request [#1753](https://github.com/w3c/epub-specs/pull/1753)._




**Wendy Reid:** can you TTS an image in a media overlay document?  
… question comes down to, in MO we make mention of the fact that there should be an audio alternative to media (i.e. audio, video, AND images)  
… and whether TTS and alt-text could fill this requirement  

> *Dave Cramer:* [https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1753/0c4b16f...692aabc.html#sec-embedded-media](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1753/0c4b16f...692aabc.html#sec-embedded-media)

**Dave Cramer:** this diff preview seems to be working better for me  
… basically it splits out embedded images separately from audio and video  
… says that audio for embedded images is optional, and TTS reading out alt-text will be used as a fallback  
… to me this makes sense  

**Brady Duga:** not sure it makes sense to me. The purpose of MO is to have narrated text associated with the elements of the MO. Not sure why you would reference the image in MO but not have narration, and instead rely on the TTS and alt-text fallback  

**Matt Garrish:** in embedded images we include ability to embed audio, video, and images. This is so that playback of audio, video, etc. could be started via the MO  

**Brady Duga:** my reading of spec is that audio element is required for every text element  

**Matt Garrish:** so what happens when the "text" element is an image?  

**Brady Duga:** well i don't think it would be valid to include a reference to an image without audio. Spec says audio must be included unless it is "content intended to be rendered via TTS". Not sure if image (within a div) counts.  

**Matt Garrish:** in HTML normally TTS would pick up the alt-text of the image  
… so why can't the TTS be used to voice the alt-text of the image in this case  

**Brady Duga:** not sure what normal TTS behaviour is. Perhaps this is why use of this fallback TTS feature is rare  

**Matt Garrish:** My guess is that the purpose of having TTS as a fallback to narration is that you might not want to get a professional narrator to narrate some parts of a MO book. e.g. backmatter  
… so you could leave this to TTS  

**Dave Cramer:** I think we should wait for more comment. From Marisa, from NNELs, from other users of MO in the wild  

</details>

marisademeglio (commented):
I think this version is clearer; I also think the TTS stuff will need a closer look, like you said.

PR #1756, opened 2021-07-09
Restore the custom attribute sections

### LINKED ISSUE #1602: Spec recommends documenting custom attributes, but zero documented in 5 years
https://w3c.github.io/epub-specs/epub33/rs/#sec-xhtml-custom-attributes says: `To facilitate interoperability of custom attributes across Reading Systems, vendors are strongly encouraged to document any extensions they implement in [AttributeExtensions].`

That links to https://idpf.github.io/epub-extensions/attributes/, dated October 2016, which says `No custom attributes have been defined at this time.`

I think we should either document all known custom attributes (on behalf of the vendors), or should remove this recommendation. In the current form it's a frustrating experience and looks embarrassing.

mattgarrish:
Also adds @iherman's suggested text about rs-specific nature of the attributes. I've made this a note, though, as it's more describing future process.

Only other change was to use "domain" instead of "authority component" in the part about the restricted strings so we can reference the URL standard instead of RFC3186.

Fixes #1602

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1602/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1602/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1756.html" title="Last updated on Aug 15, 2021, 7:18 AM UTC (7c2e047)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1756/e674219...7c2e047.html" title="Last updated on Aug 15, 2021, 7:18 AM UTC (7c2e047)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3603,14 +3603,6 @@ Spine:
 						<p>When using custom attributes, the content MUST remain consumable by a user without any
 							information loss or other significant deterioration, regardless of the Reading System it is
 							rendered on.</p>
-
-						<div class="note">
-							<p>Custom attributes are usually defined in a Reading System-specific manner and are not
-								intended for use by Reading Systems other than the one that they have been defined for.
-								For generic extensions that are to be used by multiple independent Reading Systems,
-								future versions of this specification should be extended to provide the feature
-								explicitly.</p>
-						</div>
 					</section>
 				</section>
 
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3589,6 +3589,29 @@ Spine:
 						<p>Use of the <code>trigger</code> element is <a href="#deprecated">deprecated</a>. Refer to its
 							definition in [[EPUBContentDocs-301]] for usage information.</p>
 					</section>
+
+					<section id="sec-xhtml-custom-attributes">
+						<h5>Custom Attributes</h5>
+
+						<p><a>EPUB Creators</a> MAY specify <a
+								href="https://www.w3.org/TR/epub-rs-33/#sec-xhtml-custom-attributes">custom
+								attributes</a> [[EPUB-RS-33]] in <a>XHTML Content Documents</a> to take advantage of
+								<a>Reading System</a>-specific functionality not defined in this specification.</p>
+
+						<p>Custom attributes MAY be specified on any element in an XHTML Content Document.</p>
+
+						<p>When using custom attributes, the content MUST remain consumable by a user without any
+							information loss or other significant deterioration, regardless of the Reading System it is
+							rendered on.</p>
+
+						<div class="note">
+							<p>Custom attributes are usually defined in a Reading System-specific manner and are not
+								intended for use by Reading Systems other than the one that they have been defined for.
+								For generic extensions that are to be used by multiple independent Reading Systems,
+								future versions of this specification should be extended to provide the feature
+								explicitly.</p>
+						</div>
+					</section>
 				</section>
 
 				<section id="sec-xhtml-deviations">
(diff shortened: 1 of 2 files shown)


iherman (approved):
I just wonder whether, for reasons of symmetry, we should not make a similar section (or factor out a common section) for SVG, too. After all, an RS may decide to have a RS-specific attribute in SVG content, too.

(I realize this is probably a technical purity issue of the spec rather than reflecting real usage out there, so I am o.k. if you decide to ignore this...)

mattgarrish:
> After all, an RS may decide to have a RS-specific attribute in SVG content, too.

We don't have this issue with SVG: https://www.w3.org/TR/SVG2/struct.html#ForeignNamespaces

HTML is more restrictive, so we had to explicitly allow foreign attributes.

shiestyle (approved):

mattgarrish:
I guess we should wait on a formal WG nod before merging this one, since it's undoing an earlier resolution.

dauwhe (changes requested):
I think we should be more clear about what a custom attribute is (it's not quite a standard term in HTML). As far as I can tell, what EPUB has historically allowed is attributes in namespaces other than `xhtml` or `epub`. Just adding a `foo` attribute won't validate.

iherman:
> I think we should be more clear about what a custom attribute is (it's not quite a standard term in HTML). As far as I can tell, what EPUB has historically allowed is attributes in namespaces other than `xhtml` or `epub`. Just adding a `foo` attribute won't validate.

The proposed text says:

>  To facilitate this experimentation, vendors  MAY  define custom attributes for use in  XHTML Content Documents  provided they are from a foreign namespace, which is defined as a namespace [  XML-NAMES  ] that does not incorporate either of the following strings in its  domain  [  URL  ]: 
>
> * w3.org 
> * idpf.org 

For me, that makes it explicit that just `foo` won't be acceptable because it would not be in a foreign namespace.

dauwhe:
Ah, I guess I hope there would be some explanation about the namespaces in the authoring spec as well as the reading system spec. I fear content authors will see this as a license to use `<p foo="bar">` and will then wonder why it fails EPUBCheck.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-08-13-epub#section3) on 2021-08-13

List of resolutions:

- ***[Resolution No. 2](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-08-13-epub#resolution2):  Merge PR 1756***

### 3. Custom Attributes 

_See github pull request [#1756](https://github.com/w3c/epub-specs/pull/1756)._




**Dave Cramer:** we took out the custom attributes section, but it turns out there are RS that use these  
… so now we want to put it back  

**Dan Lazin:** can we get some documentation on what exists in terms of custom attributes? So people can make use of them in their books?  

**Dave Cramer:** we can see about including some examples  

> **Proposed resolution: Merge PR 1756** *(Wendy Reid)*

> *Dave Cramer:* +1

> *Wendy Reid:* +1

> *Dan Lazin:* +1

> *Matthew Chan:* +1

> *Ivan Herman:* +1

> *Charles LaPierre:* +1

> *Ben Schroeter:* +1

> *Brady Duga:* +1

> *Masakazu Kitahara:* +1

> ***Resolution #2: Merge PR 1756***

**Dave Cramer:** okay, thanks everyone!  

---


</details>

iherman:
> Ah, I guess I hope there would be some explanation about the namespaces in the authoring spec as well as the reading system spec. I fear content authors will see this as a license to use `<p foo="bar">` and will then wonder why it fails EPUBCheck.

@dauwhe I do not think the normative statements should be changed. But I have added an extra note right after the list of domains, as follows:

> Note that this restriction also disallows the usage of custom attributes without a namespace prefix 
> (e.g., <code>&lt;p foo=""&gt;…&lt;/p&gt;</code>). Indeed, such attributes are considered to be part of
> the default namespace, i.e., either the XHTML or SVG namespaces, which are both in the <code>w3.org</code> domain.

iherman:
@dauwhe are you o.k. with https://github.com/w3c/epub-specs/pull/1756#issuecomment-899008259? We could then close this PR

dauwhe (approved):

PR #1764, opened 2021-08-03
Set up GH Actions for immediate publishing, a.k.a. live dangerously...

iherman:
The goal of this is to ensure that

- each document can be published separately (hence the series of individual yml files) 
- the yml file that is triggered is controlled via a path filter both on the pull request event and the push->main one

My understanding from the examples on spec-prod is that this will run an epubcheck on a pull request (including any push to that pull request) and will do a complete publication when the pull request is merged...

Hopefully @plehegar or @deniak can look at this and ensure that the changes are o.k...


iherman:
Well the CI-s have already run:

- The fact that the path filters allowed the run is actually all right, because I added an ECHIDNA file to each of the source directory (the fact that the CI-s actually failed is good, because I did not want to really do an ECHIDNA publishing at this stage. I got lucky:-)
- @deniak @plehegar I do not understand the error message in spec-prod runs. There is a 'SOURCE' variable set in the yml files, pointing at the right folders, and each of those has an 'index.html' file. I thought that was enough to say that this is a respec job...

iherman:
I have now added an explicit TOOLCHAIN entry for all actions (although the [description](https://github.com/w3c/spec-prod/blob/main/docs/options.md) suggests that this is not necessary).

iherman:
The latest run also started up CI (why?). I looked at the details of the failures, and it included the line

```
$ respec -s "http://localhost:3000/epub-a11y-eaa-mapping/index.html" ...
```

And that of course went wrong: the source for running is in `epub33/epub-a11-eaa-mapping`. However, the relevant yml file says:

```
        with:
            SOURCE: epub33/epub-a11y-eaa-mapping/
```

What went wrong? 

---

Also, I do not understand why, this time, the actions ran in the first place. There is a filter for paths in the pull_request, I thought that would avoid running the action unless the files that changed are indeed in a specific directory...

@marcoscaceres I also ping you for help, I believe you were also involved with all this...

dauwhe (approved):
I approve but can offer no insight into whether it is set up correctly :)

iherman:
> I approve but can offer no insight into whether it is set up correctly :)

:-) I am still waiting for some inputs on these cryptic thingies...

tidoust:
I raised the path normalization issue as a bug against spec-prod, see: https://github.com/w3c/spec-prod/issues/92

A workaround is to complete the `SOURCE` settings with the filename, for instance to have `epub33/core/index.html` instead of `epub33/core/`.

epub33/core/ECHIDNA
```diff
@@ -0,0 +1,5 @@
+# ECHIDNA configuration
```
sidvishnoi: The Echidna manifest files are not required with spec-prod, so feel free to remove them.

marcoscaceres:
> @marcoscaceres I also ping you for help, I believe you were also involved with all this...

@tidoust comment above about SOURCE addresses this, right?

epub33/core/ECHIDNA
iherman: Thanks, I will. But is this also true when the document includes external files like figures (some of the specs we have do that)?

iherman:
> > @marcoscaceres I also ping you for help, I believe you were also involved with all this...
> 
> @tidoust comment above about SOURCE addresses this, right?

I believe so :-) but I have not yet tested it (hope to do it sometimes this week).

I still have to understand this spec-prod mystery whereby the same action would run in the PR but only doing a document check and would indeed publish via ECHIDNA when merged into main. Mystery in the sense that I would like to understand how that happens; is it a clever thingy in spec-prod itself?

I really appreciate all this help, guys. Thanks.

marcoscaceres:
@sidvishnoi is the brains behind spec-prod, so he might be able to answer any questions (or point you to the appropriate code... I think spec-prod is written in TypeScript, so should be fairly easy to figure out any magic bits).

tidoust:
> > @tidoust comment above about SOURCE addresses this, right?
> 
> I believe so :-) but I have not yet tested it (hope to do it sometimes this week).

It should, and note bug was fixed in spec-prod, so workaround should no longer be necessary.


> I still have to understand this spec-prod mystery whereby the same action would run in the PR but only doing a document check and would indeed publish via ECHIDNA when merged into main. Mystery in the sense that I would like to understand how that happens; is it a clever thingy in spec-prod itself?

Yes, that is a spec-prod feature. When run on a PR, spec-prod builds the document, runs tests on them, but does not publish it anywhere. That is why it is useful to run spec-prod on PR (what would be the point otherwise?).

From a code perspective, the action checks whether it was triggered by a `push` event or a `pull_request` event: https://github.com/w3c/spec-prod/blob/main/src/prepare-deploy.ts#L69

epub33/core/ECHIDNA
sidvishnoi: Yes. It'll [find them automatically](https://github.com/w3c/spec-prod/blob/87f0bf4eac6e0132a4c4b64ae5419e4f3f78cdf7/src/build.ts#L169) and copy them, as long as they're in same repo or are images hosted on githubusercontent.com (images in GitHub issues).
You can verify that all images/embedded content is picked in the "build" step of logs.

iherman:
> @sidvishnoi is the brains behind spec-prod, so he might be able to answer any questions (or point you to the appropriate code... I think spec-prod is written in TypeScript, so should be fairly easy to figure out any magic bits).

I am pleased to see that I am not the only one who got bitten by the TypeScript bug :-)

iherman:
> Yes, that is a spec-prod feature. When run on a PR, spec-prod builds the document, runs tests on them, but does not publish it anywhere. That is why it is useful to run spec-prod on PR (what would be the point otherwise?).
> 
> From a code perspective, the action checks whether it was triggered by a `push` event or a `pull_request` event: https://github.com/w3c/spec-prod/blob/main/src/prepare-deploy.ts#L69

Ah! All seems to be clear then. I will try to update things soon and see whether things are now working or not.

Thanks @tidoust @marcoscaceres @sidvishnoi for your help!

iherman:
Well... I removed the ECHIDNA files, and I also merged the content of 'main' into this branch to get the content synced. However, none of the actions have run... which seems to be in contrast with what we said before.

The ultimate test would be to merge this into main, but that may trigger real /TR publication, and I prefer @mattgarrish (the real, main editor of the specs :-) to be present...

iherman:
Ah! I have made an explicit change on one of the spec files (overview) and the action has run successfully.

Will wait for @mattgarrish for a merge but it seems that things are o.k.!

tidoust:
> Well... I removed the ECHIDNA files, and I also merged the content of 'main' into this branch to get the content synced. However, none of the actions have run... which seems to be in contrast with what we said before.

Hmm, I'm not sure where you see a contradiction? The workflows are setup to run on a pull request when it contains changes to files that match the `paths` condition. They initially ran because the ECHIDNA files were touching the appropriate folders. When you removed the ECHIDNA files from the pull request, you removed all changes on these folders from the pull request, so workflows no longer had any reason to run. The change under `epub33/overview/` reintroduced a file change in the pull request that made the relevant workflow run (and only that one).

I note that it could be considered good practice to add in `paths` a condition that makes the workflow run when the workflow file itself is updated, as in https://github.com/w3c/webcodecs/blob/main/.github/workflows/deploy-samples.yml#L5. This is useful to force execution of the workflow when you need to tweak something in it. Otherwise you'll have to touch one of the files under the right folders to trigger the new workflow.

iherman:
> When you removed the ECHIDNA files from the pull request, you removed all changes on these folders from the pull request, so workflows no longer had any reason to run. The change under `epub33/overview/` reintroduced a file change in the pull request that made the relevant workflow run (and only that one).

Absolutely right.

iherman:
> I note that it could be considered good practice to add in `paths` a condition that makes the workflow run when the workflow file itself is updated, as in https://github.com/w3c/webcodecs/blob/main/.github/workflows/deploy-samples.yml#L5. This is useful to force execution of the workflow when you need to tweak something in it. Otherwise you'll have to touch one of the files under the right folders to trigger the new workflow.

Good idea. Will do it...

PR #1780, opened 2021-08-24
Add last modification date to examples of metadata to check for uniqueness

### LINKED ISSUE #1772: [EPUB 3.3 RS] [suggestion] Unique Identifier
The specs say: 

> Determining whether two EPUB Publications with the same Unique Identifier represent different versions of the same publication, or different publications, may require inspecting other metadata, such as the titles or authors. (https://www.w3.org/TR/epub-rs-33/#sec-pkg-doc-pub-identifiers)

Perhaps the date metadata can also be flagged as significant in distinguishing a publication, in particular the [last modified date](https://www.w3.org/TR/epub-33/#sec-metadata-last-modified).

mattgarrish:
Fixes #1772

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -319,7 +319,7 @@
 				<p>Reading Systems SHOULD NOT depend on the <code>unique-identifier</code> attribute being unique to one
 					and only one EPUB Publication. Determining whether two EPUB Publications with the same Unique
 					Identifier represent different versions of the same publication, or different publications, may
-					require inspecting other metadata, such as the titles or authors.</p>
+					require inspecting other metadata, such as the last modification date, titles, and authors.</p>
 			</section>
 
 			<section id="sec-pkg-doc-metadata">


epub33/rs/index.html
```diff
@@ -319,7 +319,7 @@ <h3>Unique Identifier</h3>
 				<p>Reading Systems SHOULD NOT depend on the <code>unique-identifier</code> attribute being unique to one
 					and only one EPUB Publication. Determining whether two EPUB Publications with the same Unique
 					Identifier represent different versions of the same publication, or different publications, may
-					require inspecting other metadata, such as the titles or authors.</p>
+					require inspecting other metadata, such as the last modification date, titles, and authors.</p>
 			</section>
```
iherman, line 323: Just to be picky:-)... shouldn't this be ", or" instead of ", and"?

(I must admit I am often mixed up by the usage of "and" vs. "or" in these types of lists in English...)

iherman (approved):

epub33/rs/index.html
mattgarrish: It's a list of possible choices that can all be used, not an exclusive list of options in which you pick only one. The original wording using "or" makes it sound like you pick one.

iherman: I trust you on that one:-)

PR #1783, opened 2021-08-25
Added alt text on the svg images

iherman:
@clapierre, I would appreciate if you looked at these. You have much more experience than I have. 

The figures themselves are in section D.4.1.1.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1783.html" title="Last updated on Sep 2, 2021, 1:00 PM UTC (27179e4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1783/3cfe077...27179e4.html" title="Last updated on Sep 2, 2021, 1:00 PM UTC (27179e4)">Diff</a>

### DIFF
--- a/epub33/common/css/common.css
+++ b/epub33/common/css/common.css
@@ -350,3 +350,12 @@ div.caution-title > span {
 	color: rgb(140,40,0);
 	text-transform: uppercase;
 }
+
+details.desc {
+	margin-top: -2.5rem !important;
+	margin-left: 3rem !important;
+}
+
+details.desc > summary {
+	text-align: left;
+}
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -80,21 +80,6 @@
 		}
             };//]]>
       </script>
-	  <style>
-		details.figure {
-			margin-bottom: 0;
-		}
-		details.figure > p {
-			text-align: left;
-			font-size: 80%;
-		}
-		details.figure > summary {
-			list-style: none;
-		}
-		details.figure > summary::-webkit-details-marker {
-			display: none;
-		}
-	  </style>
 	</head>
 	<body>
 		<section id="abstract">
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -80,6 +80,21 @@
 		}
             };//]]>
       </script>
+	  <style>
+		details.figure {
+			margin-bottom: 0;
+		}
+		details.figure > p {
+			text-align: left;
+			font-size: 80%;
+		}
+		details.figure > summary {
+			list-style: none;
+		}
+		details.figure > summary::-webkit-details-marker {
+			display: none;
+		}
+	  </style>
 	</head>
 	<body>
 		<section id="abstract">
--- a/epub33/core/vocab/rendering.html
+++ b/epub33/core/vocab/rendering.html
@@ -76,10 +76,8 @@
 			<p>The <code>rendition:flow</code> property MUST NOT be declared more than once.</p>
 
 			<figure id="fig-flow-paginated-single">
-				<figcaption>
-					<p>Rendering of an EPUB publication with a single spine item, and with the
-						<code>rendition:flow</code> set to <code>paginated</code>.</p>
-				</figcaption>
+				<figcaption>Rendering of an EPUB publication with a single spine item, and with the
+					<code>rendition:flow</code> set to <code>paginated</code>.</figcaption>
 				<img src="images/Rendering_Paginated_Single_Spine.svg" width="600" aria-details="flow-paginated-single-diagram"
 					alt="The continuous progression of paginated content produced for a single document."/>
 			</figure>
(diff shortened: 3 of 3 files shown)


mattgarrish:
These read a bit more like extended descriptions than alt text to me.

iherman:
> These read a bit more like extended descriptions than alt text to me.

To be honest, I have difficulties to differentiate these in practice... 

An alternative is to move these texts into the `<desc>` element of the SVG (together with a `<title>`. This is a bit uneasy, because if we want to make even the slightest change on the figure (the original is a google doc image) then this step has to be repeated, but that should be o.k. The question is whether accessibility tools would pick up the description from the SVG code; I simply do not know.

Alternatively, one could use the `<details>` element in HTML for this (which we should then do systematically for all possible figures in our specs).

Also, what would be the proper alt text? What else than what is already in the figcaption element?

mattgarrish:
Alt text is normally a brief sentence-length replacement for the image that says what it is without going into all the details of describing it. The captions are, indeed, closer to alt text than the current alt attribute values. The alt is only supposed to allow the user to decide if they want to listen to that fuller explanation, not require they listen to the whole thing before being able to move on (I don't believe you can escape from alt text once it starts being read).

We could use the `details` element to provide the descriptions and wire them up to the images with `aria-details`. That would make them more accessible to anyone, and also avoid their being buried in the svg markup.

But I'll let Charles weigh in with his opinions.

iherman:
Here is a new version. The structure I have used is as follows:

```
<figure role="group">
    <details class="figure">
        <summary>				
           <img src="...." width="600" alt="Rendering…" aria-describedby="paginated_multiple"/>
        </summary>
        <p id="paginated_multiple">Three column-like rectangles…</p>
    </details>
    <figcaption>
        Rendering ...
    </figcaption>
</figure>
```

- the alt text is identical to the `figcaption` text. This violates the DRY principles, but I did not know better; I presume that a missing `alt` or an `alt=""` is not really acceptable, despite `aria-describedby`. 
    - I was actually wondering whether an empty `alt `and an `aria-labelledby` pointing to the `figcaption` would be acceptable 
- I have also added some CSS into the `index.html` file whereby the detail's trigger (usually a small triangle) is hidden. (Clicking on the empty space where it would be still makes it appear, though, except on Safari, where a specific css must be used to get it completely gone. Thanks MDN...:-)

How does this look now? Visually, it is identical to what was before...

clapierre:
Unfortunately I am unable to see the completely rendered page.
looking at the code on my Mac all the svg images are referenced by:
``
which shows a broken image
when I changed the code to (adding "../" in front of the images directory it works
``

Is there a specific Branch I can pull to see this and do we need to add the ../ ?
Hard to figure this all out without it working.
Thanks.

mattgarrish:
@clapierre If you're opening the core spec locally you need to use a web server to view it properly. Respec imports the vocabulary page into the core doc, so the images are not a level above when the document is fully combined (i.e., we'd break our published docs if we changed).

If you set the server root to the /epub33 directory, for example, you should be able to view the full specification using http://localhost/core (swapping /core for any of the other of the specs)

It doesn't look like the aws preview can handle images, either, so if you're stuck without a local server you'll just have to hack the vocabulary page like you've done by referencing up a level.

mattgarrish:
It looks like our low-tech [preview](https://cdn.statically.io/gh/w3c/epub-specs/alt-text-on-svg-images/epub33/core/index.html) option still works.

(Edit: the w3c diff is also busted, so removed that link.)

clapierre:
I agree we Matt we can't repeat the Fig caption in the alt-text description here.  Also since this image is wrapped within a Summary element we are not getting the aria-describedby announced anyways at least not with Voice Over on my Mac.
I say we remove the aria-describedby and change the alt-text description to instead say "Expand / collapse extended description for Figure 1"

I would also move the Caption above the image so that it appears first and then we don't have to repeat anything and can only refer to this extended description by Figure 1 which will make this image description unique.

EG:

> 			<figure>
> 				<figcaption>
> 					Rendering of an EPUB publication with a single spine item, and with the <code>rendition:flow</code> set to <code>paginated</code>.
> 				</figcaption>
> 				<details class="figure">
> 					<summary>				
> 						Expand / Collapse extended description for Figure 1.
> 					</summary>
> 					<p>
> 						Three column-like rectangles linked left-to-middle and middle-to-right with respective arrows,
> 						with a text flowing from one rectangle to the next one. The text is sectioned with headers
> 						figuring 'Chapter 1', '2', and '3'. The leftmost rectangle is enclosed in a schematic view of
> 						a tablet.
> 					</p>
> 				</details>
> 			</figure>

Now if we don't want to hard code "Figure 1" in the alt text maybe we could make it something like
"Expand / collapse extended description for Single Paginated Spine"

GeorgeKerscher:
Recall that having the details inside the figure does not work for Jaws, and IMO they should fix that.

Our recommendation is to have the details outside the figure.

mattgarrish:
I've rearranged the figures to put the captions first and move the description in the details element outside the closing tags.

@clapierre I've rewritten the alt attributes to try and work as non-repetitive alt text, but let me know if you think this is unique enough from the captions or if you have any other ideas on how to phrase them.

clapierre:
The Summary/details should come after the figure not before, if I am seeing it correctly via the diff.  Otherwise rest looks good.

mattgarrish:
Ya, they are after their respective figures now, not before. The diff is a bit weird in showing a piece of the old captions being replaced by the details elements, but the elements are for the figures before.

clapierre:
I think I understand so to be clear we now have the following right?

```
<Figure>
  <Fig Caption/>
  <Image/>
</Figure>
<Summary / Details>
```

mattgarrish:
Right, with aria-details linking from the image to the description.

clapierre:
Great works for me!  Thanks Matt.

clapierre (approved):

PR #1784, opened 2021-08-25
Reading system accessibility section improvements

### LINKED ISSUE #1770: [EPUB 3.3 RS] [question] Relationship to HTML
The specifications say:

> This specification does not require EPUB Reading Systems to support scripting, HTML forms or the HTML DOM. Reading Systems conformant with this specification are only expected to be able to process a conforming EPUB Content Document. As support for scripting and HTML forms is not compulsory, a conformant Reading System may not be a fully-conformant HTML user agent. (https://www.w3.org/TR/epub-rs-33/#rel)

I wonder: can not supporting HTML DOM create problems for accessibility? Maybe for ARIA attributes? Maybe HTML DOM would need to be supported if the application has a viewport?

### LINKED ISSUE #1773: [EPUB 3.3 RS] [suggestion] Metadata lang
I think the [metadata section](https://www.w3.org/TR/epub-rs-33/#sec-pkg-doc-metadata) should say that each metadata element can have an `xml:lang` attribute and that this information should be honored when the metadata is shown to the user (e.g. in the bookshelf view of the application).

mattgarrish:
This PR expands the bullet on bookshelf accessibility to also mention exposing the language of the text and adds a new bullet about exposing the DOM. We already had a bullet about exposing the accessibility tree to the platform APIs, so I think the primary request was already covered in a different manner. But no harm is also suggesting to allow access to the DOM itself, since it provides different information.

Fixes #1770 
Fixes #1773 

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/editorial/rs-a11y-improvements/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/editorial/rs-a11y-improvements/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1787,7 +1787,7 @@
 					properties [[WAI-ARIA-11]], is exposed to the underlying operating system's accessibility API so
 					that users can fully interact with the content when using assistive technologies.</li>
 				<li>Document Object Model (DOM) &#8212; Provide access to the [[DOM]] of EPUB Content Documents so that
-					users can better understand the semantics and structure.</li>
+					users can investigate the semantics and structure used in the source.</li>
 				<li>Table of Contents &#8212; Ensure that the rendering of the table of contents is accessible to
 					assistive technologies (e.g., that the link text can be read and the links activated).</li>
 			</ul>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -73,7 +73,7 @@
 				lint: {
 					"wpt-tests-exist": true,
 				}
-			};//]]></script> 
+			};//]]></script>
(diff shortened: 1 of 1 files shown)


GeorgeKerscher:
I think this is fine.

gregoriopellegrino (approved):
Fine for me. Thanks

avneeshsingh:
looks fine.

PR #1794, opened 2021-09-06
Adapted the A11y structure for imaged to the overview

iherman:
I have adopted to structure used in #1783 to the diagram in the overview document.

However... I wonder whether it would not be cleaner to push all the image descriptions to an appendix, instead of leaving the image descriptions intertwined with the text. For a visual user it does not add anything; for a non-visual user the aria attribute refers to the description, so the description could just be anywhere (and there is no need for a details element).

See:

* For EPUB 3 Overview:
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/overview-diagram-a11y/epub33/overview/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/overview/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/overview-diagram-a11y/epub33/overview/index.html)


iherman:
@mattgarrish how did #1783 go through the publication? The Ci error message says that area-details is not appropriate for an img element???

mattgarrish:
s/area-/aria-/ ;)

iherman:
> s/area-/aria-/ ;)

Sigh... :-)

(This github action setup is actually good. It kicks you in the butt early in the process...)

mattgarrish:
> For a visual user it does not add anything

I don't think this is necessarily true. We're assuming our diagrams will be understandable because they are to us, but not every visual reader may make sense of them. Descriptions aren't only for blind and low vision readers.

I kind of like the current pattern for a few reasons: it's easy to forget there are descriptions that need updating when they're far away from their source; it's not easy to get to/from the descriptions as aria-details isn't perfectly supported and not available outside of AT, plus we'd need backlinking from the appendix in case that's where you've started; they take up less space being collapsed than if they're all written out in an appendix.

iherman:
> > For a visual user it does not add anything
> 
> I don't think this is necessarily true. We're assuming our diagrams will be understandable because they are to us, but not every visual reader may make sense of them. Descriptions aren't only for blind and low vision readers.
> 
> I kind of like the current pattern for a few reasons: it's easy to forget there are descriptions that need updating when they're far away from their source; it's not easy to get to/from the descriptions as aria-details isn't perfectly supported and not available outside of AT, plus we'd need backlinking from the appendix in case that's where you've started; they take up less space being collapsed than if they're all written out in an appendix.

Ok. I won't lie down the road on it :-) but I do find that lonely "Description" text a bit strange...

mattgarrish:
> I do find that lonely "Description" text a bit strange...

What if we gave it a friend and called it "Image description"?

The problem with making it too expressive is that if we use something like "Expand for image description" then it becomes kind of silly after expanding unless we write a bunch of code to toggle the summary text.

Italicizing the label to match the caption, instead of bolding it, might also make it visually less awkward.

clapierre (changes requested):
I like Matt's suggestions, the other option could be to use a Clickable Image of an [Accessibility Icon from the APA](https://www.w3.org/WAI/teach-advocate/accessibility-icon-set) which would be the Summary instead of "Image Description" as Matt suggests.
Also if we move this description somewhere else we would also need to duplicate the image as you need to have the image in proximity to its description to help with cognitive understanding of what is being described.

I am also not thrilled with this image description either.

I would suggest the following.

A component diagram with 'EPUB (OCF) Container' as the outer most component which encapsulates the 'EPUB Publication' containing two documents 'Publication Document' and 'Navigation Document' along with a inner component labelled 'Publication Resources' which contains multiple 'Content Documents' (XHTML, SVG), and multiple 'Other resources' (CSS, png, mp3, mov, ...).

iherman:
> I like Matt's suggestions, the other option could be to use a Clickable Image of an [Accessibility Icon from the APA](https://www.w3.org/WAI/teach-advocate/accessibility-icon-set) which would be the Summary instead of "Image Description" as Matt suggests.

@clapierre the page on the accessibility icons say "These icons are drafts in-progress; do not use them yet." So we should not :-(

I have changed the summary text and the page description itself. If we all agree with the approach, I will also change the relevant parts in the core spec.

@mattgarrish do you what sets the summary bold? Is it in common.css (I did not find it) or is it a default HTML style?

mattgarrish:
> the other option could be to use a Clickable Image of an [Accessibility Icon from the APA](https://www.w3.org/WAI/teach-advocate/accessibility-icon-set) which would be the Summary instead of "Image Description"

I'm not sure those would fit, as they aren't specifically about descriptions. We'd probably need something more like the info icon, but then we get into hunting for royalty free icons or having to attribute use.

> do you what sets the summary bold? Is it in common.css (I did not find it) or is it a default HTML style?

It looks like it's getting it from the W3C base.css style sheet that respec imports. You'll likely need to add an !important rule to common.css to set it back to normal. You should be able to reuse the style rule for summaries at the end of the file.

clapierre:
Agreed on all accounts here would be good for them to finalize those images says it should have happend 6 months ago ;)

iherman:
I have made the changes. Note that the PR includes a change on common.css as well as changes in the core spec, so merging will trigger a complete republication...

iherman:
@mattgarrish an editorial issue: the mechanism to copy the common.css into the core index file via a script did not seem to work in the Overview file for reasons that I was unable to find. (The reference to the js was actually missing, but I have added it now.). What is a bit discomforting is that, when looking at the developers' console, I see the following run-time error message:

> css-inline.js:20 [Deprecation] Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check https://xhr.spec.whatwg.org/.

We may have to update the script (using fetch?) or re-think the way we use this. Would it be possible to rely on respec's data include?

gregoriopellegrino (approved):

iherman:
@mattgarrish @clapierre are we o.k. merging this?

clapierre:
Did we adjust the image description?
IE:
A component diagram with 'EPUB (OCF) Container' as the outer most component which encapsulates the 'EPUB Publication' containing two documents 'Publication Document' and 'Navigation Document' along with a inner component labelled 'Publication Resources' which contains multiple 'Content Documents' (XHTML, SVG), and multiple 'Other resources' (CSS, png, mp3, mov, ...).

iherman:
> Did we adjust the image description?
> IE:
> A component diagram with 'EPUB (OCF) Container' as the outer most component which encapsulates the 'EPUB Publication' containing two documents 'Publication Document' and 'Navigation Document' along with a inner component labelled 'Publication Resources' which contains multiple 'Content Documents' (XHTML, SVG), and multiple 'Other resources' (CSS, png, mp3, mov, ...).

Yes Charles, I have copied over your text

clapierre (approved):

PR #1796, opened 2021-09-07
Add refines attribute to certifier metadata examples

### LINKED ISSUE #1789: Clarify certifier needs to be associated with conformance statement
We corrected that reports need to be associated with the certifier in #1410, but we don't indicate who has certified the conformance claim. We should similarly show an example with the certifier linked:

```
<meta property="dcterms:conformsTo" id="a11y-conf">EPUB-A11Y-11_WCAG-21-AA</meta>
<meta property="a11y:certifiedBy" refines="#a11y-conf">Acme Certifiers</meta>
```

The problem is we'll introduce warnings/errors if we require this normatively, so it has to only be an example.

I wonder if it would make sense to say refining is optional if only one conformance claim/certifier is included in the metadata, but otherwise should be used? (But also keep this an informative statement.)

It is needlessly cumbersome unless there are multiple conformance claims or certifiers.

mattgarrish:
This pull request only adds the attribute to the examples. It does not add a requirement that the metadata be linked. (Doing so might be problematic for epub 2, for example, since there isn't a way to make explicit links between metadata.)

I also noticed that the old idpf conformance URLs were still used in some of the examples, so I went through and cleaned them out.

Fixes #1789 

- Accessibility [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/certifier-refines/epub33/a11y/index.html)
- Accessibility [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/certifier-refines/epub33/a11y/index.html)

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -939,7 +939,7 @@
 						<pre>&lt;package …>
    &lt;metadata>
       …
-      &lt;meta property="dcterms:conformsTo">EPUB-A11Y-11_WCAG-21-AA&lt;/meta>
+      &lt;meta property="dcterms:conformsTo" id="conf">EPUB-A11Y-11_WCAG-21-AA&lt;/meta>
       …
    &lt;/metadata>
    …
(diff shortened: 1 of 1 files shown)


clapierre (approved):
Great cleanup Matt and will definitely help with conformance and certification.

gregoriopellegrino (approved):

PR #1797, opened 2021-09-07
More thorough review of the overview document.

iherman:
This is the first review of the Overview document since its first publication. In needed refresh in view of the various documents that we have published since (both editorially, due to dangling links, as well as content wise). As main changes, I have

- added a reference to the list of documents we have now as part of the introduction; otherwise the average reader does not really find its way through the maze
- the temporary historical part about epub 33 is now more stable
- the historical section has been turned into an appendix

I have also added some editorial notes that are worth discussing. But, overall, I wanted to keep the existing structure, nothing revolutionary...

See:

* For EPUB 3 Overview:
    * [Preview](https://raw.githack.com/w3c/epub-specs/refresh-overview/epub33/overview/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/overview/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/refresh-overview/epub33/overview/index.html)


iherman:
(@mattgarrish as an explanation for the large number of checks: I have removed a temporary reference to the tts file from the common biblio.js, and this will trigger the republication of all documents...)

epub33/overview/index.html
```diff
@@ -110,7 +152,7 @@ <h2>Package Document</h2>
 					defines a reading order for linear consumption, and associates metadata and navigation
 					information.</p>
 
-				<p>The Package Document represents a significant improvement on a typical Web site. A Web site, for
+				<p>The Package Document represents a significant extension to a typical Web site. A Web site, for
```
mattgarrish: I'm not sure "extension" is the right word here. The original wording is controversial, but we're neither improving nor extending web sites. We're sort of building a kind of better site map. I'd have to think about this some more, but my first thought is maybe it should say something along the lines of: "The Package Document represents a unique approach to the traditional structuring of a typical Web site to facilitate the authoring of digital publications."

epub33/overview/index.html
```diff
@@ -127,6 +169,10 @@ <h2>Package Document</h2>
 					preview content, or assemble an index or dictionary from its constituent XHTML Content
 					Documents.</p>
 
+				<p class=ednote>
+					The <code>collection</code> element is rarely used in practice; is it worth keeping its reference in the overview document?
```
mattgarrish: I would leave it since the section is defining the major sections of the package document. It might be useful to turn this into a proper note and give some context, though. Maybe:

> The `collection` element is not currently used in any specifications that are actively maintained by the EPUB 3 Working Group. Its future use in authoring EPUB Publications is uncertain.

epub33/overview/index.html
```diff
@@ -142,15 +188,16 @@ <h3>Reading Order</h3>
 
 					<p>Many types of publication have an obvious reading order, or logical progression through their
 						content. A novel is an example of a highly sequential document — it typically has a beginning,
-						middle and end — but not all publications are so ordered: a cookbook or collection of
+						middle, and end — but not all publications are so ordered: a cookbook or a collection of
 						photographic images might be considered to be more like a database. All documents do, however,
 						have at least one logical ordering of all their top-level content items, whether by date, topic,
-						location, or some other criteria (e.g., a cookbook is typically ordered by type of recipe).</p>
+						location, or some other criteria (e.g., a cookbook is typically ordered by the type of a recipe).</p>
```
mattgarrish: "arranged by" probably makes more sense here than "ordered by". Could also be shortened to:

> (e.g., a cookbook is typically arranged by recipe type).

epub33/overview/index.html
```diff
@@ -142,15 +188,16 @@ <h3>Reading Order</h3>
...
 
 					<p>Each EPUB Publication defines at least one such logical ordering of all its top-level content
-						(the <a href="https://www.w3.org/TR/epub-33/#sec-spine-elem">spine</a> [[EPUB-33]]), as well as
-						a declarative table of contents (the <a>EPUB Navigation Document</a> [[EPUB-33]]). EPUB
-						Publications make these data structures available in a machine-readable way <em>external</em> to
+						(in the <a href="https://www.w3.org/TR/epub-33/#sec-spine-elem">spine</a> [[EPUB-33]] of 
+						the <a href="#sec-package-file">Package Document</a>), as well as
+						through a declarative table of contents (the <a href="#sec-nav-nav-doc">EPUB Navigation Document</a> [[EPUB-33]]).
```
mattgarrish: Not sure this works grammatically. If you remove the parentheticals you now have:

> Each EPUB Publication defines at least one such logical ordering of all its top level content, as well as through a declarative table of contents.

Maybe split these two apart so the parentheticals can be dropped:

> Each EPUB Publication defines at least one such logical ordering of all its top-level content in the `<a...>`spine`</a>` of the `<a...>`Package Document`</a>` [[EPUB-33]]. Each also defines a declarative table of contents  in the `<a...>`Package Document`</a>` [[EPUB-33]].

epub33/overview/index.html
```diff
@@ -163,14 +210,20 @@ <h3>Reading Order</h3>
 					<h3>Navigation Document</h3>
 
 					<p>Each EPUB Publication contains a special XHTML Content Document called the <a>EPUB Navigation
-							Document</a>, which uses the [[HTML]] <code>nav</code> element to define human- and
-						machine-readable navigation information.</p>
+							Document</a>, which uses the [[HTML]] <a data-cite="html#the-nav-element"><code>nav</code></a>
+						element to define human- and machine-readable navigation information. Most Reading Systems make use of the
```
mattgarrish: /Most/All/ ?

Are there any that don't? It's a requirement to make the toc available per the spec.

epub33/overview/index.html
```diff
@@ -163,14 +210,20 @@ <h3>Reading Order</h3>
...
+					<p>The Navigation Document contains baseline accessibility and navigation support, and features
 						to enhance navigation for all users. Prime among these are support for internationalization (for
-						example, as an XHTML document itself, the Navigation Document natively supports ruby
-						annotations) and support for embedded grammars (MathML and SVG can be included within navigation
-						links).</p>
+						example, as an XHTML document itself, the Navigation Document natively supports 
+						<a data-cite="html#the-ruby-element">ruby annotations</a>) and support for embedded grammars
+						(MathML and SVG can be included within navigation links).</p>
+
+					<p>Note, however, that EPUB Reading Systems are not <em>required</em> to use these advanced XHTML features, such as
```
mattgarrish: Minor nit, but I'd remove "however" from this sentence as the next one also starts with "However,".

epub33/overview/index.html
```diff
@@ -237,15 +290,19 @@ <h2>Fixed Layouts</h2>
 						spreads</a> [[EPUB-33]], and <a href="https://www.w3.org/TR/epub-33/#page-spread">how to
 						position pages</a> [[EPUB-33]] within those spreads, providing a broad range of control over the
 					presentation of EPUB Publications.</p>
-			</section>
+
+				<p>The structure and semantics of EPUB Documents with Fixed Layouts are defined in the dedicated <a
+						href="https://www.w3.org/TR/epub-33/#sec-fixed-layouts">section</a> of [[EPUB-33]].</p>
```
mattgarrish: Accessibility nitpick, but linking the word "section" is ambiguous. This could be:

> The `<a...>`structure and semantics of Fixed-Layout Documents`</a>` are defined in [[EPUB-33]].

epub33/overview/index.html
```diff
@@ -267,16 +324,21 @@ <h2>Rendering and CSS</h2>
...
 					types</a> [[EPUB-33]].</p>
 
-				<p>Another key multimedia feature in EPUB 3 is the inclusion of <a>Media Overlay Documents</a>
+				<p id="para-media-overlay">Another key multimedia feature in EPUB 3 is the inclusion of <a>Media Overlay Documents</a>
 					[[EPUB-33]]. When pre-recorded narration is available for an EPUB Publication, Media Overlays
 					provide the ability to synchronize that audio with the text of a Content Document (see also <a
 						href="#sec-access-overlays"></a>).</p>
-			</section>
+
+				<p>The structure and semantics of EPUB Documents with Media Overlays are defined in the dedicated <a
```
mattgarrish: Same suggestion about the linking as above.

epub33/overview/index.html
```diff
@@ -364,11 +431,18 @@ <h2>Container</h2>
...
 
 			</section>
-
 		</section>
+
 		<section id="sec-gls">
 			<h1>Global Language Support</h1>
 
+			<p class=ednote>
+				I wonder about this section overall. The only thing that is EPUB Specific is the section of metadata. Wouldn't it be better to
```
mattgarrish: You might want to open a proper issue to consider this. I don't have a strong opinion on the GLS stuff.

epub33/overview/index.html
```diff
@@ -477,10 +555,10 @@ <h1>Accessibility</h1>
 				this requirement. This section reviews these features, detailing some established best practices for
 				ensuring that EPUB Publications are accessible where applicable.</p>
 
-			<p>EPUB 3 also includes an Accessibility specification [[EPUB-A11Y-10]] that leverages the extensive work
-				done to make Web content accessible in [[WCAG20]]. The specification defines requirements to produce
+			<p>EPUB 3.3 also includes an Accessibility specification [[EPUB-A11Y-11]] that leverages the extensive work
```
mattgarrish: The accessibility specification isn't directly tied to 3.3, but that's what it sounds like here. I would leave off the minor revision number, as in the original, or see if there's some other way to phrase that it applies to EPUB 3.3 without making it appear to be a product of this revision.

epub33/overview/index.html
iherman: I am not sure the 'unique approach' is the right wording there, it somehow smells some sort of an exclusivity. But I agree that the word extension isn't good either. What about

> The Package Document defines a layer on top of the traditional structuring of a typical Web site to facilitate the authoring of digital publications

epub33/overview/index.html
iherman: Agree. I made the change on my local copy, will be part of the next commit.

epub33/overview/index.html
iherman: Yep.

epub33/overview/index.html
iherman: I have removed the minor revision number. I am not sure there is any change that can be done in the section...

epub33/overview/index.html
iherman: Ok. I have opened #1809, and removed the editor's note from the current version.

epub33/overview/index.html
mattgarrish: works for me.

mattgarrish (approved):

iherman:
@dauwhe @wareid @shiestyle  is it o.k. to merge this?

shiestyle (approved):
OK, we should remember to restore EPUB-tts entry in biblio.js.

PR #1820, opened 2021-09-23
Add section on internationalization and techniques

### LINKED ISSUE #1790: Informative note for Unicode in EPUB accessibility
Using Unicode text is many times considered as implicit requirement. Furthermore it does not effect English speaking countries.
But the fact is that you can find non-Unicode web content in non-English speaking parts of the world. The problem is even more in E-Publications because you can embed fonts in the publications.
So, we should at least have an informative text in EPUB Accessibility specifications which remind the publishers to use Unicode text.

mattgarrish:
As discussed on the call today, here's what I'm thinking in terms of addressing internationalization.

The PR adds a basic subsection to the existing success techniques section that explains that there will be differences in the best practices necessary to make content accessible in any given language, but that it's an expected part of how the guidelines are crafted.

I don't think we need to go deeper than this for the accessibility specification, but it opens the door to adding more specific details to the techniques.

I've worked in mention that Unicode is a requirement of EPUB, @avneeshsingh, but not exactly in the way you may have imagined. I can always look at rewriting it if you think the point you're trying to get across might not be fully captured.

Fixes #1790 

- Accessibility [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-1790/epub33/a11y/index.html)
- Accessibility [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-1790/epub33/a11y/index.html)

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -64,7 +64,8 @@
 			}
 			.varlist {
 				margin-left: 3rem;
-			}</style>
+			}
+		</style>
 	</head>
 	<body>
 		<section id="abstract">
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -141,6 +141,25 @@
 							layout publications</a>. When that document is formally published, it should be mentioned in
 						this section.</p>
 				</div>
+
+				<section id="sec-sc-i18n">
+					<h4>Internationalization</h4>
+
+					<p>This specification is also designed to address the accessibility needs of users independent of
+						what languages they read. The same is true for the principles and success criteria defined in
+						[[WCAG2]]. The goal is to ensure that users can fully consume the information of a publication
+						regardless of their preferred reading modality.</p>
+
+					<p>At the same time, the language and writing conventions of the authored text will influence the
+						techniques necessary to meet the accessibility requirements. The fact that EPUB 3 requires
+						support for Unicode text, for example, is not enough on its own to ensure that the text in any
+						given language is fully accessible even though it ensures the correct character data can be
+						used.</p>
+
+					<p>That there will be language-specific best practices for meeting accessibility requirements is
+						consequently an expected outcome, even if these best practices are generally outside the scope
+						of this specification and its techniques (except where specific to EPUB).</p>
+				</section>
 			</section>
 
 			<section id="sec-application" class="informative">
@@ -190,6 +209,7 @@
 						playback.</p>
 				</div>
 			</section>
+
 			<section id="conformance"></section>
 		</section>
 		<section id="sec-discovery">
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -64,8 +64,7 @@
 			}
 			.varlist {
 				margin-left: 3rem;
-			}
-		</style>
+			}</style>
 	</head>
 	<body>
 		<section id="abstract">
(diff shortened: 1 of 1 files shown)


GeorgeKerscher:
This looks fine. I was expecting to see mention of UTF 8, but if you don't think that is needed, I am fine with it.

clapierre (approved):

mattgarrish:
> I was expecting to see mention of UTF 8, but if you don't think that is needed, I am fine with it.

@GeorgeKerscher I figured it would be better just to say Unicode, since you can also use UTF-16 even if it's no longer recommended. If we want to be more specific, though, that's also fine with me.

murata2makoto:
I am not quite happy with the last paragraph.  First, it is not clear whether best practices are good enough for capturing requirements specific to some languages and cultures.  Second, we simply do not know whether or not additions to 
this specification or the techniques are required for capturing such requirements.

iherman:
> I am not quite happy with the last paragraph. First, it is not clear whether best practices are good enough for capturing requirements specific to some languages and cultures.  Second, we simply do not know whether or not additions to this specification or the techniques are required for capturing such requirements.

The way I read the text is that the best practices _may not_ be enough to capture specific languages, and that to have them solved may go outside this specification. I.e., I do think that the message you are looking for is there.

But I do agree that the sentence may be a bit difficult to parse for non-native English speakers. How about:

> As a consequence, there may be language- or culture-specific practices for meeting accessibility requirements. These are generally outside the scope of this specification and its techniques (except where specific to EPUB).

Note that I have added 'culture-specific' which may be important. Language itself may not be enough (see the cultural differences between France and Quebec:-).

I actually wonder whether the last paragraph should not be merged with the previous one.

avneeshsingh:
Thanks Matt. It captures the core of our discussion. I am suggesting rephrasing the Unicode specific part, and also incorporating Ivan's improvement. For me also parsing the last sentence took couple of passes. You can adapt it according to your writing style.
EPUB requires support for Unicode text which ensures that correct character data can be used. However it may not be enough on its own to ensure that the text in any given language is fully accessible.
As a consequence, there may be language- or culture-specific practices for meeting accessibility requirements. These are generally outside the scope of this specification and its techniques (except where specific to EPUB).

murata2makoto:
@iherman 

>The way I read the text is that the best practices may not be enough to capture specific languages, and that to have them solved may go outside this specification. I.e., I do think that the message you are looking for is there.

Best practices may be enough or may not be enough.  The A11Y spec may be enhanced or may not be enhanced.  The techniques may be enhanced or may not be enhanced.;  Every possibility is still open especially when people have not seriously studied I18N issues for accessibility.  I do not think that the proposed wording is neutral.

mattgarrish:
We're not going to be defining best practices for problems with web content in general. There are going to be practices developed to meet WCAG, but whether or not those are problems we solve in this group depends on whether they are unique to EPUB. That's all the last paragraph is trying to capture.

The majority of issues that we encounter are web-general which is why I said these are mostly outside our scope. We don't want to set ourselves up as an alternative to WCAG for problems that should go through that group.

That's why we have to leave flexibility both between what we're doing and what others are doing. We can't make a definitive statement about what practices will work or who will develop or standardize them when we don't know what they are.

Is it the use of "even if" that you find problematic, though? If it helps we could rewrite that sentence (incorporating @iherman's suggestions) to:

> As a consequence, there may be language- or culture-specific practices for meeting accessibility requirements. Whether these practices are defined within this specification and its techniques or elsewhere (e.g., in WCAG techniques or language-specific best practice recommendations) will depend on whether the issues are specific to EPUB or broadly affect all Web content.

murata2makoto:
The revised wording is fine to me.

mattgarrish:
> For me also parsing the last sentence took couple of passes.

@avneeshsingh what if we expand on your proposal to add a little extra context so it becomes:

> EPUB [requires support for Unicode text](https://www.w3.org/TR/epub-33/#confreq-xml-enc) [EPUB-3], for example, which ensures the correct character data can be used (i.e., EPUB Creators do not have to use images of text). Although this is an important feature, it is often not enough on its own to ensure that the text is fully accessible in any given language (e.g., additional information about directionality, emphasis, pronunciation, etc. may also be neeed).

avneeshsingh:
Amendment looks good Matt.

PR #1824, opened 2021-09-24
Update list of optimized publications and faq

### LINKED ISSUE #1817: EPUB Accessibility FAQ
The FAQ is referenced in a note at the end of section 1.1, but is still an IDPF document: https://idpf.github.io/epub-guides/a11y-faq/

It's also needs revision, at least with respect to the answer about using URLs for the conformance identifier.

Looks like we may need to port this over to a note.

### LINKED ISSUE #1819: References to IDPF optimization standards registry
Going through the FAQ entries, I noticed there's a section devoted to explaining optimized publications and how to register standards. That made me go back and check the accessibility specification and sure enough there's a link to the same registry in a note in the section on optimizations.

Any thoughts on what we do with this?

Could we move it under DAISY since it's tangential to the actual purpose of the specification? (i.e., it is just a list of potential other standards a publication might conform to in place of this one.)

mattgarrish:
As discussed on the telecon yesterday, I've moved the list of optimization standards to a wiki page. This PR updates the specification to link to the new page and drops "registry" from the name since that's not technically what we have.

It also adds a new document for the FAQ.

Fixes #1819 
Fixes #1817 

- Accessibility [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-1819/epub33/a11y/index.html)
- Accessibility [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-1819/epub33/a11y/index.html)
- FAQ [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-1819/docs/a11y-faq/index.html)


iherman:
I must admit, I am not a fan of using this wiki page. It is a bit disturbing: the repo wiki has a bunch of other pages, and they are listed on the right hand side menu. Just by getting to this page one wonders whether those listed in the right column are part of this document or not (they aren't).

I think an alternative is to move the same content to a separate github markdown page. It would look essentially the same, it is almost as easy to maintain, and there is no ambiguity.

mattgarrish:
Hm, looks like I forgot to switch branches for the faq, but then these are tied together so I'm going to leave both in this one PR.

mattgarrish:
> I must admit, I am not a fan of using this wiki page.

It should look pretty informal given that it's not central to the specification or the work of the group. There was concern about this document looking too much like a registry when it's just intended as an informative list.

The whole concept of optimized publications only exists to avoid the problem of not conforming to the spec leading people to think the content isn't accessible to anyone.

clapierre (approved):
This is great Matt

iherman:
> It should look pretty informal given that it's not central to the specification or the work of the group. There was concern about this document looking too much like a registry when it's just intended as an informative list.

I understand that. My only worry is to use the Wiki for this rather the same content in a markdown file that can also be pointed to via github.io.

But I won't lie down on the road over this :-)

mattgarrish:
> My only worry is to use the Wiki for this rather the same content in a markdown file that can also be pointed to via github.io.

The wiki worries me less than why the whole discussion of optimized publications isn't informative...

Should we define rules for how to identify conformance to specifications that we don't control?

I look at this section and strongly wonder why our guidance is normative, as all we originally set out to do in this section is say that just because you don't claim conformance to our standard doesn't mean the publication is inaccessible. We can informatively note how you would use the same element for other standards, but what is the end goal of having this section be normative?

If we make the section informative, I don't see why we need to maintain a list. The only entry in that document is the same standard we use in the example, so it's not like we're losing anything.

To go back to the discussions yesterday, what other standards you might conform to could well just be a topic for best practice sites like the DAISY KB.

avneeshsingh:
The FAQ page looks great!
For list of optimized specs, I am also feeling like @iherman. With screen reader also wiki looks too informal. 
I completely understand that it should not look like W3C registry. To address this confusion should we state in the top that this document is informative it is not a WG note, W3C registry or a CG report, and place it as a mark down file in repo.

mattgarrish:
I've moved the guide to a [markdown file](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-1819/docs/optimized-pubs/index.md).

But I'm still uncomfortable with the section being normative. Why are we requiring a URL, for example, when we don't use them ourselves? We can keep all the guidance on what you should be doing, but having it normative is problematic.

mattgarrish:
Put the wrong extension in the last post: https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-1819/docs/optimized-pubs/index.md

The only difference from the wiki is an extra sentence in the overview disclaiming that it is not actually a registry. (You'll only get the raw markdown until we merge.)

GeorgeKerscher:
Hi,

Totally agree that this should not be normative in the specification.

It says:

This page provides information about standards for creating Optimized EPUB Publications, which are publications that have been optimized for a specific reading modality.

Perhaps replace the second optimized  with specially designed

mattgarrish:
I'm going to merge this pull request and handle making the optimized publications section informative in a separate one.

PR #1821, opened 2021-09-23
Adding directionality test references to RS

iherman:
This is to be checked in case https://github.com/w3c/epub-tests/pull/31 is approved (and merged): it adds the references to the tests in that PR to the RS spec.

PR #1826, opened 2021-09-25
Change respec test references to use generated report

dlazin:
Will this actually work in Respec? I don't know! (Because I don't know how to check.) But let's find out.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1826.html" title="Last updated on Sep 28, 2021, 5:58 PM UTC (1959c5d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1826/9c8c5db...1959c5d.html" title="Last updated on Sep 28, 2021, 5:58 PM UTC (1959c5d)">Diff</a>

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -69,7 +69,7 @@
 				localBiblio: biblio,
 				preProcess:[inlineCustomCSS, fixDefinitionCrossrefs],
 				postProcess:[addConformanceLinks],
-				testSuiteURI: "https://w3c.github.io/epub-tests/results/tests.html#",
+				testSuiteURI: "https://w3c.github.io/epub-tests/results/tests.html",
 				lint: {
 					"wpt-tests-exist": true,
 				}
@@ -200,11 +200,12 @@
 				<h4>Core Media Types</h4>
 
 				<p id="confreq-rs-epub3-images"
-					data-tests="cmt-gif,cmt-jpg,cmt-png,cmt-svg,cmt-webp">If a Reading System has a <a>Viewport</a>, it MUST
-					support the <a href="https://www.w3.org/TR/epub-33/#cmt-grp-image">image Core Media Type Resources</a>
+					data-tests="#cmt-gif,#cmt-jpg,#cmt-png,#cmt-svg,#cmt-webp">If a Reading System has a <a>Viewport</a>, it
+					MUST support the
+					<a href="https://www.w3.org/TR/epub-33/#cmt-grp-image">image Core Media Type Resources</a>
 				    [[EPUB-33]].</p>
 
-				<p id="confreq-rs-epub3-mp3-aac" data-tests="cmt-mp3,cmt-mp4-aac,cmt-ogg-opus">If it has the capability to
+				<p id="confreq-rs-epub3-mp3-aac" data-tests="#cmt-mp3,#cmt-mp4-aac,#cmt-ogg-opus">If it has the capability to
 					render pre-recorded audio, it MUST support the <a
 						href="https://www.w3.org/TR/epub-33/#cmt-grp-audio">audio Core Media Type Resources</a>
 					[[EPUB-33]] and SHOULD support Media Overlays [[EPUB-33]].</p>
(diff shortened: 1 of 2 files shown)


iherman:
It does not seem to work :-(

Here is what I did: I used raw.githack.com to look at the respec version you have created:

https://raw.githack.com/w3c/epub-specs/test-repo/epub33/rs/index.html

If you then go to the [base direction tests](https://raw.githack.com/w3c/epub-specs/test-repo/epub33/rs/index.html#sec-pkg-doc-base-dir), and you click on a test, you get:

https://w3c.github.io/epub-tests/results/confreq-rs-pkg-dir_rtl-001

which is of course wrong. The mechanism removes the 'index.html#' from the URL.

I see several solutions:

1. we keep to what we have officially now 
2. I try to hack this; respec gives the possibility to run a javascript _after_ it has done its processing, which may be used to hack all these URL-s. I do not know whether it is easy to get some proper hooks for those URL-s that I can hang a script on but I am willing to try
3. we come up with a totally different mechanism altogether

what say you? 

Cc @mattgarrish @wareid 

---

P.S. The reason you could not see the problem is because the preview up there shows the document _after_ respec has done its work, which is the equivalent of what lands as an official document. But respec, deliberately, cuts the test references out of that version. In other words, the test references are live only in the editor's draft, which of course makes sense.

dlazin:
> Here is what I did: I used raw.githack.com to look at the respec version you have created:
> 
> https://raw.githack.com/w3c/epub-specs/test-repo/epub33/rs/index.html

Ah! Thanks for that tip. Check again; now it works (I have moved the hash from the base URL into the data-tests attributes), except for one notable loss: with anchors, it no longer shows the names of the tests, only "test". But it does link to the same place, and I think this is much more useful, because now you can see the test description.

epub33/rs/index.html
```diff
@@ -303,15 +304,19 @@ <h4>Parsing Relative URLs</h4>
...
-					and indicates a base direction of <code>ltr</code> or <code>rtl</code>, Reading Systems MUST
-					override the bidi algorithm per <a data-cite="bidi#Higher-Level_Protocols">the higher-level
-						protocols</a> defined in [[BIDI]], setting the paragraph embedding level to 0 if the base
-					direction is <code>ltr</code>, or 1 if the base direction is <code>rtl</code>.</p>
-
-				<p id="confreq-rs-pkg-dir-auto" data-tests="confreq-rs-pkg-dir-auto.epub">Otherwise the base direction
-					is <code>auto</code>, in which case Reading Systems MUST determine the text's direction by applying
-					the Unicode Bidi Algorithm, beginning with <a data-cite="bidi#P2">Rule P2</a> of [[BIDI]].</p>
+				<p id="confreq-rs-pkg-dir"
+				    data-tests="#confreq-rs-pkg-dir_rtl-001,#confreq-rs-pkg-dir_rtl-004,#confreq-rs-pkg-dir_rtl-005,#confreq-rs-pkg-dir_rtl-006">
```
dlazin: @iherman here and below on line 316, have I corrected interpreted which of your dir tests corresponds to which statement? (Please remember to update the spec with these references after you commit tests.)

iherman:
> Check again; now it works (I have moved the hash from the base URL into the data-tests attributes), except for one notable loss: with anchors, it no longer shows the names of the tests, only "test".

Neat trick and it indeed works. It is probably pretty error-prone: people may forget those '#' signs. But I guess it is not worse than adding (or not) the '.epub' in the old scheme...

epub33/rs/index.html
iherman: Ehem. I have submitted https://github.com/w3c/epub-specs/pull/1821 which waits for your approval :-)

dlazin: Oh no! I'm the worst. Sorry I missed that — epub-specs is so busy that I almost never check. I will need to figure out a better email filtering system so that I actually see the ones meant for me.

dlazin:
Is this OK to commit now?

PR #1829, opened 2021-09-27
Reformat examples to avoid scrolling

### LINKED ISSUE #1822: A11y 1.1 - Scrolling in 2 dimensions caused by long metadata
Zooming into the A11y 1.1 recommendation causes a horizontal scrollbar that persists at the same time as the vertical scrollbar. This issue will affect general users in small desktops and tablet devices, but more importantly people with low vision using browser zoom and screen magnifiers will experience difficulty navigating the content. Additionally, people modifying text properties may experience the same issue.

### Why is there a horizontal scrollbar?
The metadata code snippet contains lots of joining characters, which are conveyed to the user agent that this is one word. I appreciate this is only the symptom to the underlying problem that `<code>` tags aren't being styled with `white-space: break-spaces`.

### Expected results

- (Preferred) The `<code>` tag is styled to break spaces or break words, OR;
- The metadata is written in a way that takes little horizontal room, with each attribute on a new line. This might have readability improvements in general:
```
<meta 
  property="a11y:certifiedBy" 
  id="certifier">
    Accessibility Testers Group
</meta>
<link 
  rel="a11y:certifierReport" 
  refines="#certifier" 
  href="http://example.com/a11y/reports/9780000000001" />
```

mattgarrish:
Fixes #1822 

- Accessibility [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-1822/epub33/a11y/index.html)
- Accessibility [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-1822/epub33/a11y/index.html)

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -55,6 +55,11 @@
 				localBiblio: biblio,
 				preProcess:[addConformanceLinks]
 			};</script>
+		<style>
+			pre,
+			code {
+				white-space: break-spaces !important;
+			}</style>
 	</head>
 	<body>
 		<section id="abstract">
@@ -177,8 +182,14 @@
 				<aside class="example">
 					<p>The following example shows the metadata entries for an EPUB 3 Publication that has textual and
 						visual access modes.</p>
-					<pre>&lt;meta property="schema:accessMode"&gt;textual&lt;/meta&gt;
-&lt;meta property="schema:accessMode"&gt;visual&lt;/meta&gt;</pre>
+					<pre>&lt;meta
+    property="schema:accessMode">
+   textual
+&lt;/meta>
+&lt;meta
+    property="schema:accessMode">
+   visual
+&lt;/meta></pre>
 				</aside>
 
 				<p>Note that the access modes of the content do not reflect any adaptations that have been provided. For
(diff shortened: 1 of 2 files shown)


clapierre:
@mattgarrish 
Reformatting of scrolling looks great.
I did notice a few &lt; and &gt; that looked like they were missed?
Also wondering why the I18N info was removed?

mattgarrish:
> I did notice a few < and > that looked like they were missed?

Do you mean the tags are missing? Could you point me to a specific case?

> Also wondering why the I18N info was removed?

This branch was created last week before I integrated the i18n section. Merging it won't remove the changes that have already been integrated. (I could force a merge from main on the branch to get the latest commits to appear, but it's extra work for no appreciable gain.)

clapierre:
```
						<pre>&lt;html …&gt;
   &lt;head&gt;
      &lt;title&gt;
						<pre>&lt;html …>
   &lt;head>
      &lt;title>
         III. The Night Shadows —
         Book the First—Recalled to Life — 
         A Tale of Two Cities
      &lt;/title&gt;
      &lt;/title>

So some cases &gt; became > but not all same for &lt; became <

```

mattgarrish:
> So some cases &gt; became > but not all same for &lt; became <

Oh, yes, I hate trying to read the source code examples when the closing brackets have been converted to greater than entities. It makes it headache to decipher the tags when you have stuff like `&lt;/a&gt;`, so I switched all those back to the actual '>' character.

You can't do the same with opening brackets as then they denote new tags within the pre element. I have to live with the lt entities for those.

PR #1833, opened 2021-09-28
Change the section on optimized publications to informative

mattgarrish:
As originally mentioned in https://github.com/w3c/epub-specs/pull/1824#issuecomment-926744548, this pull request changes the section on optimized publications to informative.

I've rewritten the latter part of the section to focus on explaining how this specification can be a model for identifying optimizations so that we avoid making requirements on guidelines we don't control. More specifically, I've grouped all the recommendations together into a single list and updated the prose so that we're not requiring specific practices (i.e., that URLs must be used when we no longer require them ourselves).

The recommendations remain largely the same, so other than making the section informative the guidance is largely unchanged. We're effectively just leaving some room for specific guidelines to make these recommendations requirements themselves.

- Accessibility [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/informative-opt-pubs/epub33/a11y/index.html)
- Accessibility [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/informative-opt-pubs/epub33/a11y/index.html)

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -1283,7 +1283,7 @@
 				</section>
 			</section>
 		</section>
-		<section id="sec-optimized-pubs">
+		<section id="sec-optimized-pubs" class="informative">
 			<h2>Optimized Publications</h2>
 
 			<p>Although WCAG [[WCAG2]] provides a general set of guidelines for making content broadly accessible,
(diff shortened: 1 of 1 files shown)


clapierre (approved):

GeorgeKerscher:
I think this section looks great.

PR #1834, opened 2021-09-29
Add explainer about applicability of SC 2.4.6

### LINKED ISSUE #1810: Add technique explaining WCAG 2.4.6 passes for all publication headings
2.4.6 is the AA success criterion that says "Headings and labels describe topic or purpose".

While this makes sense for labels, and for books with topic-based chapters, how a novel chapter heading describes its topic or purpose isn't immediately clear in a lot of cases. It's questionable if a chapter number alone describes either a topic or purpose (short of an argument that the Xth chapter's purpose is to be in position X).

I finally got around to opening an issue about this in the [WCAG tracker](https://github.com/w3c/wcag3/issues/170), and it seems the intent of the success criterion doesn't exactly match its wording, but it's also complicated to change the SC text now.

The general response I've gotten, though, is that the presence of a heading establishes the context of its chapter (if not a specific topic or purpose) and that's enough to satisfy the SC, so it's never actually an issue for publications.

Should we document this in the techniques document, though, so we have it as a reference?

mattgarrish:
This PR adds a new section to the titles and headings section of the techniques document to explain that the wording of SC 2.4.6 does not require the heading to describe a topic, only that it be unique to identify the section.

Fixes #1810 

- Accessibility [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-1810/epub33/a11y-tech/index.html)
- Accessibility [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-1810/epub33/a11y-tech/index.html)

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -1207,8 +1207,8 @@
 						currently states that all headings must describe their topic or purpose. The implication of this
 						wording is that all chapters in a novel, for example, have a topic or purpose and that that
 						topic or purpose is always clearly reflected by the title of the chapter. Not only is this not
-						the case, but this success criterion also complicates the use of chapter numbers as headings
-						since these do not establish a topic.</p>
+						always the case, but this success criterion also complicates the use of chapter numbers as
+						headings since these do not establish a topic.</p>
 
 					<p>After discussion in the <a href="https://github.com/w3c/wcag/issues/2039">Accessibility
 							Guidelines Working Group's issue tracker</a>, it is clear that the <a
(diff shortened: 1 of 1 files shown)


mattgarrish:
Somehow this PR opened before I had a chance to edit the title or description, but I've corrected everything in the first comment. The preview and diff links should both work now.

clapierre (approved):

avneeshsingh:
Looks good Matt. You can fix double "that":
"and that that topic or purpose"

PR #1836, opened 2021-10-04
Remove registered trademark symbols

mattgarrish:
Fixes #1835

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1836.html" title="Last updated on Oct 4, 2021, 11:04 AM UTC (d2ee249)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1836/f366b81...d2ee249.html" title="Last updated on Oct 4, 2021, 11:04 AM UTC (d2ee249)">Diff</a>

PR #1840, opened 2021-10-07
Added the ® sign to the first appearance of 'EPUB'

### LINKED ISSUE #1835: Removing registered trademark symbol
IDPF legally required that the ® symbol appear after the first instance of "EPUB" in a specification that wasn't part of a specification name.

The symbol was removed from the overview document, so does that mean it's no longer a requirement and can be removed from the other specifications, or are we still required to keep it per IDPF practice and it should be restored?

iherman:
This is an alternative to #1836 to fix #1835: re-adding the ® sign to the Overview document's abstract.

### DIFF
--- a/epub33/overview/index.html
+++ b/epub33/overview/index.html
@@ -82,7 +82,7 @@
 	<body>
 		<section id="abstract">
 			<p>This document provides a starting point for content authors and software developers wishing to understand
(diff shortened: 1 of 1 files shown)


mattgarrish (approved):

mattgarrish:
This is probably the better approach for legacy reasons. I'll close my other PR.

PR #1847, opened 2021-10-12
Added an editors' draft for the separate Structural Semantics Vocabulary WG Note

### LINKED ISSUE #1763: Should all Structural Vocabulary Items be normative?
Discussion started on the [mailing list thread](https://lists.w3.org/Archives/Public/public-epub-wg/2021Jul/0040.html). The relevant part of the discussion (quoting from [my original mail](https://lists.w3.org/Archives/Public/public-epub-wg/2021Jul/0052.html)):

>
> We generally have an issue at W3C on how we would define testing and implementation for vocabulary terms […]. The generally accepted approach is that for vocabulary items the term "implementation" is a a misnomer and it should be considered to be an alias to the term "usage" for our CR process. Indeed, there is no RS behavioral requirement for any of the SSV entries, so traditional requirements on implementations would not apply. Instead we may, for example, define "usage" of a specific vocabulary term, for our CR process, that there are at least two publishers out there who use these terms in production (hoping that at least some reading systems do something sensible with it). Such, or similar, ways for "testing" (per W3C process) for vocabularies was used in other specifications in the past.
>
> However: would all SSV terms pass such a requirement? All the answers on John's mail suggest that the answer would be no […], because many terms have been introduced to the spec as part of an aspiration for something. Ie, we may be creating a problem for ourselves. Shouldn't we mark the SSV vocabulary as non-normative overall with, possibly, explicitly mark a few terms as normative because we know they are accepted by the community (e.g., landmarks)? Note that if we keep all the terms as normative and then we fail on the CR tests, the usual expectation would be to remove them altogether, which we do not want (I presume).
>

iherman:
Following [WG Resolution](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-10-08-epub#resolution1).

Few practical notes:

- In order to make a reference to the new document, and avoid invalid links that might create problems in the publication process, I have
    - Added a temporary entry to `biblio.js`
    - Used, for referenced values, the `github.io` URL for the document

    Both of these should be settled once the note is published, ie, the proper reference becomes part of specref.
- I have used the version number 1.1 to differentiate from the IDPF document. One practical reason is that the EPUB-SSV is part of specref referring to the IDPF document and, for historical reason, I do not think we should modify this. Also, we may want to change something on the vocabulary, so it is better to keep this one separate.

The readable version of the new note is [here](https://raw.githack.com/w3c/epub-specs/ssv-note/epub33/ssv/index.html).

Fix #1763

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1847.html" title="Last updated on Oct 13, 2021, 9:46 AM UTC (b7b3d8d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1847/d6d6d3b...b7b3d8d.html" title="Last updated on Oct 13, 2021, 9:46 AM UTC (b7b3d8d)">Diff</a>


iherman:
@mattgarrish an editorial note: any change on biblo.js triggers a new publishing round on all documents, which is a bit of a drag. The question is: do we really need biblio.js? Afaik, only the Overview document uses extra bibliographic entries for the history part (and some of those may actually be in specref), and otherwise it may be unnecessary. Just to reduce the load...

WDYT?

mattgarrish:
> The question is: do we really need biblio.js?

It's used by more than the overview. I did a quick test and stripping it from the core spec resulted in 14 broken references. I know the accessibility specs will also break without it, and almost certainly the rs spec if the core breaks.

Maybe we should make per-specification biblio files so we know what's being used where? It would also avoid triggering every spec on a change. It's not like we add to it very often anymore, and additions now would most likely be restricted to one document.

iherman:
> Maybe we should make per-specification biblio files so we know what's being used where? It would also avoid triggering every spec on a change. It's not like we add to it very often anymore, and additions now would most likely be restricted to one document.

I would be happy with that, too.

iherman:
Actually, in some cases, 1-2 entries would suffice, which does not even warrant a separate file.

I will take a stab at the Overview for a start and see where it goes. I will also remove it right away from this ssv draft.

wareid (approved):

iherman:
I have carried over the effects of #1848: no reference to the common biblio file from the ssv note, and picked up the new version of the `core/index.html` file that relies on a local biblio file.

iherman:
The readable version of the new note is [here](https://raw.githack.com/w3c/epub-specs/ssv-note/epub33/ssv/index.html).

mattgarrish (approved):

mattgarrish:
The only weird thing is giving it a version number, as it was intended as a living document. That said, I don't think there's any way to go back through all the old iterations, so it's not really a blocker to me.

dauwhe (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-10-14-epub#section1) on 2021-10-14

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-10-14-epub#resolution1):  Publish the SSV as a working group note, using ECHIDNA, using the short name "epub-ssv"***

### 1. Publication of Structural Semantics Vocabulary as Working Group Note 

_See github pull request [#1847](https://github.com/w3c/epub-specs/pull/1847)._




_See github issue [#1763](https://github.com/w3c/epub-specs/issues/1763)._




**Dave Cramer:** last week we resolved on separating SSV into a wg note  
… we just need a formal resolution to publish it, and to use echidna  

> **Proposed resolution: Publish the SSV as a working group note, using ECHIDNA, using the short name "epub-ssv"** *(Wendy Reid)*

**Murata Makoto:** the link in Ivan's message doesn't work, so I can't see the note  

> *Dave Cramer:* [https://github.com/w3c/epub-specs/blob/main/epub33/ssv/index.html](https://github.com/w3c/epub-specs/blob/main/epub33/ssv/index.html)

> *Wendy Reid:* See [The editor's draft for the note](https://w3c.github.io/epub-specs/epub33/ssv/)

> *Dave Cramer:* +1

> *Wendy Reid:* +1

> *Matthew Chan:* +1

> *Matt Garrish:* +1

> *Toshiaki Koike:* +1

> *Masakazu Kitahara:* +1

> *Shinya Takami (高見真也):* +1

**Murata Makoto:** does this document have to do with the IMS Global Consortium? Do they need to see this?  

**Matt Garrish:** no, i think this originally came from DAISY  

**Murata Makoto:** what about the content under [sec. 10](https://w3c.github.io/epub-specs/epub33/ssv/#educational)?  

**Matt Garrish:** this is all stuff that came from us, we don't have dependence on any other party  

> ***Resolution #1: Publish the SSV as a working group note, using ECHIDNA, using the short name "epub-ssv"***

> *Dave Cramer:* 

</details>

PR #1848, opened 2021-10-12
Removed dependency on common biblio file

iherman:
As discussed on https://github.com/w3c/epub-specs/pull/1847#issuecomment-940991574: removed the dependency on the common biblio file for overview, core, rs, epubcfi, multi-redn, tts. 

In most of the cases only 1-2 entries were necessary and I added them to the respec structure directly; only overview and core warrants, in my view, the extra drag of a separate biblio file locally.

(It seems that the only IDPF document that is in specref is the ssv one; no idea how this happened. Somebody needed it... We could add the historical references but I do not think it is worth it. These are all cast in concrete...)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1848.html" title="Last updated on Oct 13, 2021, 9:31 AM UTC (caed15f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1848/20faad5...caed15f.html" title="Last updated on Oct 13, 2021, 9:31 AM UTC (caed15f)">Diff</a>

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -4,7 +4,7 @@
 		<meta charset="utf-8" />
 		<title>EPUB Accessibility Techniques 1.1</title>
 		<script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
-		<script src="../common/js/biblio.js" class="remove"></script>
+		<script src="biblio.js" class="remove"></script>
 		<script src="../common/js/permalink-a11y.js" class="remove"></script>
 		<script class="remove">
 			var respecConfig = {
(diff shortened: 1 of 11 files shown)


iherman:
@mattgarrish, I did not touch the A11y files...

If we decide to merge it before the SSV note one (ie, #1847) then I will ammend that PR to remove the dependency, too.

mattgarrish:
There were only 6-7 entries for each of the a11y documents, so I've inlined them.

I would opt to merge this now as it's pretty straightforward and reduces clutter. Haven't had a chance to look at the ssv document yet, but will try to get to it later.

I'm pretty sure I submitted SSV because of the DPUB-ARIA module, though any actual memories of doing so are long gone.

mattgarrish (approved):

PR #1856, opened 2021-10-15
Change SSV Short name from `epub-ssv-11` to `epub-ssv`

iherman:
The WG [resolution on publication](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-10-14-epub#resolution1) specified the shortname as `epub-ssv` and not as `epub-ssv-11`. This PR makes the necessary changes.

B.t.w.: if we do this (potentially yielding a small clash with the `epub-ssv` already used for the IDPF version in specref) then we may want to remove the 1.1 from the title, too. This PR preempted this and does that, too, but I can roll that back.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1856.html" title="Last updated on Oct 15, 2021, 9:58 AM UTC (a52c59a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1856/7339274...a52c59a.html" title="Last updated on Oct 15, 2021, 9:58 AM UTC (a52c59a)">Diff</a>

### DIFF
--- a/epub33/ssv/index.html
+++ b/epub33/ssv/index.html
@@ -66,7 +66,7 @@
 		</section>
 		<section id="sotd"></section>
 
-		<section id="introduction">
+		<section id="sec-introduction">
 			<h2>Introduction</h2>
 
 			<p>Structural semantics add additional meaning about the specific structural purpose an HTML (or SVG) element plays.
@@ -113,7 +113,7 @@
 			</section>
 		</section>
 			
-		<section id="partitions">
+		<section id="sec-partitions">
 			<h4>Document Partitions</h4>
 			<dl>
 				<dt id="cover">cover</dt>
@@ -175,7 +175,7 @@
 				</dd>
 			</dl>
 		</section>
-		<section id="divisions">
+		<section id="sec-divisions">
 			<h4>Document Divisions</h4>
 			<dl>
 				<dt id="volume">volume</dt>
@@ -255,7 +255,7 @@
 				</dd>
 			</dl>
 		</section>
-		<section id="sections">
+		<section id="sec-sections">
 			<h4>Document Sections and Components</h4>
 			<p>Sections and components that typically occur in the publication bodymatter.</p>
 			<dl>
@@ -455,7 +455,7 @@
 				</dd>
 			</dl>
 		</section>
-		<section id="navigation">
+		<section id="sec-navigation">
 			<h4>Document Navigation</h4>
 			<dl>
 				<dt id="toc-1">toc</dt>
@@ -546,7 +546,7 @@
 				</dd>
 			</dl>
 		</section>
-		<section id="document-references">
+		<section id="sec-document-references">
 			<h4>Document Reference Sections</h4>
 			<dl>
 				<dt id="appendix">appendix</dt>
@@ -1158,7 +1158,7 @@
 				</dl>
 			</section>
 		</section>
-		<section id="preliminary">
+		<section id="sec-preliminary">
 			<h4>Preliminary Sections and Components</h4>
 			<p>Preliminary sections and components, typically occurring in the publication frontmatter.</p>
 			<dl>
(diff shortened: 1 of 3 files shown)


epub33/ssv/index.html
```diff
@@ -13,7 +13,7 @@
 				group: "epub",
 				wgPublicList: "public-epub3",
 				specStatus: "ED",
-				shortName: "epub-ssv-1.1",
```
mattgarrish, line 16: This was the issue that @dauwhe noted on the call. You have "epub-ssv-1.1" here which won't match "epub-ssv-11" in the document references.

I don't believe we were resolving to drop the version identifier, only that "epub-ssv" is the stem for the document.

Is that your recollection, @dauwhe @wareid ?

iherman: The resolution says "using the short name “epub-ssv”"...

mattgarrish: Well, referencing a static version of the SSV is problematic, as the specifications are always supposed to be pointing at the latest version (even if it never changes again).

So, thinking through the implications of this, we shouldn't have versioned references in the specifications, just as we don't version HTML, SVG, etc. They should stay as [[EPUB-SSV]] and the specref entry needs to be fixed to use a new W3C URL that always references the latest document. If we don't do this, we're going to break any document using the current EPUB-SSV identifier.

As to whether we should be versioning the document, that I don't really care strongly about. We didn't in the past and there was no way of going back to earlier versions short of plugging the URL into something like the wayback machine, and no one ever complained.

But I assume a new version is necessary for the way TR publications work (or at least will be for any future update), so it seems safer to make this a "1.1" (even if it's technically v. 1.57 or something) and patch the epub-ssv specref entry to be a latest link entry.

mattgarrish: Or do we get epub-ssv as a default latest link if we create epub-ssv-11? If that's the case, all we'd need to do is delete the manual entry in the specref database. I'm thinking we only get a latest link URL, though, not an automatic specref entry.

iherman: > Or do we get epub-ssv as a default latest link if we create epub-ssv-11?

I do not think so. These things work in other specs because there are some special settings that are done by the systeam in TR, but this does not include IDPF documents.

I believe if we go ahead with this, then specref will use the w3c document and, if not, we can get it to do this. The question for me is whether there is any problem with this. Ie, are there documents using respec and the abbreviation 'ssv' for references that would suddenly change. DPUB-ARIA comes to mind but, actually, this might be a good thing: newer versions of DPUB-ARIA would switch to the W3C note which is what would have to happen anyway.

I guess I am arguing to go to ssv and forget about versioning, ie, that I should not have created this tempest in a tea pot... :-)

mattgarrish: I looked at my PRs to the specref database and there's actually nothing there to indicate I added it. I even tried searching on the reference and there are no commits, so it must be really old.

That said, I just tried a site search on w3.org and only the EPUB and DPUB-ARIA specs came back, so I think we'd be safe getting the manual entry deleted from specref. That would allow a new entry based on publishing this document without a version.

But if we update the document again in the future, is it okay to reissue it with the same "epub-ssv" short name or are we required to give it a unique short name each time? That's the only concern I have. If it's not a problem, let's go with "epub-ssv" as it just makes life simpler.

iherman: A WG Note can be updated as frequently as we want. It can also be turned into a rec-track document (not that I believe we would ever do it with the ssv) using the same short name. I do not think that would be a problem.

I have no idea how epub-ssv landed in specref... but if the W3C version silently overwrites the IDPF version I guess that is actually all right.

wareid:
We did the same shortname scheme as for previous documents ("epub-[name]"), and we included the version number later, I think it's ok to continue to do so for this too.

iherman:
> We did the same shortname scheme as for previous documents ("epub-[name]"), and we included the version number later, I think it's ok to continue to do so for this too.

The boss has spoken! :-) Let us keep to what we have without my own mistake.

PR #1859, opened 2021-10-18
Clarify where remote resources can be referenced

### LINKED ISSUE #1857: Clarify where remote resource references are allowed
We currently only say what resources can live outside the container, but for improved clarity (and security) we should probably also explicitly say where these references are allowed (see also https://github.com/w3c/epub-specs/issues/1061#issuecomment-943227255).

The proposal would be to:

- restrict remote audio to the `audio` + `source` elements and the MO `audio` element
- restrict remote video to the `video` + `source` elements
- restrict remote fonts to CSS `@font-face` rules, `@import` rules, and the html `link` element
- restrict remote data to scripting API calls (XHR and Fetch)

Data blocks can't be external per html ("When used to include data blocks, the data must be embedded inline") so we don't need to allow references from `script`.

mattgarrish:
Implements the proposal in https://github.com/w3c/epub-specs/issues/1857#issue-1027275195

Fixes #1857

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1859.html" title="Last updated on Oct 19, 2021, 11:58 AM UTC (2532c2f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1859/38bd812...2532c2f.html" title="Last updated on Oct 19, 2021, 11:58 AM UTC (2532c2f)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -993,7 +993,7 @@
 									href="https://html.spec.whatwg.org/multipage/media.html#the-audio-element"
 										><code>audio</code> element</a> (either from its <code>src</code> attribute or a
 								child <a href="https://html.spec.whatwg.org/multipage/media.html#the-track-element"
-										><code>track</code> element</a>) or from a Media Overlays <a
+										><code>track</code> element</a>) and from the Media Overlays <a
 									href="#sec-smil-audio-elem"><code>audio</code> element</a>.</p>
 						</li>
 						<li>
@@ -1011,7 +1011,10 @@
 						<li>
 							<p id="sec-resource-locations-fonts"><a href="#cmt-grp-font">Font resources</a> referenced
 								from <a href="https://www.w3.org/TR/css-fonts-5/#font-face-rule"><code>@font-face</code>
-									rules</a> [[CSS-Fonts-5]].</p>
+									rules</a> [[CSS-FONTS-5]], <a href="https://www.w3.org/TR/css-cascade-4/#at-import"
+										><code>@import</code> rules</a> [[CSS-CASCADE-4]], and the [[HTML]] <a
+									href="https://html.spec.whatwg.org/multipage/semantics.html#the-link-element"
+										><code>link</code> element</a>.</p>
 						</li>
 					</ul>
 
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -988,16 +988,29 @@
 
 					<ul class="conformance-list">
 						<li>
-							<p id="sec-resource-locations-audio"><a href="#cmt-grp-audio">Audio resources</a></p>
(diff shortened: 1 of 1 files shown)


dauwhe (approved):

mattgarrish:
I'm going to critique my own pull request and question if I've forgotten Media Overlays. Although hosting the synched audio outside the container doesn't strike me as a great idea, as there's no guarantee the reading system will grab the files and it could delay/interrupt playback, it has been allowed up until now.

epub33/core/index.html
```diff
@@ -988,16 +988,29 @@ <h4>Resource Locations</h4>
 
 					<ul class="conformance-list">
 						<li>
-							<p id="sec-resource-locations-audio"><a href="#cmt-grp-audio">Audio resources</a></p>
+							<p id="sec-resource-locations-audio"><a href="#cmt-grp-audio">Audio resources</a> referenced
+								from the {{HTML]] <a
```
```suggestion
							<p id="sec-resource-locations-audio"><a href="#cmt-grp-audio">Audio resources</a> referenced
								from the [[HTML]] <a
```

iherman (approved):
I agree with you that we have to be consistent v.a.v. the media overlay references. I believe we should add those to the mix.

epub33/core/index.html
mattgarrish: Yikes, time to get my eyes checked...

mattgarrish:
One other option not covered is using `@import` or the `link` element to import Google Fonts. We should allow these, too.

I think that covers everything, though.

iherman:
One issue that did come up during the call: do we have to mention `<object>`? I can imagine that there are older epub content out there that rely on it rather than using the `<video>` or the `<audio>` elements.

I do not think that we should start into allowing that, too; all this may become too much of a spaghetti code. But we may want to think about a note saying that users of `<object>` for video and/or audio should use the more recent `<video>/<audio>` elements instead.

Just a thought. I won't have sleepless nights over just leaving this unmentioned.

mattgarrish:
> I can imagine that there are older epub content out there that rely on it rather than using the `<video>` or the `<audio>` elements.

If they do, they probably have bigger problems than the object tag. To my knowledge, object was only ever used to embed flash videos, and flash video is all but dead at this point. I don't believe they even run in browsers anymore because of the security issues. It was also used to reference YouTube flash videos back when you could pull out the url to the swf file, but those days are long gone now, too.

That's why I don't think we should be concerned about allowing remote resources from object (or embed). The days of audio/video requiring plugins to work are pretty much over.

No strong opinion about a note, but I'd tend to think most authors would naturally progress old content to `audio/video` even if we say nothing.

iherman:
> No strong opinion about a note, but I'd tend to think most authors would naturally progress old content to `audio/video` even if we say nothing.

I am fine with this.

PR #1868, opened 2021-10-25
Remove generic technique identifiers

### LINKED ISSUE #1866: Remove technique numbering
I noticed in the techniques document that we now have both a section number and a short identifier for every technique. For example: 

> 5.1.1 PAGE-001: Provide page break markers

We used to suppress the list numbering in IDPF, but I'm not sure how easy that is to do in respec. Regardless, I'm wondering if the short identifiers really serve any purpose? Does anyone actually refer to them by those generic identifiers, or can we simply do away with them and use the section numbers for identification purposes?

mattgarrish:
Also worth noting is that we have permalinks for each section, so having these labels that match the IDs isn't helpful anymore in terms of linking to sections.

Fixes #1866 

- Accessibility [preview](https://cdn.statically.io/gh/w3c/epub-specs/editorial/issue-1866/epub33/a11y-tech/index.html)
- Accessibility [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/editorial/issue-1866/epub33/a11y-tech/index.html)

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -182,7 +182,7 @@
 			<h2>Discovery Metadata Techniques</h2>
 
 			<section class="suppress-numbering" id="meta-001">
-				<h3>META-001: Identify primary access modes</h3>
+				<h3>Identify primary access modes</h3>
 
 				<p>An access mode is defined as a "human sense perceptual system or cognitive faculty through which a
 					user may process or perceive the content of a digital resource." [[ISO24751-3]] For example, if an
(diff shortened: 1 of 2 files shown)


clapierre (approved):
Looks great Matt

iherman:
Looks o.k. to me. Just out of curiosity: is the class "suppress-numbering" defined by respec? Because I do not see it in the [respec doc](https://respec.org/docs/)...

mattgarrish:
> is the class "suppress-numbering" defined by respec?

No, that was what we used in IDPF. I didn't spot those on the section tags, but will remove them, too.

PR #1889, opened 2021-10-31
Added text on the behavior of RS when the same item is referenced multiple times in the spine.

### LINKED ISSUE #1686: Define what RS should do when the manifest has duplicate entries
The EPUB core spec forbids two `item` elements to identify the same resource. But the RS spec could (and should IMO) define how an RS must handle such non-conforming EPUBs.

iherman:
See issue https://github.com/w3c/epub-specs/issues/1686, in particular the text
proposed by @bduga in https://github.com/w3c/epub-specs/issues/1686#issuecomment-954839703. 

I have also made a bit more explicit in the core text that multiple references MUST NOT occur (semantically it is already in the text, but a little bit between the lines...)

See:

See:

* For EPUB 3.3:
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/multiple-item-1686/epub33/core/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/core/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/multiple-item-1686/epub33/core/index.html)
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/multiple-item-1686/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/multiple-item-1686/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1889.html" title="Last updated on Nov 9, 2021, 11:30 AM UTC (3b2d4e3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1889/7c865e5...3b2d4e3.html" title="Last updated on Nov 9, 2021, 11:30 AM UTC (3b2d4e3)">Diff</a>


dauwhe (approved):

rdeltour (commented):
Looks OK to me, with the caveat that it only solves one aspect of #1686 (see [my comment to #1686](https://github.com/w3c/epub-specs/issues/1686#issuecomment-956282869)).

epub33/rs/index.html
```diff
@@ -548,6 +548,15 @@ <h3>Spine</h3>
 				<p id="confreq-rs-pkg-spine-unknown" data-tests="#confreq-rs-pkg-spine-unknown">Reading Systems MUST
 					ignore all values expressed in spine <code>itemref</code>
 					<code>properties</code> attributes that they do not recognize.</p>
+
+				<p>
+					When presented with a single manifest item that is repeated multiple times in the linear flow of the spine,
```
bduga: Maybe replace `When presented with a single manifest item [...]` with `When presented with a single <a href="https://www.w3.org/TR/epub-33/#dfn-publication-resource">Publication Resource</a> [...]`

That should resolve the concern of @rdeltour since it won't limit this to duplicate spine references to a single item, and instead broadens the scope to resources. So it should implicitly cover two spine items with the same itemref, and two items (with different itemrefs) that resolve to the same resource.

iherman: I like that, @bduga, I have made the change.

epub33/core/index.html
```diff
@@ -3181,7 +3181,7 @@ <h6>The <code>itemref</code> Element</h6>
 							<p id="attrdef-itemref-idref">Each <code>itemref</code> element MUST reference the ID
 								[[XML]] of a unique <a href="#elemdef-package-item"><code>item</code></a> in the
 									<a>manifest</a> via the IDREF [[XML]] in its <code>idref</code> attribute (i.e., two
-								or more <code>itemref</code> elements cannot reference the same <code>item</code>).</p>
+								or more <code>itemref</code> elements MUST NOT reference the same <code>item</code>).</p>
```
mattgarrish: We shouldn't have normative requirements in parentheses. Is there a reason why we need to state the same requirement both positively and negatively, though? A lowercase "must not" seems fine here.

epub33/rs/index.html
```diff
@@ -548,6 +548,15 @@ <h3>Spine</h3>
 				<p id="confreq-rs-pkg-spine-unknown" data-tests="#confreq-rs-pkg-spine-unknown">Reading Systems MUST
 					ignore all values expressed in spine <code>itemref</code>
 					<code>properties</code> attributes that they do not recognize.</p>
+
+				<p>
+					When presented with a single <a>Publication Resource</a> that is repeated multiple times in the linear flow 
+					of the spine, Reading Systems SHOULD do their best to display that content in the correct location of that linear flow.
```
mattgarrish: "do their best" isn't a testable assertion, so perhaps this could be removed so it reads "Reading Systems MUST display content in the correction location".

Progressing linearly through the spine shouldn't present a challenge to a reading system - it's only when trying to locate the document to render when linked into - so I would make this a "must" not a "should".

epub33/rs/index.html
```diff
@@ -548,6 +548,15 @@ <h3>Spine</h3>
 				<p id="confreq-rs-pkg-spine-unknown" data-tests="#confreq-rs-pkg-spine-unknown">Reading Systems MUST
 					ignore all values expressed in spine <code>itemref</code>
 					<code>properties</code> attributes that they do not recognize.</p>
+
+				<p>
+					When presented with a single <a>Publication Resource</a> that is repeated multiple times in the linear flow 
+					of the spine, Reading Systems SHOULD do their best to display that content in the correct location of that linear flow. 
+					The Reading System SHOULD treat these as distinct pages for UI purposes (for example, each occurrence
```
mattgarrish: "distinct pages" sounds like it's referring to fixed layout pages / spreads. Maybe "items" or "instances" would be more appropriate, given that we're talking about spine items?

Isn't this also a "must" requirement, though? Why wouldn't it treat them as distinct items? The problem is that the user may not be at the document they expect, but the RS can't control that.

epub33/rs/index.html
```diff
@@ -548,6 +548,15 @@ <h3>Spine</h3>
...
+					could be independently bookmarked or annotated), but when following an internal link to that
```
mattgarrish: "internal link" is a ambiguous when talking about the reading system. Maybe make this clearer by adding the user action: "When a user activates a hyperlink to a resource that is referenced multiple times in the spine, the Reading System SHOULD ..."

mattgarrish (changes requested):

epub33/core/index.html
bduga: I'm a little confused by this - is it really the same statement repeated in reverse? If M is a unique ID in the manifest and spine item S references M, then it references a unique item in the manifest. If T also references M then it *also* references a unique item.  That is, the uniqueness of M in the manifest doesn't change when referenced multiple times. I don't think we care about the uniqueness of M (that is handled elsewhere), what we care about is the uniqueness of the references to M. Do we really mean:

```
Each itemref element MUST reference the ID of an item in the manifest via the IDREF
in its idref attribute, and item IDs MUST NOT be referenced more than once.
```

This seems to capture both requirements: you have to reference an item in the manifest, and you can't reference the same item multiple times.

epub33/rs/index.html
bduga: I didn't like MUST since it seems to mandate UI. For instance, what if a RS allows for reordering of spine items by the end user (for instance, a cookbook where you could create your own recipe order). But the term "correct location" is so vague maybe it is ok to make this a must, I just don't want to make RS authors think they are not allowed to reshuffle content.

epub33/rs/index.html
bduga: Agreed, pages is a bad word here (assumes pagination, etc). Also fine with making this a MUST.

epub33/rs/index.html
bduga: Agreed "internal link" is not a great choice, but this is less about hyperlink activation and more about resolving the destination of a hyperlink (a RS may not allow following a hyperlink). Maybe:

> When a Reading System follows a hyperlink to a resource referenced multiple times in the spine, the Reading System MUST [...]


That also lets us turn the SHOULD into a MUST.

epub33/core/index.html
mattgarrish: Ya, I see your point. I believe we added "unique" in the not-too-distant past in a previous attempt to try and say spine items cannot reference the same manifest items, along with the parenthetical as extra clarification. That's why I read them as the same statement. Your reformulation works fine for me, though.

epub33/rs/index.html
mattgarrish: That's kind of confusing, as we don't ever mention reordering the spine. It also talks about the linear flow of the spine and then refers back to "that linear flow".

I agree with you in principle about reordering, but I'm not sure the current wording is clear how the "should" relates to the default spine ordering.

Could we maybe flesh this out more like: "When presented with a single Publication Resource that is repeated multiple times in the spine, Reading Systems MUST render that content in each instance where it is referenced when presenting the default linear flow. If the Reading System allows reordering of the spine, it SHOULD attempt to retain a logical ordering of the repeated items relative to their default positions."

I'm not sure that last sentence makes a lot of sense, but maybe it doesn't need to?

bduga: I don't really want to mention reordering, that was just an example of what we might be making unintentionally illegal. Maybe the problem is we are mandating how this particular content must be displayed, but instead we should be making it clear this content is no different from any other linear content. So something like:

> For rendering purposes when presented with a single Publication Resource that is repeated multiple times in the linear flow of the spine, Reading Systems MUST treat those items as though they were unique item references.

Hmm... that's terrible. But I think we are getting in trouble trying to explain what to do with these, and instead should make clear that whatever a RS does, it should be the same as what they do for correct references.

mattgarrish: > But I think we are getting in trouble trying to explain what to do with these

Right. Could we maybe generalize it to:

> Reading Systems MUST NOT skip spine references to duplicate manifest items when presenting the linear flow, unless such references are marked as non-linear.

Is that the gist of what we're after?

(No discussion is ever complete without bringing in non-linear content!)

bduga: That looks good to me! But I am not sure we need to specify the exception for non-linear since it says linear flow, so just `Reading Systems MUST NOT skip spine references to duplicate manifest items when presenting the linear flow.` Unless you want to bring up the fact that "linear flow" is never defined.

mattgarrish: > Unless you want to bring up the fact that "linear flow" is never defined.

True, and flow gets confusing with the rendering property. It'd be more precise to say "when rendering the linear reading order."

There are a number of other "linear flows" in this PR that should probably be fixed, then, too.

epub33/core/index.html
iherman: Works for me, too, will be part of my next commit.

iherman:
@bduga @mattgarrish I believe I have incorporated all the changes you discussed in the latest commit (https://github.com/w3c/epub-specs/pull/1889/commits/202058d6b887bb3285d4baf80753f9d0d71b66f8). Just to be on the safe side, could you look at the changes again?

(I have also added `<span>` elements with ID-s around the three statements in the RS which are now all MUST-s and, therefore, will require three tests at least.)

wareid (approved):

mattgarrish:
Generally looks fine to me now. Only last thing is the RS spec should probably have a change log entry since we're adding new requirements.

mattgarrish (approved):

iherman:
> Generally looks fine to me now. Only last thing is the RS spec should probably have a change log entry since we're adding new requirements.

It is there; have been added with the date of the 1st of November (when the PR was created).

mattgarrish:
Ah, so it is! I must have still had the core spec open when I looked for it. 😄

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-11-11-epub#section4) on 2021-11-11

List of resolutions:

- ***[Resolution No. 3](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-11-11-epub#resolution3):  merge 1889.***

### 4. Approve PRs.

_See github pull request [epub-specs#1889](https://github.com/w3c/epub-specs/pull/1889)._




**Dave Cramer:** this is writing language for when should happen when same item is referenced multiple times in the spine.  
… there are a bunch of approvals from reviewers over in github.  

**Matt Garrish:** okay with me.  

> **Proposed resolution: merge 1889.** *(Dave Cramer)*

> *Dave Cramer:* +1.

> *Matt Garrish:* +1.

> *Brady Duga:* +1.

> *Shinya Takami (高見真也):* +1.

> *Victoria Lee:* +1.

> *Matthew Chan:* +1.

> *Toshiaki Koike:* +1.

> *Masakazu Kitahara:* +1.

> ***Resolution #3: merge 1889.***

</details>

PR #1891, opened 2021-11-04
Updates to structure vocab

mattgarrish:
Prompted by the email question about the reference to IDPF specs, I took a deeper look at the vocabulary and fixed a few other things:

- I think we should keep the explanation of the "usage details" sections, but we don't need to mention IDPF in it. In this PR, I've tried to better explain, too, that those fields only apply when following the respective specification. You can use any term in any way if you don't intend to conform (e.g., using `index` doesn't require following the indexes spec).
- I've added proper bracketed references for all the IDPF specs and added them to `localBiblio`
- I've added usage details and links to EPUB 3.3 for toc, page-list, and landmarks
- Surprisingly, none of the dictionary terms were added after that specification became a recommendation. Do you know if there was a reason for that @rdeltour or if no one got around to it? I've added a section to cover them.

### DIFF
--- a/epub33/ssv/index.html
+++ b/epub33/ssv/index.html
@@ -98,6 +98,7 @@
 			<p>This vocabulary, a part of EPUB® 3, defines a set of properties relating to the description 
 				of structural semantics of written works.</p>
 		</section>
+		
 		<section id="sotd"></section>
 
 		<section id="sec-introduction">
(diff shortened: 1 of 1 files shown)


rdeltour (approved):
Looks good.

The ordering of the terms, and also subsections of Document Reference Sections, may look odd. I understand it follows the order of appearance in the related specifications, but may not be easy to lookup.
I'd mildly suggest to follow alphabetical order, but can certainly live as-is.

> Surprisingly, none of the dictionary terms were added after that specification became a recommendation. Do you know if there was a reason for that @rdeltour or if no one got around to it? 

I don't remember any particular reason. That no one got around to it is the most plausible explanation 🙂

mattgarrish:
> I understand it follows the order of appearance in the related specifications, but may not be easy to lookup.

Ya, I was about to reorder them alphabetically, but none of the structure vocabulary is alphabetized. The terms are all ordered with a pedagogical aspect to them.

For example, we have "cover, frontmatter, bodymatter, backmatter" instead of "backmatter, bodymatter, cover, frontmatter".

I don't think this "teaching" aspect is particularly necessary, and where it actually matters there are specs to define the full scope of what has to be done, so it doesn't help in that regard, either. We alphabetized all the other vocabularies in the EPUB spec, so I'm fine doing that here, too.

rdeltour:
> Ya, I was about to reorder them alphabetically, but none of the structure vocabulary is alphabetized. The terms are all ordered with a pedagogical aspect to them.

Good point. I'm fine whatever you decide. The pedagogical aspect is interesting, it may work better with short lists than bigger ones.
The terms that made me react about ordering were `part-of-speech*`, which are not all grouped together.

mattgarrish:
Aside from the fact that some ordering doesn't make much sense, and some sections are alphabetized, trying to instruct about use through ordering assumes someone can make sense of why the ordering exists. If you've never encountered the terms front, body and back matter, you won't understand the ordering until after you've read the definitions.

Alphabetizing avoids having a mix of instructive ordering and meaningless ordering, which further confuses people. I'm going to commit a reorder in a minute.

iherman (approved):
I have reviewed the original changes, ie, https://github.com/w3c/epub-specs/pull/1891/commits/6f5051bc0604f3e16cf397b58c152c8036b0f735. The full PR inckudes a reoreding which is pretty difficult to review on my iPad mini, so I trust you guys on that (too:-))

PR #1895, opened 2021-11-05
Restore reference to Unique Identifier

### LINKED ISSUE #1894: RS spec section on unique-identifier should talk about dc:identifier instead/as well
https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-pub-identifiers says:

> Reading Systems SHOULD NOT depend on the `unique-identifier` attribute being unique to one and only one EPUB Publication. 

This sounds like it's referring to `<package ... unique-identifier="pub-id">` , not `<dc:identifier id="pub-id">confreq-rs-pkg-unique-id</dc:identifier>`. But in practice, `unique-identifier` is basically never unique while `dc:identifier` usually is.

This is better in the format spec (https://w3c.github.io/epub-specs/epub33/core/#elemdef-opf-dcidentifier), which says:
> The EPUB Creator MUST provide an identifier that is unique to one and only one EPUB Publication — its Unique Identifier — in an `identifier` element. This `identifier` element MUST specify an `id` attribute whose value is referenced from the package element's `unique-identifier` attribute.

I think the RS spec should be edited to make clear that we're mostly talking about `dc:identifier`, not `unique-identifier`. Or maybe both of them.

mattgarrish:
s/unique-identifier attribute/Unique Identifier/ in caution against reading systems depending on uniqueness.

Fixes #1894

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1895.html" title="Last updated on Nov 6, 2021, 4:36 PM UTC (d6a5d31)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1895/fb98744...d6a5d31.html" title="Last updated on Nov 6, 2021, 4:36 PM UTC (d6a5d31)">Diff</a>

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -264,21 +264,21 @@
 
 				<ul class="conformance-list">
 					<li>
-						<p id="confreq-rs-xml-nval" data-tests="#confreq-rs-xml-nval_invalid,#confreq-rs-xml-nval_unclosed">a
-							<a href="https://www.w3.org/TR/2008/REC-xml-20081126/#proc-types"
-								>conformant non-validating processor</a> [[XML]].</p>
+						<p id="confreq-rs-xml-nval"
+							data-tests="#confreq-rs-xml-nval_invalid,#confreq-rs-xml-nval_unclosed">a <a
+								href="https://www.w3.org/TR/2008/REC-xml-20081126/#proc-types">conformant non-validating
+								processor</a> [[XML]].</p>
 					</li>
 					<li>
-						<p id="confreq-rs-xml-ns" data-tests="#confreq-rs-xml-ns">a
-							<a href="https://www.w3.org/TR/2009/REC-XML-NAMES-20091208/#ProcessorConformance"
+						<p id="confreq-rs-xml-ns" data-tests="#confreq-rs-xml-ns">a <a
+								href="https://www.w3.org/TR/2009/REC-XML-NAMES-20091208/#ProcessorConformance"
 								>conformant processor</a> as defined in [[XML-NAMES]].</p>
 					</li>
 				</ul>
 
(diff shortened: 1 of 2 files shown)


epub33/rs/index.html
```diff
@@ -364,10 +364,11 @@ <h4>Base Direction</h4>
 				<h3>Unique Identifier</h3>
 
 				<p id="confreq-rs-pkg-unique-id" data-tests="#confreq-rs-pkg-unique-id">Reading Systems SHOULD NOT
-					depend on the <code>unique-identifier</code> attribute being unique to one and only one EPUB
-					Publication. Determining whether two EPUB Publications with the same Unique Identifier represent
-					different versions of the same publication, or different publications, may require inspecting other
-					metadata, such as the last modification date, titles, and authors.</p>
+					depend on the <a href="https://www.w3.org/TR/epub-33/#dfn-unique-identifier">Unique Identifier</a>
```
dlazin, line 367: Here or somewhere else, could we link to https://www.w3.org/TR/epub-33/#sec-opf-dcidentifier? That section is much more explicit about the fact that the unique-identifier is not actually the Unique Identifier itself :)

dlazin (approved):

epub33/rs/index.html
mattgarrish: I've updated the definition for Unique Identifier to be more specific that it is the value of the dc:identifier specified by the unique-identifier attribute (and added linking). Let me know if this works better for you now.

iherman (approved):

epub33/rs/index.html
dlazin: Much better :)

dlazin (approved):

dlazin:
Merging now to avoid conflict with my test-renaming mega-PR.

Issue #1464, opened 2021-01-12
What does it mean to "support" a foreign resource? 


dauwhe:
So, I wrote a test for manifest fallbacks. I made a JSON content document (`media-type="application/json"`) with an XHTML fallback.  Apple Books, Thorium, and Calibre display the JSON directly. ADE 4.5 crashes on opening the file. 

I suspect this is because browsers/web views will try to open JSON and render it as text.Then I made an EPUB with an XML content doc (`media-type="application/dtc+xml"`). Apple Books said the book was corrupt, although there were no EPUBCheck errors. When I opened it in Thorium, I was presented with a dialog box allowing me to download the XML file. Calibre rendered the XML as text.

dauwhe:
Made a third test with `media-type="application/xml"`. Thorium displayed the XML tree. 
<img width="968" alt="Thorium-manifest-fallback-003" src="https://user-images.githubusercontent.com/5687700/104343388-e14cd280-54c9-11eb-9cd6-d3f0afd4cf15.png">

teytag:
An experience I wanted to share. I prepared my EPUB3 content with "cryptocurrency-blockchain" (JSON). The book is read when logged in with the "reading key" prepared for the reader. ReadiumDesktop and Thorium viewed it successfully. And in Thorium the "bookmarks" feature worked fine.

dauwhe:
Reading system developers: does your RS use manifest fallbacks? Do you check mime types to decide when to use the fallback?

danielweck:
No.

/hides in shame

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-05-27-epub#section2) on 2021-05-27

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

### 2. What does it mean to "support" a foreign resource?  (issue epub-specs#1464)

_See github issue [#1464](https://github.com/w3c/epub-specs/issues/1464)._




**Dave Cramer:** When writing some spec tests, one of the foundational aspects is a core media type  
… ie something that does not need a fallback  
… so manifest fallbacks are also foundational  
… Wrote some fanciful tests with docbook, binary files, etc  
… _**No**_ reading system implemented manifest fallbacks  
… Some of the behavior was not ideal for the end user  
… for instance a .dmg file was downloaded to the local system [Scribe note: !!!]  
… Looks like everyone just throws it at a webview and let the webview handle it  
[...]

PR #1911, opened 2021-11-15
Clarifications to manifest fallbacks and examples

### LINKED ISSUE #1877: Manifest fallbacks
The [Manifest Fallbacks](https://w3c.github.io/epub-specs/epub33/core/#sec-foreign-restrictions-manifest) section is (still) difficult to understand.

One reason is that the [examples related to fallbacks](https://w3c.github.io/epub-specs/epub33/core/#sec-item-elem-examples) are currently placed before this section. My first proposal is therefore to move examples after the fallbacks section, at the same headings level. 

We could clean it up a little bit more: the specification of fallbacks is split between the section about the _item_ element (fallbacks from core resources) and the fallbacks section (fallbacks from foreign resources). It would IMHO be better to move the sentence "EPUB Creators MAY provide fallbacks for Core Media Type Resources (e.g., to provide a static alternative to a Scripted Content Document)." to the fallbacks section and rephrase the introductory sentence of the _item_ section as "The fallback attribute takes an IDREF [XML] that identifies a fallback for the Publication Resource referenced from the item element, as defined in "§ 2.3.2.4.3 Manifest Fallbacks". 

Also, [example 29](https://w3c.github.io/epub-specs/epub33/core/#example-29) is very difficult to understand by reading  the explanation. Looking at the code I understand the following:  "In the following example, an EPUB Creator inserts a reference to a local audio file directly into the spine. For this reference to be valid, a fallback to a Core Media Type Resource has to be provided, in this case an XHTML document containing a reference to the audio file.". 

In this example, the _item_ identified by "#audio01-xhtml" is not included. This does not help understanding the situation. And moving the spine extract to the start of the sample would also be helpful for the reader. 

Plus, this example does not correspond to any valid use case: we've never seen audio resources in the spine, audiobooks are created in a different way. Replacing the audio file by something that EPUB creators are inclined to put in the spine would be clearer (but what?).

mattgarrish:
This PR covers what was mentioned in #1877 plus some other improvements I thought would help. Specifically, it:

- moves the manifest fallbacks section under the item element definition
- takes the explanation of fallbacks from 2.2.1.3 Foreign Resources to better introduce the section - leaving a more basic explanation of how they're expressed in that section
- moves the sentence about fallbacks for core media types from the item element definition and reworded per the issue
- reorders the examples so the two about manifest fallbacks are together and the two about listing resources are together
- adds titles to the examples to help explain their relationships
- replaces the example with a docbook/z3986/xhtml fallback chain with a JPEG/SVG
- tries to simplify the example descriptions to make the point of each example more understandable
- adds additional markup to clarify the examples, like including the xhtml manifest entry for a remote audio reference to show it needs the remote-resources properties attribute

Fixes #1877

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1911.html" title="Last updated on Nov 20, 2021, 1:51 PM UTC (a6ee952)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1911/1182db6...a6ee952.html" title="Last updated on Nov 20, 2021, 1:51 PM UTC (a6ee952)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -87,20 +87,16 @@
 		<section id="sotd"></section>
 		<section id="sec-introduction">
 			<h2>Introduction</h2>
-
 			<section id="sec-intro-overview" class="informative">
 				<h3>Overview</h3>
-
 				<p>EPUB 3 has been widely adopted as the format for digital books (ebooks), and this revision continues
 					to increase the format's capabilities to better support a wider range of publication requirements,
 					including complex layouts, rich media and interactivity, and global typography features. The
 					expectation is that publishers will utilize the EPUB 3 format for a broad range of content,
 					including books, magazines, and educational, professional, and scientific publications.</p>
-
 				<p>This specification represents the core of EPUB 3 and includes the conformance requirements for
 						<a>EPUB Publications</a> — the product of the standard. The other specifications that comprise
 					EPUB 3 are as follows:</p>
-
 				<ul>
 					<li>
 						<p><a href="https://www.w3.org/TR/epub-rs-33/">EPUB 3 Reading Systems</a> [[EPUB-RS-33]] —
@@ -112,25 +108,20 @@
 							accessibility conformance and discovery requirements for EPUB Publications.</p>
 					</li>
 				</ul>
-
 				<p>These specifications represent the formal list recognized as belonging to EPUB 3 and that contain
 					functionality normatively referenced as part of the standard. The development of extension
 					specifications periodically adds new functionality to EPUB Publications. Features and functionality
 					defined outside of core revisions to the standard, while not formally recognized in this
 					specification, are nonetheless available for <a>EPUB Creators</a> and Reading System developers to
 					use.</p>
-
 				<p>The informative <a href="https://www.w3.org/TR/epub-overview-33/">EPUB 3 Overview</a>
 					[[EPUB-OVERVIEW-33]] provides a general introduction to EPUB 3. A list of technical changes from the
 					previous version is also available in the <a href="#change-log">change log</a>.</p>
 			</section>
-
 			<section id="sec-intro-spec-org" class="informative">
 				<h3>Organization</h3>
-
 				<p>This section reviews the organization of the EPUB specifications through the central product they
 					define: the <a>EPUB Publication</a>.</p>
-
 				<p>An EPUB Publication is typically represented by a single <a>Package Document</a>. This document
 					includes metadata used by <a>Reading Systems</a> to present the content to the user, such as the
 					title and author for display in a bookshelf as well as rendering metadata (e.g., whether the content
@@ -138,12 +129,10 @@
 						<a>spine</a> that lists the default sequence in which to render documents as a user progresses
 					through the content. Refer to <a href="#sec-package-doc"></a> for the requirements for the Package
 					Document.</p>
-
 				<p>An EPUB Publication also includes another key file called the <a>EPUB Navigation Document</a>. This
 					document provides critical navigation capabilities, such as the table of contents, that allow users
 					to navigate the content quickly and easily. Refer to <a href="#sec-nav"></a> for more information
 					about this document.</p>
-
 				<!--
 				<figure>
 					<figcaption> The following example visually represents the structure of the EPUB format. </figcaption>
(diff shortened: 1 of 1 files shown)


epub33/core/index.html
```diff
@@ -2873,128 +2932,87 @@ <h6>Examples</h6>
...
-          href="chap1.xhtml" 
-          media-type="application/xhtml+xml"/&gt; 
-    … 
+								<aside class="example" id="example-manifest-flbk"
+									title="Foreign Resource with Fallback in Spine">
+									<p>The following example shows the <a href="#sec-foreign-restrictions-manifest"
+											>fallback chain mechanism</a> allowing a <a>Foreign Resource</a> (JPEG) to
+										be listed in the spine with fallback to an SVG Content Document.</p>
+									<pre>Spine:
+&lt;itemref idref="page-001"/&gt;
```
iherman: I am bothered (here and the other examples) by the style of the example. In all the examples, we provide an extract of a bona fide package XML file with, usually, the spine following the manifest part. In this example (and the other examples in the section) we separate these, and it took me a certain time to realize that. What I would have preferred is something like:

```xml
<package …>
    <manifest>
        …
        <item id="page-001" …>
        <item id="page-001-svg" …>
        …
    </manifest>
    <spine>
        …
        <itemref idref="page-001" />
        …
    </spine>
</package>

File: epub33/core/index.html [hunk 2]
```diff
@@ -2873,128 +2932,87 @@ <h6>Examples</h6>
...
-										in the XHTML Content Document, so the file is a Publication Resource).</p>
+								<aside class="example" title="Foreign Resource with Fallback">
+									<p>The following example shows a hyperlink to a JPEG that will render as a
+											<a>Top-Level Content Document</a>. As a result, EPUB Creators must list the
+										JPEG as a non-linear item in the <a href="#sec-spine-elem">spine</a>
+										<em>and</em> specify a fallback to an EPUB Content Document in its manifest
+										declaration (in this case, an SVG).</p>
 									<pre>XHTML:
-&lt;audio src="http://www.example.com/book/audio/ch01.mp4" controls="controls"/&gt;
+&lt;img src="images/infographic.jpg" alt="&#8230;" />
```
iherman: What is the role of this entry as far as the example is concerned? My understanding is that by virtue of referring to the image via a `<a>` the item must be added to the spine as non-linear, etc. I.e., the presence of `` seems to be irrelevant for the example...

mattgarrish: It was asked in the issue for a realistic example of it in use. The image shows the context - that it's not embedding that requires a spine entry. I can expand on that in the description.

iherman: Well, let us see what the expansion leads to :-) because I do not see why having the `` makes it more realistic... As far as I am concerned, the idea that a content links "out" to an image, ie, the author wants the image to be the "main" content on the screen is perfectly realistic...

mattgarrish: The examples were put in to instruct people on when resources have to be listed in the manifest/spine and when they don't. Showing a hyperlink to an image without showing how the image is rendered within the current content document is cutting off half the issue. You're always expanding an image that's been displayed, but it's not the image itself in the content that is the issue, it's the use of a hyperlink. It's better to be explicit about that than leave people guessing about whether it only applies to the hyperlink.

iherman: Ok, I take that, but then let us make this explicit in the example caption... I think main issue to emphasize is

>  it's not the image itself in the content that is the issue, it's the use of a hyperlink

dauwhe: Example 27 bothers me. Why would we require a fallback? The image is not referenced from the spine. 

In most reading systems, if you click or double-click on an image it will open in its own window anyway. We don't need a manifest fallback in those cases. Why do we need one here?

mattgarrish: > The image is not referenced from the spine.

The image has to be referenced from the spine if you hyperlink to it. It's a top-level content document in that case.

> In most reading systems, if you click or double-click on an image it will open in its own window anyway

Ya, but this is an effort to come up with something authors might plausibly do that would load in the spine. It used to be a link to open an audio file, but that's more unrealistic. In this case, presumably the idea is that this ensures the image opens as a separate document in all reading systems regardless of extra functionality.

I found at least one old epubcheck complaint about this: https://github.com/w3c/epubcheck/issues/324

And I'm sure if I had the inclination to search the old IDPF forums there would be threads there, too, as this is a nuisance for authors that has popped up from time-to-time over the years.

But if you have a better example of non-standard stuff being linked to in the spine, I'm open to anything else.

mattgarrish: > let us make this explicit in the example caption

I did a rewrite earlier this morning, but I've tweaked it some more, including the example title. If there's something you think still needs saying, let me know.

iherman: @mattgarrish the example looks o.k. to me, and I am happy to approve the PR.

However... I wonder whether we should not separate this PR from the genuine issue @dauwhe has raised about the usefulness of this case (see also https://github.com/w3c/epub-specs/issues/1464#issuecomment-970504419). What about changing the example for a link to an external data file like, say, a CSV file or a Javascript raw code? That is of course not a realistic example for a novel, but it is for a scholarly book. In both cases providing a fallback does make more sense (although I am not sure any RS implements it) and it separates from the issue whether a fallback on a core media type should be required in the first place... And it would allow us to merge this otherwise fine PR.

iherman (approved):

epub33/core/index.html
mattgarrish: > I wonder whether we should not separate this PR from the genuine issue @dauwhe has raised about the usefulness of this case

Ya, but are we likely to change this revision? I'm not a fan of fallbacks, but allowing CMTs in the spine is a big change in direction to squeeze into the last few weeks before going to CR, especially after already going through the accessibility horizontal review.

If we change to CSV, then the example would no longer be able to explain why embedded content is different from hyperlinked content, since you can't embed a CSV without a fallback, either. The only things you can embed are CMTs, so we'd just be shuffling media types.

iherman: On the CSV issue: for me, the essence of the example is that if a direct link is done to a CSV file from the EPUB, and that file is part of the EPUB instance, then a fallback should be provided (at least this is the way I understand things). Maybe that should be a separate example, which does not mix in the hyperlink vs. embedding and helps people understand the issue.

B.t.w., I do not think @dauwhe is considering allowing an image in the spine for a linear content without fallback.

mattgarrish: > if a direct link is done to a CSV file from the EPUB, and that file is part of the EPUB instance, then a fallback should be provided

Arguably, no, but we never solved "downloadable" content from an EPUB.

Linking to a CSV within the EPUB is kind of unrealistic, as getting a mess of raw data dumped into the reading system display area probably doesn't help users, especially if the RSes ignore it and render an HTML fallback, which I presume would format the data as a table. But as a second example I don't really care that strongly about realism, since we're out on the edges anyway.

> I do not think @dauwhe is considering allowing an image in the spine for a linear content without fallback.

I'm not sure that changes the issue much, as we don't definitively define behaviour for linear v. non-linear.

mattgarrish: Are you okay with the new example, @iherman ? (i.e., can we merge this?)

iherman: Yes, I am, that is why I gave a thumbs up :-)

mattgarrish: Okay, wasn't sure if that was for the idea of adding the example or for what I committed.

llemeurfr:
Thanks Matt, Ivan, this is much clearer now. I'm still not convinced Readium toolkits have to support the fallbacks feature (because we don't see any support of this feature in producted EPUB 3) but if we were to implement it, we would at least understand it ;-)

PR #1927, opened 2021-11-21
Some editorial changes following PR 1898

### LINKED ISSUE #1919: Declarative alternative to the file path definition
Currently, the [File Path](https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-manifest) term is defined in a [separate section](https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-manifest) in an algorithmic fashion, which makes this a bit different from the rest of the document that favours declarative definitions. The question is whether there is a good alternative for the definition that would replace the algorithmic section.

@mattgarrish @rdeltour

### LINKED ISSUE #1923: Example 46 formatting
Looks like there's an example embedded in the example, so there are two example headings: https://www.w3.org/TR/epub-33/#example-46

It's also kind of strange to start the description with "for example".

### LINKED ISSUE #1924: Mixing plural/singular
The third paragraph in 3.4 of the RS specification begins:

> It SHOULD NOT use non-Publication Resources in the rendering of an EPUB Publication

https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-manifest

But the paragraphs before use "reading systems". Either switch to "They" or "Reading Systems".

iherman:
These are issues found by @mattgarrish.

It fixes #1923 and #1924.

It also fixes the editorial comment of https://github.com/w3c/epub-specs/issues/1919#issuecomment-974495911, but the closure of #1919 shouldn't be done automatically by merging this PR.

Finally, there was an invalid cross-reference to a parsing section.

See:

* For EPUB 3.3:
    * [Preview](https://raw.githack.com/w3c/epub-specs/post-1898-minor-editorials/epub33/core/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/core/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/post-1898-minor-editorials/epub33/core/index.html)
    * [Preview](https://raw.githack.com/w3c/epub-specs/post-1898-minor-editorials/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/post-1898-minor-editorials/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1927.html" title="Last updated on Nov 24, 2021, 6:18 AM UTC (a9fd23b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1927/a3a2d65...a9fd23b.html" title="Last updated on Nov 24, 2021, 6:18 AM UTC (a9fd23b)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5837,9 +5837,7 @@ No Entry</pre>
 								<var>url</var>, with the <a>container root URL</a> as <a
 								data-cite="url#concept-base-url"><var>base</var></a>.</p>
 
-						<aside class="example" title="Relative URL in a container.xml file">
-							<p>If <code>META-INF/container.xml</code> has the following content:</p>
-
+						<aside class="example" title="URLs in a container.xml file are relative to the OCF Abstract Container">
 							<pre>
 	&lt;?xml version="1.0"?&gt;
 	&lt;container version="1.0" xmlns="urn:oasis:names:tc:opendocument:xmlns:container"&gt;
@@ -5849,7 +5847,7 @@ No Entry</pre>
 		&lt;/rootfiles&gt;
 	&lt;/container&gt;
 				</pre>
-							<p>then the path <code>EPUB/Great_Expectations.opf</code> is relative to the root directory
+							<p>The path <code>EPUB/Great_Expectations.opf</code> is relative to the root directory
 								for the OCF Abstract Container and not relative to the <code>META-INF</code>
 								directory.</p>
 
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5744,7 +5744,7 @@ No Entry</pre>
 				</section>
 
 				<section id="sec-file-names-to-path-names">
-					<h4>Deriving File Paths of Files</h4>
+					<h4>Deriving File Paths</h4>
 
 					<p>To derive the <a>File Path</a>, given a file or directory <var>file</var> in the <a
 							href="#sec-container-abstract">OCF Abstract Container</a>, apply the following steps
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5837,7 +5837,7 @@ No Entry</pre>
 								<var>url</var>, with the <a>container root URL</a> as <a
 								data-cite="url#concept-base-url"><var>base</var></a>.</p>
 
-						<aside class="example">
+						<aside class="example" title="Relative URL in a container.xml file">
 							<p>If <code>META-INF/container.xml</code> has the following content:</p>
 
 							<pre>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -6035,7 +6035,7 @@ No Entry</pre>
 				<section id="sec-file-names-to-path-names">
 					<h4>Deriving File Paths</h4>
 
-					<p>To derive the <a>File Path</a>, given a file or directory <var>file</var> in the <a
+					<p>To <dfn id="dfn-derive-the-file-path">derive the File Path</dfn>, given a file or directory <var>file</var> in the <a
 							href="#sec-container-abstract">OCF Abstract Container</a>, apply the following steps
 						(expressed using the terminology of [[INFRA]]):</p>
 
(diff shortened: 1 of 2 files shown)


epub33/rs/index.html
```diff
@@ -475,7 +475,7 @@ <h3>Manifest</h3>
 						id="confreq-rs-pkg-manifest-unknown" data-tests="#pkg-manifest-unknown">Reading Systems MUST
 						ignore values of the <code>properties</code> attribute they do not recognize.</span></p>
 
-				<p id="confreq-rendition-rs-manifest" data-tests="#pkg-manifest-unlisted-resource">It SHOULD NOT use
+				<p id="confreq-rendition-rs-manifest" data-tests="#pkg-manifest-unlisted-resource">Reading Systems SHOULD NOT use
```
dlazin, line 478: Were lines 1268-86 supposed to be committed?

dlazin (approved):

epub33/rs/index.html
iherman: I do not understand your question @dlazin

rdeltour (changes requested):
I'm suggesting making the changes [suggested by @mattgarrish](https://github.com/w3c/epub-specs/issues/1919#issuecomment-975480790) to close #1919.

Other than that looks OK to me!

rdeltour (approved):

epub33/core/index.html
```diff
@@ -5746,11 +5746,12 @@ <h4>File Paths and File Names</h4>
 				<section id="sec-file-names-to-path-names">
 					<h4>Deriving File Paths of Files</h4>
```
mattgarrish: Can "of Files" be stripped here? It's kind of redundant with "File Paths" and also doesn't account for directories.

epub33/core/index.html
```diff
@@ -5746,11 +5746,12 @@ <h4>File Paths and File Names</h4>
...
 
-					<p>To derive the <a>File Path</a> of a file or directory <var>file</var> in the <a
-							href="#sec-container-abstract">OCF Abstract Container</a> apply the following steps
+					<p>To derive the <a>File Path</a>, given a file or directory <var>file</var> in the <a
```
mattgarrish: I believe the convention is to put dfn tags around the label to be able to reference the algorithm and get the list of references to it (even if it seems to not be referenced in this document right now):

`<dfn>derive the <a>File Path</a></dfn>`

The link to the definition of File Path makes this a bit weird, though. It works, but depending on which half of the definition you click you either get the references list or go the terminology.

You could maybe assume readers will know its terminology from the capital letters and omit the link.

epub33/core/index.html
iherman: Right. Done.

epub33/core/index.html
iherman: Hm. I may not understand what you say, but... the convention is, as far as I could see, the the `<dfn>` is in the terminology section, linking to the section that contains a more detailed definition (if applicable). This follows the same pattern...

mattgarrish: > the the `<dfn>` is in the terminology section

Right, I mean the convention for defining the algorithm.

If you go to the URL standard, etc., or even to pub-manifest, the name of the algorithm is defined in `dfn` tags. This lets you reference the algorithm by name, so you can later write stuff like:

"In this case, you need to `<a>`derive the File Path`</a>` of ...".

Here's an example for the "normalize data" algorithm in pub-manifest: https://www.w3.org/TR/pub-manifest/#normalize-data

If you click on the permalink it will show you where else in the document the algorithm is referenced.

iherman: I understand that. My problem is whether this (sound!) convention is not something we should use elsewhere in the document, too.

What I would do is to follow this style for this section (I prefer not to add the link because it would make things confusing) and we may want to sweep through the text at some point to see if there are other places to update accordingly.

iherman: (Done in the latest commit)

mattgarrish: Right, I hate links within links, so all good now from my perspective.

mattgarrish (approved):

PR #1928, opened 2021-11-21
Added the test references for the rendition:flow property

iherman:
This is the counterpart of https://github.com/w3c/epub-tests/pull/90

### DIFF
--- a/epub33/core/vocab/rendering.html
+++ b/epub33/core/vocab/rendering.html
@@ -39,11 +39,11 @@
 			
 			<dl class="variablelist">
 				<dt id="paginated">paginated</dt>
-				<dd>
+				<dd id="paginated-dd" data-tests="#pkg-flow-paginated">
 					<p>Dynamically paginate all overflow content.</p>
 				</dd>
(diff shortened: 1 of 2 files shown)


iherman:
We may have an editorial problem with respec, though: the test references do not appear, although the code is there. I suspect the following: the section on flow properties is not _in_ the HTML code for the core spec, but it is in a different file (`core/vocab/rendering.html`) that is "included" in the main text through a respec tool. I suspect that the code related to `data-tests` is not executed on that included code. I will have to inquire. (That should not affect the PR itself, though.)

Cc @mattgarrish

iherman:
> We may have an editorial problem with respec, though: the test references do not appear, although the code is there. I suspect the following: the section on flow properties is not _in_ the HTML code for the core spec, but it is in a different file (`core/vocab/rendering.html`) that is "included" in the main text through a respec tool. I suspect that the code related to `data-tests` is not executed on that included code. I will have to inquire. (That should not affect the PR itself, though.)
> 
> Cc @mattgarrish

Strike that. It was a nasty cache issue.

PR #1931, opened 2021-11-22
Typo

### LINKED ISSUE #1924: Mixing plural/singular

mattgarrish:
s/it/reading systems/

Fixes #1924

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -202,8 +202,8 @@
 		<section id="sec-pub-resources">
 			<h3>Publication Resource Processing</h3>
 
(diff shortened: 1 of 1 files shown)


mattgarrish:
I see this is in #1927, too. Closing.

Issue #1933, opened 2021-11-23
Do we want to keep the container root URL heuristics?


iherman:
(This is an [editors' note in the spec](https://w3c.github.io/epub-specs/epub33/rs/#sec-container-iri) converted into a github issue.)

The definition for URL's of the Root directory in previous versions of the RS spec contained the following note:

> This specification does not mandate any particular implementation technique for the creation of a unique origin in the absence of a reliable Web origin (e.g., HTTP URL scheme + host + port). The necessary heuristics may include the combination of the Publication's Unique Identifier (although, in practice, these identifiers may in fact not guarantee globally-unique identification, that is why it is recommended to combine multiple techniques), filesystem path, OCF Zip Container checksum, etc.

The new, post #1898 version on the text has replaced the original [with a more precise definition](https://w3c.github.io/epub-specs/epub33/rs/#sec-container-iri) which relies on the RS to use a root URL that relies on some features it must adhere to. The question is whether that note should be still kept in the text, or it is o.k. to remove it.

(Note: in the current version, i.e., the one produced by #1898, that note has been removed.)

iherman:
After some thoughts, my proposal is to keep that note removed (ie, no change on the spec as it stands today).

@rdeltour @mattgarrish wdyt?

rdeltour:
OK for keeping the note removed 👍

dauwhe:
Let's keep the note removed.

Issue #1932, opened 2021-11-23
Should we have to define "EPUB Publication Instance"?


iherman:
(This is an [editors' note in the spec](https://w3c.github.io/epub-specs/epub33/rs/#sec-container-iri) converted into a github issue.)

The [definition for URL's of the Root directory](https://w3c.github.io/epub-specs/epub33/rs/#sec-container-iri) says:

> The origin of the container root URL is unique for each EPUB Publication instance.

Do we have to define what "EPUB Publication instance" is? Or should we consider as some sort of common knowledge?

mattgarrish:
If the text elaborating on what it means weren't in a note, it probably would be fine to omit any further explanation.

Maybe just strip the note formatting?

iherman:
> If the text elaborating on what it means weren't in a note, it probably would be fine to omit any further explanation.
> 

My problem is not that it is in a note; the term  appears twice in normative texts: [§4.4](https://w3c.github.io/epub-specs/epub33/rs/#sec-container-iri) and [§7.1.1](https://w3c.github.io/epub-specs/epub33/rs/#sec-container-iri). If you think that it is fine to omit any further explanation, then the third occurrence in the note does not change anything...

The question is whether it does require further explanation or not. (Personally, I am fine leaving the text without it.)

mattgarrish:
> the term appears twice in normative texts

But we can solve that with a dfn tag on the first instance. We don't have to go to the terminology section for a technical definition like this, especially when it has limited application. I'd keep the terminology section for terms that are widely used throughout the specifications.

But I don't like having to read a note to find out what an "instance" is. It should be clearer in the normative text, otherwise it is easily confused with being nothing more than a ten cent way of saying "an EPUB Publication".

I wonder if there's some other way of phrasing it in the third bullet, though, to make its meaning more apparent. This is maybe a bit wordy, but something like:

> The origin of the container root URL is unique for each user-specific instance of an EPUB Publication in a Reading System.

I could live with the elaboration in a note, then, as it would follow more directly from the bullet.

iherman:
I sort of like "user-specific instance of ab EPUB Publication" and I would be o.k. exchanging all occurrences in §4.4 and §7.1.1. It is a bit verbose, but it makes the intention right and leaves it clearer that, in some sense, this "instance" is very much dependent on the particular RS.

[...]

PR #1898, opened 2021-11-10
Proposal for base URLs to be used for URL parsing

iherman:
This is a PR version of @rdeltour's https://github.com/w3c/epub-specs/issues/1888#issuecomment-961502739 proposal.

Notes:

- I kept the term "Path Name" (for now). There is a symmetry with the usage of "File Name", and the term is referred to all over the place. It would require an extensive search-replace to get them all settled. We may do that later if we want to.
- I modified some statements to make them more spec-text, e.g., by using MUST if appropriate
- I agree with @rdeltour that the order of the sections 6.1.3 and 6.1.4. should be switched. I have not done that, though, because it would make the diff file much messier. We can do that once all the dust is settled...
- I have modified the scripting section on the RS spec (§4.4) by adding a reference to the new section on the root directory URL and removing the unique origin requirement from the bullet list. There were also two notes that appeared in that section; I moved one of the two to the new section and added the other as an editorial note; the latter does not sound to be relevant anymore, but better check.
- @mattgarrish, I have added an explicit reference to the two new terms in the RS text, because the magic script did not work well (I guess it would work once everything is published). I have done that so that ECHIDNA would not complain, we can improve that later.

Fix #1888
Fix #1374

See:

* For EPUB 3.3:
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/spooky-base-urls/epub33/core/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/core/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/spooky-base-urls/epub33/core/index.html)
* For EPUB 3.3 Reading Systems:
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/spooky-base-urls/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/spooky-base-urls/epub33/rs/index.html)

***
[...]

PR #1934, opened 2021-11-23
Removing editor's notes

iherman:
Two editors' notes have been added in #1898. Those two notes have been turned into explicit issues (see #1932 and #1933), and this PR removes those two notes from the spec.


dauwhe (changes requested):
I think there's a close `div` that should be removed, too?

mattgarrish (approved):
Just what @dauwhe said, otherwise fine by me.

iherman:
> I think there's a close `div` that should be removed, too?

Oops.... Thanks!

PR #1937, opened 2021-11-24
Remove old processing

mattgarrish:
This PR drops the hack I made to make the permalinks more accessible. Similar code is now part of respec.

I also removed the last couple of references to the common biblio.js file. Locators didn't need it and the FXL only references WCAG2 that isn't in the specref database.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1937.html" title="Last updated on Nov 25, 2021, 3:07 PM UTC (81f6a02)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1937/5f769de...81f6a02.html" title="Last updated on Nov 25, 2021, 3:07 PM UTC (81f6a02)">Diff</a>


iherman (approved):
Good to push whatever we can to respec :-)

Another feature that I try to use (and we may want to see if we want a systematic change) is the `data-cite` feature of respec for terms. It may make things simpler to type, but it also has an added bonus, namely that it checks whether the fragment id is correct. It did catch some errors for me in the past. (I also do not know whether it could make the extra .js file on terminology cross-links obsolete, too.)

mattgarrish:
> is the `data-cite` feature of respec for terms

For _terms_ or for linking to other specs?

I'd actually like to do both, but I believe `xref` is for linking terminology. It would be a lot simpler to use `xref` than have to add a hyperlink to the core spec every time. The old docbook process used to do this kind of easy linking for us the same way, and I really miss it!

The challenge of `data-cite` is that it means having to clean up all the references after the links, as I don't believe you can stop respec from auto-injecting a citation after the tags.

But given it's generally quiet over US holidays, I'll take a look at cleaning these up today.

mattgarrish:
> than have to add a hyperlink to the core spec every time

Oh, I see why we had to write custom code to do this. Only a limited set of whatwg specs are supported by xref.

But I guess we can still do `data-cite`.

wareid (approved):

mattgarrish:
I've fixed all the links to known specs to use `data-cite` now.

One thing I noticed while doing this is that when we dropped our central biblio file, we lost our custom undated svg reference. We had reverted to a 1.0 reference in a few specs as a result. I've fixed this again so it's back to what it was. One nuisance this causes, though, is that we can't use `data-cite` for SVG, as we'd have to pick a specific version.

I ran into the same problem with the latest CSS2 links. If you put in `data-cite="css2/...` the link comes out hardcoded to 2.1.

But other than that, it should be good now. I didn't try to change the documents that were done in markdown as I have no idea how this would work. (I expect there's probably some mystical set of thingy characters to express it!)

Unfortunately, the net result appears to be no improvement, as there's no checking of the fragments that I've noticed. Respec only notices if the spec reference is invalid. It appears to just make more work, as you have to use spec references in a non-standard attribute and still find the path/fragment to add to it. Dumping a URL into an href attribute is a lot simpler. I could see this being more useful for cross-linking your own specs, or if you know you'll be copying and pasting the same links over and over. But, whatever, I can live with it for now.

iherman:
Hm. I did run into respec yelling at me when I used data-spec with a wrong fragment... At least I think I did :-(

iherman:
Also, in my editor at least, when I add a new data-cite somewhere, I get a pull-down menu for the references I already used. I am not sure whose magic it is, but it is helpful...

mattgarrish:
> I get a pull-down menu for the references I already used.

Ya, I don't get anything like that in oxygen. Must be specific to your editor.

Checking fragments would have been nice, but I've tried mangling all sorts of them and not a peep.

mattgarrish:
Merging this caused the SSV to fail. I don't know what this "Token check error" is:

https://github.com/w3c/epub-specs/runs/4325920059?check_suite_focus=true#step:3:655

If I follow the details link to the json, I'm told for a fact that I'm not authorized to publish...

```
"fact": "You are not authorized to publish"
```

(That made me smile!)

Any thoughts on how to fix this @iherman ?

iherman:
Hm. I probably know "where" I would look, but I am not sure why. I see this is the first time echidna would be used for SSV (the TR version is still of the 4th of November); earlier attempts failed due to the process changes.

Will look at this tomorrow.

iherman:
Settled. The shortname in the action script was erroneous: it was `epub-ssv` as opposed to `epub-ssv-11`...

PR #1938, opened 2021-11-25
Replacing "Publication instance" with "user specific instance of…"

### LINKED ISSUE #1932: Should we have to define "EPUB Publication Instance"?

iherman:
Fixes #1932, see details there...

See:

    * [Preview](https://raw.githack.com/w3c/epub-specs/editorial/issue-1932/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/editorial/issue-1932/epub33/rs/index.html)


wareid (approved):

epub33/rs/index.html
```diff
@@ -1172,13 +1172,13 @@ <h4>URL of the Root Directory</h4>
 							with the <a>container root URL</a> as <a data-cite="url#concept-base-url"
 								><var>base</var></a> is the <a>container root URL</a>.</li>
 						<li>The <a href="https://url.spec.whatwg.org/#origin">origin</a> of the <a>container root
-								URL</a> is unique for each EPUB Publication instance.</li>
+								URL</a> is unique for each user-specific instance of an EPUB Publication.</li>
```
rdeltour: In #1932 @mattgarrish proposed:
> each user-specific instance of an EPUB Publication **in a Reading System**

The addition of "in a Reading System" implies that the the origin will not be the same for two different RS owned by the same user.

I'm not sure if it was intentionally left out or not. I don't have a definitive opinion on whether it's better with or without.

rdeltour (approved):
I made one non-blocking comment, but otherwise looks OK!

epub33/rs/index.html
iherman: Nop, it was my mistake, simply...

mattgarrish (approved):

lordt4ever (approved):

dauwhe (approved):
Yeah, I think this is an improvement. Thanks!

bduga (approved):
I don't think this is any worse and probably better than what we used to have, so lgtm. But I still find this vague concept of a user-specific instance to be fairly vague for a conformance requirement (and I am not sure how useful this section is in all honesty).

PR #1939, opened 2021-11-25
Discourage using "out-of-container" relative URLs

### LINKED ISSUE #1912: should we disallow "out-of-container" relative URLs?
## Definitions

By **out-of-container relative URL string**, I mean a relative URL string that has a number of double-dot path segments ("`..`") high enough to conceptually go outside the container.

For instance, "`../../../../EPUB/content.xhtml`" is an **out-of-container relative URL string** given the following container:

```
/
├── mimetype
├── META-INF
│   └── container.xml
└── EPUB
    └── content.xhtml
```

## Problem

In previous versions of EPUB, the URL definition was unclear (see #1888), but I believe the intent was to disallow them.
In the #1898 proposal, **out-of-container URL string** are conforming, but the base URL of the container is defined such that an **out-of-container URL string** is necessarily parsed into a in-container URL.

For instance, after #1898, the URL string "`../../../../EPUB/content.xhtml`" will be parsed to the same URL as the URL string "`EPUB/content.xhtml`".

But as we added to a note in the #1898 proposal, using an **out-of-container URL string** will likely lead to interoperability issues with legacy or non-conforming RS.
In addition, as I said earlier, I believe the intent in previous versions of EPUB was to disallow them.

## Proposal

I think we should forbid **out-of-container URL strings**.

Here's a proposal (assuming #1898 is merged).
Replace:

> In the OCF Abstract Container, when a file uses a URL string to reference another file in the container, the string MUST be a path-relative-scheme-less-URL string, optionally followed by `U+0023 (#)` and a URL-fragment string. 

by something along the lines of:

> In the OCF Abstract Container, a [relative-URL string](https://url.spec.whatwg.org/#relative-url-with-fragment-string) MUST be a **container relative URL string**.
> 
> A URL string _url_ is a **container relative URL string** if it is a [path-relative-scheme-less-URL string](https://url.spec.whatwg.org/#path-relative-scheme-less-url-string) and the following steps return true:
> 
>   1. let _testURLRecord_ be the result of applying the URL parser to _url_ with "`https://example.org/A/`".
>   2. let _testURLString_ be the result of applying the URL Serializer to _testURLRecord_.
>   3. if _testURLString_ does not start with "`https://example.org/A/`", then return false.
>   4. set _testURLRecord_ to the result of applying the URL parser to _url_ with "`https://example.org/B/`".
>   5. let _testURLString_ be the result of applying the URL Serializer to _testURLRecord_.
>   6. if _testURLString_ does not start with "`https://example.org/B/`", then return false.
>   7. Return true.

## Explanation

The proposal above intends to override the URL standard definition of [**relative-URL string**](https://url.spec.whatwg.org/#relative-url-string), so that:
  * **scheme-relative-** and **path-relative-** URL strings are not allowed (in other words, URL strings starting with "`/`" are not allowed)
  * "exceeding" or "leaky" URL strings are not allowed (in other words, URLs with enough "`..`" path segments to go "outside" the container are not allowed)

The intent is even if we refer to a broader "category" of URL strings, like a [**relative-URL-with-fragment string**](https://url.spec.whatwg.org/#relative-url-with-fragment-string), our restrictions on **relative-URL string** apply.

In some way, it is monkey patching the URL standard definition. Monkey patches are usually not considered a good thing. But I do not see how to do otherwise: for the document formats we own (e.g. Package Document), we can easily define what is a valid URL string; but for other formats used in EPUB (e.g. HTML), they directly refer to the URL standard so I don't see an alternative to tweaking the definition.

## Editorial consequences

We will be able to replace all our use of:
> path-relative-scheme-less-URL string, optionally followed by `U+0023 (#)` and a URL-fragment string
by 
> [**relative-URL-string with fragment string**](https://url.spec.whatwg.org/#relative-url-with-fragment-string) (which is a bit more readable).

We may no longer need to assume the [properties of the container root URL](https://raw.githack.com/w3c/epub-specs/spooky-base-urls/epub33/core/index.html#sec-container-iri) in the core spec, as they really only apply to out-of-container URLs.
We still need those in in the RS spec, to specify how reading systems must process non-conforming URLs.

iherman:
This is the translation into a PR of the equilibrium point in the discussion in #1912.

Fixes #1912

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1939.html" title="Last updated on Dec 7, 2021, 5:34 AM UTC (e90a01f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1939/8fc5fe4...e90a01f.html" title="Last updated on Dec 7, 2021, 5:34 AM UTC (e90a01f)">Diff</a>


wareid (approved):

rdeltour (commented):
It seems I misunderstood our equilibrium point 😅

I believe the algorithm should be normative, and the explanation ("more double-dot segments than needed") informative in a note.

I think we can tweak the algorithm so that it becomes 100% correct, even if not 100% explicit.

iherman:
Hm. My initial thoughts were that if we include the base element into the determination of the current context then we may get to the right algorithm but then I got the impression that there are some other details that evade me. I am happy to be stand corrected if we can turn the algo 100% with that.

However, we indeed seem to have misunderstood one another, though, I believe, what is there mostly reflect the views of @mattgarrish and @bduga...

iherman:
Note also that if we take `<base>` into account, then, if we want to be 100% precise, the concept of out-of-container URL strings becomes iffy. Indeed, a the `url` string, by itself, may not be out-of-container, it is _the combination of `url` and `<base>`_ that becomes out-of-container. No doubt this can also be spec-d, but it would really mean a complex algorithm in the spec, and I am not sure it is worth it...

rdeltour:
@iherman I was thinking something along these lines could work:

---
For any URL string _url_ found in the OCF Abstract Container, the following steps should return true:
1. Set the **container root URL** to `https://a.example.org/A/`.
2. Let _base_ be the [base URL](https://url.spec.whatwg.org/#concept-base-url) that must be used to parse _url_, if any, as defined by the context (document or environment) where _url_ is used, according to the **container root URL**.
3. Let _testURLRecord_ be the result of applying the [URL parser](https://url.spec.whatwg.org/#concept-url-parser) to _url_, with _base_.
4. Let _testURLStringA_ be the result of applying the [URL Serializer](https://url.spec.whatwg.org/#concept-url-serializer) to _testURLRecord_
5. Set the **container root URL** to `https://b.example.org/B/`.
6. Set _base_ to the [base URL](https://url.spec.whatwg.org/#concept-base-url) that must be used to parse _url_, if any, as defined by the context (document or environment) where _url_ is used, according to the **container root URL**.
7. Set _testURLRecord_ to the result of applying the [URL parser](https://url.spec.whatwg.org/#concept-url-parser) to _url_, with _base_.
8. Let _testURLStringB_ be the result of applying the [URL Serializer](https://url.spec.whatwg.org/#concept-url-serializer) to _testURLRecord_
9. If _testURLStringA_ does not [start with](https://infra.spec.whatwg.org/#byte-sequence-starts-with) `https://a.example.org/` or _testURLStringB_ does not [start with](https://infra.spec.whatwg.org/#byte-sequence-starts-with) `https://b.example.org/`, return true.
10. If _testURLStringA_ [starts with](https://infra.spec.whatwg.org/#byte-sequence-starts-with) `https://a.example.org/A/` and _testURLStringB_ [starts with](https://infra.spec.whatwg.org/#byte-sequence-starts-with) `https://b.example.org/B/`, return true.
11. Return false.
---

To summarize, the idea is to test any URL string (relative or not, it doesn't matter), with two test container root URL, parse it with their host-defined base, and inspect the two resulting strings:
- If any of the result does not share the test URL host, it means the test URL was absolute, or the base was absolute and outside the container (e.g. `base` element with `href` set to `https://w3.org`).
- Otherwise, both results should start with the respective test root URL (the presence of the first test path segment —`A` or `B`—, indicates that the URL doesn't leak outside the container).

I believe this algorithm fully achieves what we want to test.
The only part subject to some level of interpretation is "the base URL that must be used to parse _url_, if any, as defined by the context (document or environment) where _url_ is used". The base URL concept being dependent on the host language, I don't see how to phrase it otherwise. An companion informative note could explain that bit.

What do you think?

rdeltour:
Also, note that contrary to what I initially suggested in #1912, the above does not touch the URL standard definitions (of what is a **relative-URL string**, specifically).

It just adds a SHOULD criteria to all URL strings, in addition to the individual syntactic criteria (MUST) that we already define elsewhere.

iherman:
@rdeltour yes, that works I believe.

That will affect the preceding paragraph that should, probably, disappear or be merged with another note that gives some background of what is to be achieved here. I can figure out something and we can take it from there.

it is better than what is currently in the PR; it does look a bit funny to have an algorithm in a note like that.

iherman:
Ok, I have made the changes.

I know that, from a content point of view, the constraints on the container root URL are not strictly necessary, and we could rely on the Reading System spec only. However, I believe it is better to keep them here; it gives a complete picture for the author and may be an important aspect for someone wanting to add some more complex scripts to the publication. So I kept this for now.

I also used a trick that @mattgarrish and I used in the Publication Manifest spec, namely to use the `<details>` element to add explanation to some of the not-so-obvious algorithmic steps. I think this is better than adding a separate, big note that may disturb the flow of the spec. I have added only a smaller note following the algorithmic part.

I have, finally, moved the example and also added the example in the original problem setting of @rdeltour to make it clearer what happens with out-of-container cases.

rdeltour:
Sounds good! I’ll review as soon as possible. 👍

epub33/core/index.html
```diff
@@ -6082,9 +6082,90 @@ <h4>URLs in the OCF Abstract Container</h4>
 					</div>
 
 					<p> In the <a>OCF Abstract Container</a>, when a file uses a URL string to reference another file in
-						the container, the string MUST be a <a data-cite="url#path-relative-scheme-less-url-string"
-							>path-relative-scheme-less-URL string</a>, optionally followed by <code>U+0023 (#)</code>
-						and a <a data-cite="url#url-fragment-string">URL-fragment string</a>. </p>
+						the container, the string MUST be a
```
rdeltour: I think SHOULD is enough here (see also the related discussion with @mattgarrish in #1912)

epub33/core/index.html
```diff
@@ -6082,9 +6082,90 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+
+					<p>For a URL string <var>url</var> found in the <a>OCF Abstract Container</a>, the following steps MUST also return <var>true</var>:</p>
+
+					<ol class="algorithm" id="algo-out-of-container">
+						<li>
+							Set the <a>container root URL</a> to <code>https://a.example.org/A/</code>.
+						</li>
+
+						<li>
+							Let <var>base</var> be the <a data-cite="url#concept-base-url">base URL</a> that must be used to <a data-cite="url#concept-url-parser">parse</a> <var>url</var> as defined by the context (document or environment) where <var>url</var> is used, and according to the <a>container root URL</a>.
```
rdeltour: I wouldn't link "parse" to the URL standard here. We're not calling the algorithm, and it does not give more info on the base URL anyway.

epub33/core/index.html
```diff
@@ -6082,9 +6082,90 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+					<ol class="algorithm" id="algo-out-of-container">
+						<li>
+							Set the <a>container root URL</a> to <code>https://a.example.org/A/</code>.
+						</li>
+
+						<li>
+							Let <var>base</var> be the <a data-cite="url#concept-base-url">base URL</a> that must be used to <a data-cite="url#concept-url-parser">parse</a> <var>url</var> as defined by the context (document or environment) where <var>url</var> is used, and according to the <a>container root URL</a>. 
+							<details>
+								<summary>Explanation</summary>
+								<p class="note">In the case of a URL in the package document the <var>base</var> variable is set to the <a>container root URL</a>. In the case of a URL in a content document, the <a>content URL</a> of the enclosing document will be used. Furthermore, this step also takes into account a possible <a data-cite="html#document-base-url"><code>base</code></a> element in an XHTML content document, although the usage of that element is <a href="#sec-xhtml-deviations-base">discouraged</a>.</p>
```
rdeltour: > In the case of a URL in the package document the base variable is set to the container root URL.

s/container root URL/content URL of the Package Document/

Also, we could link to section "Parsing URLs in the Package Document" so that the reader can navigate to where it is exactly defined.

> In the case of a URL in a content document, the content URL of the enclosing document will be used. Furthermore, this step also takes into account a possible base element in an XHTML content document, although the usage of that element is discouraged.

Likewise, I would link to [where it is defined in HTML](https://html.spec.whatwg.org/#resolving-urls), and maybe rephrase to something like:

> In the case of a URL in an XHTML Content Document, the base URL used for parsing is defined by the [HTML standard](https://html.spec.whatwg.org/#resolving-urls). Typically, it will be the content URL of the content document (if the discouraged `base` element is not used).

epub33/core/index.html
```diff
@@ -6082,9 +6082,90 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+						<a data-cite="url#path-relative-scheme-less-url-string">path-relative-scheme-less-URL string</a>, 
+						optionally followed by <code>U+0023 (#)</code> and a 
+						<a data-cite="url#url-fragment-string">URL-fragment string</a>. 
+					</p>
+
+					<p>For a URL string <var>url</var> found in the <a>OCF Abstract Container</a>, the following steps MUST also return <var>true</var>:</p>
+
+					<ol class="algorithm" id="algo-out-of-container">
+						<li>
+							Set the <a>container root URL</a> to <code>https://a.example.org/A/</code>.
```
rdeltour, line 5997: Maybe here we could explain:
- why we use a test URL
- that this test URL willfully violates the required properties of container root URLs

I don't know…

epub33/core/index.html
```diff
@@ -6082,9 +6082,90 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+
+						<li>
+							Let <var>testURLStringA</var> be the result of applying the <a data-cite="url#concept-url-serializer">URL Serializer</a> to <var>testURLRecord</var>.
+						</li>
+
+						<li>
+							Set the <a>container root URL</a> to <code>https://b.example.org/B/</code>.
+							<details>
+								<summary>Explanation</summary>
+								<p class="note">The reasons to repeat the same steps twice with different, and artifical, settings of the container root URL is to avoid collision which may occur if the <var>url</var> string also includes <code>/A/</code>. Consider, for example, the case where <var>url</var> is of the form <code>"../../A/doc.xhtml</code>.</p>
```
rdeltour: - s/artifical/artificial/
- s/of the form//
- remove the opening quote in the URL

epub33/core/index.html
```diff
@@ -6082,9 +6082,90 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+							</details>
+						</li>
+
+						<li>
+							Let <var>base</var> be the <a data-cite="url#concept-base-url">base URL</a> that must be used to <a data-cite="url#concept-url-parser">parse</a> <var>url</var> as defined by the context (document or environment) where <var>url</var> is used, and according to the <a>container root URL</a>.
```
rdeltour: s/Let _base_ be/Set _base_ to/
(the _base_ variable has already been defined)

epub33/core/index.html
```diff
@@ -6082,9 +6082,90 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+								<p class="note">The reasons to repeat the same steps twice with different, and artifical, settings of the container root URL is to avoid collision which may occur if the <var>url</var> string also includes <code>/A/</code>. Consider, for example, the case where <var>url</var> is of the form <code>"../../A/doc.xhtml</code>.</p>
+							</details>
+						</li>
+
+						<li>
+							Let <var>base</var> be the <a data-cite="url#concept-base-url">base URL</a> that must be used to <a data-cite="url#concept-url-parser">parse</a> <var>url</var> as defined by the context (document or environment) where <var>url</var> is used, and according to the <a>container root URL</a>. 
+						</li>
+
+						<li>
+							Let <var>testURLRecord</var> be the result of applying the <a data-cite="url#concept-url-parser">URL parser</a> to <var>url</var>, with <var>base</var>.
```
rdeltour: s/Let _testURLRecord_ be/Set _testURLRecord_ to/

epub33/core/index.html
```diff
@@ -6082,9 +6082,90 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+
+						<li>
+							Let <var>testURLStringB</var> be the result of applying the <a data-cite="url#concept-url-serializer">URL Serializer</a> to <var>testURLRecord</var>.
+						</li>
+
+						<li>
+							If <var>testURLStringA</var> does not start with <code>https://a.example.org/</code> or <var>testURLStringB</var> does not start with <code>https://b.example.org/</code>, return <var>true</var>.
+							<details>
+								<summary>Explanation</summary>
+								<p class="note">If any of the result does not share the test URL host, it means the test URL, or the <code>base</code> element, was <em>absolute</em> and point outside the container. This is acceptable.</p>
```
rdeltour: I would replace:
> or the `base` element

by

> or its base URL (for instance, in HTML, if it is explicitly set with the `base` element)

epub33/core/index.html
```diff
@@ -6082,9 +6082,90 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+							</details>
+						</li>
+
+						<li>
+							If <var>testURLStringA</var> starts with <code>https://a.example.org/A/</code> and <var>testURLStringB</var> starts with <code>https://b.example.org/B/</code>, return <var>true</var>.
+							<details>
+								<summary>Explanation</summary>
+								<p class="note">The presence of the first test path segments (<code>A</code>, respectively <code>B</code>) indicate that the URL doesn't leak outside the container.</p>
```
rdeltour, line 6074: I'm not sure that note explains a lot (I know it comes from one of my comment 😅).

epub33/core/index.html
```diff
@@ -6082,9 +6082,90 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+
+						<li>Return <var>false</var>.</li>
+					</ol>
+
+					<div class="note">
+						<p>	
+							The properties of the <a>container root URL</a> are such that whatever the amount
+							of <a data-cite="url/#double-dot-path-segment">double-dot path segments</a> in a URL string (for
+							example, <code>../../../secret</code>), it is parsed to a <a>content URL</a> (and not "leak"
+							outside the container). This avoids potential run-time security issues. However, the additional
```
rdeltour: What if we moved this note before the algorithm, to introduce why we define this extra requirement (the algorithm)?

epub33/core/index.html
```diff
@@ -6101,14 +6182,26 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+							│   └── container.xml
+							└── EPUB
+							    └── content.xhtml
+						</pre>						
+
+						<p>
+							The URL `../../../../EPUB/secret.xhtml`, appearing in `content.xhtml`, is out of container
+							and would be flagged as non acceptable by a checker tool based on the relevant <a href="algo-out-of-container">algorithm</a>. Furthermore, following the 
+							<a href="#confreq-root-url">constraints</a> on the <a>container root URL</a> the Reading System parses
+							this relative URL into <a>content URL</a> with a path `EPUB/secret.xhtml`.
```
rdeltour: I would present this the other way around, something like:

> A URL `../../../../EPUB/secret.xhtml` appearing in content.xhtml would be parsed by a Reading System into a content URL with a path `EPUB/secret.xhtml`, following the constraints on the container root URL. However, as the URL could be perceived as one of a resource outside the container, and create interoperability issues, it would be reported as a warning by a checker tool.

rdeltour (commented):
Looks good! 👍
I prefer this version to the previous in-note algorithm 😊. Also, the inline explanations are helpful.

See comment details for some editorial suggestions.

lordt4ever:
Sure from the UK government and bank of England and conduct authority cg.

epub33/core/index.html
iherman: I agree it reads better, but then 

> the test URL

must also be changed to

> that <var>url</var>

The 'test URL' can be misread as being <var>testURLStringA</var> and that is not the point...

epub33/core/index.html
iherman: It works for me; not sure what I would put in its place (unless writing a very long text, and I do not think that is the idea here).

epub33/core/index.html
iherman: Not sure. I am a bit bothered if there is a note preceding what is being formally defined; it somehow breaks the (normative) narrative for me. I guess it is a bit of a taste issue; why don't I leave it as is, and see how others react?

rdeltour: Sure, works for me!

iherman: Thinking about it further, and also related to your note on adding an explanation to the start of the algorithm, I agree and have moved the first paragraph of the note.

epub33/core/index.html
iherman: Well... it is not obvious to write it down succinctly. I have made a try, I am sure you will help to formulate it better!

epub33/core/index.html
```diff
@@ -6082,9 +6082,112 @@ <h4>URLs in the OCF Abstract Container</h4>
 					</div>
 
 					<p> In the <a>OCF Abstract Container</a>, when a file uses a URL string to reference another file in
-						the container, the string MUST be a <a data-cite="url#path-relative-scheme-less-url-string"
-							>path-relative-scheme-less-URL string</a>, optionally followed by <code>U+0023 (#)</code>
-						and a <a data-cite="url#url-fragment-string">URL-fragment string</a>. </p>
+						the container, the string SHOULD be a
```
rdeltour: It wasn't here that s/MUST/SHOULD/.
Here it should still be a MUST 😊

epub33/core/index.html
```diff
@@ -6082,9 +6082,112 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+					<p class="note">	
+						The properties of the <a>container root URL</a> are such that whatever the amount
+						of <a data-cite="url/#double-dot-path-segment">double-dot path segments</a> in a URL string (for
+						example, <code>../../../secret</code>), it is parsed to a <a>content URL</a> (and not "leak"
+						outside the container). This avoids potential run-time security issues. 
+						Furthermore, the additional constraint and <a href="#algo-out-of-container">algorithm</a> below 
+						ensure that such potentially problematic URLs can also be detected when checking the EPUB Document.
+					</p>
+
+					<p>For a URL string <var>url</var> found in the <a>OCF Abstract Container</a>, the following steps MUST also return <var>true</var>:</p>
```
rdeltour: s/MUST/SHOULD/

epub33/core/index.html
```diff
@@ -6082,9 +6082,112 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+									The goal of the algorithm is to detect whether <var>url</var> could be seen as
+									"leaking" outside the container. 
+									To do that, the official <a data-cite="url#concept-url-parser">URL parsing algorithm</a> 
+									is used with an artificial root URL; the detection of the "leak" is done by 
+									comparing the result of the parsing with the presence of the first test path 
+									segment (<code>A</code>). 
+									(Note that the artificial container root URL wilfully violates, for the purpose of this
+									algorithm, the <a href="#confreq-root-url">required properties</a> by using 
+									that first test path segment.)     
+								</p>
```
rdeltour: Sounds good!

Just to nitpick: s/official/standard/ 😊

epub33/core/index.html
```diff
@@ -6082,9 +6082,112 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+							</details>
+						</li>
+
+						<li>
+							Let <var>base</var> be the <a data-cite="url#concept-base-url">base URL</a> that must be used to parse <var>url</var> as defined by the context (document or environment) where <var>url</var> is used, and according to the <a>content URL</a> of the <a>Package Document</a> (see <a href="#sec-parse-package-urls"></a>). 
+							<details>
+								<summary>Explanation</summary>
+								<p class="note">
+									In the case of a URL in the package document the <var>base</var> variable is set 
+									to the <a>container root URL</a>. In the case of a URL in an XHTML Content Document, the
```
rdeltour: Copying a previous comment, that was left out:

s/container root URL/content URL of the Package Document/

Also, we could link to section "Parsing URLs in the Package Document" so that the reader can navigate to where it is exactly defined.

epub33/core/index.html
iherman: Ok, I misunderstood, that is where you pointed at... so the algorithm is a SHOULD and not a MUST, right?

I must admit when I wrote the text I hesitated, and I then went for a MUST...

epub33/core/index.html
iherman: On the extra link: it is already there in the 'standard' part of the text, I did not want to make the note too heavy.

rdeltour (approved):

epub33/core/index.html
```diff
@@ -6082,9 +6082,112 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+					<p class="note">	
+						The properties of the <a>container root URL</a> are such that whatever the amount
+						of <a data-cite="url/#double-dot-path-segment">double-dot path segments</a> in a URL string (for
+						example, <code>../../../secret</code>), it is parsed to a <a>content URL</a> (and not "leak"
+						outside the container). This avoids potential run-time security issues. 
+						Furthermore, the additional constraint and <a href="#algo-out-of-container">algorithm</a> below 
+						ensure that such potentially problematic URLs can also be detected when checking the EPUB Document.
+					</p>
+
+					<p>To validate a URL string <var>url</var> found in the <a>OCF Abstract Container</a>, the following steps SHOULD return <var>true</var>:</p>
```
rdeltour: I'm not sure I like "to validate a URL".

There's more to validation than that single algorithm. A URL is valid if and only if it parses without validation errors, **and** (in EPUB only) it returns true to for this algorithm.

If we say "to validate a URL", it almost sounds like we only consider the EPUB-specific part.

Also, taken as a whole,  "to validate a URL, the following steps SHOULD return true" sound odd to me. Like the "SHOULD" only applies if your objective is to validate the URL. 🤔

mattgarrish: Ya, this is why I didn't think it was exactly right when I mentioned it in the issue. The best I could get was that this validates any URL is valid to use in an EPUB, not that the URL is valid to the URL standard.

You're not parsing a URL exactly, or strictly confirming that URLs resolve within the container, so you've got me kind of stumped what would be a more apt descriptor for it.

mattgarrish: Why does this algorithm need to be used with absolute URLs, out of curiosity? Are you just trying to make it input agnostic?

Otherwise, why not restrict it to checking relative paths? You could pin a stronger definition name on it and make it a MUST that it return true. If a URL fails, whether that constitutes an error or warning should be a separate statement from how to check conformance.

Again, just for example, that earlier paragraph about relative paths could additionally say : "Path-relative-scheme-less-URL strings SHOULD resolve to resources in the OCF Container. To verify, the result of running **resolve the location in the OCF Container** with a path-relative-scheme-less-URL string MUST be true."

And the algorithm becomes: "To **resolve the location in the OCF Container** of give a path-relative-scheme-less-URL string _url_, perform the following steps. They return a boolean value."

iherman: I actually like your proposal, @mattgarrish, but I will wait for @rdeltour's comment before I foolishly change it again (I should not have done it).

As for:

> Why does this algorithm need to be used with absolute URLs, out of curiosity? Are you just trying to make it input agnostic?

The problem is that step 9 may lead to true even if _url_ is relative, namely if the value of `base` is in fact absolute. This is a combination we must account for but this then results in the algorithm being agnostic to the relative or absolute nature of _url_.

rdeltour: > Why does this algorithm need to be used with absolute URLs, out of curiosity? Are you just trying to make it input agnostic?

what @iherman said.

The algorithm does **need** to work with absolute URLs **as input**; but after parsing it does need to handle remote (i.e. absolute) URLs as well as container URLs. So it has to work with absolute-URL strings, as `http://example.com/my-remote-url` isn't different than `./my-remote-url` with a `http://example.com` base.

> Otherwise, why not restrict it to checking relative paths?
> You could pin a stronger definition name on it and make it a MUST that it return true.

Restricting the algorithm to relative paths or not is strictly equivalent.

The algorithm is a blanket pass for all absolute URLs. I don't understand why restricting it to relative URL strings makes it stronger.

> If a URL fails, whether that constitutes an error or warning should be a separate statement from how to check conformance.

Not convinced, but I may not understand.

> Again, just for example, that earlier paragraph about relative paths could additionally say : "Path-relative-scheme-less-URL strings SHOULD resolve to resources in the OCF Container. To verify, the result of running resolve the location in the OCF Container with a path-relative-scheme-less-URL string MUST be true."

I'm not convinced this is an improvement over the current status of the PR.

For one thing, "SHOULD resolve to resources in the OCF Container" does not work well with the `base` element (which I know is discouraged, but still).  
Also, the algorithm is only syntactic, it does not check if the URL resolves to a resource.

> And the algorithm becomes: "To resolve the location in the OCF Container of give a path-relative-scheme-less-URL string url, perform the following steps. They return a boolean value."

The above sounds odd to me. I would expect that "resolving a location" returns a location, not a boolean. And again, the algorithm does not resolution, it only checks that double-dot segments are balanced with non-double-dot segments in a URL path string.

All that said, I do not disagree with your original point that we can do better at introducing the algorithm. I thought putting the note above it would be enough, but if you skip the note it makes less sense.

mattgarrish: > The problem is that step 9 may lead to true even if url is relative

Right, I didn't mean that step 9 gets thrown out. I was just trying to figure out what single goal there is that would unite the algorithm. Simplify the input and see if there's a way of calling the algorithm rather than it just existing apart from anything else.

> does not work well with the base element (which I know is discouraged, but still)

Right, I understand. And, yes, I wasn't thinking of that outcome in the name, but I'm not intending the names as things to use at this point. I'm just trying to understand how to make a common theme, if you want, out of what this algorithm is checking.

What bothers me a lot with the current wording is that there is a "should" with no normative reason attached to it. How would epubcheck report this: [Warning] The url is not true? [Warning] The url does not validate ... to?

Looking at it some more, I'd probably recommend dropping the note formatting on the paragraph above. You could likely take that prose and the algorithm and make them a subsection of 6.1.5 for clearer separation. I'm still not sure what a good title would be -- "Checking for Invalid Paths"?

Maybe then the algorithm could have a name like "To **check a path does not leak outside the OCF Container** ..." ? I still wouldn't put a normative requirement into an algorithm introduction, though. Something else should call it with the requirement (ideally the algorithm would state to issue a warning, but I don't think we want to open that can of worms).

rdeltour: What if we unite the algorithm with the definition of the class of URL strings valid in EPUB?

> A string is a **valid absolute container URL string** if it is an absolute-URL-with-fragment string.
>
> A string is a **valid relative container URL string** if it is a path-relative-scheme-less-URL string, optionally followed by U+0023 (#) and a URL-fragment string, and if these steps return true:
>  * [insert algorithm here]
>
> In the OCF Abstract Container, any URL string MUST be a valid absolute container URL string or a valid relative container URL string.

mattgarrish: That's making a lot more sense to me! 😄

iherman: That makes a lot of sense to me as well, with one remark: why introducing the **valid absolute container URL string** term? This is just an alias, as far as I can see, of [absolute-URL-with-fragment string](https://url.spec.whatwg.org/#absolute-url-with-fragment-string). 

Besides, the name suggests that the URL is such that its "domain" is the same as the content root URL (ie, more normatively, the URL's scheme, host and port are identical to the content root URL's) but I do not think we want to say that, because we use this for any URL, and that should include external URLs, too.

So, shouldn't the algorithm say:

>
> A string is a **valid relative container URL string** if it is a path-relative-scheme-less-URL string, optionally followed by U+0023 (#) and a URL-fragment string, and if these steps return true:
>  * [insert algorithm here]
>
> In the OCF Abstract Container, any URL string MUST be a [absolute-URL-with-fragment string](https://url.spec.whatwg.org/#absolute-url-with-fragment-string) or a valid relative container URL string.

Or do I miss something?

rdeltour: @iherman sure, that works too 😊

Just so we're clear: one drawback of this approach (not @iherman’s last tweak, but the combined definition generally) is that we lose the prose that said "when a file uses a URL string to reference another file in the container, the string MUST be".
So the intra-container cross-referencing aspect is less obvious.

Also, a side comment not specific to the current PR: compared to what we had in EPUB 3.2, we removed the requirement that relative intra-container URLs MUST resolve. Should we create a new issue to discuss that?

iherman: (Just to save some time for the others; I presume, @rdeltour, you refer to the last paragraph in https://www.w3.org/publishing/epub32/epub-ocf.html#sec-container-iri). The sentence is also around in an earlier version of the current spec: https://www.w3.org/TR/2021/WD-epub-33-20211014/#sec-container-iri

rdeltour: @iherman yes! (I should have been more explicit, thanks for the pointers 😊)

iherman: As for the last comment:

> Also, a side comment not specific to the current PR: compared to what we had in EPUB 3.2, we removed the requirement that relative intra-container URLs MUST resolve. Should we create a new issue to discuss that?

You are right and I would consider it as our bug, insofar as in our editing (ie, #1898) we made a mistake; this is a "stuff happens" category.

However, would it be easy expand a little bit on what we have and add this? E.g.

> In the OCF Abstract Container, any URL string MUST be a absolute-URL-with-fragment string or a valid relative container URL string. If it is the latter, the result of parsing it with the container root URL as base MUST identify a resource within the OCF Abstract Container.

rdeltour: > You are right and I would consider it as our bug, insofar as in our editing (ie, #1898) we made a mistake; this is a "stuff happens" category.

Yeah, I was aware from the beginning that #1898 removed that. It's two separate problems IMO: #1898 and this one #1939 are more syntactic (can be checked without full knowledge of the OCF content), and the resolution issue is about actual resources (needs knowledge of all the OCF resources).
That point was mentioned here and there since then (by @mattgarrish and me), but always as side comments 😊.
 
> However, would it be easy expand a little bit on what we have and add this? E.g.
> 
> > (…) If it is the latter, the result of parsing it with the container root URL as base MUST identify a resource within the OCF Abstract Container.

Unfortunately this doesn't work well with explicit remote bases.

With the difference in nature mentioned above, I would prefer —if we want to keep that requirement— to make a separate statement.

All that said, I'm not entirely convinced this requirement is important at the OCF level. For the Package Document it probably make sense to require that relative URLs resolve. But for content documents, less so? For instance if I have an `img` with a missing image, it won't break the EPUB or its interoperability. A MUST may be too strong there. A SHOULD is probably enough.
For other OCF resources, a non-resolving URL may not even be an issue at all?

iherman: > Unfortunately this doesn't work well with explicit remote bases.

Oh sh...t, I always forget about that one...

> All that said, I'm not entirely convinced this requirement is important at the OCF level. For the Package Document it probably make sense to require that relative URLs resolve. But for content documents, less so? For instance if I have an img with a missing image, it won't break the EPUB or its interoperability. A MUST may be too strong there. A SHOULD is probably enough.
For other OCF resources, a non-resolving URL may not even be an issue at all?

O.k. You just convinced me that this is not something we should handle here, and it should be the topic of a separate issue.

I will go ahead and do the changes we agreed upon (ie, https://github.com/w3c/epub-specs/pull/1939#discussion_r761814432) with the hope that we all agree, and we can merge this PR. Would you raise a separate issue and see if we can agree to simply leave the document as is or add that statement with a _SHOULD_ (I agree with you that a _MUST_ is too strong).

iherman: I have updated the document, see  latest commit. 

Cc @rdeltour @mattgarrish may we merge?

epub33/core/index.html
```diff
@@ -5984,10 +5984,119 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+					</p>
+
+					<div class="note">
+						<p>	
+							The properties of the <a>container root URL</a> are such that whatever the amount
+							of <a data-cite="url#double-dot-path-segment">double-dot path segments</a> in a URL string (for
+							example, <code>../../../secret</code>), it is parsed to a <a>content URL</a> (and not "leak"
+							outside the container). This avoids potential run-time security issues. 
+							The additional constraint on <a>valid-relative-container-URL-strings</a>  
+							ensures that such potentially problematic URLs can also be detected when checking the EPUB Document.
```
mattgarrish: EPUB Document looks like it should be EPUB Publication here.

This is exclusive of meta-inf files, correct, or does it handle the difference in where the paths are resolved relative to?

It might need to be "OCF Abstract Container" if it applies to all file types, as we don't have a word that. The meta-inf files sit above a publication.

iherman: I have changed for Publication.

iherman: You are right about the META-INF files, though. I would not worry about it too much in the Note (that does not have to be 100% precise). But I have the impression that the algorithm may not be 100% precise either for META-INF: step (2) (and (6)) seems to work only for content document or package document URL-s. The META-INF directory is not mentioned... 

I will think about it tomorrow, it is too late now for me... (unless someone beats me into how to add that to the algorithm without making it horribly complicated...)

iherman: Actually... I could argue that, in step 2/6

> Let base be the base URL that must be used to parse url as defined by the context (document or environment) where url is used

covers the META-INF case: the context  specifies that the base URL must be the container Root URL, per section 6.1.6.2. It is a bit subtle, and may be worth adding to explanation note. But normative text may just be fine as is...

rdeltour: @iherman yes, exactly 👍 the algorithm does (intend to) cover the META-INF files.

As for EPUB Publication vs. OCF Abstract Container (as raised by @mattgarrish), I believe EPUB Publication is more readable in the note. The normative statement, however, does say "OCF Abstract Container", which is theoretically more encompassing. In short, we're all good 😊.

epub33/core/index.html
```diff
@@ -5984,10 +5984,119 @@ <h4>URLs in the OCF Abstract Container</h4>
 
 					</div>
 
-					<p> In the <a>OCF Abstract Container</a>, when a file uses a URL string to reference another file in
-						the container, the string MUST be a <a data-cite="url#path-relative-scheme-less-url-string"
-							>path-relative-scheme-less-URL string</a>, optionally followed by <code>U+0023 (#)</code>
-						and a <a data-cite="url#url-fragment-string">URL-fragment string</a>. </p>
+					<p>
+						A string is a <dfn id="dfn-valid-relative-container-url-string">valid-relative-container-URL-string</dfn> if it is
```
rdeltour: super-nitpicking-friday-night comment: 
  * if we want to be consistent with the URL standard way of naming things, no last hyphen between "URL" and "string".
  * maybe naming it "relative-container-URL-with-fragment string" would be more symmetric with "absolute-URL-with-fragment string"?

more important: we need to define the  input variable _url_ of the algorithm. s/A string/A string _url_/ should work.

epub33/core/index.html
```diff
@@ -6086,11 +6075,28 @@ <h4>URLs in the OCF Abstract Container</h4>
 						<li>Return <var>false</var>.</li>
 					</ol>
 
-					<p class="note">
-						For better interoperability with non-conforming or legacy Reading Systems and toolchains, 
-						EPUB Creators should not use more <a data-cite="url/#double-dot-path-segment">double-dot path segments</a> 
-						than needed to reach the target container file.
-					</p>							
+					<p>
+						In the <a>OCF Abstract Container</a>, any URL string MUST be an <a data-cite="url#absolute-url-with-fragment-string">absolute-URL-with-fragment-string</a> or a <a>valid-relative-container-URL-string</a>.
```
rdeltour: same nitpick here: remove the last hyphen before "string"

epub33/core/index.html
mattgarrish: > the algorithm does (intend to) cover the META-INF files

Okay, thanks for clarifying that. I had to run out when I posted this and I couldn't tell from the markup diff which was the case so wasn't sure exactly which to suggest. Publication is fine with me, as it is just a note after all.

epub33/core/index.html
iherman: yes, yes, and yes :-)

epub33/core/index.html
iherman: yes, module the change of the reference to also include "with-fragment"

iherman:
@mattgarrish @dauwhe I plan to merge this PR this (Monday) evening or tomorrow morning; I have the impression that we have now converged to an acceptable consensus (we can always come back to some details in separate issues, if we want). Any objections?

dauwhe:
Trying to understand the algorithm, using Example 53.

1. we set container root url = `https://a.example.org/A/`
2. The path to the XHTML content file is` EPUB/content.xhtml`
3. If we use the URL Parser to parse the path `EPUB/content.xhtml` with the base URL `https://a.example.org/A/` we get `https://a.example.org/A/EPUB/content.xhtml`
4. If we then look at the suspect URL In the content doc `../../../../EPUB/secret.xhtml`, with the content doc URL as the base (`https://a.example.org/A/EPUB/content.xhtml`) we get `https://a.example.org/EPUB/secret.xhtml`

5. We see that our result `https://a.example.org/EPUB/secret.xhtml` DOES NOT contain the `/A/` segment, which means that there is potentially a leak outside the container. 

Is this correct?

rdeltour:
> Is this correct?

yes, exactly 👍

PR #1945, opened 2021-11-30
Publish IDPF epub type to aria roles authoring guide

mattgarrish:
The publishing CG's accessibility task force discussed this document in its call today and decided the guide better belongs in the WG because it is specific to EPUB 3.

Opening this PR to consider publishing it in this group.

### DIFF
--- a/epub33/epub-aria-authoring/index.html
+++ b/epub33/epub-aria-authoring/index.html
@@ -1,0 +1,1286 @@
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en-US" xml:lang="en-US">
+	<head>
+		<meta charset="utf-8" />
+		<title>EPUB Type to ARIA Role Authoring Guide 1.1</title>
+		<script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove" defer="defer"></script>
+		<script class="remove">
+			var respecConfig = {
+                group: "epub",
+                wgPublicList: "public-epub-wg",
+                specStatus: "ED",
+				noRecTrack: true,
+                shortName: "epub-aria-authoring",
+				edDraftURI: "https://w3c.github.io/epub-specs/epub33/epub-aria-authoring/",
+				copyrightStart: "2021",
+				editors:[ {
+					name: "Matt Garrish",
+					company: "DAISY Consortium",
+					companyURL: "https://daisy.org",
+					w3cid: 51655
+				}],
+				processVersion: 2020,
+				includePermalinks: true,
+				permalinkEdge:     true,
+				permalinkHide:     false,
+				diffTool:          "http://www.aptest.com/standards/htmldiff/htmldiff.pl",
+				github:			 {
+					repoURL: "https://github.com/w3c/publ-a11y",
+					branch: "main"
+				},
+				localBiblio: {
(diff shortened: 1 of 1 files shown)


mattgarrish:
Preview is available here: https://cdn.statically.io/gh/w3c/epub-specs/guides/epub-type-to-aria/epub33/epub-aria-authoring/index.html

iherman:
I think that this text does indeed belong to the WG, alongside with the other notes.

One food for thoughts: would it be a possibility to merge this with the SSV vocabulary document? After all, the two are very tightly coupled. (I am just wary of maintaining too many notes, but if there are good arguments to keep them separated then I won't object.)

mattgarrish:
I don't know if it works within the SSV, I'll let others weigh in on that, but it definitely would be useful to link from the SSV to the guide if we keep it separate, and potentially also from the note in epub:type section of the core spec.

It could work as an appendix, for example, but the one disadvantage of integrating it is a potential lack of visibility compared to being a separate document. It would also make referencing it kind of awkward.

iherman:
> I don't know if it works within the SSV, I'll let others weigh in on that, but it definitely would be useful to link from the SSV to the guide if we keep it separate, and potentially also from the note in epub:type section of the core spec.
> 

That is for sure.

> It could work as an appendix, for example, but the one disadvantage of integrating it is a potential lack of visibility compared to being a separate document. It would also make referencing it kind of awkward.

Let us see what others think. At the moment, we publish (or plan to publish) 11 documents. It is doable, but we should be mindful about that.

mattgarrish:
This is probably a completely out-to-lunch idea, so feel free to say so, but is there any precedent for publishing a "compendium" note?

By that, I mean, instead of having every guide as a separate document, could we publish a single multi-part guide as a note, where each chapter/document is a separate guide?

For example, the note is "EPUB Guides 1.0" and to start it has chapters for this, the accessibility faq, and maybe the schema.org authoring guide.

It would give us a lot of freedom to write guidance without a mass proliferation of documents... but I expect it's cheating w3c process, or at least making for a publishing headache.

iherman:
> This is probably a completely out-to-lunch idea, so feel free to say so, but is there any precedent for publishing a "compendium" note?

That notion does not exist, as you probably realize, in the W3C process. What is of course possible is to create one big note with main sections for the various specific guides. It simplifies the W3C publishing process aspect (although it may be too late for that now...) but I do not know whether it is better for the community...

mattgarrish:
> What is of course possible is to create one big note with main sections for the various specific guides

What's the difference is the root of my question? How is one big guide different from the same guide with its sections split into separate documents?

There are near approximations, like the wcag techniques.

> but I do not know whether it is better for the community

Well, ya, one big document is hideous for all kinds of reasons.

iherman:
> Well, ya, one big document is hideous for all kinds of reasons.

:-)

You know what? We may be too late with this discussion; let us see if this document is approved by the WG and run with it... We will survive yet another Note!

mattgarrish:
Should we merge this to close the PR and take up whether to publish it in a future call?

iherman:
> Should we merge this to close the PR and take up whether to publish it in a future call?

Sure. And let our fearless chairs decide whether to put it on the next WG agenda!

@dauwhe @wareid @shiestyle

PR #1947, opened 2021-11-30
Add processing requirements for spine overrides

### LINKED ISSUE #1941: Should the RS spec include something about layout overrides?
The content document, in [§5.2.1.1](https://www.w3.org/TR/epub-33/#layout-overrides) includes the possibility to overide the global value for fixed layout (prepaginated or reflowable). However, nothing is said about RS requirements about this: is it a _MUST_, a _SHOULD_, or, possibly, a _MAY_ for RS-s to honor this request?

@dauwhe @mattgarrish @wareid

mattgarrish:
Here's my best attempt at adding processing requirements for the various spine overrides. I tried breaking out each override, but that led to a lot of duplication with the primary values so I ended up making it a single statement to process the overrides per the existing requirements. There's also a requirement to ignore additional overrides for the same property type.

Fixes #1941 

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1941/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1941/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -545,8 +545,8 @@
 							property</a> [[EPUB-33]], Reading Systems MUST follow the requirements for the override's
 						global value when displaying that spine item.</p>
 
-					<p>For example, a spine item that contains the <a href="def-layout-pre-paginated"
-								><code>layout-prepaginated</code> override</a> [[EPUB-33]] is rendered following the
+					<p>For example, a spine item that contains the <a href="https://www.w3.org/TR/epub-33/#layout-overrides"
+								><code>layout-pre-paginated</code> override</a> [[EPUB-33]] is rendered following the
 						requirements of the global <a href="#def-layout-pre-paginated"><code>pre-paginated</code>
 							value</a>.</p>
 
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -941,6 +941,7 @@
 
 				<section id="layout">
 					<h4>The <code>rendition:layout</code> Property</h4>
+
 					<p> The default value <code>reflowable</code> MUST be assumed by EPUB Reading Systems as the global
 						value if no meta element carrying this property occurs in the <a
 							data-cite="epub-33#elemdef-opf-metadata"><code>metadata</code> section</a> section
@@ -965,6 +966,15 @@
 								rendering.</p>
 						</dd>
 					</dl>
+
+					<p>When one of the <a href="https://www.w3.org/TR/epub-33/#layout-overrides"
+								><code>rendition:layout</code> overrides</a> [[EPUB-33]] is specified in the
+							<code>properties</code> attribute on a spine <code>itemref</code> element, Reading Systems
+						MUST apply the same rendering logic to the referenced resource.</p>
+
+					<p>If more than one <code>rendition:layout</code> override is specified in the
+							<code>properties</code> attribute, Reading Systems MUST only process the first value listed
+						in the attribute.</p>
 				</section>
 
 				<section id="orientation">
@@ -976,6 +986,15 @@
 					<p>Reading Systems that support multiple orientations SHOULD convey the intended orientation to the
 						user unless the given value is <code>auto</code>. The means by which they convey the intent is
 						implementation specific.</p>
+
+					<p>When one of the <a href="https://www.w3.org/TR/epub-33/#orientation-overrides"
+								><code>rendition:orientation</code> overrides</a> [[EPUB-33]] is specified in the
+							<code>properties</code> attribute on a spine <code>itemref</code> element, Reading Systems
+						MUST apply the same rendering logic to the referenced resource.</p>
+
+					<p>If more than one <code>rendition:orientation</code> override is specified in the
+							<code>properties</code> attribute, Reading Systems MUST only process the first value listed
+						in the attribute.</p>
 				</section>
 
 				<section id="spread">
@@ -1011,6 +1030,15 @@
 								of a <a>Content Display Area</a> utilization optimization process.</p>
 						</dd>
 					</dl>
+
+					<p>When one of the <a href="https://www.w3.org/TR/epub-33/#spread-overrides"
+								><code>rendition:spread</code> overrides</a> [[EPUB-33]] is specified in the
+							<code>properties</code> attribute on a spine <code>itemref</code> element, Reading Systems
+						MUST apply the same rendering logic to the referenced resource.</p>
+
+					<p>If more than one <code>rendition:spread</code> override is specified in the
+							<code>properties</code> attribute, Reading Systems MUST only process the first override
+						listed in the attribute.</p>
 				</section>
 
 				<section id="page-spread">
@@ -1757,11 +1785,6 @@
 					occurs in the <a data-cite="epub-33#elemdef-opf-metadata"><code>metadata</code> section</a>
 					[[EPUB-33]]. Reading Systems MAY support only this default value.</p>
 
-				<p>If a Reading Systems supports the <a href="#layout"><code>rendition:layout</code></a> property, it
-					MUST ignore the <code>rendition:flow</code> property when it has been set on a spine item that also
-					specifies the <a href="#layout"><code>rendition:layout</code></a> value
-					<code>pre-paginated</code>.</p>
-
 				<p>The <code>rendition:flow</code> property values have the following processing requirements:</p>
 
 				<dl class="variablelist">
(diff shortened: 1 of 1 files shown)


iherman (approved):
Works for me (but I have no experience with FXL...)

wareid (approved):

epub33/rs/index.html
```diff
@@ -965,6 +966,15 @@ <h4>The <code>rendition:layout</code> Property</h4>
 								rendering.</p>
 						</dd>
 					</dl>
+
+					<p>When one of the <a href="https://www.w3.org/TR/epub-33/#layout-overrides"
+								><code>rendition:layout</code> overrides</a> [[EPUB-33]] is specified in the
+							<code>properties</code> attribute on a spine <code>itemref</code> element, Reading Systems
+						MUST apply the same rendering logic to the referenced resource.</p>
+
+					<p>If more than one <code>rendition:layout</code> override is specified in the
```
dauwhe: > When one of the `rendition:layout` overrides is specified in the `properties` attribute on a spine `itemref` element, Reading Systems MUST apply the same rendering logic to the referenced resource.

Same rendering logic as the override? Maybe:

> When a value of the `properties` attribute of the `itemref` element overrides the global value of `rendition:layout`, Reading Systems MUST use the rendering logic specified on that spine item.

mattgarrish: > When a value of the properties attribute of the itemref element overrides the global value of rendition:layout

I got bogged down trying wording like this because you then have to define how to parse the attribute. The attribute has a value that is the whitespace separated list of tokens. We might get away with "contains", but I still think we need to reference the spine overrides otherwise there's no context in this document.

> Reading Systems MUST use the rendering logic specified on that spine item.

The paragraph was intended to follow on from the rendering logic for each global value in the list that immediately precedes the paragraph (i.e., use the applicable logic above). We're still missing that connection.

This might still need some more tweaking, but:

> When the value of the `properties` attribute on an `itemref` element contains an [override of the global rendition:layout property](http://www.w3.org/TR/epub-33/#) [[EPUB-33]], Reading Systems MUST follow the rendering requirements for the override's global value when displaying that spine item (e.g., a spine item that specifies `layout-prepaginated` is rendered following the requirements of the global `pre-paginated` value).

(EDIT: tweaked the text after the must to get rid of "logic")

mattgarrish: I've just committed a fairly substantive change for this. It's annoyingly duplicative to have the same prose for every rendition property, so I've created a new section under the spine processing for handling overrides. I was going to put it in the FXL properties section, but it also applies to the flow property, so keeping it with the spine seemed most appropriate.

I also noticed that the orientation property didn't expand on its values. It just had a generic statement saying to use the orientation specified, but that didn't account for the auto value. I've added a dl list to define these like the other definitions.

epub33/rs/index.html
```diff
@@ -973,9 +998,28 @@ <h4>The <code>rendition:orientation</code> Property</h4>
...
+						</dd>
+
+						<dt id="def-orientation-portrait">portrait</dt>
+						<dd>
+							<p>Reading Systems that support multiple orientations SHOULD render the content in portrait
+								orientation.</p>
+						</dd>
+					</dl>
+
+					<p>The means by which they convey the intent is implementation specific.</p>
```
mattgarrish, line 1021: Does anyone want to hazard a guess at what this sentence means? "convey the intent" sounds like relaying on to the user what the author is asking for in the metadata, but that makes no sense.

I would assume it is trying to say how the reading system puts the content in the desired orientation is implementation specific, but how is that not just stating the obvious?

I'd like to delete this sentence, but want to make sure there isn't some hidden meaning I'm not grasping here.

dauwhe: The original sentence was

> Reading Systems that support multiple orientations should, unless the given value is 'auto', convey the intended orientation to the user. The means by which the intent is conveyed is implementation-specific.

Ah. I wonder if we were imagining that FXL reading systems would have user controls to choose orientation. And so the reading system would need some way to show the user which orientation that the content creator wanted. How to do that would be decidedly implementation-specific, as it is about the user interface of the reading system.

mattgarrish: > I wonder if we were imagining that FXL reading systems would have user controls to choose orientation.

But that's what's weird about this. The first sentence is the requirement to render it in the given orientation, or so I can only assume. "convey the intended orientation to the user" sounds like Swedish for "display in the specified orientation to the user", if I were to guess. 😉 

It always could be an artefact of an early draft of the FXL document for all I know, but why would only this property define such a thing?

In any case, since we don't define user control over FXL I assume that's also a case for dropping it.

epub33/rs/index.html
```diff
@@ -536,6 +536,23 @@ <h3>Spine</h3>
...
+				<section id="spine-overrides">
+					<h4>Spine Overrides</h4>
+
+					<p>When a spine <code>itemref</code> element's <code>properties</code> attribute contains an
+						override of a <a href="https://www.w3.org/TR/epub-33/#app-rendering-vocab">global rendering
+							property</a> [[EPUB-33]], Reading Systems MUST follow the requirements for the override's
+						global value when displaying that spine item.</p>
+
+					<p>For example, a spine item that contains the <a href="def-layout-pre-paginated"
+								><code>layout-prepaginated</code> override</a> [[EPUB-33]] is rendered following the
```
```suggestion
					<p>For example, a spine item that contains the <a href="https://www.w3.org/TR/epub-33/#layout-overrides"
								><code>layout-pre-paginated</code> override</a> [[EPUB-33]] is rendered following the
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-12-09-epub#section4) on 2021-12-09

List of resolutions:

- ***[Resolution No. 3](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-12-09-epub#resolution3):  Merge PR 1947 and close issue 1941.***

### 4. Should the RS spec include something about layout overrides? (issue epub-specs#1941)

_See github issue [epub-specs#1941](https://github.com/w3c/epub-specs/issues/1941)._




_See github pull request [epub-specs#1947](https://github.com/w3c/epub-specs/pull/1947)._




**Wendy Reid:** we have a statement in content spec that it is possible to override the global layout value, but nothing is said about it in the RS requirements.  
… and some of the associated behaviours are SHOULD in the RS spec.  

> *Wendy Reid:* See [Statement in the content document](https://www.w3.org/TR/epub-33/#layout-overrides).

**Wendy Reid:** these overrides theoretically make mixed modality content (fixed + reflow) possible, but its underspecified.  

**Dave Cramer:** i'm aware of some uses, e.g. in educational content, there might be reflow text on one side, and fixed layout reference material on the other side.  

**Brady Duga:** i'd hate to lose this feature to a lack of implementations. We don't support it, but I think we should.  

**Wendy Reid:** mgarrish has made a PR where he he proposes language.  

**Shinya Takami (高見真也):** in Japan we do use partially fixed layout.  

**Dave Cramer:** does that give us enough implementations to get this feature through?.  

**Wendy Reid:** we might also have a partial implementation of this - specifically for covers for reflow books.  

**Dave Cramer:** we do seem to have consensus that its necessary to have an RS conformance statement for these overrides.  
… the fact that its missing might just be an artefact from when we split out RS requirements from core.  

> **Proposed resolution: Merge PR 1947 and close issue 1941.** *(Wendy Reid)*

> *Brady Duga:* +1.

> *Shinya Takami (高見真也):* +1.

> *Wendy Reid:* +1.

> *Matthew Chan:* +1.

> *Murata Makoto:* +1.

> *Toshiaki Koike:* +1.

> *Dave Cramer:* +1.

> *Masakazu Kitahara:* +1.

> ***Resolution #3: Merge PR 1947 and close issue 1941.***

</details>

iherman:
@wareid the merge conflict is related to the assigned tests entries (ie, data-tests attribute values); can you settle those conflicts and, then, merge this PR? I was uncomfortable doing it myself...

mattgarrish:
I've moved the data-tests attribute to the orientation landscape definition to match what's been done elsewhere for these properties, but have a look and make sure that's where you want it.

PR #1943, opened 2021-11-30
Remove element restrictions from remote resource locations

mattgarrish:
Although there's a security advantage in saying where remote resources can be used, rather than relying solely on media type to allow them (which people periodically try to fake to get through epubcheck), this PR undoes our earlier change because I don't see a simple way to account for custom elements or dynamic writing of content generally.

Trying to be overly-restrictive always has a way of coming back and biting us, too, when some new technique comes along that we haven't accounted for, so this is probably the better long-term option.

I also moved up the notes that were after the examples, as it was a weird place for them (usually we put examples last). I also removed the note formatting from the one that says the rules apply to both CMT and foreign resource types as it is more of a normative statement than an informative one.

Fixes #1913

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1943.html" title="Last updated on Dec 3, 2021, 4:19 PM UTC (f615706)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1943/3afeba7...f615706.html" title="Last updated on Dec 3, 2021, 4:19 PM UTC (f615706)">Diff</a>

Issue #1913, opened 2021-11-16
Custom elements and remote resources


mattgarrish:
Now that we reference the WHATWG HTML spec, custom elements will become valid in EPUB 3.3, which are kind of the nightmare scenario for epubcheck since they aren't evaluated until documents load.

But putting aside the general inability to enforce rules on them, @rdeltour asked in https://github.com/w3c/epubcheck/issues/1288 about their impact on the remote resource rules.

While I believe the underlying requirements remain valid, it's not easy to check how the resources will be used not knowing what the custom element extends or translates to from epubcheck.

So the possible options are:

1. Scrap the explicit restrictions and rely on media type checking and whatever heuristics epubcheck has for determining what are legitimate remote resources
2. Add another bullet that allows any of the four types of exempted resources to be referenced from custom elements

I tend to lean toward the first option for pragmatic reasons -- it's probably what the second is going to force epubcheck to do anyway -- but if we want to keep the current specificity we could go with the second option.

dauwhe:
I'm a bit wary here. Using custom elements for audio and video, aside from any other issue, is likely to be bad for accessibility. We removed the capability of doing custom audio/video controls from EPUB. 

Is there an existing reading system where a custom element with some sort of functionality would actually work?

mattgarrish:
> Using custom elements for audio and video, aside from any other issue, is likely to be bad for accessibility.

I don't know that it would be, since underlying the custom element is either a literal audio/video element or a direct inheritance from those elements.

I haven't experimented with them much, but my understanding is that either the custom element is replaced by the contents of a template element (which would hold straight html), the HTML to render is written via script, or the audio/video element inherits whatever extra functionality the author has defined via the `is` attribute.

That's why there's a sort-of/almost way of finessing the requirement into custom elements, along the lines that when referenced from a custom element, the resource must still be used as defined in the restrictions. But there's no way to enforce that and I didn't have much luck trying to turn it into a sensible requirement.

Custom elements are well supported in browser cores, so I expect that means reading system support is a mess. 😄 

We can't really get rid of them now, though.

dauwhe:
> Custom elements are well supported in browser cores, so I expect that means reading system support is a mess. 😄

Partly true... Safari doesn't support extending native elements with `is`. 

I'm going to try to see if I can get a test working.

iherman:
> We can't really get rid of them now, though.

[...]

PR #1949, opened 2021-12-01
Alternative remote resource rule for custom elements

mattgarrish:
Opening this for comparison with #1943.

Instead of throwing out all the element restrictions, this is my best attempt to word an allowance for custom elements. It's not going to be easily checkable, but we wouldn't lose the basic restrictions just to accommodate custom elements.

I'm still not sure how much restricting elements helps at the end of the day. It prevents the simple abuse case of putting a wrong media type in the manifest to get external resources past epubcheck that you can then use in an iframe to render, but it doesn't stop people from scripting the dom to add stuff at run time or using custom elements other than the way we say.

This probably also requires a reading system "should not" restriction against using remote resources outside the referenced elements if adopted. That also seems a bit unlikely, as I expect reading systems either blanket allow or disallow the classes of remote resources regardless of what element they're used in.

But worth the exercise to compare, I suppose.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1949.html" title="Last updated on Dec 1, 2021, 1:05 PM UTC (7a01257)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1949/97bae7f...7a01257.html" title="Last updated on Dec 1, 2021, 1:05 PM UTC (7a01257)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1041,8 +1041,23 @@
 						</li>
 					</ul>
 
-					<p>EPUB Creators should locate these resources inside the EPUB Container whenever feasible to allow
+					<p>EPUB Creators MAY reference any of the Remote Reources listed above from a <a
+							href="https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements">custom
+							element</a> [[HTML]] provided the custom element extends a corresponding element (e.g.,
+							<code>HTMLAudioElement</code> for remote audio) or uses the Remote Resource in a
+						corresponding element in the DOM [[DOM]].</p>
+
+					<p>The rules in this section for Publication Resource locations apply regardless of whether the
+						given resource is a <a>Core Media Type Resource</a> or a <a>Foreign Resource</a>.</p>
+
+					<p>EPUB Creators should locate all resources inside the EPUB Container whenever possible to allow
 						users access to the entire presentation regardless of connectivity status.</p>
+
+					<div class="note">
+						<p>Refer to the <a href="#remote-resources"><code>remote-resources</code> property</a> for more
+							information on how to indicate that a <a>manifest</a>
+							<a href="#sec-item-elem"><code>item</code></a> references a <a>Remote Resource</a>.</p>
+					</div>
 
 					<aside class="example" title="Referencing a local resource">
 						<p>In this example, the audio file referenced from the [[HTML]] <a
@@ -1075,17 +1090,6 @@
    &lt;/body>
 &lt;/html></pre>
 					</aside>
-
-					<div class="note">
-						<p>The rules in this section for Publication Resource locations apply regardless of whether the
-							given resource is a <a>Core Media Type Resource</a> or a <a>Foreign Resource</a>.</p>
-					</div>
-
-					<div class="note">
-						<p>The inclusion of Remote Resources is indicated via the <a href="#remote-resources"
-									><code>remote-resources</code> property</a> on the <a>manifest</a>
-							<a href="#sec-item-elem"><code>item</code> element</a>.</p>
-					</div>
 				</section>
 
 				<section id="sec-data-urls">


iherman:
> Opening this for comparison with #1943.
> 
> Instead of throwing out all the element restrictions, this is my best attempt to word an allowance for custom elements. It's not going to be easily checkable, but we wouldn't lose the basic restrictions just to accommodate custom elements.
> 

Right. Checking this would require the checker to analyze the underlying javascript code, because that where the parent element of a custom element is decided. @rdeltour is the final arbiter here, but I wouldn't be surprised if ebupcheck was not equipped to do this.

On the other hand, adding an untestable and uncheckable restriction into the standard is weird. 

> 
> But worth the exercise to compare, I suppose.

+1 to that but, personally, I would favour #1943.

rdeltour:
I much prefer #1943, not only because this PR would be difficult to implement in EPUBCheck, but also because it is still overly restrictive IMO.

It adds complexity with no clear benefit (or have we identified cases where authors abuse remote resources and that would be solved by these restrictions?).

mattgarrish:
> because this PR would be difficult to implement in EPUBCheck

But you already have this partially implemented in epubcheck... 😉 

> or have we identified cases where authors abuse remote resources and that would be solved by these restrictions

Well, this was the regular suggestion on the IDPF forums for how to embed a youtube page. Fake the media type as video/mp4 (or whatever) and away you go. It only fell apart because some major reading systems used to disallow all external resources or their use in iframes, can't remember which anymore.

But it seems I can't do this now. Faking the media type still works in epubcheck, of course, but it is only completely silent if the URL is in the src attribute for an audio or video element. If I put it anywhere else, like in an iframe, an error is thrown.

mattgarrish:
Closing this as resolved on the working group call today.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-12-03-epub#section1) on 2021-12-03

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-12-03-epub#resolution1):  Merge PR 1943 and not 1949.***
- ***[Resolution No. 2](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-12-03-epub#resolution2):  Close issue 1913 once PR 1943 is merged..***

### 1. Custom elements and remote resources (issue epub-specs#1913)

_See github issue [epub-specs#1913](https://github.com/w3c/epub-specs/issues/1913)._




**Dave Cramer:** first few issues are about remote resources.  

_See github pull request [epub-specs#1949](https://github.com/w3c/epub-specs/pull/1949)._




_See github pull request [epub-specs#1943](https://github.com/w3c/epub-specs/pull/1943)._




**Dave Cramer:** all of our previous conclusions are complicated by custom elements in HTML.  
… mgarrish tries to phrase our existing requirements in a way that fits with custom elements.  
… but this makes things impossible for epubcheck, which won't know what is actually there until scripts have run.  

**Matt Garrish:** right, there are 2 PRs, [this one](https://github.com/w3c/epub-specs/pull/1949) tries to find language that still works with our past intent, the other PR more specifically address the existence of custom elements.  
… there is really little we can do to check for this, people will try to work around it. It will ultimately be down to RS to decide whether to even support custom elements.  

**Dave Cramer:** i made a test book for this, it worked in Apple and ADE, Kindle previewer failed because it doesn't support js, didn't work in Thorium.  

**Ivan Herman:** what you tested tells me that custom elements are here, it would be foolish of us to try to restrict them. They are part of the element that we have to deal with.  
… I'm in favor of the [first PR](https://github.com/w3c/epub-specs/pull/1943), which removes restrictions. The [second PR](https://github.com/w3c/epub-specs/pull/1949) which tries to keep restrictions is harder to check anyway.  

> *Ivan Herman:* +1 to matt.

**Matt Garrish:** the simpler solution might be just to say that RS should not render remote resources in iframes.  

**Dave Cramer:** worried that people would then work around that by using a custom element to effectively make an iframe.  

**Matt Garrish:** there's more awareness of how to handle that at the RS level than at checking, because you have the DOM.  

**Brady Duga:** when it comes to scripting, it kind of the wild west. It'll be impossible to test whether scripts are being used to do something malicious.  
… there's utility to having everything be in the package (interop, archival), and if people don't care about that then there's little we can do.  
… our main concern in making these exceptions for resources are for size.  
… should the restriction be based on size then.  

**Ivan Herman:** to be clear, custom elements mostly means scripting. As soon as RS allows scripting, any restriction on custom elements is meaningless. The real restriction is on scripting.  
… the [first PR](https://github.com/w3c/epub-specs/pull/1943) simplifies that.  

**Dave Cramer:** there are 2 types of custom elements. First inherents from existing HTML element (i.e. "is" attribute). This type doesn't work in webkit, and will probably be limited in epub world..  
… this is a can of worms that we can't really do much about.  

> **Proposed resolution: Merge PR 1943 and not 1949.** *(Wendy Reid)*

> *Dave Cramer:* +1.

> *Ivan Herman:* +1.

> *Ben Schroeter:* +1.

> *Brady Duga:* +1.

> *Wendy Reid:* +1.

> *Matthew Chan:* +1.

> *Toshiaki Koike:* +1.

> *Matt Garrish:* +1.

> *Masakazu Kitahara:* +1.

> *Charles LaPierre:* \.

> *Charles LaPierre:* ].

> *George Kerscher:* +1.

> *Dan Lazin:* +1.

> *Charles LaPierre:* +1.

> ***Resolution #1: Merge PR 1943 and not 1949.***

> *Bill Kasdorf:* +1.

> **Proposed resolution: Close issue 1913 once PR 1943 is merged..** *(Wendy Reid)*

> *Ben Schroeter:* +1.

> *Brady Duga:* +1.

> *Wendy Reid:* +1.

> *Matthew Chan:* +1.

> *Masakazu Kitahara:* +1.

> *Ivan Herman:* +1.

> *Bill Kasdorf:* +1.

> *Dave Cramer:* +1.

> *Dan Lazin:* +1.

> *Matt Garrish:* +1.

> *Charles LaPierre:* +1.

> ***Resolution #2: Close issue 1913 once PR 1943 is merged..***

**Dave Cramer:** I would warn everyone to be wary of custom elements, because it's easy to miss things related to a11y. HTML elements take care of a lot of that thinking for you.  

**Ivan Herman:** that's not really our fight, leave for WHAT WG.  

**Dan Lazin:** we use custom elements quite a bit in my day job, and there are more a11y friendly ways of doing it, but yes, beware.  

</details>

PR #1950, opened 2021-12-01
Make page-spread-center an alias for spread-none

### LINKED ISSUE #1929: Is page-spread-center a synthetic spread or not?
The definition of page-spread-center is:

> The rendition:page-spread-center property specifies the forced placement of a Content Document in a Synthetic Spread.

But this doesn't really tell you anything, not to mention it would be impossible to force a centered document into a two-page spread. Then after the property definitions comes this:

> The rendition:page-spread-center property indicates to override the synthetic spread mode and render a single viewport positioned at the center of the screen.

This is much clearer about what to render, but isn't this completely contradictory to the definition? Is there a spread if it overrides spread mode? How are we forcing the document into a spread at the same time we're overriding spreads?

If there's only a single centered viewport, that's not the same as merging the two halves of a spread to make one dual-page spread.

I'd hazard a guess the property would make more sense if it were called "rendition:page-center", since it seems to have little to do with spreads other than to override them.

I'm not sure how best to make sense of the current contradictions, though. All I can think of off the top of my head is to maybe make a separate subsection to explain "page-spread-center" so it's less confusing with the -left and -right properties. Dropping the current definition about forcing a document into a synthetic spread would also help.

mattgarrish:
The only concern I have now that I'm opening the PR is whether we should formally deprecate page-spread-center or if we should "discourage" its use. I don't know if anyone has used it that we'd affect by this change.

Otherwise, the PR itself removes all traces of page-spread-center except to deprecate and make it an alias and adds the centering prose to the rendition:spread none processing requirements.

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1929/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-1929/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1950.html" title="Last updated on Dec 10, 2021, 1:32 PM UTC (e472d62)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1950/a6e27d4...e472d62.html" title="Last updated on Dec 10, 2021, 1:32 PM UTC (e472d62)">Diff</a>

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -846,12 +846,11 @@
 					</li>
 
 					<li>
-
-						<p id="confreq-rs-scripted-origin"> It MUST assign a unique <a
-								data-cite="url#origin">origin</a> [[URL]], shared by all <a
-								data-cite="epub-33#sec-scripted-spine">spine-level scripts</a> of the
-							EPUB Publication. That <a data-cite="url#origin">origin</a> [[URL]] MUST
-							be <em>unique</em> for each user-specific instance of an EPUB Publication in a Reading System. </p>
+						<p id="confreq-rs-scripted-origin"> It MUST assign a unique <a data-cite="url#origin"
+							>origin</a> [[URL]], shared by all <a data-cite="epub-33#sec-scripted-spine">spine-level
+								scripts</a> of the EPUB Publication. That <a data-cite="url#origin">origin</a> [[URL]]
+							MUST be <em>unique</em> for each user-specific instance of an EPUB Publication in a Reading
+							System. </p>
 					</li>
 
 					<li>
@@ -991,7 +990,8 @@
 					<dl class="variablelist">
 						<dt id="def-spread-none">none</dt>
 						<dd>
-							<p>Reading Systems MUST NOT incorporate spine items in a Synthetic Spread.</p>
+							<p>Reading Systems MUST NOT incorporate spine items in a Synthetic Spread. Reading Systems
+								SHOULD create a single viewport positioned at the center of the screen.</p>
 						</dd>
 						<dt id="def-spread-landscape">landscape</dt>
 						<dd>
@@ -1021,28 +1021,22 @@
 					<p>The <span id="page-spread-left"><code>rendition:page-spread-left</code></span> property indicates
 						that the given spine item SHOULD be rendered in the left-hand slot in the spread, and <span
 							id="page-spread-right"><code>rendition:page-spread-right</code></span> that it SHOULD be
-						rendered in the right-hand slot. The <span id="page-spread-center"
-								><code>rendition:page-spread-center</code></span> property indicates that the synthetic
-						spread mode SHOULD be overridden and a single viewport rendered and positioned at the center of
-						the screen.</p>
+						rendered in the right-hand slot.</p>
 
-					<p>The <code>rendition:page-spread-left</code>, <code>rendition:page-spread-right</code>, and
-							<code>rendition:page-spread-center</code> properties apply to both pre-paginated and
-						reflowable content, and they only apply when the Reading System is creating Synthetic
-						Spreads.</p>
+					<p>Reading Systems MUST handle the <span id="page-spread-center"
+								><code>rendition:page-spread-center</code></span> property as though the EPUB Creator
+						has specified the <a href="#def-spread-none"><code>spread-none</code> property</a> on the
+							<code>itemref</code> element.</p>
+
+					<p>The <code>rendition:page-spread-left</code> and <code>rendition:page-spread-right</code>
+						properties apply to both pre-paginated and reflowable content, and they only apply when the
+						Reading System is creating Synthetic Spreads.</p>
 
 					<p>The <code id="page-break-precedence">rendition:page-spread-*</code> properties MUST take
 						precedence over whatever value of the <a
 							href="https://www.w3.org/TR/CSS2/page.html#propdef-page-break-before"
 								><code>page-break-before</code> property</a> [[CSSSnapshot]] has been set for an
 							<a>XHTML Content Document</a>.</p>
-
-					<div class="note">
-						<p>The presence of <code>rendition:page-spread-center</code> does not change the viewport
-							dimensions. It does not indicate that a Reading System must create a viewport with the size
-							of the whole spread. This is important so that the scale factor stays consistent between
-							regular and center-spread pages.</p>
-					</div>
 
 					<p id="page-layout-both">When a <a href="#def-layout-reflowable">reflowable</a> spine item follows a
 							<a href="#def-layout-pre-paginated">pre-paginated</a> one, the reflowable one SHOULD start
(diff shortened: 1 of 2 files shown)


epub33/rs/index.html
```diff
@@ -1021,29 +1021,23 @@ <h4>The <code>rendition:page-spread-*</code> Properties</h4>
...
-						rendered in the right-hand slot. The <span id="page-spread-center"
-								><code>rendition:page-spread-center</code></span> property indicates that the synthetic
-						spread mode SHOULD be overridden and a single viewport rendered and positioned at the center of
-						the screen.</p>
+						rendered in the right-hand slot.</p>
+
+					<p>Reading Systems MUST handle the <span id="page-spread-center"
+								><code>rendition:page-spread-center</code></span> property as though the EPUB Creator
+						has specified the <a href="#def-spread-none"><code>spread-none</code> property</a> on the
+							<code>itemref</code> element.</p>
```
iherman: Note that if `page-spread-center` is labeled as "deprecated" then the _MUST_ statement is to be made a bit more precise. Per the [definition of deprecated](https://w3c.github.io/epub-specs/epub33/core/#app-overview-unsupported) an RS _MAY_ implement it but not more. Ie, this statement should say something like "if the RS implements `page-spread-center` then it _MUST_…"

iherman (approved):

wareid (approved):

epub33/rs/index.html
mattgarrish: I'm not sure that works for aliasing, as it seems to suggest the reading system ignore it if it only supports spread-none even though it's an alias. Or that it can support the property independent of it being an alias (i.e., you support spread-none if you support rendition:page-spread-center, since only the former really exists).

Maybe "handle" is the wrong word here, as my intention was only to recognize it as the other property and do what that property requires (if supported).

What about reversing the relationship and dropping the handling, which flows from recognizing the value as an alias, so it becomes:

"Reading Systems that support the `spread-none` property MUST recognize the `rendition:page-spread-center` property as an alias for it, otherwise they MUST ignore the `rendition:page-spread-center` property."

iherman: Yep, @mattgarrish, the reformulation works for me, but... don't we have to rework the content document as well? Namely, is the statement in  [§5.2.4](https://www.w3.org/TR/epub-33/#page-spread) on `rendition:page-spread-center` still correct? What is the intention there? Shouldn't it mean the same as the definition of `spread-none` in the previous section?

Cc @wareid

mattgarrish: > Shouldn't it mean the same as the definition of spread-none in the previous section?

It should. With this update it now reads:

> rendition:page-spread-center [DEPRECATED]
> Use of the rendition:page-spread-center property is deprecated. EPUB Creators should use the spread-none property to disable spread behavior in Reading Systems.

I followed the pattern we used for rendition:spread portrait here. (But the deprecated label needs reformatting looking at how we applied it elsewhere.)

iherman: Shoot. I should have realized that you also edited the content document... I will be silent now!

shiestyle (changes requested):
I understand this change but most EPUB 3 files in Japan use 'page-spread-center' for 1st page as cover and we should avoid that EPUBCheck will recognize it as error and warning. 

So alias is acceptable but deprecated is not good label.

mattgarrish:
> most EPUB 3 files in Japan use 'page-spread-center' for 1st page as cover

That's what I was worried about. I'm perfectly fine with only making it an alias and not deprecating.

mattgarrish:
Last few commits make the following changes:

- removes the deprecated label from page-spread-center
- adds an explanation of what the property does to its definition - I also moved up the explanations of -left and -right into their definitions as it was odd to only say they are aliases and then in a following paragraph explain what they do
- revises the note about the aliases to remove the repeated statement of what -left and -right are aliases for. I also added a new paragraph explaining why -center exists as an alias (to simplify thinking about two-page and single layouts, but check this makes sense)
- restores the example of using page-spread-center but without the confusing text about not needing to apply spread-none since its role as an alias is now clear

shiestyle (approved):
Thanks, Matt. That's fine for me.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2021-12-09-epub#section2) on 2021-12-09

- no resolutions were taken

### 2. Is page-spread-center a synthetic spread or not? (issue epub-specs#1929)

_See github issue [epub-specs#1929](https://github.com/w3c/epub-specs/issues/1929)._




_See github pull request [epub-specs#1950](https://github.com/w3c/epub-specs/pull/1950)._




**Wendy Reid:** i just got through writing the tests for all the FXL rendition properties, which lead to a question about page-spread-center.  
… this is a spine property that is not meant to override spread behaviour, but that the page should be centered in the viewport.  
… but this doesn't seem to be what is happening in practice.  
… rather, the observed behaviour seems to be identical to spread none.  
… we've discussed in the issue, and mgarrish came up with the idea of making page-spread-center an alias of spread none.  
… and not to deprecate page-spread-center.  

**Brady Duga:** not sure why page-spread-center is confusing. Imagine you are doing a synthetic spread, but you want one specific page centered in the middle of the screen. That's how you would tell the RS to do that..  
… spread none sounds like you've turned off spreads completely.  

**Wendy Reid:** an additional level of oddness was finding out that you can put page spread properties on spine level items as well.  

**Brady Duga:** spread none and spread center mean different things, because spread none doesn't tell you where to put the page.  
… and yes, it is confusing that you can put this on a spine item (you can also alternate portrait and landscape on spine level items as another weird way of interacting with spreads).  

**Dave Cramer:** is there a concern that this could affect the relative size of pages as you go from spread, to non-spread, and back to spread?.  

**Brady Duga:** i can't see people really doing this.  
… one potential use case is if one of your content documents is already a spread, so you want to disable synthetic spreads for that one only.  

**Shinya Takami (高見真也):** in Japan we use spread-center for the cover pages in many many cases, so making spread-center a depreciated feature would not be acceptable.  
… aliasing it to spread none is okay.  

**Wendy Reid:** right, so we won't deprecate page-spread-center.  
… this might be the sort of thing we need to test out.  
… when the page appears in the center of the viewport, i think there is a change in the page size.  

**Dave Cramer:** i'm not comfortable deciding this until we have more info on whether we need to better define how a RS should position the viewport when you use page-spread-center or spread none as an override of a synthetic spread epub.  

**Wendy Reid:** agree that the FXL properties section could use some visual aides.  
… we can put this off for tonight.  

**Brady Duga:** if we end up keeping page spread then we need to fix this definition.  

**Wendy Reid:** yeah, definitions in that section are not very good.  

**Dave Cramer:** it was a long time ago that we wrote those, we were balancing against implementations that already existed at the time (amazon, apple).  

</details>

dauwhe (approved):

PR #1951, opened 2021-12-03
Removing the paragraph mentioned in 1918

### LINKED ISSUE #1918: Remove a paragraph after 1898
In a post #1898 world the first paragraph in [§3.4 Manifest in the RS document](https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-manifest) seems to be obsolete now. However, there is also a test assigned to it, and the corresponding test should be "reassigned"...

Cc @rdeltour @dlazin

iherman:
This to make the statement in [§3.4](https://www.w3.org/TR/epub-rs-33/#sec-pkg-doc-manifest) more precise.

Fixes #1918

See:

    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/editorial/remove-paragraph-1918/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/editorial/remove-paragraph-1918/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -452,7 +452,7 @@
 
 				<p id="confreq-rs-pkg-manifest-url" data-tests="#pkg-manifest-url,#pkg-relative-url">When an
 						<code>href</code> attribute contains a <a data-cite="url#relative-url-string">relative-URL
-						string</a>, Reading Systems MUST use the URL of the Package Document as the <a
+						string</a>, Reading Systems MUST use the <a data-cite="epub-33#dfn-content-url">content URL</a> of the Package Document as the <a
 						data-cite="url#concept-base-url">base URL</a> when <a data-cite="url#concept-url-parser"
 						>parsing</a> to a <a data-cite="url#concept-url">URL record</a> [[URL]].</p>
 
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -314,7 +314,7 @@
 		<section id="sec-package-doc">
 			<h2>Package Document Processing</h2>
 
-			<p id="confreq-rs-epub-pub" class="support" data-tests="#pkg-manifest-url">Reading Systems MUST process the
+			<p id="confreq-rs-epub-pub" class="support" data-tests="#pkg-manifest-url,#pkg-relative-url">Reading Systems MUST process the
 					<a data-cite="epub-33#sec-package-doc">Package Document</a> [[EPUB-33]].</p>
 
 			<section id="sec-pkg-doc-base-dir">
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -450,11 +450,6 @@
 			<section id="sec-pkg-doc-manifest">
 				<h3>Manifest</h3>
 
-				<p id="confreq-rs-pkg-manifest-url" data-tests="#pkg-manifest-url,#pkg-relative-url">When an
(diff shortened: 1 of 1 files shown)


rdeltour (changes requested):
See [my comment in #1918](https://github.com/w3c/epub-specs/issues/1918#issuecomment-985485675). 

I suggest, in order of preference:
1. to remove that paragraph altogether (simpler)
2. to make it informative, if we want to be redundant (a normative statement has more maintenance cost)

iherman:
I revised my reaction @rdeltour (after some further thoughts): I agree to remove that paragraph altogether. This is what is now in the PR

rdeltour (approved):

epub33/rs/index.html
```diff
@@ -450,11 +450,6 @@ <h4>Metadata</h4>
 			<section id="sec-pkg-doc-manifest">
 				<h3>Manifest</h3>
 
-				<p id="confreq-rs-pkg-manifest-url" data-tests="#pkg-manifest-url,#pkg-relative-url">When an
-						<code>href</code> attribute contains a <a data-cite="url#relative-url-string">relative-URL
-						string</a>, Reading Systems MUST use the URL of the Package Document as the <a
-						data-cite="url#concept-base-url">base URL</a> when <a data-cite="url#concept-url-parser"
-						>parsing</a> to a <a data-cite="url#concept-url">URL record</a> [[URL]].</p>
```
dlazin, line 453: Should https://w3c.github.io/epub-specs/epub33/rs/#issue-container-generatedID-12 now be MUST instead of must? It seems strange to me that there would be no normative requirement for RSes to parse base URLs correctly.

iherman: I believe the same argument holds as in https://github.com/w3c/epub-specs/issues/1918#issuecomment-985485675 of @rdeltour: we should try to avoid redundant MUST-s. The note refers to the relevant [MUST statements in the core spec](https://w3c.github.io/epub-specs/epub33/core/#sec-parsing-urls-metainf).

A certain level of redundancy between the RS and the Core are unavoidable: these two specs are very closely related. (One could argue that the whole [section](https://w3c.github.io/epub-specs/epub33/rs/#sec-container-iri) in the RS is redundant as the constraints at the start of the section are the same as in the [content spec](https://w3c.github.io/epub-specs/epub33/core/#sec-container-iri), but sometimes the "readability" of the spec is also at stake (the whole construction of the root URL is crucial to the spec), so it is all right in my view.

iherman:
Note: if we agree with what I propose in https://github.com/w3c/epub-tests/pull/96#discussion_r762531400, ie, to use [§3](https://www.w3.org/TR/epub-rs-33/#sec-pkg-doc-manifest) as an anchor for package tests that are directly bound to a specific statement in the RS spec, then I will also add, in this PR, an anchor ID.

iherman:
If there is an agreement around https://github.com/w3c/epub-tests/pull/96#discussion_r762531400, I would reassign the removed tests to https://www.w3.org/TR/epub-rs-33/#confreq-rs-epub-pub

mattgarrish (approved):

PR #1952, opened 2021-12-03
Add doc for gathering info on accessibility spec metadata usage

mattgarrish:
Question is where does this belong?

I've created a reports folder under /epub33 for this PR, since that seemed like the most logical place to store it here. But do we want to integrate it into the epub-tests repository somewhere? Other thoughts?

### DIFF
--- a/epub33/reports/a11y-properties-use.md
+++ b/epub33/reports/a11y-properties-use.md
@@ -1,0 +1,104 @@
+# EPUB Accessibility 1.1 Metadata Usage Report
+
+## Candidate Recommendation Exit Criteria
+
+The EPUB Working Group intends to exit the Candidate Recommendation stage and submit
+the EPUB Accessibility 1.1 specification for consideration as a W3C Proposed Recommendation
+after documenting implementation of each feature.</p>
+
+For this specification to advance to Proposed Recommendation, it has to be
+proven that metadata defined and required in this specification have sufficient usage by the
+target communities. This metadata falls into two categories:</p>
+
+1. required and recommended schema.org metadata for describing the accessibility
+of publications; and
+2. EPUB-defined metadata properties for reporting aspects of conformance
+
+Usage of these properties means that they are regularly included in the Package Document
+metadata for their EPUB Publications (as appropriate for each title).
+
+## Publisher Implementations
+
+### Schema.org discovery metadata
+
+The following table provides a list of publishers who have stated that they are currently using
+the schema.org discovery metadata properties in production or who are in the process of rolling
+out their implementations.
+
+<table>
+    <thead>
+        <tr>
+            <th>Role</th>
+            <th>[Publisher]</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <th>schema:accessibilityFeature</th>
+            <td></td>
+        </tr>
+        <tr>
+            <th>schema:accessibilityHazard</th>
+            <td></td>
+        </tr>
+        <tr>
+            <th>schema:accessibilitySummary</th>
+            <td></td>
+        </tr>
+        <tr>
+            <th>schema:accessMode</th>
+            <td></td>
+        </tr>
+        <tr>
+            <th>schema:accessModeSufficient</th>
+            <td></td>
+        </tr>
+    </tbody>
+</table>
+
+### Conformance metadata
+
+The following table provides a list of publishers who have stated that they are currently using
+the conformance metadata properties in production or who are in the process of rolling out their
+implementations.
+
+<table>
+    <thead>
+        <tr>
+            <th>Role</th>
+            <th>[Publisher]</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+        	<td>dcterms:conformsTo</td>
+        	<td></td>
+        </tr>
+        <tr>
+        	<td>a11y:certifiedBy</td>
+        	<td></td>
+        </tr>
+        <tr>
+        	<td>a11y:certifierCredential</td>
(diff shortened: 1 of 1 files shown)


iherman:
@dlazin is the boss on testing, but I think the logical place is indeed the epup-tests repository.

If we do agree on that, it should be under the `docs` directory: the github pages feature has been activated on that directory, as opposed to the full repo. Within that the only constraint is not to change the `index.html` for now (which is auto generated) although we would have to decide how we link this page from the `index.html` file (which, today, lists all the available tests).

mattgarrish:
That's the thing. We're not actually testing anything for the accessibility metadata, only gathering statements. That's why I'm not sure if it fits with the automated stuff you have going for the core specs. It could live here and simply be linked to from a list of all reports?

iherman:
> It could live here and simply be linked to from a list of all reports?

Of course they could, too, but remember that the main reason of having this document (and we may need another one) is the CR phase, and that is what epub-tests concentrates on.

But... this is almost bikeshedding :-)

mattgarrish:
Would it make sense to reopen this in the tests repository under docs/a11y-report/index.md then?

avneeshsingh:
Are we expecting such reports for other EPUB 3.3 features also? If yes then better to place everything at one place.

mattgarrish:
We'll have to do the same for the epub metadata. I don't know if that will be done as one big report or if we'll need to create reports per type.

epub33/reports/a11y-properties-use.md
```diff
@@ -0,0 +1,104 @@
+# EPUB Accessibility 1.1 Metadata Usage Report
+
+## Candidate Recommendation Exit Criteria
+
+The EPUB Working Group intends to exit the Candidate Recommendation stage and submit
+the EPUB Accessibility 1.1 specification for consideration as a W3C Proposed Recommendation
+after documenting implementation of each feature.</p>
```
dlazin: Markdown now, so remove </p> :)

epub33/reports/a11y-properties-use.md
```diff
@@ -0,0 +1,104 @@
...
+
+For this specification to advance to Proposed Recommendation, it has to be
+proven that metadata defined and required in this specification have sufficient usage by the
+target communities. This metadata falls into two categories:</p>
```
dlazin: Same

epub33/reports/a11y-properties-use.md
```diff
@@ -0,0 +1,104 @@
...
+
+The following table provides a list of publishers who have stated that they are currently using
+the schema.org discovery metadata properties in production or who are in the process of rolling
+out their implementations.
+
+<table>
+    <thead>
+        <tr>
+            <th>Role</th>
+            <th>[Publisher]</th>
```
dlazin, line 32: Why square brackets? I don't see what they're trying to indicate.

epub33/reports/a11y-properties-use.md
```diff
@@ -0,0 +1,104 @@
...
+2. EPUB-defined metadata properties for reporting aspects of conformance
+
+Usage of these properties means that they are regularly included in the Package Document
+metadata for their EPUB Publications (as appropriate for each title).
+
+## Publisher Implementations
+
+### Schema.org discovery metadata
+
+The following table provides a list of publishers who have stated that they are currently using
```
dlazin, line 24: Question: what's a publisher? I self-published a book earlier this year that uses all of these attributes. It will be easy to find one more "publisher" like me ... but I am not, say, Macmillan.

dlazin:
> @dlazin is the boss on testing, but I think the logical place is indeed the epup-tests repository.

I don't have a strong opinion, but this doesn't feel like testing to me. This is info about EPUB usage, akin to "publishers, please tell us all the custom metadata you've invented" (like ibooks:specified-fonts) or "do you ever use color images?". Therefore it seems to me it belongs in the spec repo.

iherman:
> > @dlazin is the boss on testing, but I think the logical place is indeed the epup-tests repository.
> 
> I don't have a strong opinion, but this doesn't feel like testing to me. This is info about EPUB usage, akin to "publishers, please tell us all the custom metadata you've invented" (like ibooks:specified-fonts) or "do you ever use color images?". Therefore it seems to me it belongs in the spec repo.

Ok, this is really (uninteresting) bike shedding. Leave the document where it is, ie, in the spec repo under reports.

iherman (approved):

epub33/reports/a11y-properties-use.md
mattgarrish: It's a placeholder right now for an actual publisher name. Each column will list their usage. I can add an entry for Benetech's GCA program so we don't need the placeholder if that's okay with you @clapierre ?

epub33/reports/a11y-properties-use.md
mattgarrish: Isn't that like asking does a reading system count for testing if it doesn't have a significant user base? 😃 

To me, the answer is yes for both cases. We're listing implementations, not judging their value, so we shouldn't shut anyone out from reporting because they aren't deemed "good enough".

That said, I expect the people reading our implementation reports will judge the value of our implementations so if we only list a few self-publishers then we may not pass.

I'm not too worried about the accessibility metadata (with the exception of the certifierReport property) as GCA has more than two publishers implementing it. The package metadata might get more interesting.

mattgarrish:
> Ok, this is really (uninteresting) bike shedding.

And we have until the end of CR to figure out how to arrange the reports, so starting it here is as good as anywhere. It's not exactly complicated to move... 😄

PR #1953, opened 2021-12-03
Add risky class of unsupported features

### LINKED ISSUE #1944: What happens to features that do not get enough implementation in CR?
The title tells it all...

In an average W3C Recommendation process, the answer to this question would be to _remove_ this feature from the specification. However, the backward compatibility requirement of our charter forbids us to remove an existing feature (unless is is specifically 3.3 only), hence the question in the issue.

This question must have a clean answer _before_ we ask for a transition request.

I copy this issue to Ralph Swick (@swickr) because his opinion is crucial on what we propose to do.

mattgarrish:
This PR implements the "risky" class as discussed on the WG telecon today.

The most confusing part to me was what reading systems should do with these features given that they may run a strange gamut of things we want implemented in the future to things that are maybe (?) just zombies. I made it a "should" to support them for now.

Feedback welcome.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1953.html" title="Last updated on Dec 6, 2021, 4:33 PM UTC (0287bd4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1953/34cffe4...0287bd4.html" title="Last updated on Dec 6, 2021, 4:33 PM UTC (0287bd4)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -8720,8 +8720,8 @@ html.my-document-playing * {
 				<h3>Risky Features</h3>
 
 				<p>A <strong>risky</strong> feature is a feature introduced prior to EPUB 3.3 but that the Working Group
-					has not been able to establish is implemented in two or more Reading Systems as required by W3C
-					Process [[PROCESS]].</p>
+					has not been able to establish is implemented in two or more Reading Systems as required by <a
+						href="https://www.w3.org/Consortium/Process/">W3C Process</a>.</p>
 
 				<p>These features are considered important to retain despite this limitation because they are known to
 					be implemented by EPUB Creators (i.e., their deprecation would invalidate existing content) and/or
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -8731,7 +8731,7 @@ html.my-document-playing * {
 
 				<ul>
 					<li>
-						<p><a>EPUB Creators</a> MAY use the features as described in their <a>EPUB Publications</a>.</p>
+						<p><a>EPUB Creators</a> MAY use the features as described.</p>
 					</li>
 					<li>
 						<p><a>Reading Systems</a> SHOULD support the feature as described.</p>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5215,8 +5215,9 @@ No Entry</pre>
 
 					</div>
 
-					<p id="fxl-layout-duplication" data-tests="#fxl-layout-duplication">When the property is set to <code>pre-paginated</code> for a spine item, its content dimensions
-						MUST be set as defined in <a href="#sec-fixed-layouts"></a>.</p>
+					<p id="fxl-layout-duplication" data-tests="#fxl-layout-duplication">When the property is set to
+							<code>pre-paginated</code> for a spine item, its content dimensions MUST be set as defined
+						in <a href="#sec-fixed-layouts"></a>.</p>
 
 					<p>The <code>rendition:layout</code> property MUST NOT be declared more than once.</p>
 
@@ -5320,8 +5321,8 @@ No Entry</pre>
 						</dd>
 					</dl>
 
-					<p id="fxl-orientation-duplication" data-tests="#fxl-orientation-duplication">EPUB Creators MUST NOT declare the <code>rendition:orientation</code> property more than
-						once.</p>
+					<p id="fxl-orientation-duplication" data-tests="#fxl-orientation-duplication">EPUB Creators MUST NOT
+						declare the <code>rendition:orientation</code> property more than once.</p>
 
 					<aside class="example" id="fxl-ex2" title="Specifying global landscape orientation">
 						<p>In this example, the content should also render without synthetic spreads.</p>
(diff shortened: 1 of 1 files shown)


epub33/core/index.html
```diff
@@ -8711,9 +8711,49 @@ <h2>Security and Privacy</h2>
...
-				to these features and their support expectations.</p>
+			<p>This specification contains certain features that are not yet fully supported in Reading Systems, that
+				the Working Group no longer recommends for use, or that are only retained for interoperability with EPUB
+				2 Reading Systems. This section defines the meanings of the designations attached to these features and
+				their support expectations.</p>
+
+			<section id="risky">
+				<h3>Risky Features</h3>
+
+				<p>A <strong>risky</strong> feature is a feature introduced prior to EPUB 3.3 but that the Working Group
```
rdeltour: "risky" sounds odd to me, and also too close to "at risk".
But I do not have a better suggestion 😅

epub33/core/index.html
```diff
@@ -8711,9 +8711,49 @@ <h2>Security and Privacy</h2>
...
+
+				<p>These features are considered important to retain despite this limitation because they are known to
+					be implemented by EPUB Creators (i.e., their deprecation would invalidate existing content) and/or
+					they are integral to the content model on which EPUB is built.</p>
+
+				<p>If this specification designates a feature as risky, the following hold true:</p>
+
+				<ul>
+					<li>
+						<p><a>EPUB Creators</a> MAY use the features as described in their <a>EPUB Publications</a>.</p>
```
rdeltour: I would remove "as described" (here and in the next item).
I first read "features as-described-in-their-publications" instead of "features-as-described in their publication" and was confused.

rdeltour (changes requested):
Apart from editorial comments, looks OK to me!

epub33/core/index.html
mattgarrish: Ya, I see your point. Maybe it would be better to drop "in their EPUB Publications" as where else are they going to use them? The "as described" is there because you still have to use them as specced, which is a more important point.

epub33/core/index.html
mattgarrish: Other than beta, the only other thought I've had is "unproven". Technically, that is what we're saying - we haven't been able to prove the features have adoption.

Their being "risky" seems to depend on how the future unfolds. Some could stabilize not long after we publish. (At least in theory, not yet knowing what all will get labelled.)

iherman: So far, "unproven" is at the top of my list...

epub33/core/index.html
```diff
@@ -8711,9 +8711,49 @@ <h2>Security and Privacy</h2>
...
+				2 Reading Systems. This section defines the meanings of the designations attached to these features and
+				their support expectations.</p>
+
+			<section id="risky">
+				<h3>Risky Features</h3>
+
+				<p>A <strong>risky</strong> feature is a feature introduced prior to EPUB 3.3 but that the Working Group
+					has not been able to establish is implemented in two or more Reading Systems as required by <a
+						href="https://www.w3.org/Consortium/Process/">W3C Process</a>.</p>
+
```
iherman, line 8726: I would not refer to the W3C Process here.

The fact of "two or more Reading systems" is _not_, formally, part of the W3C Process. There is a separate section in the process on ["implementation experience"](https://www.w3.org/2021/Process-20211102/#adequate-implementation) but each WG has to decide how the general terms are mapped against the specificities of the WG's specification. The "two or more" is just a general practice.

I would say something more general

> has not been able to establish enough implementation experience.

And possibly point to the process document section above.

epub33/core/index.html
```diff
@@ -8711,9 +8711,49 @@ <h2>Security and Privacy</h2>
...
+					<p>Whether risky labels are removed or replaced by deprecation in a future version of the standard
+						cannot be determined at this time. EPUB Creators should strongly consider the interoperability
+						problems that may arise both now and in the future when using these features.</p>
+				</div>
+
+				<div class="note">
+					<p>The marking of features as risky is a one-time event to account for the different process under
+						which EPUB was developed prior to being brought into W3C. This label will not be used on
+						features in the future.</p>
+				</div>
```
iherman: I just stopped for a moment reading this; are we sure about this? I _think_ it is all right, but I would really like the approval of the others of this specific note; I worry that we may tie our hands unnecessarily for the future.

iherman (changes requested):

epub33/core/index.html
mattgarrish: I was assuming this is a one-time thing we can do to grandfather the spec into W3C, but for anything going forward we should be clear that we are going to follow W3C process more officially.

It was a late addition after I wrote the description for how the class works, as I was worried that by adding this class we might provoke reactions that we're trying to create our own looser process for EPUB. I assume moving forward that we'll only have new features without support, in which case they would follow the standard at-risk/removal process for going through to REC.

epub33/core/index.html
mattgarrish: You're the expert!

epub33/core/index.html
mattgarrish: Although looking at the section again, the first paragraph does say that this is for features developed prior to EPUB 3.3. The note is probably redundant except for not using the label again. I assume testing will ferret out all the features so it won't be added again in the future, but it may not be strictly necessary to say.

epub33/core/index.html
iherman: Another term I dreamt up (literally, I thought of it at dawn today) is "incubated".

epub33/core/index.html
iherman: I agree with the intention and I am all right with what is stated.

I actually did not realize it is a duplicate of the first paragraph, but it is... ie, it may be not be necessary to keep it.

mattgarrish: One thing I don't like looking at it again is that it seems to suggest we would have to remove the labels in the next revision. That may or may not be true.

I've tweaked the wording of the note to clarify that it won't be applied to new features, as they will follow w3c process. That should complement the first paragraph better.

iherman (approved):

epub33/core/index.html
mattgarrish: We should probably merge this PR without closing 1944 and take the debate on the right name back to that issue.

PR #1954, opened 2021-12-06
Editorial rephrase of epubReadingSystem property definitions

iherman:
There is a table in the definition of `epubReadingSystem` where a deprecated feature was listed as part of the required features. This is changed.

I have also changed two occurrences of _MAY_ into "may" in the table of [§A.4.1.2](https://w3c.github.io/epub-specs/epub33/rs/#app-ers-hasFeature-features). The term _MAY_ is used for the conformance clause of a Reading System, where, in this case, the "may" merely refers to what a script may or may not do, ie, it is the explanation of a property.

---

See:

    * [Preview](https://raw.githack.com/w3c/epub-specs/epubReadingSystem-updates/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/epubReadingSystem-updates/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2114,18 +2114,13 @@ partial interface Navigator {
 							<td>Returns a <code>String</code> value representing the version of the Reading System
 								(e.g., "<code>1.0</code>", "<code>2.1.1</code>").</td>
 						</tr>
-						<tr>
-							<td id="propdef-layoutStyle">
-								<code>
-									<dfn>layoutStyle</dfn>
-								</code>
-							</td>
-							<td>Use of the <code>layoutStyle</code> property is <a data-cite="epub-33#deprecated"
-									>deprecated</a> [[EPUB-33]]. Refer to its definition in [[EPUBContentDocs-301]] for
-								usage information.</td>
-						</tr>
 					</tbody>
 				</table>
+
+				<p>
+					This specification used to define a <code>layoutStyle</code> property, but it is now <a data-cite="epub-33#deprecated"
+					>deprecated</a> [[EPUB-33]]. Refer to its definition in [[EPUBContentDocs-301]] for more information.
+				</p>
 			</section>
 
 			<section id="app-ers-methods">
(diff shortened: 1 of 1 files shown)


iherman:
There is also another oddity that I have not changed, but maybe it should. The introductory paragraph of §A.4.1.2 says:

> The following table lists the set of features that Reading Systems that support the epubReadingSystem object MUST recognize (i.e., provide a return value for). Support for these features is OPTIONAL.

I _think_ that what is meant that the various options in the table are not required to be offered (ie, manipulate the dom) although the entries _MUST_ be implemented. Maybe it is only me, but I did not find the formulation very clear.

iherman:
One more question: is there any reason why this is part of an Appendix? The text says that if a RS implements scripting, this feature _MUST_ be implemented. I would propose to turn it into a bona fide normative section.

mattgarrish:
> I _think_ that what is meant that the various options in the table are not required to be offered (ie, manipulate the dom) although the entries _MUST_ be implemented.

Yes, that's the gist of it. I think it might also help to say these are booleans, in which case we don't need two normative statements. So something more like:

> When the following features are queried from the `hasFeature` method, Reading Systems MUST return a boolean value indicating their support.

mattgarrish:
> is there any reason why this is part of an Appendix?

It was clunky to drop an entire object definition into the middle of the Content Docs spec., as it probably would have had to have been jammed somewhere in the middle of the scripting section. For general readability, having RS implementation details in an appendix made sense.

Now that it's in a reading system specification, it's not strictly necessary it be an appendix.

iherman:
@mattgarrish, I have removed the appendix class, and modified the intro paragraph to the table.

mattgarrish (approved):

PR #1955, opened 2021-12-07
Hack on data-test display: show real test id

iherman:
Adding a simple javascript hack that displays the real test id-s in the pull down lists of tests instead of simply their number.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1955.html" title="Last updated on Dec 7, 2021, 8:29 AM UTC (0e7312a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1955/8d607e1...0e7312a.html" title="Last updated on Dec 7, 2021, 8:29 AM UTC (0e7312a)">Diff</a>

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -5,6 +5,7 @@
 		<title>EPUB Accessibility 1.1</title>
 		<script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
 		<script src="../common/js/css-inline.js" class="remove"></script>
+		<script src="../common/js/data-test-display.js" class="remove"></script>
 		<script class="remove">
 			var respecConfig = {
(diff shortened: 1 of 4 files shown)


iherman:
(This was a thorn in my eyes for a long time :-)

dlazin (approved):
Hurray, thanks! Just to check, though: could/should you not contribute this to respec instead? I have no idea how hard that is.

iherman:
> Just to check, though: could/should you not contribute this to respec instead? I have no idea how hard that is.

I believe what we have was a side effect of using fragment ID-s to the tests, ie, no separate path name per test references. The correct way would be to modify the respec code in this respect (and that is not what this hack does), and I think we are the exception in our testing, so I would leave it as is.

PR #1964, opened 2021-12-17
Settled a broken link

### LINKED ISSUE #1963: Broken link to URL standard definition
In the first paragraph after the definition box in [2.3.2.5.2 The item Element](https://w3c.github.io/epub-specs/epub33/core/#sec-item-elem), the link on "path-relative-scheme-less-URL" is

https://url.spec.whatwg.org/#path-relative-scheme-less-URL 

but should be:

https://url.spec.whatwg.org/#path-relative-scheme-less-url-string

It might also be good to put a reference to the URL standard at the end of the sentence, as they look like they might be internal links without one.

iherman:
Fix #1963 (editorial)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1964.html" title="Last updated on Dec 18, 2021, 8:30 AM UTC (ba1757c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1964/9305e0b...ba1757c.html" title="Last updated on Dec 18, 2021, 8:30 AM UTC (ba1757c)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3000,10 +3000,10 @@ XHTML:
 							</dl>
 
 							<p>Each <code>item</code> element identifies a <a>Publication Resource</a> by the URL
-								[[URL]] in its <code>href</code> attribute. The value MUST be a an <a
+								[[URL]] in its <code>href</code> attribute. The value MUST be an <a
 									data-cite="url#absolute-url-string">absolute-</a> or <a
-									data-cite="url#path-relative-scheme-less-URL">path-relative-scheme-less-URL</a>
-								string. EPUB Creators MUST ensure each URL is unique within the <code>manifest</code>
+									data-cite="url#path-relative-scheme-less-url-string">path-relative-scheme-less-URL</a>
+								string [[URL]]. EPUB Creators MUST ensure each URL is unique within the <code>manifest</code>
 								scope after <a href="#sec-parse-package-urls">parsing</a>.</p>
 
 							<p id="attrdef-item-media-type">The Publication Resource identified by an <code>item</code>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -997,7 +997,7 @@
 							</li>
 							<li>
 								<p><a href="#sec-foreign-restrictions-manifest">manifest fallback</a> chains defined on
-										<a href="sec-item-elem"><code>item</code> elements</a> in the <a>Package
+										<a href="#sec-item-elem"><code>item</code> elements</a> in the <a>Package
 										Document</a>.</p>
 							</li>
 						</ul>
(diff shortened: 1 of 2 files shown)


mattgarrish:
This looks fine. I only stumbled on the broken link while adding an epubcheck issue, but running the link checker it's also reporting these errors:

- http://info.vlb.de/files/wgsneuversion2_0.pdf redirected to https://vlb.de/files/wgsneuversion2_0.pdf
- https://www.w3.org/TR/epub-33/sec-item-elem (missing a #)
- https://www.w3.org/TR/epub-33/epub-a11y-11 (href needs to be data-cite on an a element in the accessibility section)
- https://www.w3.org/TR/epub-3-tts (should be epub-tts-10)
- https://www.w3.org/TR/wcag/ (should be /wcag2/, also in accessibility section)

I guess we'll have to remember to periodically run the checker now.

And, alas, the data-cite syntax makes it impossible to know if we have busted fragments in the HTML spec, as it turns the links into "https://html.spec.whatwg.org/multipage/#blah". These only resolve to the proper page when you open them, so to the link checker they're all broken fragments.

mattgarrish (approved):

iherman:
I have settled the link checker problems you found before merging, just to get them out of the way.

B.t.w., I was indeed too optimistic with respec+data-cite: in contrast to what I said, it does not check the validity of the fragment id. What it does, at least on my machine, is that whenever I add a new data-cite entry, I get a pull-down menu with the _data-cite values already used in the document_. I do find that very useful. (I am not sure how this is achieved, though; I use Visual Studio Code, but I never did anything respec specific. It may be a built-in feature of VSC and not related to respec in any way...)

mattgarrish:
> it does not check the validity of the fragment id

Ya, I'm thinking we should switch the HTML references back so we don't get all those meaningless fragment errors.

From what I see, this script parses the fragment and redirects the browser: https://html.spec.whatwg.org/link-fixup.js

But as this isn't a proper redirect, the link checker doesn't see it.

Or do you think it's worth trying to get the link checker updated to handle the redirects?

iherman:
> > it does not check the validity of the fragment id
> 
> Ya, I'm thinking we should switch the HTML references back so we don't get all those meaningless fragment errors.
> 
> From what I see, this script parses the fragment and redirects the browser: https://html.spec.whatwg.org/link-fixup.js
> 
> But as this isn't a proper redirect, the link checker doesn't see it.
> 
> Or do you think it's worth trying to get the link checker updated to handle the redirects?

What I am considering is rather to see if the github actions can be used to check the links. After all, whenever we submit a PR, those actions check the document per pubrules (but not the link checker); it should be possible to add the link checker as well to issue warnings.

I am not 100% sure how to do that, and I would have to check with the system team. Next year...

PR #1967, opened 2022-01-11
Fix unlinked definitions

mattgarrish:
Another new respec check: issues warning if a term in dfn tags isn't linked to.

Most of these errors were caused by using dfn tags on the epub-created css definitions. I changed them to use code tags to match the other element, attribute and property definitions.

The scripting context definitions (container-constrained and spine-level) had the same issue, but as these terms are only used once outside their section -- in the extended example in the appendix -- I just dropped the dfn tags.

The "Content Display Area" term was also flagged as it is not used in the core spec, so I moved it to the RS spec's terminology section (it's the only unique term to that specification). I also added a link across to the core terminology section, as not having one wasn't very helpful.

And, finally, the "derive a file path" function got flagged but I put the css class to hide the tag on it as the function should have a definition.

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/unlinked-dfn/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/unlinked-dfn/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/1967.html" title="Last updated on Jan 12, 2022, 12:09 PM UTC (a4a07ec)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/1967/503c4b7...a4a07ec.html" title="Last updated on Jan 12, 2022, 12:09 PM UTC (a4a07ec)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -4488,9 +4488,9 @@ No Entry</pre>
 						<p>EPUB 3 defines two contexts for script execution:</p>
 
 						<ul>
-							<li><a href="#sec-scripted-container-constrained">container-constrained scripts</a> &#8212;
+							<li><a href="#sec-scripted-container-constrained">container constrained</a> &#8212;
 								when the execution of a script occurs within an <code>iframe</code>; and</li>
-							<li><a href="#sec-scripted-spine">spine-level scripts</a> &#8212; when the execution of a
+							<li><a href="#sec-scripted-spine">spine level</a> &#8212; when the execution of a
 								script occurs directly within a <a>Top-level Content Document</a>.</li>
 						</ul>
 						<p>Whether EPUB Creators embed the code directly in the <code>script</code> element or reference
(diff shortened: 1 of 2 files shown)


iherman:
For some reasons I cannot generate the diff file via statically either. Any idea why?

Looking at the code I am fine with it, but I did not really find a documentation on respec on what the `lint-ignore` class does...

iherman (approved):

mattgarrish:
> Any idea why?

No, other than it must have been down. The RS diff is working again now. Don't know what happened with pr-preview, unless maybe it ran into a similar issue.

> I did not really find a documentation on respec on what the `lint-ignore` class does

Ya, I can only assume it's related to the [lint configuration option](https://respec.org/docs/#lint). It just tells the linter to ignore that there aren't any links to the definition. The warning respec was generating said to remove the dfn tags or use that css class to ignore, so I mostly opted to remove the dfn tags.

PR #1993, opened 2022-02-11
Switched the requirement level for MO 

### LINKED ISSUE #1991: Reading Systems and Media Overlays...
This may be opening the flood gates, but… at the moment, the RS Specification [says about MO](https://w3c.github.io/epub-specs/epub33/rs/#sec-media-overlays) that 

> Reading Systems MAY support [Media Overlays](https://www.w3.org/TR/epub-33/#sec-media-overlays)

why is this is so weak? Shouldn't we say, at least that they _SHOULD_ support them? Where does this statement come from?

@wareid @dauwhe @mattgarrish @avneeshsingh @marisademeglio

iherman:
See #1991, and the [relevant resolution](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-02-11-epub#resolution1)

Fix #1991

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2448,7 +2448,7 @@ alert("Feature " + feature + " supported?: " + conformTest);</pre>
 					>Working Group's issue tracker</a>.</p>
 
 			<ul>
-				<li>11-Feb-2021: Media Overlays are now required as a SHOULD instead of MAY See <a
+				<li>11-Feb-2021: Media Overlays are now required as a SHOULD instead of MAY. See <a
 						href="https://github.com/w3c/epub-specs/issues/1991">issue 1991</a>.</li>
 				<li>04-Feb-2021: Changed the informative security considerations for controlling network access, opening
 					external links, and managing local storage to normative recommendations. See <a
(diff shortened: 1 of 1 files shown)


wareid (approved):

mattgarrish:
This doesn't look like quite the right fix. Media overlays actually already were recommended; it was only the support box that wasn't right:

> If it has the capability to render pre-recorded audio, it MUST support the [audio Core Media Type Resources](https://www.w3.org/TR/epub-33/#cmt-grp-audio) [[EPUB-33](https://w3c.github.io/epub-specs/epub33/rs/#bib-epub-33)] and SHOULD support Media Overlays [[EPUB-33](https://w3c.github.io/epub-specs/epub33/rs/#bib-epub-33)].

https://w3c.github.io/epub-specs/epub33/rs/#confreq-rs-epub3-mp3-aac

Now we have contradictory statements going the other way: one conditional on supporting audio playback and the other not.

I would personally remove the support statement from 2.4, since it's not about core media types. But the support statement at the beginning of the MO section should also be clear that it's a recommendation when audio playback is supported. Otherwise, it's kind of weird to blanket recommend support when a device can't render audio.

PR #2001, opened 2022-02-18
Updated the MO text and audio element definitions

### LINKED ISSUE #1986: Usage requirement of mo audio element
I a bit bothered by the usage requirement of the [audio](https://w3c.github.io/epub-specs/epub33/core/#sec-smil-audio-elem) element in MO. The text says:

> **Usage**
> A REQUIRED child of the [par element](https://w3c.github.io/epub-specs/epub33/core/#elemdef-smil-par) unless its sibling [text element](https://w3c.github.io/epub-specs/epub33/core/#elemdef-smil-text) refers to audio or video media, or to textual content intended for rendering via [text-to-speech](https://w3c.github.io/epub-specs/epub33/core/#sec-tts), in which case it is OPTIONAL (see [7.3.2.4 Embedded Media](https://w3c.github.io/epub-specs/epub33/core/#sec-embedded-media)).

As read, this sounds like an untestable requirement to me, unless it is tested run-time. After all, how does a checker know that an HTML file, linked through a `text` element, is used with text-to-speech? We do not define normatively anything for tts after all.

Maybe EPUBCheck found a trick that I miss (@rdeltour ?). Otherwise, it may be cleaner to replace that requirement by something like:

> An OPTIONAL child of the [par element](https://w3c.github.io/epub-specs/epub33/core/#elemdef-smil-par). If the element is missing, its adjacent `text` element either refers to audio or video, or is supposed to refer to a textual content intended for rendering via text-to-speech.

I think that content-wise we are saying the same thing, but the checker is not under the obligation of checking this.

iherman:
Summarizes the possible consenus in #1986.

Fix #1986

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2001.html" title="Last updated on Feb 22, 2022, 7:54 AM UTC (da78432)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2001/5ba7886...da78432.html" title="Last updated on Feb 22, 2022, 7:54 AM UTC (da78432)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -8153,10 +8153,7 @@ No Entry</pre>
 							<p>
 								When a <a href="#elemdef-smil-text"><code>text</code></a> element references embedded 
 								audio or video, Reading Systems will intiate playback of the media in the absence of an 
-								<a href="#elemdef-smil-audio"><code>audio</code></a> element sibling. Therefore, EPUB
-								Creators should carefully examine any overlapping timed media situations and deal with
-								them at the production stage, as Reading Systems handling of many aspects of simultaneous
-								timed media is implementation-specific.
+								<a href="#elemdef-smil-audio"><code>audio</code></a> element sibling. 
 							</p>
 
 							<p><a>EPUB Creators</a> SHOULD avoid using scripts to control playback of referenced
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7672,10 +7672,6 @@ No Entry</pre>
 										</p>
 									</li>
 								</ul>
-								<p>The <a href="#elemdef-smil-audio"><code>audio</code> element</a> is OPTIONAL only if
-									its sibling <a href="#elemdef-smil-text"><code>text</code> element</a> refers to
-									audio or video media (see <a href="#sec-embedded-media"></a>), or to textual content
-									intended for rendering via <a href="#sec-tts">text-to-speech</a>.</p>
 							</dd>
 						</dl>
 					</section>
@@ -7683,9 +7679,13 @@ No Entry</pre>
 					<section id="sec-smil-text-elem">
 						<h5>The <code>text</code> Element</h5>
 
-						<p>The <code>text</code> element references an element in the <a>EPUB Content Document</a>. A
-								<code>text</code> element typically refers to a textual element but can also refer to
-							other EPUB Content Document media elements (see <a href="#sec-embedded-media"></a>).</p>
+						<p>
+							The <code>text</code> element references an element in an <a>EPUB Content Document</a>.
+							A <code>text</code> element typically refers to a textual element but can also refer to
+							other EPUB Content Document media elements (see <a href="#sec-embedded-media"></a>).
+							In the absence of a sibling <code>audio</code> element this element SHOULD refer to a 
+							textual content intended for rendering via <a href="#sec-tts">text-to-speech</a>.
+						</p>
 
 						<dl class="elemdef" id="elemdef-smil-text">
 							<dt>Element Name</dt>
@@ -7732,6 +7732,16 @@ No Entry</pre>
 								<p>Empty</p>
 							</dd>
 						</dl>
+
+						<p class="note">
+							This specification places no restriction on the <code>src</code> attribute of a <code>text</code> element. 
+							Authors should, however, refer to a content that can be styled with CSS to make 
+							the <a href="#sec-docs-assoc-style">association with style information</a> effective, i.e., 
+								<a href="https://html.spec.whatwg.org/multipage/dom.html#palpable-content">palpable content</a> for XHTML 
+								or <a href="https://www.w3.org/TR/SVG11/paths.html">paths</a>, 
+								<a href="https://www.w3.org/TR/SVG11/shapes.html">basic shapes</a>, or 
+								<a href="https://www.w3.org/TR/SVG11/text.html">text</a> elements in SVG.
+						</p>
 					</section>
 
 					<section id="sec-smil-audio-elem">
@@ -7749,11 +7759,7 @@ No Entry</pre>
 
 							<dt>Usage</dt>
 							<dd>
-								<p>A REQUIRED child of the <a href="#elemdef-smil-par"><code>par</code> element</a>
-									unless its sibling <a href="#elemdef-smil-text"><code>text</code> element</a> refers
-									to audio or video media, or to textual content intended for rendering via <a
-										href="#sec-tts">text-to-speech</a>, in which case it is OPTIONAL (see <a
-										href="#sec-embedded-media"></a>).</p>
+								<p>An OPTIONAL child of the <a href="#elemdef-smil-par"><code>par</code> element</a>.</p>
 							</dd>
 
 							<dt>Attributes</dt>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -8150,10 +8150,14 @@ No Entry</pre>
 						<section id="sec-emb-audio-video">
 							<h6>Embedded Audio and Video</h6>
 
-							<p>When a <code>text</code> element references embedded audio or video, the <a
-									href="#elemdef-smil-audio"><code>audio</code></a> sibling element is OPTIONAL.
-								Reading Systems will intiate playback of the media in the absence of an
-									<code>audio</code> element.</p>
+							<p>
+								When a <a href="#elemdef-smil-text"><code>text</code></a> element references embedded 
+								audio or video, Reading Systems will intiate playback of the media in the absence of an 
+								<a href="#elemdef-smil-audio"><code>audio</code></a> element sibling. Therefore, EPUB
+								Creators should carefully examine any overlapping timed media situations and deal with
+								them at the production stage, as Reading Systems handling of many aspects of simultaneous
+								timed media is implementation-specific.
+							</p>
 
 							<p><a>EPUB Creators</a> SHOULD avoid using scripts to control playback of referenced
 								embedded EPUB Content Document media, as this might conflict with Media Overlays
@@ -10736,6 +10740,12 @@ EPUB/images/cover.png</pre>
 					>Working Group's issue tracker</a>.</p>
 
 			<ul>
+				<li>
+					19-Feb-2021: Clarified the <a href="#elemdef-smil-audio"><code>audio</code></a> element's 
+					definition by making it optional, and adapt the specification's text elsewhere to address
+					the situation when the element is indeed not present. See <a
+					href="https://github.com/w3c/epub-specs/issues/1986">issue 1986</a>.
+				</li>
 				<li>04-Feb-2021: Expanded the section on security and privacy to include new sections on the threat
 					model for EPUB Publications and additional recommendations for ensuring security and privacy. See <a
 						href="https://github.com/w3c/epub-specs/issues/1871">issue 1871</a>, <a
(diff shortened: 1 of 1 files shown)


marisademeglio:
Proposed change to [embedded media](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2001.html#sec-emb-audio-video):

Current text:
>When a text element references embedded [audio](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2001.html#elemdef-smil-audio)[](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2001.html#dfn-epub-creator) or video, the audio sibling element is OPTIONAL. Reading Systems will intiate playback of the media in the absence of an audio element.

But the audio sibling is optional anyway. So how about instead:

When a text element references embedded [audio](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2001.html#elemdef-smil-audio) or video, Reading Systems will intiate playback of the media in the absence of an audio element sibling. Therefore, EPUB Creators should carefully examine any overlapping timed media situations and deal with them at the production stage, as Reading Systems handling of many aspects of simultaneous timed media is implementation-specific.

[EPUB Creators](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2001.html#dfn-epub-creator) SHOULD avoid using scripts to control playback of referenced embedded EPUB Content Document media, as this might conflict with Media Overlays playback behavior.

iherman:
Ah, good catch, @marisademeglio, I did not realize that the text needed changes in that section as well! I have taken over your text. 

I have also added an item to the changes' section. Although the change may be considered as editorial only, I think it is worth drawing attention to this clarification.

epub33/core/index.html
```diff
@@ -7672,20 +7672,20 @@ <h5>The <code>par</code> Element</h5>
...
 						<h5>The <code>text</code> Element</h5>
 
-						<p>The <code>text</code> element references an element in the <a>EPUB Content Document</a>. A
-								<code>text</code> element typically refers to a textual element but can also refer to
-							other EPUB Content Document media elements (see <a href="#sec-embedded-media"></a>).</p>
+						<p>
+							The <code>text</code> element references an element in an <a>EPUB Content Document</a>.
+							A <code>text</code> element typically refers to a textual element but can also refer to
+							other EPUB Content Document media elements (see <a href="#sec-embedded-media"></a>).
+							In the absence of a sibling <code>audio</code> element this element SHOULD refer to a
```
marisademeglio: I would remove the last sentence ("In the absence of a sibling...") because it makes it sound like the presence of a sibling `audio` element makes non-TTS-able content accessible, but this is not true. E.g. an image of text alongside an audio rendering of that text still would not allow a user to explore that text with their screenreader, to find out the spelling.

marisademeglio (commented):
In addition to the inline comment, I would remove the paragraph at line 8166:

>EPUB Creators should carefully examine any overlapping audio situations and deal with
them at the production stage, as Reading Systems handling of simultaneous volume levels is optional.


Because it is superseded  by the statement about simultaneous timed media around line 8157 (I just replaced "volume" with "many aspects" since there are more things than volume that could get hairy).

epub33/core/index.html
iherman: I am not sure I agree. That sentence does not say anything about the situation you describe (which is of course perfectly o.k.) and I do not see why it would suggest disallowing it. On the other hand, we have moved the reference to TTS (in case there is no sibling) here from the definition of the `audio` element; if we remove this sentence there is no reference to the usage of TTS any more from that definition section, which I think is more of a problem; it is a bit hidden in the section of embedded media which the reader may not automatically associate with the subject.

iherman:
> In addition to the inline comment, I would remove the paragraph at line 8166:
> 
> > EPUB Creators should carefully examine any overlapping audio situations and deal with
> > them at the production stage, as Reading Systems handling of simultaneous volume levels is optional.
> 
> Because it is superseded by the statement about simultaneous timed media around line 8157 (I just replaced "volume" with "many aspects" since there are more things than volume that could get hairy).

Agree. I have removed it.

epub33/core/index.html
marisademeglio: > In the absence of a sibling <code>audio</code> element this element SHOULD refer to a textual content intended for rendering via <a href="#sec-tts">text-to-speech</a>.

Can I read into this that in the presence of a sibling audio element, the text element may refer to non-TTS-able content?

E.g. 

```
<par>
  <text src="page.html#video">
  <audio src="background-music.mp3">
</par>
```

Which is potentially totally inaccessible but still legal (that is ok, this isn't the EPUB a11y spec). 

But as soon as you remove the audio, we are saying with the above sentence that then you need to have
```
<par>
  <text src="page.html#text-not-video">
</par>
```

Because why? Speaking to this specific example, an in-page video in its own `par` could make sense for an author, to initiate playback after some narration. (Even though I would still rather remove the ability to reference embedded media with `text`, that's another discussion).

iherman: We seem to disagree...

Everything you write above is of course correct, with the exception of:

> But as soon as you remove the audio, we are saying with the above sentence that then you need to have

I read this sentence as if the statement used a _MUST_. But we say _SHOULD_, which essentially say "do this, unless you have a good reason not to".  Your use case of referring to a video element has a use case indeed, which is a good reason to keep it. But I think we can agree that this use case is probably a very minor one compared to the TTS case (I almost said 'in practice', though we do not know how either TTS or embedded videos are implemented 😒). If you really dislike it, we can exchange _SHOULD_ to _MAY_, which weakens the requirement even more.

> (Even though I would still rather remove the ability to reference embedded media with text, that's another discussion)

Yep, I agree. I think the simplification of SMIL into MO went a step too far. I would prefer to re-introduce the SMIL `<ref>` element, and maybe also the `<video>` element. This would make the spec way cleaner and would help with these ambiguities. Do you think it is worth raising this to the WG? Or should we declare this as the boat that has sailed?

marisademeglio: > Everything you write above is of course correct, with the exception of:
> 
> > But as soon as you remove the audio, we are saying with the above sentence that then you need to have
> 
> I read this sentence as if the statement used a _MUST_. But we say _SHOULD_, which essentially say "do this, unless you have a good reason not to". Your use case of referring to a video element has a use case indeed, which is a good reason to keep it. But I think we can agree that this use case is probably a very minor one compared to the TTS case (I almost said 'in practice', though we do not know how either TTS or embedded videos are implemented 😒). If you really dislike it, we can exchange _SHOULD_ to _MAY_, which weakens the requirement even more.

Well, all of  MO is a _SHOULD_ anyway 🙃

I still don't think that requiring content to be TTS-able only if there is no `audio` makes sense. What problem are we solving?

> 
> Yep, I agree. I think the simplification of SMIL into MO went a step too far. I would prefer to re-introduce the SMIL `<ref>` element, and maybe also the `<video>` element. This would make the spec way cleaner and would help with these ambiguities. Do you think it is worth raising this to the WG? Or should we declare this as the boat that has sailed?

I think that boat is long gone. I would remove the ability to reference to embedded media. Make `text` mean actual text. Maybe add `image` for images? But then you either restrict `src` of image to embedded images or you have to define a viewport for them, which gets complex. 

Introducing `ref` and `video` is tempting but I don't see a practical case for it when it comes to content creation and RS implementation. Not to mention the spec would get considerably more intimidating.

We have a lot of thoughts on this over in [sync-media-pub](https://w3c.github.io/sync-media-pub/).

mattgarrish: > I would remove the ability to reference to embedded media. Make text mean actual text. Maybe add image for images? But then you either restrict src of image to embedded images or you have to define a viewport for them, which gets complex.

Using `text` for embedded media is wildly unintuitive, to put it mildly, but I'm really worried about the can of worms we'd be opening this late in the revision by changing too much, too.

There's so much about `text` without `audio` for any non-text content that is problematic to implement. We allow images, but we don't define how to determine the text to TTS. Is it only the alt text? Do you account for `aria-describedby`, `aria-label(ledby)`, and `aria-details`, too? We're assuming, I guess, that the reading system will be ARIA-aware and use its rules.

And what is the likelihood we'll find any implementations of audio/video initiation from a `text` element? I suspect that feature will get tagged "under-implemented".

iherman: > Using text for embedded media is wildly unintuitive, to put it mildly, but I'm really worried about the can of worms we'd be opening this late in the revision by changing too much, too.

Yep. Maybe, in a future version of EPUB, [sync-media-pub](https://w3c.github.io/sync-media-pub/) would take over MO, so touching MO may be unnecessary. Let us be silent on this now.

@mattgarrish could you be the decider on the (mild:-) disagreement @marisademeglio and I have on whether or not to keep the sentence

>  In the absence of a sibling audio element this element SHOULD refer to a textual content intended for rendering via [text-to-speech](https://github.com/w3c/epub-specs/pull/2001#sec-tts).

in the definition of the `<text>` element or not? This is prominently part of the current spec, though in an awkward  manner, and I still believe it is better to keep it as it is in the current PR; @marisademeglio thinks it may be misleading. 

I would like to merge and close this issue...

iherman: > Well, all of MO is a SHOULD anyway 🙃

Right, exactly... 

> I still don't think that requiring content to be TTS-able only if there is no audio makes sense. What problem are we solving?

That is a valid question but, at this point, my worry is to remove or, more exactly, to strongly downplay a feature that is in the  current spec. We could just make the `<audio>` element REQUIRED, full stop, and remove that feature, but that is not something we should do in this iteration (we are not supposed to make EPUB 3.2 instances invalid 3.3). After all, the only purpose of this PR (and its original issue) is a clean-up.

marisademeglio: Can we change the last sentence to not contain a normative statement? So like:

"In the absence of a sibling audio element, textual content referred to by this element may be rendered via [text-to-speech](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2001.html#sec-tts)."

iherman: That works for me. I have made the change.

iherman: I mark this issue as resolved; I think the whole discussion around the possible re-introduction of `<ref>` should be put aside for now.

marisademeglio:
Looking at what I think is the [latest](https://w3c.github.io/epub-specs/epub33/core/#sec-tts), I see an inconsistency. We have made the `audio` element optional but this section still says (and I think I suggested part of this, now I am un-suggesting it):

> A Media Overlay [par element](https://w3c.github.io/epub-specs/epub33/core/#elemdef-smil-par) MAY omit an [audio element](https://w3c.github.io/epub-specs/epub33/core/#elemdef-smil-audio) when its [text element](https://w3c.github.io/epub-specs/epub33/core/#elemdef-smil-text) references text or an image. In these cases, Reading Systems are expected to render the referenced fragment via TTS, so EPUB Creators MUST ensure the fragment is appropriate for TTS rendering (e.g., contains a textual EPUB Content Document element or has a text fallback).

So how about, to align with our other wording changes:

> When a Media Overlay [par element](https://w3c.github.io/epub-specs/epub33/core/#elemdef-smil-par) omits an [audio element](https://w3c.github.io/epub-specs/epub33/core/#elemdef-smil-audio),  its [text element](https://w3c.github.io/epub-specs/epub33/core/#elemdef-smil-text) may be rendered in Reading Systems via TTS.  If the text fragment is not appropriate for TTS rendering (e.g., is not a text element and/or has no text fallback), this may produce unexpected results.

iherman:
@marisademeglio see #2008

PR #2009, opened 2022-02-24
FXL Accessibility - Layout and Limits 

### LINKED ISSUE #1733: [EPUB 3.3] [suggestion] Note on Fixed Layout and accessibility
Perhaps add a note in the section on Fixed Layout EPUBs indicating that there are currently no ways to produce fully accessible (i.e. EPUB Accessibility compliant) FXL EPUBs?

https://www.w3.org/TR/epub-33/#sec-fixed-layouts

### LINKED ISSUE #1736: [EPUB 3.3] [suggestion] Note on accessibility and orientation for FXL
Perhaps it might make sense to include a note saying that blocking the orientation may create problems for content accessibility?

wareid:
Adding text to address the concerns raised in issues #1733 and #1736.

### DIFF
--- a/epub33/fxl-a11y/index.html
+++ b/epub33/fxl-a11y/index.html
@@ -262,7 +262,7 @@ If an image splits over two pages, put the full image description of both images
 
 If there are a group of images in sequence, you only have to describe details in the first image. In the proceeding images only mention what has changed. W3 Tutorial on Image Groups
 
-Not all details are needed in writing alt-text for images, and what you do describe relies heavily on context. For example, if the image is described in the surrounding text, you only need to briefly describe it in the alt-text. For more information on when, and how to describe you can go to the DIAGRAM centre, or AccessibilePublishing.ca. See resources for links.    
+Not all details are needed in writing alternative text for images, and what you do describe relies heavily on context. For example, if the image is described in the surrounding text, you only need to briefly describe it in the alt-text. For more information on when, and how to describe you can go to the DIAGRAM centre, or AccessibilePublishing.ca. See resources for links.    
 
 ### Useful Resources for describing images
 
@@ -414,7 +414,7 @@ An image of text, or images with text, are not accessible. It is important to tr
 
 The best way to create an accessible table is to present it as tabular data. This is already documented in the DAISY Knowledge Base with directions on creating [semantic tagging for tables](http://kb.daisy.org/publishing/docs/html/tables.html).
 
-If the table needs to remain an image, another option is to use alt-text and caption or ARIA roles to describe the data in logical reading order. The caption could be used as a summary of the table, and the Alt-text would dive deeper. You would start with describing what the header row is followed by what is presented in each row. You could then dive deeper and list all the data as it appears in reading order.
+If the table needs to remain an image, another option is to use alternative text and caption or ARIA roles to describe the data in logical reading order. The caption could be used as a summary of the table, and the alternative text would dive deeper. You would start with describing what the header row is followed by what is presented in each row. You could then dive deeper and list all the data as it appears in reading order.
 
 Depending on the complexity of the image this may require complex description with one of the following formats:
 
--- a/epub33/fxl-a11y/index.html
+++ b/epub33/fxl-a11y/index.html
@@ -76,7 +76,7 @@ Fixed Layout publications present some unique challenges for accessibility. The 
 
 In particular, the needs of people with low vision or learning disabilities that rely on the transformation of text are still almost impossible to accommodate in fixed layout content. Content creators concerned about this may choose not to use fixed layout. 
 
-We want to recognize these challenges for content creators, and in this document will outline some techniques for producing more accessible fixed layout content. We encourage content creators to explore the possibilities that digital publications present, even when creating fixed layout publications.
+We want to recognize these challenges for content creators, and in this document will outline some techniques for producing more accessible fixed layout content. We encourage content creators to explore the full range of options for accessibility that digital publications present, even when creating fixed layout publications.
 
 ## Reading Order {#reading-order}
 
@@ -442,17 +442,18 @@ These display characteristics include:
 * <code>rendition:spread</code> - the spread settings of the book, can be <code>none</code>, <code>landscape</code>, <code>both</code>, or <code>auto</code>
 * <code>rendition:page-spread-*</code> - spread settings for individual spine items, can be <code>left</code>, <code>right</code>, and <code>center</code>
 
+The default value for the <code>orientation</code> and <code>spread</code> properties is <code>auto</code>, which means that the reading system settings or defaults take effect. 
+
+It is recommended that content creators do not set a specific <code>orientation</code> property, as this can interfere with user device preferences. It is especially important for users who may be unable to change the orientation of their device to match the content settings. 
+
 <aside class="example" id="metadata-example-1">
+    <p>An example of package metadata for a fixed layout publication.</p>
     <pre>
         &lt;meta property="rendition:layout"&gt;pre-paginated&lt;/meta&gt;
         &lt;meta property="rendition:orientation"&gt;auto&lt;/meta&gt;
         &lt;meta property="rendition:spread"&gt;auto&lt;/meta&gt;
     </pre>
 </aside>
-
-The default value for the <code>orientation</code> and <code>spread</code> properties is <code>auto</code>, which means that the reading system settings or defaults take effect. 
-
-It is recommended that content creators do not set a specific <code>orientation</code> property, as this can interfere with user device preferences. It is especially important for users who may be unable to change the orientation of their device to match the content settings. 
 
 ## Accessibility metadata {#a11y-metadata}
 
--- a/epub33/fxl-a11y/index.html
+++ b/epub33/fxl-a11y/index.html
@@ -74,7 +74,9 @@ This note serves to help content authors and publishers try to address some of t
 
 Fixed Layout publications present some unique challenges for accessibility. The requirements laid out in [[EPUB-A11Y-11]] recommend [[WCAG2]] AA, but for many use cases in fixed layout, that might not be possible without fundamental changes to the content.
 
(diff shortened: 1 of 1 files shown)


epub33/fxl-a11y/index.html
```diff
@@ -432,6 +432,28 @@ <h3 class="Subhead">Understanding Depth of Field</h3>
...
+    <pre>
+        &lt;meta property="rendition:layout"&gt;pre-paginated&lt;/meta&gt;
+        &lt;meta property="rendition:orientation"&gt;auto&lt;/meta&gt;
+        &lt;meta property="rendition:spread"&gt;auto&lt;/meta&gt;
+    </pre>
+</aside>
+
+The default value for the <code>orientation</code> and <code>spread</code> properties is <code>auto</code>, which means that the reading system settings or defaults take effect. 
+
+It is recommended that content creators do not set a specific <code>orientation</code> property, as this can interfere with user device preferences. It is especially important for users who may be unable to change the orientation of their device to meet the content settings.
```
dauwhe: perhaps "match the content settings" rather than "meet the content settings" ?

epub33/fxl-a11y/index.html
```diff
@@ -74,7 +74,9 @@
 
 Fixed Layout publications present some unique challenges for accessibility. The requirements laid out in [[EPUB-A11Y-11]] recommend [[WCAG2]] AA, but for many use cases in fixed layout, that might not be possible without fundamental changes to the content.
 
-We want to recognize these challenges for content creators, and in this document will outline some techniques for managing these challenges. We encourage content creators to explore the possibilities that digital publications present, even when creating fixed layout publications.
+In particular, the needs of people with low vision or learning disabilities that rely on the transformation of text are still almost impossible to accommodate in fixed layout content. Content creators concerned about these particular areas are recommended to review whether the content they are producing needs to be in fixed layout.
```
dauwhe: "are recommended to review" is a bit awkward. I wonder if a slightly stronger statement is warranted:

> Content creators concerned about this may choose not to use fixed layout.

epub33/fxl-a11y/index.html
```diff
@@ -260,8 +262,7 @@
 
 If there are a group of images in sequence, you only have to describe details in the first image. In the proceeding images only mention what has changed. W3 Tutorial on Image Groups
 
-Not all details are needed in writing alt-text for images, and what you do describe relies heavily on context. For example, if the image is described in the surrounding text, you only need to briefly describe it in the Alt-text. For more information on when, and how to describe you can go to the DIAGRAM centre, or AccessibilePublishing.ca. See resources for links.
-    
+Not all details are needed in writing alt-text for images, and what you do describe relies heavily on context. For example, if the image is described in the surrounding text, you only need to briefly describe it in the Alt-text. For more information on when, and how to describe you can go to the DIAGRAM centre, or AccessibilePublishing.ca. See resources for links.
```
dauwhe: I don't think alt-text should be capitalized in the second sentence.

dauwhe (approved):

mattgarrish:
Haven't had a chance to read yet, but posting links to avoid the raw diff... 😄 

- [preview](https://cdn.statically.io/gh/w3c/epub-specs/fxl-a11y-additions/epub33/fxl-a11y/index.html)
- [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/fxl-a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fxl-a11y-additions/epub33/fxl-a11y/index.html)

mattgarrish:
Not specific to this PR, but respec is complaining about the [[CSS]] reference on line 374. Probably needs a local biblio reference.

gregoriopellegrino:
Some questions/comments for metadata:

- `displayTransformability`: how do we define an FXL made from SVG with respect to `displayTransformability`? (I am not SVG expert)
- from my point of view the `displayTransformability` value can be used only if the EPUB is based on HTML/CSS
- `displayTransformability`: it's more of a feature that's missing on the reading solutions side of EPUB FXL, but content creators should be careful (at the very least) not to hide content overflow with CSS` overflow:hidden`, avoid overlapping content, etc.

clapierre:
wouldn't you not be able to put in displayTransformability if it is a FXL book?

gregoriopellegrino:
> wouldn't you not be able to put in displayTransformability if it is a FXL book?

Theoretically if the content is in CSS and HTML the transformability of the content can be possible (if the reading solution allows it, but none does at the moment, as far as I know).

clapierre (approved):

gregoriopellegrino (approved):

mattgarrish:
> Theoretically if the content is in CSS and HTML the transformability of the content can be possible (if the reading solution allows it, but none does at the moment, as far as I know).

Right, it's more theoretically possible than practically possible given that most reading systems don't let users modify fxl display properties.

I can't see the legitimacy in setting a property for something that can't actually be accomplished. But to your question, SVG text is stylable, so if the value applies to modifying HTML it should also apply to SVG.

But at some point we need to better clarify what criteria the content has to meet to be "display transformable".

epub33/fxl-a11y/index.html
mattgarrish: It's kind of jargony to call it "alt text", too (no hyphen needed).

Might be better to write it out as "alternative text" all the time, as the document flips between the two. That or put alt text in parentheses after the first use of "alternative text" to indicate it's an abbreviation.

epub33/fxl-a11y/index.html
```diff
@@ -74,7 +74,9 @@
 
 Fixed Layout publications present some unique challenges for accessibility. The requirements laid out in [[EPUB-A11Y-11]] recommend [[WCAG2]] AA, but for many use cases in fixed layout, that might not be possible without fundamental changes to the content.
 
-We want to recognize these challenges for content creators, and in this document will outline some techniques for managing these challenges. We encourage content creators to explore the possibilities that digital publications present, even when creating fixed layout publications.
+In particular, the needs of people with low vision or learning disabilities that rely on the transformation of text are still almost impossible to accommodate in fixed layout content. Content creators concerned about this may choose not to use fixed layout. 
+
+We want to recognize these challenges for content creators, and in this document will outline some techniques for producing more accessible fixed layout content. We encourage content creators to explore the possibilities that digital publications present, even when creating fixed layout publications.
```
mattgarrish: I'm not sure I fully understand what this sentence is saying:

> We encourage content creators to explore the possibilities that digital publications present, even when creating fixed layout publications.

Do you mean something like: "explore the full range of options for accessibility"?

epub33/fxl-a11y/index.html
```diff
@@ -260,8 +262,7 @@
 
 If there are a group of images in sequence, you only have to describe details in the first image. In the proceeding images only mention what has changed. W3 Tutorial on Image Groups
 
-Not all details are needed in writing alt-text for images, and what you do describe relies heavily on context. For example, if the image is described in the surrounding text, you only need to briefly describe it in the Alt-text. For more information on when, and how to describe you can go to the DIAGRAM centre, or AccessibilePublishing.ca. See resources for links.
-    
+Not all details are needed in writing alt-text for images, and what you do describe relies heavily on context. For example, if the image is described in the surrounding text, you only need to briefly describe it in the alt-text. For more information on when, and how to describe you can go to the DIAGRAM centre, or AccessibilePublishing.ca. See resources for links.
```
mattgarrish: I'm not sure if this won't confuse people:

> For example, if the image is described in the surrounding text, you only need to briefly describe it in the alt-text.

Technically, you don't need to do anything but put an empty alt attribute in if the context of the image is already defined in the surrounding text.

It also makes the alt text sound like a trimmed description, which isn't entirely accurate, either.

It might be easier to punt these technical details to the referenced documentation, though. The first sentence tells users enough, so maybe just put the "for example" after the comma and delete the second sentence.

epub33/fxl-a11y/index.html
```diff
@@ -432,6 +432,28 @@ <h3 class="Subhead">Understanding Depth of Field</h3>
...
+
+The [package metadata](https://www.w3.org/TR/epub-33/#sec-fxl-package) used in the EPUB is the primary method for a reading system to determine whether content is fixed layout or reflowable. In addition to identifying the pagination mode with <code>rendition:layout</code>, package metadata can also allow the content creator to have some control over other display characteristics.
+
+These display characteristics include: 
+
+* <code>rendition:orientation</code> - the orientation of the book, can be <code>landscape</code>, <code>portrait</code>, or <code>auto</code>
+* <code>rendition:spread</code> - the spread settings of the book, can be <code>none</code>, <code>landscape</code>, <code>both</code>, or <code>auto</code>
+* <code>rendition:page-spread-*</code> - spread settings for individual spine items, can be <code>left</code>, <code>right</code>, and <code>center</code>
+
+<aside class="example" id="metadata-example-1">
```
mattgarrish, line 459: It might be better to put this example at the end of the section. It's not until you read the next two paragarphs that you'll understand what the auto settings are there for. (Maybe consider adding an explanation for the example, too.)

dauwhe:
> Theoretically if the content is in CSS and HTML the transformability of the content can be possible (if the reading solution allows it, but none does at the moment, as far as I know).

But it's likely such a transform (like increasing font size) would make content unreadable. It's common for every word to be in an absolutely positioned span.

epub33/fxl-a11y/index.html
wareid: Yes I do, so I will fix this!

epub33/fxl-a11y/index.html
wareid: I'm going to resolve this for now as I just want to focus on the changes in the PR, but this section probably deserves some revision.

wareid:
Even with these recommendations, we have to consider legacy FXL content, I don't think we'll ever really see font changes in FXL in reading systems.

epub33/fxl-a11y/index.html
wareid: Fixed!

PR #2013, opened 2022-02-25
Add paragraph about origin-specific security headers being hard to use in EPUB

### LINKED ISSUE #1843: How do you allow EPUBs in CORS / CSP / iframe origin policy?
In the WG discussion for https://github.com/w3c/epub-specs/issues/1061, I mentioned that one consideration is how the developer maintaining an external iframed resource could specify a CORS header that lets EPUBs (or one specific EPUB, or one specific reading system) to load the resource, while blocking some other origins.

The same basic question applies to Content Security Policy and the obsolete X-Frame-Options header: since an EPUB isn't a real webpage, it doesn't have an URL, so how does the developer maintaining an external service specify an EPUB "address" (origin) in a CORS/CSP/X-Frame-Origins directive?

Links on the topic:
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options

dlazin:
As discussed in today's meeting, here is a draft that resolves #1843.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2013.html" title="Last updated on Mar 1, 2022, 8:20 PM UTC (37d730d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2013/8af3f43...37d730d.html" title="Last updated on Mar 1, 2022, 8:20 PM UTC (37d730d)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -8818,10 +8818,11 @@ html.my-document-playing * {
 						<p>Checking for malware and exploits at distribution time is not always reliable, either, as the
 							malicious content can be swapped in any time after publication, unlike resources that come
 							embedded in the EPUB Container.</p>
-						<p>The <a href="#sec-container-iri">origin</a> of an EPUB is both opaque and inconsistent between
-							Reading Systems. Consequently, if the EPUB Creator hosts remote resources on a web server they
-							control, the server effectively cannot use security features that require specifying allowable
-							origins, such as headers for <a href="https://fetch.spec.whatwg.org/#cors-protocol">CORS</a>,
+						<p>The <a href="#sec-container-iri">origin</a> of an EPUB is both unknown to the EPUB Creator and
+							specific to each Reading System implementation. Consequently, if the EPUB Creator hosts remote
+							resources on a web server they control, the server effectively cannot use security features that
+							require specifying allowable origins, such as headers for
+							<a href="https://fetch.spec.whatwg.org/#cors-protocol">CORS</a>,
 							<a href="https://www.w3.org/TR/CSP2/#content-security-policy-header-field"><code>Content-Security-Policy</code></a>,
 							or <a href="https://www.rfc-editor.org/rfc/rfc7034"><code>X-Frame-Options</code></a>.</p>
 					</dd>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -8818,6 +8818,12 @@ html.my-document-playing * {
 						<p>Checking for malware and exploits at distribution time is not always reliable, either, as the
 							malicious content can be swapped in any time after publication, unlike resources that come
 							embedded in the EPUB Container.</p>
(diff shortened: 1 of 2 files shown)


epub33/core/index.html
```diff
@@ -8818,6 +8818,12 @@ <h3>Threat Model</h3>
 						<p>Checking for malware and exploits at distribution time is not always reliable, either, as the
 							malicious content can be swapped in any time after publication, unlike resources that come
 							embedded in the EPUB Container.</p>
+						<p>The <a href="#sec-container-iri">origin</a> of an EPUB is both opaque and inconsistent between
```
iherman: Tiny comment:  it may be only me, but the term "inconsistent" has a slightly pejorative feel to my ears. The origin (essentially, the internal URL chosen for the package content) is reading system/implementation dependent. That might be a better way of describing it.

epub33/rs/index.html
```diff
@@ -2156,6 +2156,12 @@ <h3>Threat Model</h3>
 						<p>Calls to remote resources can also be used to track information about users (e.g., through
 							server logs). Reading Systems should limit the information they expose through HTTP requests
 							to only what is essential to obtain the resource.</p>
+						<p>The <a href="#sec-container-iri">origin</a> of an EPUB is both opaque and inconsistent between
```
iherman: Same comment as above on the term "inconsistent"

iherman (approved):

epub33/core/index.html
mattgarrish: It's not correct to say that origins in EPUB are opaque, either, is it? An opaque origin is a special case where you get a null address, like for a data URL.

As @iherman says, could we just say something like "The origin of an EPUB (i.e., its scheme, hostname, and port) is implementation-specific, so will differ between Reading Systems."

Otherwise, this looks okay to me.

mattgarrish (approved):

epub33/core/index.html
dlazin: Gotcha, how's this?

I didn't realize that "opaque origin" was a term of art, but I see now that it is; I just meant unknown to the author, so I have said that now.

dlazin: Gotcha, how's this?

I didn't realize that "opaque origin" was a term of art, but I see now that it is; I just meant unknown to the author, so I have said that now.

epub33/rs/index.html
dlazin: Done too.

epub33/core/index.html
iherman: LGTM

iherman:
@dauwhe or @wareid, you have the honour of merging...

wareid (approved):

PR #2020, opened 2022-03-01
limit wrapping to pre elements

### LINKED ISSUE #2018: code element preserves line breaks
Looks like the line breaks in the `code` tags are being preserved in the first sub-bullet here: https://www.w3.org/TR/epub-33/#confreq-nav-a-href

Need to verify if that is the problem and whether it's affecting markup elsewhere.

mattgarrish:
There was never a need to include `code` in the white-space handling as `code` is only used for inline element/attribute/property names and similar. It doesn't have the line handling issues of `pre`.

This PR removes code from the CSS declaration, leaving the line breaking only for `pre` tags.

Fixes #2018

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2020.html" title="Last updated on Mar 1, 2022, 3:22 PM UTC (6eabb7c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2020/8af3f43...6eabb7c.html" title="Last updated on Mar 1, 2022, 3:22 PM UTC (6eabb7c)">Diff</a>

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -98,8 +98,7 @@
 				}
 			};</script>
 		<style>
-			pre,
-			code {
+			pre {
 				white-space: break-spaces !important;
 			}</style>
 	</head>
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -112,8 +112,7 @@
 			.varlist {
 				margin-left: 3rem;
 			}
-			pre,
-			code {
+			pre {
 				white-space: break-spaces !important;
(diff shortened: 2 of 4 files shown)


iherman (approved):
One of the CSS issues I did not even know existed :-)

mattgarrish:
Hopefully it gets resolved in the W3C stylesheets in the future so we don't have to hack it in ourselves every time. There's an issue open here: https://github.com/w3c/tr-design/issues/282

PR #2025, opened 2022-03-02
Add section on conformance checking

mattgarrish:
This PR comes out of a discussion we had in the epubcheck group about using that tool to help with CR exit criteria.

In a nutshell, the specification doesn't currently say anything about validators except for the abrupt inclusion of requirements on "validation tools" in the sections on deprecated features.

To try and better address conformance checking as an outcome of the specification, I've made the following changes:

- added a formal definition of an "EPUB Validator" and used the term to replace "validation tools" (intentionally trying to keep the definition lightweight, as I didn't want to start spewing lexical constraints and stuff like that!)
- added a new section under the conformance criteria section that explains the requirements for conformance validation

I've also tried to address various issues that come up around validation - like how not all requirements are testable and why not following recommendations does not make a publication invalid (i.e., blanket rejection of warnings isn't a good idea).

I'm open to this being both too much and not enough. More of a starting point.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2025.html" title="Last updated on Mar 16, 2022, 1:44 PM UTC (62fd1dd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2025/75fe0c3...62fd1dd.html" title="Last updated on Mar 16, 2022, 1:44 PM UTC (62fd1dd)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -10765,10 +10765,12 @@ EPUB/images/cover.png</pre>
 					>Working Group's issue tracker</a>.</p>
 
 			<ul>
-				<li> 19-Feb-2021: Clarified the <a href="#elemdef-smil-audio"><code>audio</code></a> element's
-					definition by making it optional, and adapt the specification's text elsewhere to address the
-					situation when the element is indeed not present. See <a
-						href="https://github.com/w3c/epub-specs/issues/1986">issue 1986</a>. </li>
+				<li>02-Mar-2022: Add new section on conformance checking and definition for EPUB Validator. See <a
+						href="https://github.com/w3c/epub-specs/pull/2025">pull request 2025</a>.</li>
+				<li>19-Feb-2021: Clarified the <a href="#elemdef-smil-audio"><code>audio</code></a> element's definition
+					by making it optional, and adapt the specification's text elsewhere to address the situation when
+					the element is indeed not present. See <a href="https://github.com/w3c/epub-specs/issues/1986">issue
+						1986</a>.</li>
 				<li>04-Feb-2021: Expanded the section on security and privacy to include new sections on the threat
 					model for EPUB Publications and additional recommendations for ensuring security and privacy. See <a
 						href="https://github.com/w3c/epub-specs/issues/1871">issue 1871</a>, <a
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -683,10 +683,10 @@
 						improve the interoperability of EPUB Publications across <a>Reading Systems</a>.</p>
 
 					<p>An EPUB Validator MUST be capable of checking the machine-processable normative requirements of
-						this specification and reporting on their presence. As these types of issues will generally
-						result in EPUB Publications not rendering, or rendering in inconsistent ways, it is RECOMMENDED
-						that EPUB Validators report them at the highest severity (e.g., as errors and/or critical
-						errors).</p>
+						this specification (practices identified by the keywords "MUST", "MUST NOT", and "REQUIRED") and
+						reporting on their presence. As these types of issues will generally result in EPUB Publications
+						not rendering, or rendering in inconsistent ways, it is RECOMMENDED that EPUB Validators report
+						them at the highest severity (e.g., as errors and/or critical errors).</p>
 
 					<div class="note">
 						<p>It will not be possible for an EPUB Validator to check all the requirements as some issues
@@ -694,10 +694,11 @@
 							are rendered (i.e., due to scripting), etc.</p>
 					</div>
 
-					<p>Although an EPUB Validator SHOULD also check the machine-processable normative recommendations of
-						this specification, not following recommended practices does not result in an invalid EPUB
-						Publication. It is RECOMMENDED that EPUB Validators report these types of issues at a lower
-						severity (e.g., as warnings).</p>
+					<p>An EPUB Validator SHOULD also check the machine-processable normative recommendations of this
+						specification (practices identified by the keywords "SHOULD", "SHOULD NOT", and "RECOMMENDED").
+						Not following recommended practices does not result in an invalid EPUB Publication, however. It
+						is RECOMMENDED that EPUB Validators report these types of issues at a lower severity (e.g., as
+						warnings).</p>
 
 					<div class="note">
 						<p>Vendors, distributors, and other retailers of EPUB Publications should consider the
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -379,6 +379,12 @@
 							conformant with this specification.</p>
 					</dd>
 
+					<dt><dfn id="dfn-epub-validator" data-lt="EPUB Validators">EPUB Validator</dfn></dt>
+					<dd>
+						<p>An application that verifies the requirements of this specification against <a>EPUB
+								Publications</a> and reports on their conformance.</p>
+					</dd>
+
 					<dt>
 						<dfn id="dfn-file-name" data-lt="File Names">File Name</dfn>
 					</dt>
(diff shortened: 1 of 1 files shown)


epub33/core/index.html
```diff
@@ -667,6 +673,47 @@ <h3>Conformance Criteria</h3>
...
+						that EPUB Validators report them at the highest severity (e.g., as errors and/or critical
+						errors).</p>
+
+					<div class="note">
+						<p>It will not be possible for an EPUB Validator to check all the requirements as some issues
+							may require human inspection, some may not be checkable until <a>EPUB Content Documents</a>
+							are rendered (i.e., due to scripting), etc.</p>
+					</div>
+
+					<p>Although an EPUB Validator SHOULD also check the machine-processable normative recommendations of
```
iherman: I presume this is a cut-and-paste issue; this section talks about non-normative features, right?

```suggestion
					<p>Although an EPUB Validator SHOULD also check the machine-processable non-normative recommendations of
```

iherman (changes requested):
I am a bit worried about the usage of _MUST_-s and _SHOULD_-s in this section. Do we really think this section should be normative and therefore warrant these terms? If it is normative, then we have to check them, we have to formally test them, etc. I know this is doable because we do have epubcheck, but I wonder whether we create ourselves a bunch of problems for no good reasons. Flagging the section as informative (and turn all _MUST_-s and _SHOULD_-s into must-s and should-s) might be enough...

epub33/core/index.html
mattgarrish: Isn't it:

- normative requirements = MUST/REQUIRED
- normative recommendations = SHOULD/RECOMMENDED

Non-normative recommendations are best pratices in notes and such, no?

(I don't know what to make of MAY/OPTIONAL, but I'm not sure we need to say anything about those for validation.)

mattgarrish:
> I am a bit worried about the usage of _MUST_-s and _SHOULD_-s in this section.

Right, I have the same concern, but I was trying to harmonize this with the fact that we normatively describe what validators have to do for handling deprecated features.

Maybe we should also remove the normative keywords from the deprecated features section and keep all talk of validators informative?

rdeltour:
@iherman 
> I am a bit worried about the usage of MUST-s and SHOULD-s in this section. Do we really think this section should be normative and therefore warrant these terms?

I thought the point of this section was to justify using EPUBCheck for CR? If yes then we need it to be normative, no?

@mattgarrish 
I too was originally confused by "requirements" vs. "recommendations". These are not defined anywhere as far as I can tell, specifically, not in RFC2119. 

Does this section need to discuss severity at all?

What about using only something very generic along the lines of [what is said in HTML](https://html.spec.whatwg.org/#conformance-classes):
> Conformance checkers must verify that a document conforms to the applicable conformance criteria described in this specification.

and then precise that they checking scripts is optional.

mattgarrish:
> These are not defined anywhere as far as I can tell, specifically, not in RFC2119.

I'd argue they're implicit in the definitions. A MUST is a requirement of its specification as are it's synonym (REQUIRED) and negative (MUST NOT).

But we can, of course, make this explicit.

rdeltour:
> I'd argue they're implicit in the definitions.

Yeah, it's too confusing IMO. Also, "requirements" is often use to encompass them all. For example, RFC2119 talks about varying "requirement levels", so RECOMMENDED is one specific requirement level 🙃. Another example is our own section titles ("XHTML Requirements"), or when we say things like "This specification defines the authoring requirements for EPUB Publications".

I would advise against using "requirements" vs. "recommendations" altogether.

rdeltour:
But again I'm not sure why we need to make two distinct cases anyway, unless we really want to talk about severity (but then, why?).

What you're saying is basically: validator MUST report MUSTs, and SHOULD report SHOULDs.

What I'm proposing is: validator MUST check conformance, unless it's impossible to check by a machine in which case its OPTIONAL.

mattgarrish:
> What about using only something very generic along the lines of [what is said in HTML](https://html.spec.whatwg.org/#conformance-classes):

Ya, I had a look at that when I was writing the section up, but the sticking point (for me) is that we normatively say how to report the issues arising from unsupported features.

If we don't say how to report issues generally, why normatively say how to handle them there?

Once we start dropping the normative keywords we do end up back at a simpler statement like HTML makes, but it also suggests validators are outside the scope of the specification.

iherman:
> @iherman
> 
> > I am a bit worried about the usage of MUST-s and SHOULD-s in this section. Do we really think this section should be normative and therefore warrant these terms?
> 
> I thought the point of this section was to justify using EPUBCheck for CR? If yes then we need it to be normative, no?
> 

I do not think that is necessarily a conclusion.

If we define, normatively, the behaviour of an EPUB Validator, then we have to test that as part of CR. Ie, we have to test whether EPUBCheck is a conformant validator in CR, when we use EPUBCheck as a validator for CR... this is kind of a circular argument that I am not sure how to solve.

I do not see any issue if the full section is non-normative, setting informal expectation (and I believe having that section is a great idea!) and we rely on the community usage of EPUBCheck as part of the CR process.

mattgarrish:
> If we define, normatively, the behaviour of an EPUB Validator, then we have to test that as part of CR.

Ya, this is what really worries me. We want epubcheck as a proof without it being subject to exit criteria itself.

See if the latest commit helps at all. I've made the validator section informative and rewritten it to tie it into being a proof of the enforceability of the specification requirements. I also took out the parts about reporting severity and instead tried to explain why reporting these issues is useful to authors.

I also changed the unsupported features so the statements about how to report the features are in notes.

iherman:
> 
> See if the latest commit helps at all. I've made the validator section informative and rewritten it to tie it into being a proof of the enforceability of the specification requirements. I also took out the parts about reporting severity and instead tried to explain why reporting these issues is useful to authors.
> 
> I also changed the unsupported features so the statements about how to report the features are in notes.

That works for me, thank you.

iherman (approved):

epub33/core/index.html
```diff
@@ -379,6 +379,12 @@ <h3>Terminology</h3>
 							conformant with this specification.</p>
 					</dd>
 
+					<dt><dfn id="dfn-epub-validator" data-lt="EPUB Validators">EPUB Validator</dfn></dt>
```
rdeltour: I personally prefer "conformance checker" instead of "validator". It is closer to spec language and is a bit less grim.

rdeltour (commented):
TBH if this section is informative I'm not sure I understand how it improves the specification.

To paraphrase a bit, my understanding of this sections is:

- 1st paragraph "authors should check their EPUBs" → I think that's an accepted practice already. At least to anyone reading this specification.
- 2nd paragraph "EPUB checkers are useful if thorough" → OK…
- 3d and 4th paragraph say "validator must check MUSTs and should check SHOULDs" → why not, but note that it's not the philosophy of EPUBCheck, which is more "must check everything when possible". Again I don't see how the differentiation between MUSTs and SHOULD is useful here?


Essentially I'm wondering: what are we exactly intending to clarify with this section?
Depending, maybe it can be condensed into a simpler note in the preceding "Conformance criteria" section?

wareid (approved):

mattgarrish:
> TBH if this section is informative I'm not sure I understand how it improves the specification.

It doesn't define validator requirements, but straying into those was always problematic. Short of writing rules on how to report every violation, it's pointless to sprinkle the specification with a few requirements on how to report issues. Adding a few more vague normative requirements isn't making things better.

Where I see it helping is in providing a place to surface notions about validation that we know to be true but have never been voiced. That recommended practices are not violations of the specification, for example. It may change nothing in the real world, but at least it documents that it's not the specification that is the issue when content is rejected for warnings.

It also provides a place to formally point readers to epubcheck, same as you can find your way to validator.nu from HTML's section.

Not everyone who arrives at the specification is going to understand conformance checking. The latest commit rewrites the section some more to focus it more directly on authoring concerns, so there's no focus on what a conformance checker is expected to do.

shiestyle (approved):

mattgarrish:
Can this be merged or do we need to discuss it any more?

mattgarrish:
As there hasn't been any further comment on this for a couple of weeks, I'm going to merge. If further tweaking is necessary, we can always do in another issue.

PR #2027, opened 2022-03-02
Adding a note on duration and TTS

### LINKED ISSUE #2007: 'duration' value in MO and TTS
[§ 7.3.5.2 of the content document](https://www.w3.org/TR/epub-33/#sec-mo-package-metadata) says that  each MO document _MUST_ have a duration value in the package document (via the [`media:duration`](https://www.w3.org/TR/epub-33/#sec-duration) property) and, furthermore, the duration for the whole EPUB Publication _MUST_ also be set.

On the other hand the [section on TTS in the RS document](https://www.w3.org/TR/epub-rs-33/#sec-text-to-speech) says:

> The Media Overlay text element's lifespan corresponds to the rendering time of the associated speech synthesis. The implicit duration of the text element (and by inference, of the parent par element) is therefore determined by the execution of the Text-to-Speech engine, and cannot be known at authoring time.

I.e., setting the (required) duration values in the package document sounds, in this case, quite meaningless.

Recognizing that the TTS is not a major usage of MO, I would not think we should change the aforementioned normative requirements at this point. Maybe we can just add some words in the TTS section, acknowledging that the duration values may be approximative. Ie, add something like this to the end of that paragraph:

> This also means that the `duration` values set in the package document at authoring time may be considered as approximative only, and should be considered as such by the Reading System.

cc @wareid @dauwhe @mattgarrish @dlazin @marisademeglio

iherman:
Implements the proposal in #2007

Fix #2007

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1835,7 +1835,10 @@
 						associated speech synthesis. The implicit duration of the <code>text</code> element (and by
 						inference, of the parent <code>par</code> element) is therefore determined by the execution of
 						the Text-to-Speech engine, and cannot be known at authoring time (factors like speech rate,
-						pauses and other prosody parameters influence the audio output).</p>
+						pauses and other prosody parameters influence the audio output).
+						This also means that the <a data-cite="epub-33#duration"><code>duration</code></a>
+							property values set in the package document at authoring time may be considered as approximative only, and should be considered as such by the Reading System.
+					</p>
 				</section>
 			</section>
 
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1836,8 +1836,7 @@
 						inference, of the parent <code>par</code> element) is therefore determined by the execution of
 						the Text-to-Speech engine, and cannot be known at authoring time (factors like speech rate,
 						pauses and other prosody parameters influence the audio output).
-						This also means that the <a data-cite="epub-33#duration"><code>duration</code></a>
-							property values set in the package document at authoring time may be considered as approximative only, and should be considered as such by the Reading System.
+						This also means that Reading Systems should treat the data-cite="epub-33#duration"><code>duration</code></a> property values set in the Package Document as approximative when making use of them.
 					</p>
 				</section>
 			</section>


wareid (approved):

epub33/rs/index.html
```diff
@@ -1835,7 +1835,10 @@ <h5>Text-to-Speech</h5>
 						associated speech synthesis. The implicit duration of the <code>text</code> element (and by
 						inference, of the parent <code>par</code> element) is therefore determined by the execution of
 						the Text-to-Speech engine, and cannot be known at authoring time (factors like speech rate,
-						pauses and other prosody parameters influence the audio output).</p>
+						pauses and other prosody parameters influence the audio output).
+						This also means that the <a data-cite="epub-33#duration"><code>duration</code></a>
+							property values set in the package document at authoring time may be considered as approximative only, and should be considered as such by the Reading System.
```
mattgarrish: Isn't their being approximative more definitive than the current statement suggests? I'd simplify this more along the lines of:

> This also means that Reading Systems should treat the `duration` property values set in the Package Document as approximative only when making use of them.

(Seems almost normative, except we don't define how to make use of them.)

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-03-03-epub#section1) on 2022-03-03

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-03-03-epub#resolution1):  Merge PR 2027 after wording is confirmed, and close issue 2007.***

### 1. 'duration' value in MO and TTS (issue epub-specs#2007)

_See github issue [epub-specs#2007](https://github.com/w3c/epub-specs/issues/2007)._




**Dave Cramer:** Duration values in TTS.  

_See github pull request [epub-specs#2027](https://github.com/w3c/epub-specs/pull/2027)._




**Dave Cramer:** see issue and pull request.  
… historically we required a duration value for MO, but there is a question about what happens when there is no accompanying audio.  
… which means TTS will render the paragraph.  
… We are acknowledging it isn't common, but we need to cover it in the spec.  
… Matt, summary?.  

**Matt Garrish:** Not much to say, but the spec is a little wordy/redundant.  
… I suggest shortening, but overall am happy.  

**Dave Cramer:** Some comment about making the duration a 'should', but I am against that.  

**Matt Garrish:** agreed. I am not even sure how common TTS is.  

**Dave Cramer:** And you won't have MO if you are relying on TTS anyway.  

**Wendy Reid:** Using MO for TTS is pretty much overkill.  
… European A11y act requires TTS in RSes, so TTS will be more common.  

**Matt Garrish:** We did TTS previously, but not as MO. Seems like this is more theoretical than anything. Happy to leave as is.  

**Murata Makoto:** DAISY in Japan has MO always, but it is pre-recorded.  

**Dave Cramer:** We will clean up the PR and close the issue.  

> **Proposed resolution: Merge PR 2027 after wording is confirmed, and close issue 2007.** *(Wendy Reid)*

> *Matt Garrish:* +1.

> *Dave Cramer:* +1.

> *Toshiaki Koike:* +1.

> *Wendy Reid:* +1.

> *Shinya Takami (高見真也):* +1.

> *Masakazu Kitahara:* +1.

> *Zheng Xu (徐征):* +1.

> *Brady Duga:* +1.

> *Dan Lazin:* +1.

> *Murata Makoto:* +1.

> ***Resolution #1: Merge PR 2027 after wording is confirmed, and close issue 2007.***

</details>

epub33/rs/index.html
iherman: Adopted the change

iherman:
Merging per WG resolution.

Issue #1888, opened 2021-10-30
What base URLs to use for URL parsing in EPUB?


rdeltour:
(or: The Big Mystery of Spooky EPUB Relative URLs 👻🎃)

**TL;DR**: EPUB 3.3 now normatively references the [URL Standard](https://url.spec.whatwg.org). But URL parsing is ambiguous in some cases, because base URLs are not clearly defined.

⚠️  EPUB references in the problem statement below point to a dated version of the EPUB 3.3 working draft (Oct 29, 2021). Do not copy out of context! 😉

## Current situation

In an EPUB, files reference each other via relative URL strings (see [Relative URLs, in Open Container Format](https://www.w3.org/TR/2021/WD-epub-33-20211029/#sec-container-iri)).
In the URL standard, to parse a relative URL string into URL records, the [URL parser](https://url.spec.whatwg.org/#concept-url-parser) needs a base URL.

The base URL used to parse a URL string is defined by host languages (like in [CSS](https://www.w3.org/TR/css-values-4/#relative-urls), or [HTML](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#resolving-urls)). Typically, it is the URL of the document containing the URL string.

EPUB defines what base URL to use for URL parsing in two cases:
- relative URL strings found in documents located in the `META-INF` directory
- relative URL strings in the Package Documents

### Parsing a URL in documents located in the `META-INF` directory

For documents in the `META-INF` directory, URL strings must be parsed using the root directory as the base URL (see [Relative URLs, in Open Container Format](https://www.w3.org/TR/2021/WD-epub-33-20211029/#sec-container-iri)).

The problem is that [*Root Directory*](https://www.w3.org/TR/2021/WD-epub-33-20211029/#dfn-root-directory) is not defined as a URL, but quite abstractly as "the base of the OCF Abstract Container". The spec also says the root directory is "virtual in nature". In fact, RS may or may not generate a physical directory for the root directory (see [OCF ZIP Container RS processing](https://www.w3.org/TR/2021/WD-epub-rs-33-20211029/#confreq-zip-rootdir)).

### Parsing a URL in the Package Document

For Package Documents, URL strings must be parsed uses the URL of the Package Document as the base URL (see [Parsing Relative URLs, in Package Documents RS processing](https://www.w3.org/TR/2021/WD-epub-rs-33-20211029/#sec-pkg-doc-relative-urls)).

[...]

PR #2028, opened 2022-03-03
Restore requirement that relative URLs resolve to resources

### LINKED ISSUE #2024: Clarify that resources must be present for relative paths
It appears in our rewrite of URL processing that we've lost the requirement for there to be a resource at the end of relative paths. In 3.2, we had this statement:

> All relative IRI references MUST resolve to resources within the OCF Abstract Container (i.e., at or below the Root Directory).

This was meant to cover both that the URL not leak outside the container and that there must be a resource at the end of it.

Our new rewrite covers URLs not leaking outside the container better, but we no longer appear to have a requirement that there be a resource at the destination. Without that bit, someone could argue that missing resources in the container aren't really errors.

We should add back a statement about the need for a resource to be present so that we're consistent again with previous versions of EPUB 3.

mattgarrish:
I've added a requirement that valid-relative-container-URL-with-fragment strings resolve to resources in the OCF Abstract Container to the end of 6.1.5.

Fixes #2024

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2028.html" title="Last updated on Mar 10, 2022, 11:30 AM UTC (e574bae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2028/ec5a4a3...e574bae.html" title="Last updated on Mar 10, 2022, 11:30 AM UTC (e574bae)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -6045,12 +6045,13 @@ No Entry</pre>
 						<li>Return <var>false</var>.</li>
 					</ol>
 
-					<p> In the <a>OCF Abstract Container</a>, any URL string MUST be an <a
+					<p>In the <a>OCF Abstract Container</a>, any URL string MUST be an <a
 							data-cite="url#absolute-url-with-fragment-string">absolute-URL-with-fragment-string</a> or a
-							<a>valid-relative-container-URL-with-fragment string</a>. </p>
+							<a>valid-relative-container-URL-with-fragment string</a>.</p>
 
-					<p>In addition, all <a>valid-relative-container-URL-with-fragment strings</a> MUST resolve to
-						resources within the OCF Abstract Container.</p>
+					<p>In addition, all <a data-cite="url#relative-url-with-fragment-string">relative-URL-with-fragment
+							strings</a> [[URL]] MUST, after <a data-cite="url#concept-url-parser">parsing</a>, be equal
+						to the <a>Content URL</a> of an existing file in the OCF Abstract Container.</p>
 
 					<div class="note">
 						<p> The properties of the <a>container root URL</a> are such that whatever the amount of <a
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -6049,6 +6049,9 @@ No Entry</pre>
 							data-cite="url#absolute-url-with-fragment-string">absolute-URL-with-fragment-string</a> or a
 							<a>valid-relative-container-URL-with-fragment string</a>. </p>
 
+					<p>In addition, all <a>valid-relative-container-URL-with-fragment strings</a> MUST resolve to
+						resources within the OCF Abstract Container.</p>
+
 					<div class="note">
 						<p> The properties of the <a>container root URL</a> are such that whatever the amount of <a
 								data-cite="url#double-dot-path-segment">double-dot path segments</a> in a URL string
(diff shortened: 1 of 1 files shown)


iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-03-03-epub#section3) on 2022-03-03

List of resolutions:

- ***[Resolution No. 2](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-03-03-epub#resolution2):  Merge PR 2028, close issue 2024.***

### 3. Clarify that resources must be present for relative paths (issue epub-specs#2024)

_See github issue [epub-specs#2024](https://github.com/w3c/epub-specs/issues/2024)._




**Dave Cramer:** Our other issue is similar. When we did this we lost a requirement that there be a resource at the end of the path.  
… originally we had 1. not leak and 2. must exist.  

_See github pull request [epub-specs#2028](https://github.com/w3c/epub-specs/pull/2028)._




**Dave Cramer:** We lost part 2. We should put it back. There is already a PR.  

**Matt Garrish:** We just need to put the existence text back.  

**Dave Cramer:** Yes, and it sounds like the PR does just that.  
… I am in favor of merging.  

**Dan Lazin:** I don't know if it follows that there must always be a thing that is referenced.  
… For instance a book on how to write epub that shows errors (eg this is what it looks like when a resource is missing).  
… The cases are rare, but I can imagine some.  
… and there are plenty of ways to create bad content.  

**Wendy Reid:** Going with must on this, because nothing stops people from ignoring epubcheck.  
… In general epubcheck is heavily relied on by publishers to tell them of problems.  
… and this will more often than not help those publishers find problems.  

**Matt Garrish:** I basically agree with wendyreid.  
… epubcheck does look for this now.  
… There are other things epubcheck will flag.  
… This seems like a can of worms to change, and error has worked really well to date.  

**Dan Lazin:** Those are persuasive arguments.  
… I thought epubcheck gave warnings on shoulds, isn't that good enough?.  

**Wendy Reid:** Errors scare them more.  

**Dan Lazin:** I am ok with must, but odd that epubcheck is preflight and a conformance checker.  

> **Proposed resolution: Merge PR 2028, close issue 2024.** *(Wendy Reid)*

**Dave Cramer:** Propose merging 2028.  

> *Dave Cramer:* +1.

> *Matt Garrish:* +1.

> *Dan Lazin:* +1.

> *Shinya Takami (高見真也):* +1.

> *Brady Duga:* +1.

> *Wendy Reid:* +1.

> *Zheng Xu (徐征):* +1.

> *Toshiaki Koike:* +1.

> ***Resolution #2: Merge PR 2028, close issue 2024.***

**Dave Cramer:** AOB?.  

</details>

iherman (approved):

mattgarrish:
@rdeltour I'm wondering if this wording doesn't really work? The algorithm for determining if you have a valid-relative-container-URL-with-fragment string gives you a true or false value in return, not a URL that you can use to check if a resource exists at the location.

A simple check of the string as relative path could leak back outside the container.

rdeltour:
> @rdeltour I'm wondering if this wording doesn't really work?

Yeah, what "resolve" means for a URL string is not strictly defined.

Alternatively, we could use something along the lines of what I proposed back in https://github.com/w3c/epub-specs/issues/1888#issuecomment-961502739:

> All relative-URL-with-fragment strings MUST, after parsing, be equal to the [container URL](#) of an existing file in the OCF Abstract Container.

or more precisely:

> The result of parsing  relative URL string found in the OCF Abstract Container, with the base URL defined by the context where it is found, must be the [container URL](#) of an existing file in the OCF Abstract Container.

although it is a bit of a mouthful 🤔

mattgarrish:
> although it is a bit of a mouthful 🤔

Ya, that's a tough read! But what is a container URL? We don't define that in the spec, unless I'm missing something (the link doesn't go anywhere).

rdeltour:
> But what is a container URL? We don't define that in the spec, unless I'm missing something (the link doesn't go anywhere)

Sorry I copied that from the old proposal 🤦‍♂️. We now define this as [Content URL](https://www.w3.org/TR/epub-33/#dfn-content-url).

mattgarrish:
I've updated the PR with the first proposed text. I'm hoping the reference to parsing is enough, and we don't have to get into the details of resolving base URLs... 😄

PR #2029, opened 2022-03-04
Extended the note on URL-s

### LINKED ISSUE #2023: Adding a note to the section on OCF URL resolutions
The specification disallows, in the [section on URLs in OCF](https://w3c.github.io/epub-specs/epub33/core/#sec-container-iri) the usage of URLs like `/root/relative/urls` or `../../../crazy/path/urls`. This is done through the statement

> In the [OCF Abstract Container](https://w3c.github.io/epub-specs/epub33/core/#dfn-ocf-abstract-container), any URL string MUST be an [absolute-URL-with-fragment-string](https://url.spec.whatwg.org/#absolute-url-with-fragment-string) or a [valid-relative-container-URL-with-fragment string](https://w3c.github.io/epub-specs/epub33/core/#dfn-valid-relative-container-url-with-fragment-string).

The fact that this is restricted has been around for a while, and reflects the current status on RS implementations. However, to realize this is the case requires to either some links to the URL spec, or follow through the algorithm that precedes that statement. Both are a difficult read.

Maybe we should add an editorial note (or extend the note that follows this statement) making this restrictions explicit.

iherman:
This is a proposal to fix #2023.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2029.html" title="Last updated on Mar 8, 2022, 1:30 PM UTC (c7e472e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2029/47e0f2e...c7e472e.html" title="Last updated on Mar 8, 2022, 1:30 PM UTC (c7e472e)">Diff</a>

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1348,19 +1348,14 @@
 						copy of the same EPUB Publication, the origins will be different for the two users on those
 						copies even if the same Reading System is used.</p>
 
-					<p class="note">
-						The properties of the <a>container root URL</a> are such that a conforming Reading System will parse any relative URL
-						string to a <a>content URL</a>. In other words, relative links do not "leak" outside the container content, which is an important
-						feature for security. This includes URLs strings starting with a
-						<code>/</code> character (for example, <code>/EPUB/content.xhtml</code>) or containing
-						more <a data-cite="url#double-dot-path-segment">double-dot path segments</a> than needed to reach the target
-						(for example, <code>EPUB/../../../../config.xml</code>). (Although those links are
-						<a data-cite="epub-33#urls-in-ocf-constraints">non-conformant</a> in an EPUB Publications.)
-					</p>
-
 					<div class="note">
-						<p>The required properties of the container root URL are such that it behaves similarly to a URL
-							defined as follows:</p>
+						<p>
+							The properties of the <a>container root URL</a> are such that a conforming Reading System will parse any relative URL
+							string to a <a>content URL</a>. In other words, relative links do not "leak" outside the container content, which is an important
+							feature for security.
+						</p>
+
+						<p>In practice, the container root URL behaves similarly to a URL defined as follows:</p>
 
 						<table class="zebra">
 							<thead>
@@ -1406,16 +1401,16 @@
 									</td>
 								</tr>
 								<tr>
-									<td> Package Document </td>
+									<td>Package Document</td>
 									<td>
-										<code>OPS/package.opf</code>
+										<code>EPUB/package.opf</code>
 									</td>
 									<td>
-										<code>http://localhost:49152/OPS/package.opf</code>
+										<code>http://localhost:49152/EPUB/package.opf</code>
 									</td>
 								</tr>
 								<tr>
-									<td> Content Document </td>
+									<td>Content Document</td>
 									<td>
 										<code>HTML/file name.xhtml</code>
 									</td>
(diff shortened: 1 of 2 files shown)


iherman:
I did not refer to security issues in the paragraph on path-absolute-URL strings (I hope that is the right reference) because... I am not sure there is one. In other words, in the light of the definitions earlier in the section, the backward compatibility with RS-s seems to be (to me at least) the only reason why we have this restriction...

rdeltour (commented):
The proposed change is fine.

But I'm wondering if we can even rework the entire note more aggressively to make it more readable:

> NOTE
> These constraints on URL strings mean that:
> - relative URL strings starting with a `/` (U+002F) (for example, `/EPUB/content.xhtml`) are disallowed
> - relative URL strings containing more double-dot path segments than needed to reach the target file (for example, `EPUB/../../../../config.xml`) are disallowed
> - any other absolute or relative URL string is allowed
>
> Note that in any case, to avoid potential run-time security issues, the properties of the container root URL are such that a conforming Reading System will parse any relative URL string to a content URL. In other words, even the disallowed URL strings described above will not "leak" outside the container. They are still disallowed for better interoperability with non-conforming or legacy Reading Systems.

rdeltour:
I'm also thinking that the part about run-time security could be duplicated over in the note of the RS spec, to shed more light on the rationale for the container root URL properties for RS folks.

iherman:
@rdeltour I am fine with both proposals (you are bolder than I am for making changes:-). I am happy to do it if @mattgarrish agrees.

mattgarrish:
I find the list approach a lot more readable. Nothing wrong with stating the truth in no uncertain terms...

iherman:
Ok, I will adopt Romain's text, and add a copy to the reading system

iherman:
@mattgarrish could you have a quick look at the new text (including the RS) to check whether I made some mistake? I guess if things are fine, per the WG discussion, this could be merged.

mattgarrish:
Do we need to repeat the note in the RS specification? Isn't it repeating the what the list just above it defines?

iherman:
> Do we need to repeat the note in the RS specification? 

I believe it is a good idea to keep it. The detailed constraint on URL is _not_ present in the RS spec, it is only defined in the content spec so, in some sense, it is even more important for readability to have it in the RS spec than in the core.

> Isn't it repeating the what the list just above it defines?

I do not think so. The definition just above the note describes the behaviour of the Container Root URL, while the note refers the type of URLs we allow to have in an EPUB Spec.

rdeltour:
> Do we need to repeat the note in the RS specification?

I think the most relevant for RS implementors is to explain the rationale for the root URL constraints: i.e. a relative URL string must always be parsed such as it doesn't leak out of the container, for security reasons (= the second part of the note).

The note might also mention that "`/blahblah`" and "`../../blahblah`" URLs are non-conforming (= first part of the note), but it is not crucial info for RS implementors as long as they abide by the container root URL constraints.

So I wouldn't duplicate the note exactly as-is, but slightly re-order or rephrase to make the RS-centric part stand out.

iherman:
I see your point. I have significantly shortened the note.

rdeltour:
> I see your point. I have significantly shortened the note.

Looks better to me 👍😊

Also, what about merging that with the next note? Both start with "the properties are such that…": they're both clarifying the same concept.

I'd put the new note content first, and then follow with the `localhost` comparison with `s/The required properties of the container root URL are such that it /Practically, the container root URL /`.
(I hope that's clear enough, lmk if I should edit the PR directly 😅).

I'm now wondering if counter-examples of invalid root URLs would be helpful too 🤔

iherman:
Actually, I did a bit more. I merged the two notes but, adapting to the style of what used to be the second note, I moved the example of the tricky relative URL-s into a separate table using the `localhost` used already.

epub33/rs/index.html
```diff
@@ -1415,6 +1420,45 @@ <h4>URL of the Root Directory</h4>
 								</tr>
 							</tbody>
 						</table>
+
+						<table class="zebra">
+							<thead>
+								<tr>
+									<td style="font-weight: bold; text-align: center;">URL<br/>(relative to the package document)</td>
```
rdeltour: I'd call this: "URL string (found for example in the package document)"

rdeltour (approved):

PR #2033, opened 2022-03-05
Consolidate the usage requirements for manifest item properties

### LINKED ISSUE #2030: Manifest properties requirements duplicative and contradictory
This statement in the item element's definition contradicts the requirements in the vocabulary:

>EPUB Creators MUST declare all applicable descriptive metadata properties for each Publication Resource in this attribute (e.g., if the resource is the [cover image](https://w3c.github.io/epub-specs/epub33/core/#sec-cover-image), contains [scripting](https://w3c.github.io/epub-specs/epub33/core/#sec-scripted), or references [remote resources](https://w3c.github.io/epub-specs/epub33/core/#sec-remote-resources)).

It's true of all but the `cover-image`, which doesn't have to be identified. This is the problem that typically arises when we make the same requirements in different places. It feels like we should make this informative (possible a note?) and refer to the vocabulary for the requirements.

Same goes for the next sentence that requires the nav property be set, which technically means we require it be set three times (the paragraph above requires it as does its definition).

mattgarrish:
This PR creates a Resource Properties section under the manifest item element's definition. I've consolidated the many statements requiring the use of these properties into this one section, but otherwise nothing significant changes.

Fixes #2030

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2033.html" title="Last updated on Mar 7, 2022, 1:29 PM UTC (914790e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2033/2283319...914790e.html" title="Last updated on Mar 7, 2022, 1:29 PM UTC (914790e)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -2963,7 +2963,7 @@ XHTML:
 										</li>
 										<li>
 											<p>
-												<a href="#attrdef-properties">
+												<a href="#sec-item-resource-properties">
 													<code>properties</code>
 												</a>
 												<code>[optional]</code>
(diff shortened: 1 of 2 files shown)


iherman (approved):
I like it!

PR #2038, opened 2022-03-08
Reword restriction on multiple overrides

### LINKED ISSUE #2037: Make restriction on multiple overrides more direct
For the rendering property spine overrides we say:

> EPUB Creators MAY use only one of these overrides on any given spine item.

But this seems like it would be better worded as a clear restriction:

> EPUB Creators MUST NOT use more than one of these overrides on any given spine item.

mattgarrish:
Fixes #2037

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2038.html" title="Last updated on Mar 8, 2022, 1:00 PM UTC (e397aa0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2038/2283319...e397aa0.html" title="Last updated on Mar 8, 2022, 1:00 PM UTC (e397aa0)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -2444,9 +2444,9 @@
 								contributor expression by defining the role of the person.</li>
 						</ul>
 
-						<p>EPUB Creators MAY use subexpressions to refine the meaning of other subexpressions,
-								thereby creating chains of information. Refinement chains MUST NOT contain circular
-								references or self-references.</p>
+						<p>EPUB Creators MAY use subexpressions to refine the meaning of other subexpressions, thereby
+							creating chains of information. Refinement chains MUST NOT contain circular references or
+							self-references.</p>
 
 						<p class="note">All the DCMES [[DCTERMS]] elements represent primary expressions, and permit
 							refinement by meta element subexpressions.</p>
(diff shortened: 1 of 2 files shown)


epub33/core/index.html
```diff
@@ -5501,8 +5501,8 @@ <h4>Spread Placement</h4>
 						properties on the spine items for the two adjacent EPUB Content Documents, and omit the
 						properties on spine items where one-up or two-up presentation is equally acceptable.</p>
 
-					<p>EPUB Creators MAY only declare one <code>page-spread-*</code> property on any given spine
-						item.</p>
+					<p>EPUB Creators MUST NOT declare more than one <code>page-spread-*</code> property on any given
+						spine item.</p>
```
iherman, line 5506: The github interface does not allow me to comment on older lines but: respec reports missing links for the terms `page-spread-left` and `page-spread-right` properties in the description list at the beginning of §5.2.4. I suspect they should all be without `page-`.

iherman (approved):

epub33/core/index.html
mattgarrish: I think it may have been a problem with however you're viewing the document, as I'm not getting any errors. The links work fine for me. Was the import of the vocabulary maybe failing?

At any rate, I'm going to merge and see what happens, as the build should have failed if the links were broken.

PR #2043, opened 2022-03-08
FXL A11y Issues 

### LINKED ISSUE #2012: Accessibility of image-dominated FXL publications
In Japan, most EPUB publications contain almost no text.  I propose to add some remarks about such image-dominated FXL publications,  Here is my proposal.

Some FXL publications contain lots of images and almost no text. Moreover, these images contain rasterized text.  Most EPUB
publications in Japan are such FXL publications.

Image-dominated FXL publications are used for mangas, picture books, and even scholarly publications containing complicated layouts.  Unfortunately, image-dominated FXL publications have almost no accessibility.

One reason for using image-dominated FXL publications is the difficulty to achieve layout fidelity.  There are at least two reasons. First, font embedding is difficult due to the sheer number of characters in Japan.  Second, the mixture of vertical-writing and horizontal-writing makes layout fidelity particularly difficult. In fact, FXL publications containing lots of text have almost never been tried in Japan.

Another reason is that manga is fundamentally image-dominated.  Even if all words are read aloud, manga is not very enjoyable for blind people.  For this reason, some attempts for accessible manga introduce aural descriptions of scenes. and adopt purely aural representations rather than EPUB.  However, it remains to be seen if it is possible to make EPUB manga accessible for dyslexic people, low-vision people, and so forth.  Representing words as text without rasterization in EPUB publications might be good enough.

### LINKED ISSUE #2014: Hidden Text
I know we mention this elsewhere but for this section:

> 3.3 HTML
> For this reason, descriptions are typically hidden from view using a variety of HTML, ARIA and CSS techniques. Descriptions can be hidden, clipped, made opaque, layered under an image, etc. The [knowledge base page on hidden content](http://kb.daisy.org/publishing/docs/html/hidden.html) delves into these possibilities in more detail.

Should we mention here that these techniques for hiding text although an option is not recommended as this would be only accessible by using assistive technology and not everyone who would benefit from these descriptions use these tools.

### LINKED ISSUE #2015: Tables and Hidden Descriptions
In this section

> 7.Tables
> Using hidden description (see example 4)

Should we add a note?

(Note: Least desirable option as this could only be accessed to those using assistive technologies, and a person who may have a cognitive disability would not benefit from this hidden description)

### LINKED ISSUE #2016: AccessModeSufficient Example
In this Example

> 9.3 AccessModeSufficient Example 7
> 

> ```
>  <meta property="schema:accessModeSufficient">
>      textual
>  </meta>
> 
> ```

I think we also need to include, as it implies if a book with all images described is only “textual”.
Add:
```
<meta property="schema:accessModeSufficient">
    visual,textual
</meta>
```

### LINKED ISSUE #2017: AccessibilitySummary typo's and improved description
In this section

> 9.5 AccessibilitySummary 
> 
> The schema.org property accessibilitySummary is a human readable statment on how accessible or inacessible this book is.
> 

(there are two typos: statment and inacessible)

Also add to the example: Here is where you can also include any exceptions to the overall accessibility of the book (eg. All complex images are well described other than a few reference materials in the Appendix.)

wareid:
Addressing the issues raised with the FXL A11y document in issue #s #2012, #2014, #2015, #2016, and #2017.

### DIFF
--- a/epub33/fxl-a11y/index.html
+++ b/epub33/fxl-a11y/index.html
@@ -178,23 +178,25 @@ Fixed-layout documents can be presented as ‘synthetic spreads’ when a left a
 
 If the text must be read in this way, the only solution to maintain the correct reading order is to convert the double page spread in to a single landscape page that contains the entire content of the spread and for the EPUB to be rendered as single pages.
 
-## Alternative text {#alt-text}
+## Images in Fixed Layout #images
+
+### Overview
+    
+Images are often an integral part of fixed layout publications. Fixed layout publications sometimes consist entirely of images, in the case of comics, or images may be used as backgrounds to a story, as in children's books. Fixed layout content where images serve as both the content and the layout pose a particular challenge for content creators interested in accessibility. 
+
+Ensuring that the information conveyed in the images is available to users who may not be able to perceive the background or may have difficulty processing it, is consequently of high priority in making fixed layouts as accessible as possible.
+
+The [basic requirements for all images](http://kb.daisy.org/publishing/docs/html/images.html) are to provide alt text and extended descriptions when they contain information necessary to understanding the publication, which applies to fixed layouts. For example, while a reader may be able to follow the dialogue of a story when it is overlaid as text, only the placement on image might give context to what character is saying what.
+
+One challenge with fixed layouts is finding ways to describe the image and provide context given that there is no extra area on the page users can access in which to place a description. The other challenge is for content where the text is rasterized as part of the image. As much as possible, we recommend making the text on the page it's own layer, using technologies such as SVG and CSS to achieve the desired styling and placement, while also making the text more accessible to the user. 
+
+### Alternative Text and Image Descriptions
 
 Describing images within a fixed layout book will somewhat depend on the type of book these images are within.  For example describing a childrens picture book will be quite different than if this is a fixed layout graphic novel such as a comic book.
 
 <aside class="ednote">
 For the latest best practices regarding image descriptions it is recommended to visit the [DAISY Knowledge Base](http://kb.daisy.org/publishing/docs/fxl/img.html) as sections from there fixed layout image descriptions has been copied here.   
 </aside>
-    
-### Overview
-    
-Images are often an integral part of fixed layout publications. Fixed layout publications sometimes consist entirely of images, in the case of comics, or images may be used as backgrounds to a story, as in children's books.
-
-Ensuring that the information conveyed in the images is available to users who may not be able to perceive the background or may have difficulty processing it, is consequently of high priority in making fixed layouts as accessible as possible.
-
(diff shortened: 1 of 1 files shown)


epub33/fxl-a11y/index.html
```diff
@@ -178,23 +178,25 @@
...
 
 The [basic requirements for all images](http://kb.daisy.org/publishing/docs/html/images.html) are to provide alt text and extended descriptions when they contain information necessary to understanding the publication, which applies to fixed layouts. For example, while a reader may be able to follow the dialogue of a story when it is overlaid as text, only the placement on image might give context to what character is saying what.
 
-The challenge with fixed layouts is finding ways to describe the image and provide context given that there is no extra area on the page users can access in which to place a description.
+One challenge with fixed layouts is finding ways to describe the image and provide context given that there is no extra area on the page users can access in which to place a description. The other challenge is for content where the text is rasterized as part of the image. As much as possible, we recommend making the text on the page it's own layer, using technologies such as SVG and CSS to achieve the desired styling and placement, while also making the text more accessible to the user.
```
mattgarrish: s/it's/its/

mattgarrish (dismissed):
Noticed one typo, otherwise seems fine. I'll let @clapierre determine if it fixes all the issues he opened, though.

clapierre (approved):
much better!

murata2makoto:
Text rasterization causes big A11Y problems, but it is considered unavoidable.  It would be better if we can have some wording about why text rasterization cannot be easily avoided and what we should do to overcome this problem.

murata2makoto (changes requested):
Text rasterization causes big A11Y problems, but it is considered unavoidable. It would be better if we can have some wording about why text rasterization cannot be easily avoided and what we should do to overcome this problem.

wareid:
@mattgarrish @murata2makoto Please review the revised text please!

murata2makoto (changes requested):
Thank you so much, Wendy.  But this is such an important and long-lasting problem, especially in Japan, I would like to be very thorough.

First, I would like to mention rasterized text in the first paragraph.  For example:

- Images may contain rasterized text and sophisticated layout (scholarly
publishing, for example) (Note: Do we have a list of genres for FXL publications?)

Second, I do not believe that alt text is good enough for image-dominated FXL publications.  Several attempts have been made (e.g, longdesc of HTML and multiple-renditions of EPUB) but none of them are successful, unfortunately.  I would like to explicitly mention this shortcoming in our family of specifications and encourage future improvements.

wareid:
@murata2makoto I've added a little more detail. I think more work needs to be done to this section, but I'd like to close this PR for now. 

We do have a list of genres for FXL publications, we're working on accompanying examples and most of that work will go there, including some work around different methods for image description.

PR #2047, opened 2022-03-09
Add ncx doctype to allowed external identifiers

### LINKED ISSUE #2045: Allow NCX doctype external identifier?
The doctype was allowed in EPUB 2:

> `<!DOCTYPE ncx PUBLIC "-//NISO//DTD ncx 2005-1//EN" "http://www.daisy.org/z3986/2005/ncx-2005-1.dtd">`

We don't have an exception for it in Appendix B, though, so it would still be invalid in EPUB 3 even after the loosening for SVG/MathML.

I don't know how much it was used in EPUB 2, but should we add to simplify any porting forward of content?

mattgarrish:
Fixes #2045

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2047.html" title="Last updated on Mar 10, 2022, 11:27 AM UTC (1b1de14)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2047/f8a9622...1b1de14.html" title="Last updated on Mar 10, 2022, 11:27 AM UTC (1b1de14)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1202,9 +1202,10 @@
 				<section id="sec-parse-package-urls">
 					<h4>Parsing URLs in the Package Document</h4>
 
-					<p id="pkg-parse-package-url" data-tests="#ocf-url_link_relative,#ocf-url_relative"> To parse a URL string <var>url</var> used in the Package Document, the <a
-								data-cite="url#concept-url-parser">URL Parser</a> [[URL]] MUST be applied to
-								<var>url</var>, with the <a>content URL</a> of the Package Document as <var>base</var>.</p>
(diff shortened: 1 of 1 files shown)


iherman:
Purely editorial comment: the table looks a bit strange, with the mathml media types in bullet points and the other not (see below). Maybe it is better if

- All left hand items are rendered via a `<li>`
- Set the class of the table as "zebra" like in the [examples in RS](https://www.w3.org/TR/epub-rs-33/#sec-container-iri), that should help readability (and would keep the three mathml items together) t




Screen Shot 2022-03-09 at 13 09 43

iherman (approved):

mattgarrish:
The zebra class appears to suppress the bullets, which is probably better than having one-item lists.

PR #2058, opened 2022-03-10
Revise statement about embedded resources in the spine

### LINKED ISSUE #2057: Embedded resources in the spine
This statement sounds convincingly normative on first read:

> Embedded Publication Resources (e.g., via the [[HTML](https://www.w3.org/TR/epub-33/#bib-html)] [iframe](https://html.spec.whatwg.org/multipage/#the-iframe-element) element) must not be listed in the spine.

I was going to suggest this should have a capital "MUST NOT", but why do we care? If you want to also put an iframed document into the spine as a non-linear document, that seems fine.

In older versions of the specification, this used to say "do not have to be listed". I think this passive advice is better than using either "must" or "should" here, even in their non-normative senses.

mattgarrish:
Fixes #2057 as described in the issue.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2058.html" title="Last updated on Mar 11, 2022, 7:00 PM UTC (80f5e07)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2058/c86702e...80f5e07.html" title="Last updated on Mar 11, 2022, 7:00 PM UTC (80f5e07)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3451,11 +3451,10 @@ No Entry</pre>
 								are not subject to the requirement to include in the spine (e.g., Web pages and
 								resources).</p>
 
+							<p>Publication Resources used in the rendering of <code>spine</code> items (e.g., referenced
+								from [[HTML]] <a data-cite="html#embedded-content-2">embedded content</a>) similarly do
+								not have to be included in the <code>spine</code>.</p>
 						</div>
-
-						<p>EPUB Creators do not have to list embedded Publication Resources (e.g., via the [[HTML]] <a
-								data-cite="html#the-iframe-element"><code>iframe</code></a> element) in the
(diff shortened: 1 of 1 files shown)


iherman (approved):

mattgarrish:
I've made a couple of changes to try and improve this. The text now reads:

> Publication Resources used in the rendering of spine items (e.g., referenced from [[HTML](http://localhost:120/core/#bib-html)] [embedded content](https://html.spec.whatwg.org/multipage/#embedded-content-2)) similarly do not have to be included in the spine.

And I've moved this up into the note that preceded it, as it was also explaining things that don't have to go in the spine.

PR #2062, opened 2022-03-11
Make hidden attribute nav section informative

### LINKED ISSUE #2061: Is nav hidden attribute guidance non-normative?
I'm failing to see how [this section](https://www.w3.org/TR/epub-33/#sec-nav-def-hidden) is even specific to the navigation document, let alone normative.

All it says is to control rendering of the nav elements in the spine in reading systems without a viewport, use the `hidden` attribute because CSS won't be as reliable. (The "must" is unenforceable.)

But by the same logic, if you want to control rendering of any HTML content in the spine in a reading system without a viewport you should use the `hidden` attribute.

It seems like this might almost slot in better as an informative section under CSS.

mattgarrish:
I've changed the hidden attribute section to informative and renamed it "Using in the Spine". I also rewrote the first paragraph a bit so it doesn't jump straight into hiding content.

I kept the guidance about using the `hidden` attribute for RSes without viewports as-is, but added a note that also recommends pairing with the CSS property for maximum interoperability.

Hopefully, that's enough to cover the issue.

(I also took advantage of the moment to strip another unnecessary "XXX Definition" wrapper section that nothing even links to.)

Fixes #2061

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2062.html" title="Last updated on Mar 14, 2022, 12:58 PM UTC (c05d8d2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2062/52c424b...c05d8d2.html" title="Last updated on Mar 14, 2022, 12:58 PM UTC (c05d8d2)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5029,7 +5029,7 @@ No Entry</pre>
 			</section>
 
 			<section id="sec-nav-doc-use-spine" class="informative">
-				<h3>Using the Navigation Document in the Spine</h3>
+				<h3>Using in the Spine</h3>
 
 				<p>Although it is possible to reuse the EPUB Navigation Document in the <a>spine</a>, it is often the
 					case that not all of the navigation structures, or branches within them, are needed. <a>EPUB
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5039,9 +5039,9 @@ No Entry</pre>
 
 				<p>While the <a href="https://www.w3.org/TR/CSS2/visuren.html#propdef-display"><code>display</code>
 						property</a> [[CSSSnapshot]] controls the visual rendering of EPUB Navigation Documents in
-					Reading Systems with <a>Viewports</a>, not all Reading Systems provide such an interface. To control
-					rendering across all Reading Systems, EPUB Creators should use the [[HTML]] <a
-						data-cite="html#the-hidden-attribute"><code>hidden</code></a> attribute to indicate which (if
+					Reading Systems with <a>Viewports</a>, Reading Systems without Viewports may not support CSS. To
+					better ensure the proper rendering in these Reading Systems, EPUB Creators should use the [[HTML]]
+						<a data-cite="html#the-hidden-attribute"><code>hidden</code></a> attribute to indicate which (if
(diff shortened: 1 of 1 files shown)


epub33/core/index.html
```diff
@@ -5031,34 +5024,41 @@ <h5>Other <code>nav</code> Elements</h5>
...
-						<pre>&lt;nav
+				<p>Although it is possible to reuse the EPUB Navigation Document in the <a>spine</a>, it is often the
+					case that not all of the navigation structures, or branches within them, are needed. <a>EPUB
+						Creators</a> will often want to hide the <a href="sec-nav-pagelist">page list</a> and <a
+						href="sec-nav-landmarks">landmarks</a> navigation elements, or trim the branches of the table of
+					contents for books that have many levels of subsections.</p>
+
+				<p>While the <a href="https://www.w3.org/TR/CSS2/visuren.html#propdef-display"><code>display</code>
+						property</a> [[CSSSnapshot]] controls the visual rendering of EPUB Navigation Documents in
+					Reading Systems with <a>Viewports</a>, not all Reading Systems provide such an interface. To control
```
iherman: I am not sure that the 'interface' is the usual terminology for CSS implementation (I looked at, for example, the [CSS21 Conformance part](https://www.w3.org/TR/CSS21/conform.html#q3.0). Maybe:

```suggestion
					Reading Systems with <a>Viewports</a>, not all Reading Systems implement it. To control
```

iherman (approved):

epub33/core/index.html
mattgarrish: The interface here is the viewport, not the CSS rendering. I've revised this as follows to hopefully make it more clear:

> While the [display property](https://www.w3.org/TR/CSS2/visuren.html#propdef-display) [CSSSnapshot] controls the visual rendering of EPUB Navigation Documents in Reading Systems with Viewports, Reading Systems without Viewports may not support CSS. To better ensure the proper rendering in these Reading Systems ...

iherman: Ah! Yes, that is much clearer, thanks!

shiestyle (approved):

wareid (approved):

PR #2080, opened 2022-03-13
Add section explaining the requirements of the property table fields

### LINKED ISSUE #2073: Metadata property field explanations
The meta properties vocabulary has this explanation at the beginning:

> In each property definition, the Allowed values field indicates the expected type of value (using [[XMLSCHEMA-2](https://w3c.github.io/epub-specs/epub33/core/#bib-xmlschema-2)] datatypes), the Cardinality field indicates the number of times the property MAY be attached to another property, and the Extends field identifies the properties it MAY be attached to.

This doesn't appear again until the manifest properties vocabulary, but in this one it confusingly says cardinality is the number of times a property MUST appear in the package document (can you have a must for 0?).

I'm wondering, though, if we should make this an informative explainer section for all the vocabularies, as it's odd how it comes and goes? (A relic of these all being in separate documents and the prose not getting copied around, undoubtedly.)

mattgarrish:
Fixes #2073

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2080.html" title="Last updated on Mar 14, 2022, 6:30 PM UTC (a506e35)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2080/58d400b...a506e35.html" title="Last updated on Mar 14, 2022, 6:30 PM UTC (a506e35)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5122,97 +5122,114 @@ No Entry</pre>
 				</section>
 			</section>
 		</section>
-		<section id="sec-fixed-layouts">
-			<h2>Fixed Layouts</h2>
+		<section id="sec-rendering-control">
+			<h2>Layout Rendering Control</h2>
 
-			<section id="fxl-intro" class="informative">
+			<section id="sec-general-rendering-intro" class="informative">
 				<h3>Introduction</h3>
 
-				<p>EPUB documents, unlike print books or PDF files, are designed to change. The content flows, or
-					reflows, to fit the screen and to fit the needs of the user. As noted in <a
-						data-cite="epub-overview-33#sec-rendering">Rendering and CSS</a> "content presentation adapts to
-					the user, rather than the user having to adapt to a particular presentation of content."
-					[[EPUB-OVERVIEW-33]]</p>
+				<p>Not all rendering information can be expressed through the underlying technologies that EPUB is built
+					upon. For example, although HTML with CSS provides powerful layout capabilities, those capabilities
+					are limited to the scope of the document being rendered.</p>
 
-				<p>But this principle does not work for all types of documents. Sometimes content and design are so
-					intertwined it is not possible to separate them. Any change in appearance risks changing the meaning
-					or losing all meaning. <a>Fixed-Layout Documents</a> give <a>EPUB Creators</a> greater control over
-					presentation when a reflowable EPUB is not suitable for the content.</p>
-
-				<p>EPUB Creators define fixed layouts using a <a href="#sec-fxl-package">set of Package Document
-						properties</a> to control the rendering in <a>Reading Systems</a>. In addition, they set <a
-						href="#sec-fxl-package">the dimensions of each Fixed-Layout Document</a> in its respective EPUB
-					Content Document.</p>
-
-				<div class="note" id="note-mechanisms">
-					<p>EPUB 3 affords multiple mechanisms for representing fixed-layout content. When fixed-layout
-						content is necessary, the EPUB Creator's choice of mechanism will depend on many factors
-						including desired degree of precision, file size, accessibility, etc. This section does not
-						attempt to dictate the EPUB Creator's choice of mechanism.</p>
-
-				</div>
+				<p>This section defines properties that allow EPUB Creators to express package-level rendering
+					intentions (i.e., functionality that can only be implemented by the <a>EPUB Reading System</a>). If
+					a Reading System supports the desired rendering, these properties enable the user to be presented
+					the content as the EPUB Creator optimally designed it.</p>
 			</section>
 
-			<section id="sec-fxl-package">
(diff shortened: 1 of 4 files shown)


iherman:
I am not at my usual machine, so I cannot easily check, but in the preview the vocabulary definitions are missing; for some reasons the data inclusion mechanism of respec didn't work. Is this a problem with preview or is there a problem with the html source?

mattgarrish:
It's a flaw of the previews. For whatever reason, they don't handle respec's file imports. My guess would be that the static file is generated from a local file system, rather than served over http, so the script that imports the html doesn't have access to the vocabulary files.

mattgarrish:
If you view the source, it's full of errors where the imports are:

```
<meta charset="utf-8">
<title>Error</title>
<pre aria-busy="false"><code class="hljs">Cannot GET /uploads/ILiUfH/tmp/vocab/meta-property.html</code></pre>
 
<meta charset="utf-8">
<title>Error</title>
<pre aria-busy="false"><code class="hljs">Cannot GET /uploads/ILiUfH/tmp/vocab/link.html</code></pre>
```

iherman:
> It's a flaw of the previews.

SIgh. Anyway, here is a URL that I could use for preview:

https://raw.githack.com/w3c/epub-specs/editorial/issue-2073/epub33/core/index.html#app-rendering-vocab

iherman:
I am fine with the changes, +1 for this.

There is, however, a slight oddity that I am not sure how to solve (and this is a different issue, we may choose to spawn off one). We have a number of vocabulary alternatives, and we usually have, at the beginning of a respective section a neat statement to help to "locate" that vocabulary. E.g., §D.6 is on Manifest Properties Vocabulary, and the section begins by:

> The properties in this vocabulary are usable in the manifest [item element's](https://raw.githack.com/w3c/epub-specs/editorial/issue-2073/epub33/core/index.html#sec-item-elem) [properties attribute](https://raw.githack.com/w3c/epub-specs/editorial/issue-2073/epub33/core/index.html#attrdef-item-properties).

And that is great. However, there is an odd man out, namely §D.5 on Package Rendering Vocabulary, which is introduced with a longer explanation on rendering. The difficulty is that the section defines a bunch of properties for the `meta` element, but it also has a definition for overrides on the `itemref` elements. (At first I thought of moving the whole section as a subsection of §D.,3, but that dual set of properties makes this a bit awkward.)

Maybe we can make this explicit and have an introductory sentence of the form 

> The core properties in the vocabulary are usable in the [meta element's](https://raw.githack.com/w3c/epub-specs/editorial/issue-2073/epub33/core/index.html#sec-meta-elem) property attribute, with overrides in the spine [itemref element's](https://raw.githack.com/w3c/epub-specs/editorial/issue-2073/epub33/core/index.html#sec-itemref-elem) [properties attribute](https://raw.githack.com/w3c/epub-specs/editorial/issue-2073/epub33/core/index.html#attrdef-itemref-properties)

It is a bit more convoluted than the others, but it is still good to have imho.

iherman (approved):

mattgarrish:
> The difficulty is that the section defines a bunch of properties for the `meta` element, but it also has a definition for overrides on the `itemref` elements.

And align-x-center which only defines overrides and doesn't define a global value...

It's always been awkward how we ended up with these two properties. They sort of spawned out of the effort to integrate the FXL properties without being for FXL.

The table in D.4.2 for the FXL properties hides that most of the properties are also overrides, not properties in the `meta` element sense.

> At first I thought of moving the whole section as a subsection of §D.,3, but that dual set of properties makes this a bit awkward.

They rightfully should be defined in the body in the same way that the FXL properties are. The ideal would be that we only maintain the "vocabulary" for referencing purposes.

The problem is where? FXL makes a neat package, but these are just two stray properties not even connected to each other.

I guess we could squash them onto the end of the Content Documents section as some kind of general rendering control section, or maybe it would make sense to have them follow the FXL section, as FXL at least prepares you for them?

mattgarrish:
I've committed a change that restructures the document so there's now a parent "Layout Rendering Control" section. Under this is the FXL section plus a new section for the reflow properties.

The vocabulary in the appendix now just provides a mapping to the sections, with a note explaining its differences from the other vocabularies. I've also split out the overrides into a new column.

Let me know if this makes more sense.

mattgarrish:
You can review a working preview here: https://cdn.statically.io/gh/w3c/epub-specs/editorial/issue-2073/epub33/core/index.html

iherman:
> Let me know if this makes more sense.

YES! This is really way better and more logical.

Looking at me made me realized that we may want to have some figures for the fixed layout properties like spread and the like. I will look at this at some point, although I may need some help from @wareid, because I've never really dived into FXL. But that may be a good occasion...

Anyway. That is by no means something to handle in this PR.

wareid (approved):

PR #2082, opened 2022-03-14
Added a note to the URL constraints

### LINKED ISSUE #2075: Two properties in 6.1.5 of EPUB 3.3 look like tautologies
Two properties in [6.1.5 URLs in the OCF Abstract Container](https://w3c.github.io/epub-specs/epub33/core/#sec-container-iri) look like tautologies to me.

The result of [parsing](https://url.spec.whatwg.org/#concept-url-parser) "/" with the [container root URL](https://w3c.github.io/epub-specs/epub33/core/#dfn-container-root-url) as [base](https://url.spec.whatwg.org/#concept-base-url) is the [container root URL](https://w3c.github.io/epub-specs/epub33/core/#dfn-container-root-url).
The result of [parsing](https://url.spec.whatwg.org/#concept-url-parser) ".." with the [container root URL](https://w3c.github.io/epub-specs/epub33/core/#dfn-container-root-url) as [base](https://url.spec.whatwg.org/#concept-base-url) is the [container root URL](https://w3c.github.io/epub-specs/epub33/core/#dfn-container-root-url).

I learned from Romain that these do not always hold.  But I would like to see a note explaining why these properties do not always hold and what restrictions we are imposing by having these properties.

iherman:
Added a note in §6.1.5 on the background of the constraints on URLs

Fix #2075

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2082.html" title="Last updated on Mar 17, 2022, 12:25 PM UTC (d5ecbd9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2082/c601d10...d5ecbd9.html" title="Last updated on Mar 17, 2022, 12:25 PM UTC (d5ecbd9)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5968,13 +5968,13 @@ No Entry</pre>
 							The <a>container root URL</a> is the URL assigned by the Reading System to the root of the container. It typically depends on how the reading system internally implements the container file system.
 						</p>
 						<p>
-							However, a Reading System cannot arbitrarily use any URL, but one that honor the constraints defined above. These constraints ensure that any relative URL string found in the EPUB will always be parsed to a URL of a resource within the container (which may or may not exist).
+							However, a Reading System cannot arbitrarily use any URL, but one that honors the constraints defined above. These constraints ensure that any relative URL string found in the EPUB will always be parsed to a URL of a resource within the container (which may or may not exist).
 							The primary reason for these constraints is to avoid potential run-time security issues that would be caused by parsed URLs "leaking" outside the container files.
 						</p>						
 						<p>
 							For example, URLs like <code>https://localhost:12345/</code> or <code>https://www.example.org:12345/</code> honor these properties. But URLs like <code>https://localhost:12345/path/to.epub/</code>, 
 							<code>file:///path/to.epub#path=/</code>, or <code>jar:file:/path/to.epub!/EPUB/</code> do not 
-							(for example, parsing the URL string "<code>..</code>" with these three examples as base would return <code>https://localhost:12345/path/</code>, <code>file:///path/</code>, and a parsing error, respectively).
+							(parsing the URL string "<code>..</code>" with these three examples as base would return <code>https://localhost:12345/path/</code>, <code>file:///path/</code>, and a parsing error, respectively).
 							It is the responsibility of the Reading System to assign a URL to the root directory that complies with the properties defined above.
 						</p>	
 					</div>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5963,13 +5963,20 @@ No Entry</pre>
 						of <a data-cite="url#concept-url-parser">parsing</a> the file's <a>File Path</a> with the
 							<a>container root URL</a> as <a data-cite="url#concept-base-url"><var>base</var></a>.</p>
 
-					<div class="note">
-						The constraints on a <a>container root URL</a> are automatically honored for URLs like,
-						for example,  <code>https://localhost:12345/</code> or <code>https://www.example.org:12345/</code>; however,
-						a Reading System may choose a different strategy to identify the files of an EPUB container like, for example,
-						<code>file://path/to.epub#path=/EPUB/content.opf</code> or <code>jar:file:/path/to.epub!/EPUB/content.opf</code>.
-						In such a case, the Reading System has to implement its own version of URL parsing to conform to the conceptual
-						model of a <a>container root URL</a> as defined in this specification.
+					<div class="note" id="note-cru-explanation">
+						<p>
+							The <a>container root URL</a> is the URL assigned by the Reading System to the root of the container. It typically depends on how the reading system internally implements the container file system.
+						</p>
+						<p>
+							However, a Reading System cannot arbitrarily use any URL, but one that honor the constraints defined above. These constraints ensure that any relative URL string found in the EPUB will always be parsed to a URL of a resource within the container (which may or may not exist).
+							The primary reason for these constraints is to avoid potential run-time security issues that would be caused by parsed URLs "leaking" outside the container files.
+						</p>						
+						<p>
+							For example, URLs like <code>https://localhost:12345/</code> or <code>https://www.example.org:12345/</code> honor these properties. But URLs like <code>https://localhost:12345/path/to.epub/</code>, 
+							<code>file:///path/to.epub#path=/</code>, or <code>jar:file:/path/to.epub!/EPUB/</code> do not 
+							(for example, parsing the URL string "<code>..</code>" with these three examples as base would return <code>https://localhost:12345/path/</code>, <code>file:///path/</code>, and a parsing error, respectively).
+							It is the responsibility of the Reading System to assign a URL to the root directory that complies with the properties defined above.
+						</p>	
 					</div>
 
 					<div class="note">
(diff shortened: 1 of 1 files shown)


epub33/core/index.html
```diff
@@ -5963,6 +5963,15 @@ <h4>URLs in the OCF Abstract Container</h4>
 						of <a data-cite="url#concept-url-parser">parsing</a> the file's <a>File Path</a> with the
 							<a>container root URL</a> as <a data-cite="url#concept-base-url"><var>base</var></a>.</p>
 
+					<div class="note">
+						The constraints on a <a>container root URL</a> are automatically honored for "usual" URLs like,
+						for example,  <code>https://localhost:12345/</code> or <code>https://www.example.org:12345/</code>. However,
+						a Reading System may chose a different strategy to identify the files of an EPUB container like, for example,
+						<code>file://path/to.epub#path=/EPUB/content.opf</code> or <code>jar:file:/path/to.epub!/EPUB/content.opf</code>;
+						in such a case the Reading System have to implement its own version of URL parsing to be conform to the conceptual
```
mattgarrish: A few more nitpicks:
- better to use a period than a semicolon to separate (it's a hard stop from the "however" clause)
- comma after "case"
- s/have/has/
- s/to be/to/

epub33/core/index.html
```diff
@@ -5963,6 +5963,15 @@ <h4>URLs in the OCF Abstract Container</h4>
 						of <a data-cite="url#concept-url-parser">parsing</a> the file's <a>File Path</a> with the
 							<a>container root URL</a> as <a data-cite="url#concept-base-url"><var>base</var></a>.</p>
 
+					<div class="note">
+						The constraints on a <a>container root URL</a> are automatically honored for "usual" URLs like,
+						for example,  <code>https://localhost:12345/</code> or <code>https://www.example.org:12345/</code>. However,
+						a Reading System may chose a different strategy to identify the files of an EPUB container like, for example,
```
mattgarrish: s/chose/choose/

iherman: Yep

epub33/core/index.html
iherman: Yep and yep and yep and yep.

epub33/core/index.html
```diff
@@ -5963,6 +5963,15 @@ <h4>URLs in the OCF Abstract Container</h4>
 						of <a data-cite="url#concept-url-parser">parsing</a> the file's <a>File Path</a> with the
 							<a>container root URL</a> as <a data-cite="url#concept-base-url"><var>base</var></a>.</p>
 
+					<div class="note">
+						The constraints on a <a>container root URL</a> are automatically honored for "usual" URLs like,
+						for example,  <code>https://localhost:12345/</code> or <code>https://www.example.org:12345/</code>; however,
```
rdeltour: "usual" sounds slightly odd to me, as one could assume that `https://localhost:12345/epub/` isn't less usual at first sight.

epub33/core/index.html
```diff
@@ -5963,6 +5963,15 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+						a Reading System may choose a different strategy to identify the files of an EPUB container like, for example,
+						<code>file://path/to.epub#path=/EPUB/content.opf</code> or <code>jar:file:/path/to.epub!/EPUB/content.opf</code>.
+						In such a case, the Reading System has to implement its own version of URL parsing to conform to the conceptual
+						model of a <a>container root URL</a> as defined in this specification.
```
rdeltour: I'm reading this as the root URLs above ("such case") are conforming, but then the RS has then to work around by using a custom parsing logic.

This looks wrong to me (or at least confusing, if I did not understand correctly). The RS must always abide by the parsing rules defined in the URL standard, which implies the two URLs above are not conforming container root URLs.

wareid (approved):

rdeltour (changes requested):
Parts of the note sound confusing to me (see inline comments).

Also, I'm wondering if part of the other note defined later in the section should not be moved here now, as it helps to clarify the container root URL concept.

I'm talking about the part saying:
> Note that in any case, to avoid potential run-time security issues, the properties of the [container root URL ](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2082/c601d10...c00558b.html#dfn-container-root-url)are such that a conforming Reading System will parse any relative URL string to a [content URL ](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2082/c601d10...c00558b.html#dfn-content-url). In other words, even the disallowed URL strings described above will not "leak" outside the container.

Here's a possible rewording (to be refined):

1st note:
> The container root URL is the URL assigned by the Reading System to the root of the container. It typically depends on how the reading system internally implements the container file system.
>
> However, a Reading System cannot arbitrarily use any URL, but one that honor the constraints defined above. These constraints ensure that any relative URL string found in the EPUB will always be parsed to a URL of a resource within the container (the resource may effectively not exist, but the parsed URL conceptually points to something inside the container).
> The primary reason for these constraints is to avoid potential run-time security issues that would be caused by parsed URLs "leaking" outside the container files.
> 
> For example, URLs like `https://localhost:12345/` or `https://www.example.org:12345/` honor these properties. But URLs like `https://localhost:12345/moby-dick.epub/`, `file://path/to.epub#path=/`, or `jar:file:/path/to.epub!/EPUB/` do not (for example, parsing the URL string "`..`" with these three examples as base would respectively return `https://localhost:12345/`, `file:///`, and a parsing error).
> It is the responsibility of the Reading System to assign a URL to the root directory that complies with the properties defined above.
  

And then in the second note the final paragraph could be simplified to something like:
> Note that in any case, even the disallowed URL strings described above will not "leak" outside the container after parsing (as explained in the first note of this section). They are still disallowed for better interoperability with non-conforming or legacy Reading Systems and toolchains.

epub33/core/index.html
iherman: Agree. I have removed it.

epub33/core/index.html
iherman: Well... In theory I could imagine using these URL-s and then the RS would have to re-write the URL parsing library to abide to the behavioural specification of URL-s. Becuse the container root URL-s are, in some way, hidden to the everyday user (e.g., for RS-s that do not use scripting), this could be done and thereby making the RS conformant. I would agree that it would be mad to do it this way, but I shied away to declare these RS-s non-conformant.

That being said, it makes it probably easier to read, in a note, that these are non-conformant, even though a specific RS may find a way around it...

iherman: Looking at your generic reformulation, though, this issue is becoming moot anyway...

iherman:
@rdeltour I agree with your change proposal. I have adopted it with minimal changes.

epub33/core/index.html
```diff
@@ -5963,13 +5963,20 @@ <h4>URLs in the OCF Abstract Container</h4>
...
-						a Reading System may choose a different strategy to identify the files of an EPUB container like, for example,
-						<code>file://path/to.epub#path=/EPUB/content.opf</code> or <code>jar:file:/path/to.epub!/EPUB/content.opf</code>.
-						In such a case, the Reading System has to implement its own version of URL parsing to conform to the conceptual
-						model of a <a>container root URL</a> as defined in this specification.
+					<div class="note" id="note-cru-explanation">
+						<p>
+							The <a>container root URL</a> is the URL assigned by the Reading System to the root of the container. It typically depends on how the reading system internally implements the container file system.
+						</p>
+						<p>
+							However, a Reading System cannot arbitrarily use any URL, but one that honor the constraints defined above. These constraints ensure that any relative URL string found in the EPUB will always be parsed to a URL of a resource within the container (which may or may not exist).
```
rdeltour: s/honor/honors/

epub33/core/index.html
```diff
@@ -5963,13 +5963,20 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+							The primary reason for these constraints is to avoid potential run-time security issues that would be caused by parsed URLs "leaking" outside the container files.
+						</p>						
+						<p>
+							For example, URLs like <code>https://localhost:12345/</code> or <code>https://www.example.org:12345/</code> honor these properties. But URLs like <code>https://localhost:12345/path/to.epub/</code>, 
+							<code>file:///path/to.epub#path=/</code>, or <code>jar:file:/path/to.epub!/EPUB/</code> do not 
+							(for example, parsing the URL string "<code>..</code>" with these three examples as base would return <code>https://localhost:12345/path/</code>, <code>file:///path/</code>, and a parsing error, respectively).
```
rdeltour: s/for example, //

(redundant with the first occurence outside the parens)

rdeltour (approved):
👍 (just noticed a  typo and a repetition, coming from my proposal, my bad 😅)

iherman:
@murata2makoto I would like to merge this PR before the end of the week. Is that ok with you? Thanks.

murata2makoto (approved):

PR #2089, opened 2022-03-17
Fix requirements on authoring semantics in MO documents

### LINKED ISSUE #2066: Media overlay semantics
The third bullet in [7.2.1](https://w3c.github.io/epub-specs/epub33/core/#sec-overlay-req) says:

> [A Media Overlay Document] SHOULD use semantic markup where appropriate.

Aside from this being overly vague, I can't find anywhere where semantics could be applied where they are recommended. The skippability and escapability sections don't recommend semantics be set, and these would be the key reasons for them.

Should we remove this statement and determine in the future whether to recommend semantics be set for skippability and escapability, or even leave this for the accessibility specification to require? (It also only has semantics as an optional requirement right now.)

mattgarrish:
This is my best attempt to solve #2066 as minimally as possible. The PR:

- removes the bullet about adding semantics from the requirements section because the first bullet already allows semantics to be added since the method of adding them is part of the schema definition (specifying things twice is how these mismatches always arise). Section 7.3.3 also already covers adding semantics with the correct normative requirements.
- adds "MAY" statements for using the lists of semantics we previously called out for skippability and escapability. This improves our eventually raising support of skippability and escapability to recommendations in the accessibility specification
- moves the lists of semantics up above the examples in the skippability and escapability sections. It's odd to reach an example of implementing the features before knowing what the semantics to use are.
- adds a paragraph to the RS specification for skippability to require that reading systems suppress playback of par/seq elements with matching semantics. All we were requiring was that reading systems should look and see if there are semantics to determine whether to enable skippability. I'm not sure if I have the best wording for this, though, so feedback welcome.

<strike>I also had to make some fixes to the markup of the RS spec, as I was getting errors. There was a missing end tag and part of an opening tag was missing, so we were getting an attribute in the source.</strike> (Fixed these in main.)

Anyway, let me know if I've got this right or if there's more that needs fixing.

Fixes #2066 

- Reading Systems [preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-2066/epub33/rs/index.html)
- Reading Systems [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-2066/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2089.html" title="Last updated on Mar 25, 2022, 9:45 AM UTC (e525a8b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2089/55d5466...e525a8b.html" title="Last updated on Mar 25, 2022, 9:45 AM UTC (e525a8b)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7278,10 +7278,6 @@ No Entry</pre>
 								more than one Media Overlay Document MUST NOT reference the same EPUB Content
 								Document.</p>
 						</li>
-						<li>
-							<p id="confreq-mo-docprops-semantics">SHOULD use <a href="#sec-docs-structural-semantic"
-									>semantic markup</a> where appropriate.</p>
-						</li>
 					</ul>
 				</section>
 
@@ -8488,6 +8484,29 @@ html.my-document-playing * {
 							<a href="#sec-docs-structural-semantic"><code>epub:type</code></a> attribute to determine
 						when to offer users the option of skippable features.</p>
 
+					<p>EPUB Creators MAY use the following semantics to enable skippability:</p>
+
+					<ul>
+						<li>
+							<p>footnote</p>
+						</li>
+						<li>
+							<p>endnote</p>
+						</li>
+						<li>
+							<p>pagebreak</p>
+						</li>
+					</ul>
+
+					<p>This list is non-exhaustive, however. It represents terms from the Structural Semantics
+						Vocabulary [[?EPUB-SSV-11]] for which Reading Systems are most likely to offer the option of
+						skippability.</p>
+
+					<div class="note">
+						<p>Reading System are not required to support for skippability based on <code>epub:type</code>
+							values.</p>
+					</div>
+
 					<aside class="example" title="Media Overlay with a page break">
 						<p>In this example, a Reading System could offer the user the option of turning on and off the
 							page break/page number announcements, which are often cumbersome to listen to.</p>
@@ -8554,25 +8573,6 @@ html.my-document-playing * {
    &lt;/body>
 &lt;/html></pre>
 					</aside>
-
-					<p>The following non-exhaustive list represents terms from the Structural Semantics
-						Vocabulary [[?EPUB-SSV-11]] for which Reading Systems may offer the option of skippability:</p>
-
-					<ul>
-						<li>
-							<p>footnote</p>
-						</li>
-						<li>
-							<p>endnote</p>
-						</li>
-						<li>
-							<p>pagebreak</p>
-						</li>
-					</ul>
-					<div class="note">
-						<p>Reading System are not required to support for skippability based on <code>epub:type</code>
-							values.</p>
-					</div>
 				</section>
 
 				<section id="sec-escapability">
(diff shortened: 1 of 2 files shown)


marisademeglio:
So overall, using structural semantics to enable RS behaviors is a [MAY](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-2066/epub33/rs/index.html#sec-structural-semantics), but in MO it's a SHOULD. Do these need to align? 

Typo in the core spec in 2 places ([skip note](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2089.html#sec-skippability) and [escape note](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2089.html#sec-escapability)):
" to support for skippability" => "to support skippability"

All in all, though, it looks much better!

mattgarrish:
> So overall, using structural semantics to enable RS behaviors is a [MAY](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-2066/epub33/rs/index.html#sec-structural-semantics), but in MO it's a SHOULD.

I guess the argument is we don't define anything for reading systems to do with structural semantics in the content, so that can't change without standardizing something.

But for MO we do define at least two features (with table mode in a note), but it's optional to author the semantics. There's a usual argument there for making the RS requirement match the authoring.

Then again, the difference between a MAY and SHOULD is pretty small. You're just supposed to have a good reason not to implement a SHOULD, but who's checking?

There may be implications for the Accessibility spec in the future if authoring of the semantics increases to a recommendation, but presumably we'd could update both documents.

I can live with either, in other words, so I'll let others weigh in on preferences.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-03-17-epub#section1) on 2022-03-17

- no resolutions were taken

### 1. Media overlay semantics (issue epub-specs#2066)

_See github issue [epub-specs#2066](https://github.com/w3c/epub-specs/issues/2066)._




**Wendy Reid:** you may have noticed a flurry of new issues logged due to people reviewing spec.  
… we have a normative statement that MO SHOULD use semantics when appropriate, and we do not clarify what we mean by semantics, or what semantics to use.  

_See github pull request [epub-specs#2089](https://github.com/w3c/epub-specs/pull/2089)._




**Wendy Reid:** mgarrish has discussed with danielwreck and marissa, and put together a PR.  
… to clean up the language in that section, particularly dropping SHOULD to MAY.  
… we can't drop the statement altogether, but SHOULD is a little strong in this case.  
… and to specify that we are talking about footnote, endnote, page break.  
… now we have something that could technically be tested (although it is not a required feature, we are discouraging use of epub:type).  
… there is also a small addition to the RS spec to make similar clarifications there.  
… questions?.  

**Murata Makoto:** this changes both Core and RS, right?.  

**Wendy Reid:** yes.  
… the change to RS is barely a change, it keeps the normative statements but offers a little more information about what the RS should be doing.  

**Murata Makoto:** mgarrish dropped one bullet in EPUB 33 core?.  

**Wendy Reid:** he changed one SHOULD to MAY, and both escapability and skippability are better defined. It should be in the PR.  

**Murata Makoto:** seems sensible, but I haven't reviewed the details.  

**Wendy Reid:** to clarify, he dropped the SHOULD, and made the following statement a MAY.  

**Murata Makoto:** no objections, but 4 reviewers have proof read the proposed change?.  

**Wendy Reid:** yes.  
… I didn't know there was a going to be a PR in time for the meeting today, mgarrish was ahead of schedule.  
… we will have to close this issue soon (within the next week or so) so if you have an interest in this please review the PR.  

</details>

avneeshsingh:
I am concerned with the following note, because we have still not completely moved Media Overlays from epub:type to DPUB ARIA roles. Plus EPUB accessibility provides recommendation to implement skippability and escapability.
"Reading System are not required to support for skippability based on epub:type values."

In this statement, which is present in both skippability and Escapability section, we should also direct people to section in EPUB accessibility, which provides additional recommendations for Skippability and Escapability.

mattgarrish:
> I am concerned with the following note, because we have still not completely moved Media Overlays from epub:type to DPUB ARIA roles.

I wasn't sure why we had the note for escapability but not for skippability, so included with both. Can we drop it entirely?

I don't think it's about how we support the features, but just seems to be a caution that even if you do add semantics it's not required RSes do anything. But that's true for lots of things.

avneeshsingh:
+1 to removing the notes. No need to add text for EPUB Accessibility if notes are removed.

mattgarrish:
> No need to add text for EPUB Accessibility if notes are removed.

No, this PR better sets up the recommendations of the accessibility spec. We don't need to change anything in that document.

mattgarrish:
Are you okay with the changes in this pull request now @avneeshsingh ?

PR #2090, opened 2022-03-17
Change epub:type note to a caution and elaborate on applications

### LINKED ISSUE #2070: epub:type restriction that's outlived its usefulness
These statements date from a time when we thought assistive technologies might use epub:type semantics like ARIA roles:

> The inflected semantic MUST express a subclass of the semantic of the carrying element. In the case of semantically neutral elements, such as the [HTML] [div](https://html.spec.whatwg.org/multipage/#the-div-element) and [span](https://html.spec.whatwg.org/multipage/#the-span-element) elements, the inflected semantic MUST NOT attach a meaning that is already conveyed by an existing element (e.g., that a div represents a paragraph or section).

We have recommendations on what elements to use the semantics on in the structure vocabulary, but these have never been enforced and were always meant as guidance. In other words, they don't even bear out these requirements.

I don't see the harm in making these statements informally, so switch these to lowercase should/should not.

mattgarrish:
For this PR, I've changed the note in the `epub:type` section to a caution in removing the unchecked restrictions on its use. I believe it's worth extra emphasis that `epub:type` is not at all like ARIA `role`.

I've also tried to explain the differences better, so the caution now reads:

> Although the epub:type attribute is similar in nature to the [role attribute](https://html.spec.whatwg.org/multipage/#attr-aria-role) [HTML], the attributes serve different purposes. The values of the epub:type attribute do not enhance the accessibility of EPUB Publications, for example, as they do not map to the accessibility APIs used by assistive technologies. This means that adding epub:type values to semantically neutral elements like [HTML] [div](https://html.spec.whatwg.org/multipage/#the-div-element) and [span](https://html.spec.whatwg.org/multipage/#the-span-element) does not make them any more accessible. Only ARIA roles influence how assistive technologies understand such elements.
> 
> The epub:type attribute is consequently only intended for publishing semantics and Reading System enhancements. Refer to [Digital Publishing WAI-ARIA Module 1.0](https://www.w3.org/TR/dpub-aria-1.0/#) [DPUB-ARIA-1.0] for more information about accessible publishing roles.

Fixes #2070

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2090.html" title="Last updated on Mar 20, 2022, 11:51 AM UTC (17f21b4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2090/8d4a9d6...17f21b4.html" title="Last updated on Mar 20, 2022, 11:51 AM UTC (17f21b4)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -9533,17 +9533,22 @@ html.my-document-playing * {
 				<div class="caution">
 					<p>Although the <code>epub:type</code> attribute is similar in nature to the <a
 							data-cite="html#attr-aria-role"><code>role</code> attribute</a> [[HTML]], the attributes
-						serve different purposes. The values of the <code>epub:type</code> attribute do not enhance the
-						accessibility of EPUB Publications, for example, as they do not map to the accessibility <abbr
-							title="Application Programming Interfaces">APIs</abbr> used by assistive technologies. This
-						means that adding <code>epub:type</code> values to semantically neutral elements like [[HTML]]
-							<a data-cite="html#the-div-element"><code>div</code></a> and <a
+						serve different purposes. The values of the <code>epub:type</code> attribute do not enhance
+						access through assistive technologies like screen readers as they do not map to the
+						accessibility <abbr title="Application Programming Interfaces">APIs</abbr> used by these
+						technologies. This means that adding <code>epub:type</code> values to semantically neutral
+						elements like [[HTML]] <a data-cite="html#the-div-element"><code>div</code></a> and <a
 							data-cite="html#the-span-element"><code>span</code></a> does not make them any more
-						accessible. Only ARIA roles influence how assistive technologies understand such elements.</p>
+						accessible to assistive technologies. Only ARIA roles influence how assistive technologies
+						understand such elements.</p>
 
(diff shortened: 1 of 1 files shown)


clapierre (approved):

iherman (approved):

avneeshsingh:
We should avoid stating that EPUB-type is not for accessibility. Reading systems can implement accessibility features using the EPUB-type attribute. We should change the text a little. The example follows, you can polish it further.

Although the epub:type attribute is similar in nature to the role attribute [HTML], the attributes serve different purposes. The values of the epub:type attribute do not enhance the access through assistive technologies like screen readers as they do not map to the accessibility APIs used by assistive technologies. This means that adding epub:type values to semantically neutral elements like [HTML] div and span does not make them any more accessible to assistive technology. Only ARIA roles influence how assistive technologies understand such elements.
The epub:type attribute is consequently only intended for publishing semantics and Reading System enhancements. It is worth mentioning that reading systems may use epub:type for providing accessibility enhancements like in-built ReadAloud or Media Overlays functionality where interaction with assistive technology is not essential. Please refer to Digital Publishing WAI-ARIA Module 1.0 [DPUB-ARIA-1.0] for more information about accessible publishing roles.

PR #2092, opened 2022-03-18
Added figures to the FXL section to illustrate spreads

### LINKED ISSUE #2087: Misplaced example in FXL
[Example 50](https://w3c.github.io/epub-specs/epub33/core/#fxl-ex4) on overriding the global spread behaviour should be moved into the relevant section: [5.2.2.4 Spread placement](https://w3c.github.io/epub-specs/epub33/core/#page-spread).

iherman:
Unfortunately, the usual preview does not work with images. This [preview does](https://raw.githack.com/w3c/epub-specs/editorial/fxl-spread-images/epub33/core/index.html#spread)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2092.html" title="Last updated on Mar 23, 2022, 10:55 AM UTC (f3f54f3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2092/7915c13...f3f54f3.html" title="Last updated on Mar 23, 2022, 10:55 AM UTC (f3f54f3)">Diff</a>


iherman:
@clapierre, I would appreciate your extra review on the diagram organization and texts from an a11y point of view...

epub33/core/index.html
```diff
@@ -5498,6 +5498,79 @@ <h5>Synthetic Spreads</h5>
 								of local page-progression-direction within content documents.</p>
 						</div>
 
+						<figure id="spread-none-figure">
+							<figcaption>
+								Rendering of an EPUB Fixed Layout Publication with three pages in portrait and in landscape modes,
```
mattgarrish: "EPUB Fixed Layout Publication" isn't a formal term of the specification. You could make this and the other below "a fixed-layout EPUB Publication" or join the pagination part to make it more concisely "Rendering of three Fixed-Layout Documents..."

But my primary comment here is whether these descriptions make sense as titles. They're trying to explain the metadata, but wouldn't it make sense to merge the figure with actual package document markup that explains the views? The title of the figure could then be reduced to something like "Rendering of Fixed Layout Documents when the EPUB Creator specifies no spreads."

iherman: Just to understand your proposal: let us have explicit code examples (or modify existing ones) that are in line with the figures and then cross-reference them? I can definitely try to do that.

mattgarrish: > Just to understand your proposal: let us have explicit code examples (or modify existing ones) that are in line with the figures and then cross-reference them?

I was actually thinking you could put the example markup inside the figure tags, or perhaps it would be better to add the images to an example div with package markup. Sort of like how we have some examples that show the package document markup together with content document markup (or in pub manifest I think we showed markup with internationalization examples).

iherman: I ended up creating codes as in a usual example, and added the respective figure into the `aside`. Doing it the other way round clashed a bit with the core css, and placed the code into the center with a reduced width, which looked odd.

I left a (adapted) figure caption in, although I wonder whether it is necessary to keep them... wdyt?

iherman: B.t.w., there is a relationship with #2087; it would be good to agree on that one and make the changes as part of this PR. I am worried of unnecessary merge complications...

iherman: I have added a new figure for the local override example (to have a drawing for all four examples). As an aside, this also takes care of #2087.

Fix #2087

iherman:
@clapierre @wareid could you look at this PR? It would be good to get it out of the way before the deluge of new PR-s...

wareid (approved):

PR #2101, opened 2022-03-20
Expand and fix resource explanations

### LINKED ISSUE #2083: (editorial) on Core Media Type Resources
Currently the terminological section says:

> **Core Media Type Resource**
> A [Publication Resource](https://w3c.github.io/epub-specs/epub33/core/#dfn-publication-resource) that does not require the provision of a [fallback](https://w3c.github.io/epub-specs/epub33/core/#sec-foreign-restrictions) (cf. [Foreign Resource](https://w3c.github.io/epub-specs/epub33/core/#dfn-foreign-resource)).

and

> **Foreign Resource**
A [Publication Resource](https://w3c.github.io/epub-specs/epub33/core/#dfn-publication-resource) that is not a [Core Media Type Resource](https://w3c.github.io/epub-specs/epub33/core/#dfn-core-media-type-resource). Foreign Resources are subject to the fallback requirements defined in [2.2.1.3 Foreign Resources](https://w3c.github.io/epub-specs/epub33/core/#sec-foreign-restrictions).

These are almost circular definitions, except for the second which does branch out to a section in the spec.

I think it would be helpful to change the terminology of Core Media Type Resource to something like:

> A [Publication Resource](https://w3c.github.io/epub-specs/epub33/core/#dfn-publication-resource) that conforms to one of the MIME media types [[RFC2046](https://w3c.github.io/epub-specs/epub33/core/#bib-rfc2046)] listed in [2.2.1.2 Supported Media Types](https://w3c.github.io/epub-specs/epub33/core/#sec-cmt-supported) and, therefore, does not require the provision of a [fallback](https://w3c.github.io/epub-specs/epub33/core/#sec-foreign-restrictions) (cf. [Foreign Resource](https://w3c.github.io/epub-specs/epub33/core/#dfn-foreign-resource)).

### LINKED ISSUE #2160: Clarify "SHOULD NOT use non-Publication Resources"
https://w3c.github.io/epub-specs/epub33/rs/#sec-pkg-doc-manifest says:

> Reading Systems SHOULD NOT use non-[Publication Resources](https://www.w3.org/TR/epub-33/#dfn-publication-resource) in the rendering of an EPUB Publication due to the inherent limitations and risks involved (e.g., lack of information about the resource and how to process it, security risks from remotely-hosted sources, lack of fallbacks, etc.).

My understanding is that remote resources *can* be publication resources if they are listed in the manifest; therefore, I believe this is saying that Reading Systems should only render resources listed in the manifest.

It kinda reads, though, like it's saying "don't use remote resources at all" (especially because of the examples; I don't think security risks are any less if a remote resource is listed in the manifest).

Can we clarify, and limit the parenthetical to things that are different between publication resources and non-publication resources?

mattgarrish:
With this PR, I've tried to rationalize all the categories of resources and exemptions we've layered onto the spec.

Probably the biggest addition is to add formal terms for things we've left unnamed:

- I've added Linked Resource as a formal term for anything that isn't a Publication Resource to cover everything you can find in the container (at the highest level). We were already informally using this term in the `link` element definition, so it's more a standardization effort.
- I've added Foreign Content Document to cover anything in the spine that isn't an EPUB Content Document. We slipped into a bad habit of calling these Foreign Resources in places, when that's only a subset. Most of the Core Media Type Resources are Foreign Content Documents when used in the spine.

I've also tried to explain how all these resources come together in a new introduction to the Publication Resources section. The groupings generally break down like this:

- EPUB Container  = Publication Resources + Linked Resources
- Spine = EPUB Content Documents + Foreign Content Documents
- EPUB/Foreign Content Documents contain Core Media Type Resources and Foreign Resources

But there are overlaps and exemptions that make not quite so simple.

I've also gone through and tried to clean up all the inconsistencies that have arisen because we've tended to drop into using the wrong terminology in various places.

From a structural standpoint, this PR gets rid of the "Core Media Types" section and renames the section "Supported Formats" to "Core Media Types", as this is really where we define them. It also moves the Manifest Fallbacks section up into Publication Resources, although I'm still not sure if this is the best location given how heavy it is on explaining the item element.

At any rate, hopefully this helps improve the ambiguities about why core media types are not always fallback exempt, why foreign resources are not the only foreign things in the spine, etc.

It's also a starting point, so if there are other improvements we can make here please feel free to comment.

(The changes so far should not normatively change anything.)

Also fixes #2083
Also fixes #2160

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html" title="Last updated on Mar 31, 2022, 11:39 PM UTC (a3deaaf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2101/017a47c...a3deaaf.html" title="Last updated on Mar 31, 2022, 11:39 PM UTC (a3deaaf)">Diff</a>


epub33/core/index.html
```diff
@@ -691,277 +732,410 @@ <h4>Conformance Checking</h4>
...
+						Primary in this group are Publication Resources, which are all the resources used in rendering
+						an EPUB Publication to the user. <a>EPUB Creators</a> always have to list these resources in the
+						Package Document <a>manifest</a>. Most Publication Resources have to be included in the EPUB
+						Container, but EPUB 3 makes <a href="#sec-resource-locations">exceptions for a few large data
+							formats</a> (audio, video, font and script data resources).</p>
+
+					<p>Publication Resources are not the only resources that may be in the EPUB Container. EPUB 3 also
+						allows EPUB Creators to reference <a>Linked Resources</a>. Linked Resources can be hosted either
+						inside or outside the EPUB Container as they are not used in rendering the EPUB Publication.
+						When included in the EPUB Container, they are not bound to the any of the rules of
```
```suggestion
						When included in the EPUB Container, they are not bound to any of the rules of
```

epub33/core/index.html
```diff
@@ -691,277 +732,410 @@ <h4>Conformance Checking</h4>
...
+							<a>Publication Resources</a> &#8212; they are not listed in the manifest, for example, as
+						their nature and purpose is already defined by their <code>link</code> element, and they are not
+						subject to fallback requirements. (Refer to <a href="#sec-link-elem"></a> for more
+						information.)</p>
```
iherman: We may want to add a note or simply a side remark in the text making it clear that resources that are referred to by an HTML `link` element _are_ Publication Resources and not Linked Resources.

epub33/core/index.html
```diff
@@ -691,277 +732,410 @@ <h4>Conformance Checking</h4>
...
+
+					<p>Moving below these broad categorizations, how resources are designated is tied to how they are
+						used within an EPUB Publication. Consequently, the next logical categorization to consider is
+						how resources are used in the <a href="#sec-spine-elem">spine</a>. The spine defines the default
+						reading order of resources, which includes both <a href="#attrdef-itemref-linear">linear and
+							non-linear content</a>. The spine instructs Reading Systems on how to load these documents
+						as the user progresses through the publication. Although many resources may be bundled in an
+						EPUB Container, they are not all allowed by default in the spine.</p>
+
+					<p>EPUB 3 defines a special class of resources called EPUB Content Documents that EPUB Creators can
```
```suggestion
					<p>EPUB 3 defines a special class of resources called <a>EPUB Content Documents</a> that EPUB Creators can
```

epub33/core/index.html
```diff
@@ -691,277 +732,410 @@ <h4>Conformance Checking</h4>
...
+						use in the spine without any restrictions. EPUB Content Documents encompass both <a>XHTML
+							Content Documents</a> and <a>SVG Content Documents</a>.</p>
+
+					<p>To use any other type of resource in the spine, called a <a>Foreign Content Document</a>,
+						requires including a fallback to an EPUB Content Document. This extensibility model allows EPUB
+						Creators to experiment with formats while ensuring that Reading Systems are always able to
+						render something for the user to read, as there is no guarantee of support for Foreign Content
+						Documents.</p>
+
+					<p>Fallbacks for spine items are provided through a mechanism called manifest fallbacks. When a
```
```suggestion
					<p>Fallbacks for spine items are provided through a mechanism called <a href="#sec-manifest-fallbacks">manifest fallbacks</a>. When a
```

epub33/core/index.html
```diff
@@ -691,277 +732,410 @@ <h4>Conformance Checking</h4>
...
+
+					<div class="caution">
+						<p>EPUB Creators are strongly discouraged from relying on manifest fallbacks both in the spine
+							and, as discussed next, for Foreign Resources. They have limited support in Reading Systems,
+							which means that unsupported formats may simply fail to render. They also do not provide
+							accessibility benefits (e.g., falling back to an accessible XHTML alternative when images
+							are used in the spine), as users have no choice over what the Reading System renders.</p>
+					</div>
+
+					<p>Moving below <a>EPUB Content Documents</a> and <a>Foreign Content Documents</a> comes the
```
iherman: Why "below"?

The section, as far as I can see, defines three different "axes" (or "facets"?) of resource categorization:

- *manifest axis* 
	- Publication Resource: must be listed in the manifest
	- Linked Resource: is not to be listed in the manifest
- *spine axis*
	- EPUB Content Document: can appear in the spine without fallback, these are XHTML and SVG
	- Foreign Content Document: can appear in the spine only if a fallback is defined, all non-EPUB Content Documents are in this category
- *rendering axis*
	- Core Media Type Resource: RS can be expected to support (see the core media types) do not necessarily require fallbacks
	- Foreign Resources: require fallbacks under all circumstances

If this understanding is correct, then there is no "below" but three different axes. (Maybe it is worth spelling this out in the text more explicitly). It is just "another"

epub33/core/index.html
```diff
@@ -691,277 +732,410 @@ <h4>Conformance Checking</h4>
...
+						resources used in their rendering. These fall into two broad categories: <a>Core Media Type
+							Resources</a> and <a>Foreign Resources</a>.</p>
+
+					<p>A Core Media Type Resource is one that Reading Systems are expected to support so they can be
+						used without restriction in EPUB Content Documents or Foreign Content Documents. (For more
+						information about Core Media Type Resources, refer to <a href="sec-core-media-types"></a>.)</p>
```
```suggestion
						information about Core Media Type Resources, refer to <a href="#sec-core-media-types"></a>.)</p>
```

epub33/core/index.html
```diff
@@ -691,277 +732,410 @@ <h4>Conformance Checking</h4>
...
+
+					<p>Note that being a Core Media Type Resource does not mean that Reading Systems will always render
+						a resource, however, as not all Reading Systems support all features of EPUB 3. A Reading System
+						without a <a>Viewport</a>, for example, will not render visual content such as images.</p>
+
+					<p>Another common point of confusion is the listing of XHTML and SVG Content Documents as Core Media
```
iherman: If I look at this from the point of view of the three different axis, then it is not really a confusion. An XHTML document is, obviously, a Core Media Type Resource on the rendering axis and also a content document on the spine axis...

epub33/core/index.html
```diff
@@ -691,277 +732,410 @@ <h4>Conformance Checking</h4>
...
+						format or via manifest fallbacks.</p>
+
+					<p>The preferred method is to use the fallback capabilities of the parent format. Many HTML
+						elements, for example, have intrinsic fallback capabilities. One example is the
+							<code>picture</code> element, which allows EPUB Creators to specify multiple alternative
+						image formats.</p>
+
+					<p>If an intrinsic fallback method is not available, it is also possible to use manifest fallbacks,
+						but this method, as described above, is strongly discouraged. (For more information about
+						Foreign Resources, refer to <a href="sec-foreign-resources"></a>.)</p>
```
```suggestion
						Foreign Resources, refer to <a href="#sec-foreign-resources"></a>.)</p>
```

epub33/core/index.html
```diff
@@ -691,277 +732,410 @@ <h4>Conformance Checking</h4>
...
+					<div class="note">
+						<p>Inclusion as a Core Media Type Resource does not mean that all Reading Systems will support
+							the rendering of a resource. Reading System support also depends on the capabilities of the
+							application (e.g., a Reading System with a <a>Viewport</a> must support image Core Media
+							Type Resources, but a Reading System without a Viewport does not). Refer to <a
+								data-cite="epub-rs-33#sec-epub-rs-conf-cmt">Core Media Types</a> [[EPUB-RS-33]] for more
+							information about which Reading Systems rendering capabilities require support for which
+							Core Media Type Resources.</p>
 
 						<p>The Working Group typically only includes formats as Core Media Type Resources when they have
```
iherman, line 1139: Not sure about this one: it _may_ be better to refer to W3C not to a WG; per its nature, a WG has a time-limited existence compared to W3C...

```suggestion
						<p>W3C typically only includes formats as Core Media Type Resources when they have
```

epub33/core/index.html
```diff
@@ -691,277 +732,410 @@ <h4>Conformance Checking</h4>
...
 
-						<p><a>EPUB Creators</a> MAY include <a>Publication Resources</a> that conform to the following
-							MIME media type [[RFC2046]] specifications in <a>EPUB Publications</a> without
-							fallbacks.</p>
+					<p>A <a>Foreign Resource</a>, unlike a <a href="#sec-core-media-types">Core Media Type Resource</a>
+						is one which is not guaranteed <a>Reading System</a> support when used in an <a>EPUB Content
+							Document</a> or <a>Foreign Content Document</a>.</p>
 
-						<p>The columns in the following table represent the following information:</p>
+					<p id="confreq-cmt">Except as detailed in <a href="sec-foreign-resource-exemptions"></a>, EPUB
```
```suggestion
					<p id="confreq-cmt">Except as detailed in <a href="#sec-foreign-resource-exemptions"></a>, EPUB
```

epub33/core/index.html
```diff
@@ -973,38 +1147,90 @@ <h5>Foreign Resources</h5>
...
-							scientific journal might include a data set with instructions on how to extract it from the
-							EPUB Container).</p>
+						<p>This exception allows EPUB Creators to include resources in the <a>EPUB Container</a> that
+							are not for use by EPUB Reading Systems. The primary case for this exception is to allow
+							data files to travel with an EPUB Publication, whether for scripts to use in their
+							constituent EPUB Content Documents or for external applications to use (e.g., a scientific
+							journal might include a data set with instructions on how to extract it from the EPUB
+							Container).</p>
+
+						<p>It also EPUB Creators to use Foreign Resources in Foreign Content Documents without Reading
```
```suggestion
						<p>It also allows EPUB Creators to use Foreign Resources in Foreign Content Documents without Reading
```

epub33/core/index.html
```diff
@@ -973,38 +1147,90 @@ <h5>Foreign Resources</h5>
...
+							Systems or <a>EPUB Conformance Checkers</a> having to understand the fallback capabilities
+							of those resources (i.e., the requirement for a fallback for the Foreign Content Document
+							covers any rendering issues).</p>
+
+						<p>This specification also exempts some elements from fallback requirements. In these cases,
+							Foreign Resources MAY be referenced without a fallback. For more information, refer to <a
+								href="#sec-xhtml-fallbacks"></a>.</p>
```
iherman: I am a bit bothered by the separate §3.1.4.4 on Foreign Resource Restrictions. 

First of all, there are redundancies: the first paragraph of §3.1.4.4. seems to be a duplication of requirements that are defined in these sections. Also, §3.1.4.4 appears in a section on "HTML Deviations and Constraints", but the features described in that section are not really a constraint on HTML per se, but on how an EPUB specific feature binds to some HTML features.

Would it be feasible to move the content of 3.1.4.4. to §2.2.4 in some way or others? That would mean that all fallback-related definitions would be at the same place...

iherman (changes requested):
I must admit that I originally had a slightly different mental model for the resources than this ["axes" view](https://github.com/w3c/epub-specs/pull/2101#discussion_r832119056). I have tried to make a diagram for a categorization of resources in 

https://docs.google.com/drawings/d/1HFke3nLDSOmZZtjwV38SFrNbA_omRtTgafK51PmEsWA/edit

where the issues around manifest storage or fallbacks appear as "characterizations" of the various resource types. I am just throwing this into the discussion, because I am still not 100% convinced that this "axes" view is the more readable...

mattgarrish:
> I am just throwing this into the discussion, because I am still not 100% convinced that this "axes" view is the more readable...

There a few problems with your view:

1. EPUB and Foreign Content Documents are inverted. They are not instances of core media type resources, but are, from a container perspective (or all publication resource perspective), a separate class of resources.
2. We could disallow all CMTs and you'd still have EPUB Content Documents allowed in the spine because they are the central building blocks of EPUB. They don't get their legitimacy from CMTs.
3. Core Media Type Resources and Foreign Resources are only used in rendering EPUB Content Documents (and Foreign Content Documents but we tend to care less about these uses).
4. EPUB Content Documents are only considered a Core Media Type because it allows their embedding in other EPUB Content Documents. Technically, what we allow as CMTs are XHTML and SVG documents that are also required to conform to the XHTML and SVG Content Document definitions.
5. Foreign Content Documents are everything _in the spine_ minus EPUB Content Documents. They don't flow from Core Media Type Resources or even from Foreign Resources.

I understand what you're trying to achieve by coaxing the core media types and foreign resources beyond their defined uses, but it doesn't work. There aren't neat lines because a single resource can, for example, be both a Foreign Content Document and Core Media Type depending on the context in which it is used.

epub33/core/index.html
mattgarrish: Do you want to take this up separately? I bet there are a lot of instances where we mention the "Working Group". One of the caution boxes even mentions what the working group thinks w3c will do...

epub33/core/index.html
mattgarrish: Sure, I like this idea!

Sometimes we _can_ agree on resources... ;)

mattgarrish:
> There a few problems...

I guess this comment is kind of moot if we agree about axes. Maybe I should have read the full review first...

epub33/core/index.html
iherman: :-)

epub33/core/index.html
iherman: ok, forget about this one, and will come back to it when the current flurry of PR-s are done...

iherman:
> > There a few problems...
> 
> I guess this comment is kind of moot if we agree about axes. Maybe I should have read the full review first...

I think we agree on those. Not sure whether to use the term "facet" or "axis", though. I leave that choice to you.

epub33/core/index.html
```diff
@@ -691,277 +732,462 @@ <h4>Conformance Checking</h4>
...
-							Foreign Resources does not impact on the ability of the user to consume the content.</p>
-
-						<p>This section lists the <a href="#sec-cmt-supported">set of Core Media Type Resources</a> and
-							identifies <a href="#sec-foreign-restrictions">fallback mechanisms</a> used to satisfy the
-							consumability requirement.</p>
+					<ul>
+						<li>The <a>publication axis</a> &#8212; The publication axis describes the overarching
+							requirements for all the resources of the EPUB Publication (namely, <a>Publication
+								Resources</a> and <a>Linked Resources</a>).</li>
+						<li>The <a>spine axis</a> &#8212; The spine axis describes the content documents used in
```
```suggestion
						<li>The <a>spine axis</a> &#8212; The spine axis describes the resources used in
```

We use the term 'resources' in all the other places, probably better here, too (although we also use the term 'document' extensively...)

epub33/core/index.html
```diff
@@ -691,277 +732,462 @@ <h4>Conformance Checking</h4>
...
-									Publications.</p>
-							</li>
-							<li><strong>Content Type Definition</strong>—The specification to which the given Core Media
-								Type Resource must conform.</li>
-							<li><strong>Applies to</strong>—The Publication Resource type(s) that the Media Type and
-								Content Type Definition applies to.</li>
+							<li>Video resources are neither Core Media Type Resources nor Foreign Resources. EPUB
+								Creators may include any video format without fallbacks, but Reading Systems are not
+								guaranteed to support any video format.</li>
+							<li>Font resources do not require fallbacks because CSS rules will always ensure Reading
```
iherman: Actually... Isn't it true that a font resource is a Core Media Type Resource? After all, the media types for WOFF, OpenType, etc, are listed in §2.2.2. If so then, per definition, Font files do not need fallbacks, and there is no reason to list them as exceptions...

iherman: Or do we say that fonts _MUST NOT_ have fallbacks? I don't think so...

mattgarrish:
You're jumping in before I'm done rewriting... ;)

I'm actually still pondering some things, like the names of the axes.

I had trouble making "manifest axis" work because listing in the manifest is more a byproduct of the type of resource. You could equally call it a "container axis" because the type of resource also gets reflected in whether it is in the container or not. I opted to go with "publication axis" to cover that these are all the possible resources.

I also had trouble with "rendering axis" because the spine is also a rendering axis. CMTs and Foreign Resources sort of fall into the subrendering axis of the spine rendering axis. That's why I have it as content document axis right now.

These may still change.

I'll have a look at the comments you've made, though.

epub33/core/index.html
mattgarrish: > Isn't it true that a font resource is a Core Media Type Resource?

Technically, no. Not for all font resources, only the ones listed in the CMT section.

It's kind of weird that we still list those, but I guess they're referenced for the obfuscation section.

> Or do we say that fonts MUST NOT have fallbacks?

No, I don't think we could go this far without a small risk causing backwards compatibility issues. You used to have to provide fallbacks for foreign resource fonts, so there may be content somewhere out there that does this.

iherman: > Technically, no. Not for all font resources, only the ones listed in the CMT section.

Ie, those whose media type is listed. I got that, but maybe it is worth emphasizing it then... I will not be the only one asking this questions

(I thought that the four font types listed cover the vast majority of font types, so the leftover are not necessarily relevant in practice, but I may be very wrong...)

epub33/core/index.html
mattgarrish: I'm working at this, too. The part about exempt elements can fit in the new 2.2.3.1, but squashed in the middle is a set of requirements for the img element that have nothing to do with exemptions from fallbacks.

I'm still trying to find someplace to put that text. It may end up as a subsequent subsection to 2.2.3.1, but it's a weird mix of using CMTs and foreign resources.

epub33/core/index.html
iherman: Thinking about it further: the formulation of those four exceptions is still bothering me. I wonder whether it is not cleaner to say that, on the Content Document Axis (as it is called now) the categorization between Core Media Type and Foreign Resources are not exhaustive, but the division along that axis is three-way: CMT, Foreign, and, say (just for the sake of this comment) Extras. Extras are characterized by the fact that fallbacks are not to be defined for these files for various reasons. These include then videos, files that are linked or used via tracks, and fonts. It sounds different from 'exceptions'... Just a thought.

mattgarrish: Ya, we technically have a kind of "Exempt Resources" class that are neither CMTs not Foreign Resources. That's what video falls into, plus fonts and anything in the list of exempt elements.

> I thought that the four font types listed cover the vast majority of font types, so the leftover are not necessarily relevant in practice

Probably, but do we know for certain no one has used any others? Do we know that no one had to use fallbacks to work around our not having a font media type in the CMT list in the past. If we explicitly disallow fallbacks for fonts we're potentially invalidating all these cases.

Still, I don't understand how we can exempt all fonts from fallbacks and also have a set of CMTs. My gut says if we need to explain what fonts are in the obfuscation section, we should list the table there.

murata2makoto:
I have thought that I have a sound understanding of EPUB, but @iherman and @mattgarrish shook my confidence!

Is Ivan's diagram now consistent with the proposed text?

epub33/core/index.html
mattgarrish: Although, before I get too hasty to change things, CMTs are supposed to be supported, so I guess technically we expect support for the CMT fonts while also allowing any other fonts without fallbacks even though there is no requirement of support for them. It's kind of a weird in-between situation. That's what's confusing me about the exemption.

mattgarrish: If we make an "exempt" class, though, we could move the font exemption explanation to a new section and explain its purpose better there. It would at least make clearer the distinction between fonts that Reading Systems are expected to support and ones they aren't.

mattgarrish:
> Is Ivan's diagram now consistent with the proposed text?

No, the diagram linked above has a number of problems with it that I described in an earlier comment.

The planes approach is what we're fleshing out right now, which is like this:

> manifest plane
> Publication Resource: must be listed in the manifest
> Linked Resource: is not to be listed in the manifest
> 
> spine plane
> EPUB Content Document: can appear in the spine without fallback, these are XHTML and SVG
> Foreign Content Document: can appear in the spine only if a fallback is defined, all non-EPUB Content Documents are in this category
>
> content plane
> Core Media Type Resource: RS can be expected to support (see the core media types) do not necessarily require fallbacks
> Foreign Resources: require fallbacks under all circumstances
> Exempt Resources: not supported but no fallback requirements

There are overlaps between these planes, of course. Everything in the lower two is also a publication resource, for example, and a core media type resource may also be a foreign content document in the spine.

But this should be consistent with how we've always defined resources in the epub. It's just never been made clear.

mattgarrish:
I still have to do some more cleanup and consistency checking on this, but I've added the "exempt resource" class. Let me know if this new structure is along the lines of what you were thinking, @iherman.

iherman:
> I have thought that I have a sound understanding of EPUB, but @iherman and @mattgarrish shook my confidence!
> 

Welcome in the club!

> Is Ivan's diagram now consistent with the proposed text?

No. Forget about that one...

iherman:
> I still have to do some more cleanup and consistency checking on this, but I've added the "exempt resource" class. Let me know if this new structure is along the lines of what you were thinking, @iherman.

Very much so! Thanks.

(I am still thinking about an example or diagram, or both, that may be helpful, but I do not have a really good idea yet. One thing I am considering is some sort of a real-life-looking OPF file example that contain most of the variations.)

mattgarrish:
> I am still thinking about an example or diagram, or both, that may be helpful

Yes, I'm not opposed to having a diagram, too, it's just that we disagreed earlier what it should look like. :)

mattgarrish:
I've renamed the axes again in the lastest commit. I've gone back to "manifest axis" because I like the parallel with the manifest, even if it doesn't completely work. I've added some prose to the definition to try and place the link elements as a kind of extension of the manifest, at least from an "axis" point of view.

I also shortened "content document axis" to "content axis". It's less of a mouthful and should still convey that it's what is within content documents.

mattgarrish:
Switched to calling them "planes" after an offline discussion with @iherman as "axes" is not a good fit conceptually.

epub33/core/index.html
```diff
@@ -691,320 +742,636 @@ <h4>Conformance Checking</h4>
...
+								of no support.</p>
+							<p>Refer to the <a data-cite="epub-rs-33#confreq-css-rs-fonts">Reading System support
+									requirements for fonts</a> [[EPUB-RS-33]] for more information.</p>
+						</dd>
 
-						<p id="confreq-foreign-no-fallback">EPUB Creators MAY include <a>Foreign Resources</a> without
-							fallbacks provided they:</p>
+						<dt id="exempt-links">Linked resources</dt>
+						<dd id="confreq-resources-cd-fallback-link">
+							<p>Any resource referenced from the [[HTML]] <a data-cite="html#the-link-element"
```
iherman, line 1261: I think that, just as in the case of Fonts, we should say "All resources not already covered Core Media Types...". Otherwise CSS, used via a link element, has a dual categorization...

Maybe a general statement for all exempt resources may be better to cover all cases.

mattgarrish:
I was just going over the new example in more detail and while it's generally fine, xhtml documents were being listed on the content plane when they weren't applicable there. An xhtml document has to be embedded inside an epub content document for it to be a core media type resource on the content plane. Most xhtml will only exist on the spine plane, and in the case of a nvigation document not listed in the spine, only on the manifest plane.

To try and help explain this, I've added an iframe to the example with "widget.xtml" explaining that it doesn't exist on the spine plane and does on the content plane.

Also, fallbacks are not listed in the spine unless you also want them to render to the reader, but that would be kind of weird. The fallback chain means they automatically slot into the spine item of the resource that couldn't be rendered (which makes annotating/bookmarking spine items with fallbacks flaky, but that's for another day). The text alternative to the image in the spine isn't listed in the spine.

epub33/core/index.html
mattgarrish: Yes, good catch!

I tried to come up with a general explanation, but it really only applies to fonts and html links. I found it much clearer just to clearly exclude CMTs in both cases. My best attempts at a general explanation sound like a lot of confusing bloviation, as the exemptions aren't consistent (fonts aren't tied to a specific mechanism while links are).

The latest commit adds an exclusion on CMTs and also includes a link from the intro down to the new example.

murata2makoto:
@mattgarrish 
>Core Media Type Resource: RS can be expected to support (see the core media types) do not necessarily require fallbacks

But images are not allowed as spine times without fallbacks.  Right?

iherman:
> @mattgarrish
> 
> > Core Media Type Resource: RS can be expected to support (see the core media types) do not necessarily require fallbacks
> 
> But images are not allowed as spine times without fallbacks. Right?

That is correct.

iherman:
> Also, fallbacks are not listed in the spine unless you also want them to render to the reader, but that would be kind of weird. The fallback chain means they automatically slot into the spine item of the resource that couldn't be rendered (which makes annotating/bookmarking spine items with fallbacks flaky, but that's for another day). The text alternative to the image in the spine isn't listed in the spine.

I was wondering about that myself. If that document should not be on the spine (which I agree with) then we may have a problem with the categorization. In the (new) §2.2.5 on manifest fallbacks it says:

> For Foreign Content Documents, the chain MUST contain at least one EPUB Content Document

However, EPUB Content Documents are defined on the spine plane. Ie, just like the nav example, if an XHTML document does not appear in the spine, it is not an EPUB Content Document...

This means that either we have to define some sort of subclass of Core Media Type Resources for the XHTML/SVG contents (which may be over the top but is conceptually clean) or, in that clause, we make it explicit that the manifest chain contains at least one XHTML or SVG document...

mattgarrish:
> However, EPUB Content Documents are defined on the spine plane.
> Ie, just like the nav example, if an XHTML document does not appear in the spine, it is not an EPUB Content Document.

I'm not sure I'm following you here. Any fallback has to be on the spine plane, otherwise how could it be a fallback to a spine item? If it's not another Foreign Content Document or EPUB Content Document, then it can't be a fallback.

It isn't _directly_ listed as a unique spine item, but shares it's place in the spine through the fallback chain.

If anything, you can sort of say that it exists outside all but the manifest plane, like the nav, until called upon in the spine, but that's not a distinction that's going make the model comprehensible (or maybe even logical). If we try to split hairs too finely like this, then what is a manifest fallback for an `img` element? It's not on the content plane, right, because it's not directly listed in any epub content document.

We can maybe make explicit in the intro that resources in the fallback chain are always on the same plane, but that's as far as I'd go.

mattgarrish:
> But images are not allowed as spine times without fallbacks. Right?

Right, one of the big improvements in this model is filling in the gaps in terminology. We shouldn't be talking about Core Media Type Resources in the spine. Core Media Types are only what are allowed in EPUB Content Documents.

What this PR introduces is Foreign Content Documents to talk about things in the spine that require fallbacks. Foreign Content Documents are everything that is not an XHTML or SVG Content Document. Or put another way, Foreign Content Documents encompass all of Foreign Resources plus almost all Core Media Types.

Talking about CMTs in the spine has always been confusing as that designation was created for formats you can use _within_ content documents. As I tried to explain in the new intro, XHTML and SVG documents are CMTs because it allows them to be embedded in content documents via iframe, object, etc. They also happen to be interchangeable as XHTML Content Documents and SVG Content Documents because we require them to conform to the same definitions, but they are never "Core Media Types in the spine".

iherman:
I am fine with the extra paragraph you added to the spine plane:

> Although they are not directly listed in the spine, all of the resources in the fallback chain are considered part of the spine, and by extension part of the spine plane, since any may be used by a Reading System.

I actually also liked your terminology in the example:

> The resource is an XHTML document. It is the "target" of a manifest fallback so is not listed in the spine (i.e., it replace the existing spine reference when needed).

Ie, I think we are fine with this.

epub33/core/index.html
```diff
@@ -10505,6 +10758,212 @@ <h3>Media Overlays Schema</h3>
...
+							the <a>manifest plane</a>, a <a>Foreign Content Document</a> on the <a>spine plane</a>, and
+							a <a>Core Media Type Resource</a> on the <a>content plane</a>. As a <a>Foreign Content
+								Document</a> a fallback is required, and is provided via a <a
+								href="#sec-manifest-fallbacks">manifest fallback</a>.</p>
+					</dd>
+
+					<dt><code>image_desc.xhtml</code></dt>
+					<dd>
+						<p>The resource is an XHTML document. It is the "target" of a manifest fallback so is not listed
+							in the spine (i.e., it replace the existing spine reference when needed). It is a
```
```suggestion
							in the spine (i.e., it replaces the existing spine reference when needed). It is a
```

mattgarrish:
Not critical to this PR, but it might be good to rearrange the sections of the RS to match the core. For some reason, it has Foreign Resources, Remote Resources, and Data URLs before Core Media Types.

Also, what still bugs me about manifest fallbacks in section 2 is that all the reading system requirements on their processing are combined into the Manifest section under Package Document Processing. I guess we don't have to exactly parallel the core specification, but that exemplifies the disconnect I have with it. It's all about the mechanism in the package document moreso than about resources themselves. I can live with it if no one else finds it problematic, though.

mattgarrish:
To help anyone who wants to review this PR (which would be helpful for a change this size!), here's a recap of what has been updated, with links into the preview.

Of note is that this PR __does not change the way that you include or use resources__ in a publication, the manifest, or the spine. It's all about better explaining why we have all the classes and what their uses are.

- There are three new terms to fill in the logic gaps: [Exempt Resource](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html#dfn-exempt-resource), [Foreign Content Document](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html#dfn-foreign-content-document), and [Linked Resource](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html#dfn-linked-resource)
- There is an extensive new [introduction to the Publication Resources section](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html#sec-pub-res-intro), this includes the definitions of the [manifest plane](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html#sec-manifest-plane), [spine plane](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html#sec-spine-plane), and [content plane](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html#sec-content-plane). (These "planes" are purely informative and only exist within the intro for descriptive purposes.)
- The "Core Media Types" parent section is removed, which puts Foreign Resources as a sibling section of the [new Core Media Types section](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html#sec-core-media-types) (this used to be called "Supported Types").
- The specific details we had for HTML fallbacks are now defined in two new subsections of Foreign Resources: [HTML `audio` Fallbacks](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html#sec-fallbacks-audio) and [HTML `img` Fallbacks](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html#sec-fallbacks-img)
- There is a new [Exempt Resources section](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html#sec-exempt-resources). This section collects the various exemptions we've defined (for HTML elements, Fonts, and resources in the container but not used in the direct rendering).
- [Manifest Fallbacks](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html#sec-manifest-fallbacks) is moved up to section 2 from where it was under the Manifest section of the Package Document. 
- There is a new [detailed example of publication resources](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2101.html#publication-resources-example) explaining how resources in an example epub fit into the different planes depending on their uses.

Everything else is just linking cleanup, rewordings to use the new terminology, and various minor fixes where we were using the wrong terms to explain concepts.

iherman:
@dauwhe @wareid @shiestyle we would really need extra eyes on this PR. It may be worth adding it on the agenda of this week's call, too.

epub33/core/index.html
```diff
@@ -691,320 +744,646 @@ <h4>Conformance Checking</h4>
...
+						<h5>The Content Plane</h5>
+
+						<p>The <dfn>content plane</dfn> classifies resources based on their use in rendering <a>EPUB
+								Content Documents</a> and <a>Foreign Content Documents</a>. These resources fall into
+							three categories: <a>Core Media Type Resources</a>, <a>Foreign Resources</a>, and <a>Exempt
+								Resources</a>.</p>
+
+						<p>A Core Media Type Resource is one that <a>Reading Systems</a> have to support, so it can be
+							used without restriction in EPUB or Foreign Content Documents. For more information about
+							Core Media Type Resources, refer to <a href="sec-core-media-types"></a>.</p>
```
```suggestion
							Core Media Type Resources, refer to <a href="#sec-core-media-types"></a>.</p>
```

epub33/core/index.html
```diff
@@ -691,320 +744,646 @@ <h4>Conformance Checking</h4>
...
+								System without a <a>Viewport</a>, for example, will not render visual content such as
+								images.</p>
+						</div>
+
+						<p>The opposite of Core Media Type Resources are Foreign Resources. These are resources that
+							Reading Systems are not guaranteed to support the rendering of. As a result, similar to how
+							using Foreign Content Documents in the spine requires fallbacks to ensure their rendering,
+							using Foreign Resources in content documents also requires fallbacks. These fallbacks are
+							provided in one of two ways: using the capabilities of the parent format or via manifest
+							fallbacks.</p>
```
iherman, line 892: The term "host" format is used elsewhere (e.g., §2.2.3) instead of "parent format". I can live with any of the two (although I have a slight preference for "host") but we should be consistent.

iherman (approved):

epub33/core/index.html
mattgarrish: Ya, "parent" is kind of weird, as that would make the markup its... child? Changed to host.

murata2makoto:
When will this issue be closed?  I think that this issue deserves discussion in a WG meeting and 1-week review.

dauwhe (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-03-31-epub#section1) on 2022-03-31

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-03-31-epub#resolution1):  Merge Pull Request #2101.***

### 1. Expand and fix resource explanations (pr epub-specs#2101)

_See github pull request [epub-specs#2101](https://github.com/w3c/epub-specs/pull/2101)._




> *Dave Cramer:* [https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2101/017a47c...6a76ddc.html#sec-publication-resources](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2101/017a47c...6a76ddc.html#sec-publication-resources).

**Dave Cramer:** this is a very comprehensive re-working of how we define and explain publication resources.  

**Matt Garrish:** ivan had some diagrams in the PR, and this was also a useful exercise for identifying areas where terminology had gotten kind of wonky.  
… it looks big, but the bulk of the PR is adding an introduction and some examples.  
… ultimately we broke it down to 3 "planes".  
… the top one is _manifest plane_: everything in the container.  
… we had a term called publication resources, which described anything used in rendering, and then we had non-publication resources for everything else.  
… what we've done is to formalize the term linked resource to describe these other resources.  
… moving down from that we have the _spine plane_ and epub content documents.  
… it was kind of a mix, sometimes we called things in the spine publication resources, sometimes we referred to foreign resources.  
… so we've made it either epub content document or foreign content document.  
… finally the lowest level is the _content plane_: everything that gets rendered in an epub content document.  
… this is where CMT and foreign resource live as concepts.  
… and we created a 3rd category of exempt resources that collects together all these exemptions to CMT/fallback rules, e.g. fonts.  
… this PR basically just sets up this understanding and these terminology.  
… ivan went and did an extensive example in the appendix.  
… and how each of the parts of the example would match up to the 3 planes.  
… the rest is just clean-up, to make sure that our use of terminology is consistent, try to weed out places where we previously lacked defined terminology so meaning was muddier.  

**Dave Cramer:** I wish I could wipe my memory of epub and start over with this.  

**Matt Garrish:** yes, it was a really helpful exercise, and discussing without this was very convoluted.  

**Dave Cramer:** yes, I've witnessed confusion because of the lack of terminology.  
… about remote resources, does that concept stay the same as part of this picture?.  

**Matt Garrish:** we described where remote resources exist in the planes, but they aren't a class unto themselves.  
… i don't want to make it so complex in explaining it that we go back into the state of confusion where we were before.  

**Dave Cramer:** right, both remote resources and manifest fallbacks need to serve these other roles.  

**Murata Makoto:** i've been a bit skeptical about the publication resources. Now we have a class of linked resources and publication resources.  
… so now SSML and PLS are publication resources, right?.  

**Matt Garrish:** yes.  

**Murata Makoto:** so rendering means not just visual rendering, but any sort of rendering.  

**Matt Garrish:** publication resources are just the big giant bucket of everything used in rendering the epub, c.f. linked resources which are not directly part of the publication.  

**Dave Cramer:** ancillary to the publication.  

**Murata Makoto:** wondering if we really need the term publication resources, maybe just resource, or resources which are not linked.  

**Matt Garrish:** yeah, but then aren't we just going back to not have a direct term for it? would it conflict with other contexts where people talk about resources?.  

**Murata Makoto:** so you are saying resources within the zip file are one category, and that resources outside the zip file are something else.  

**Matt Garrish:** resources is just a very general term, so i'm worried that it becomes too broad.  
… and part of it is that publication resources is just always the term we've used.  

**Dave Cramer:** resources as a term is so overloaded, i don't think it gains us clarity.  

**Murata Makoto:** not happy with the definition of publication resources because it depends on rendering, which is not well defined in itself.  
… can we say that it is all non-linked resources in the epub?.  
… if some data is used for braille, are they publication resources? I guess so, but it is not so clear.  

**Dave Cramer:** i think we're clear that rendering is anything presented to the user.  

**Murata Makoto:** is that in the spec?.  

**Dave Cramer:** i think it's a term of art in the web/graphics world.  

**Murata Makoto:** before i got involved in a11y, i never thought of braille as rendering.  

**Matt Garrish:** it's still presented to the user.  
… to be honest, use of rendering was intentional, we specifically avoided use of display.  
… not sure how we would define that beyond the common definition of rendering.  

**Murata Makoto:** how about adding a note on the scope of rendering?.  
… specifically referring to braille terminals.  

**Matt Garrish:** that's fine with me, not sure where it would fit, but we could do this.  

**Dave Cramer:** maybe in the terminology section?.  

**Matt Garrish:** we could make a defined term out of it, like 'encompasses all the ways content may be presented, whether visual, tactile, or auditory'.  
… we do define some common terms, so probably no harm in providing some basic definition.  
… just worried about having to link it then (we are supposed to link each term the first time it is used).  
… but we don't have to settle this now.  

**Dave Cramer:** in general i think this is a massive improvement to the legibility of the spec.  
… it's also one of these things that is hard to sort out all the issues before merging it.  
… with big PRs there is always the possibility of typos sneaking in.  
… given broad support for the goals and language here, I propose we merge, and then we can file issues and fix as necessary, as with the definition of rendering above.  

> **Proposed resolution: Merge PR 2101.** *(Dave Cramer)*

**Matt Garrish:** the bulk of this is informative, so likely nothing normative changes as a result of this.  
… doesn't really block move to CR.  

> *Dave Cramer:* +1.

**Murata Makoto:** will we close this issue too?.  

> *Shinya Takami (高見真也):* +1.

**Dave Cramer:** yes, and then we can open more detailed issues on subsets of this.  

**Matt Garrish:** there's no actual issue, the PR is based on discussions i had with ivan.  

**Murata Makoto:** so nothing wrong with raising smaller issues related to this?.  

**Dave Cramer:** yes.  

**Matt Garrish:** and it's hard to raise issues off a PR.  

> *Matt Garrish:* +1.

> *Murata Makoto:* +1.

**Dave Cramer:** anyone else, please vote now.  

> *Matthew Chan:* +1.

> *Toshiaki Koike:* +1.

> *Masakazu Kitahara:* +1.

> ***Resolution #1: Merge Pull Request #2101.***

**Dave Cramer:** mgarrish please merge at your convenience.  

</details>

PR #2105, opened 2022-03-21
Update MO fragment reference recommendations

### LINKED ISSUE #2096: MO "Referencing Document Fragments" should probably not refer to HTML's "target element" concept
In section ["Referencing Document Fragments"](https://www.w3.org/TR/2022/WD-epub-33-20220317/#sec-media-overlays-fragids) EPUB 3.3 says (since #1663):

> For XHTML and SVG Content Documents, the fragment SHOULD be a reference to a [target element](https://html.spec.whatwg.org/multipage/#target-element) [[HTML](https://www.w3.org/TR/2022/WD-epub-33-20220317/#bib-html)] (…)

Referring to "target element" looks wrong to me. If I understand correctly, the _target element_ is a property of a Document, defined algorithmically. So speaking about **a** _target element_ is not appropriate IMO.

HTML calls the thing identified by the fragment "[the indicated part of the document](https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-indicated-part-of-the-document)". It then says that the mapping semantics depends on the document's MIME type, and for XHTML (being XML), this is defined in RFC7303, which in turns refers to the XPointer Framework.

So it looks like if we want to be precise we're back to what we wanted to avoid in #1586? 🙃

Could we maybe simplify this by saying:
> the fragment SHOULD indicate an HTML element or be an SVG Fragment Identifier, respectively

🤔

mattgarrish:
I'm not sure this is right yet, but opening so we can tweak the PR rather than talk about it.

The text right now is:

> The fragment SHOULD resolve to the [body element](https://html.spec.whatwg.org/multipage/#the-body-element) or an instance of [palpable content](https://html.spec.whatwg.org/multipage/#palpable-content) [[HTML](http://localhost:120/core/#bib-html)] for [XHTML Content Documents](http://localhost:120/core/#dfn-xhtml-content-document), or be an [SVG Fragment Identifier](https://www.w3.org/TR/SVG/linking.html#SVGFragmentIdentifiers) [[SVG](http://localhost:120/core/#bib-svg)] for [SVG Content Documents](http://localhost:120/core/#dfn-svg-content-document).

My one concern is calling out the specific allowed html elements. Will this be a source of ongoing headaches for epubcheck, given that the list changes over time?

Fixes #2096

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2105.html" title="Last updated on Mar 21, 2022, 4:49 PM UTC (f38c2d3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2105/d6c4632...f38c2d3.html" title="Last updated on Mar 21, 2022, 4:49 PM UTC (f38c2d3)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5775,23 +5775,23 @@ No Entry</pre>
 				<p>Although control over the rendering of <a>EPUB Content Documents</a> to create <a
 						href="#sec-fixed-layouts">fixed layouts</a> is an obvious need not handled by other
 					technologies, there are also considerations for reflowable content that are unique to EPUB
-					Publications (e.g., how to handle the flow of content in the <a>Viewport</a>). This section
-					defines properties that allow <a>EPUB Creators</a> to control presentation aspects of reflowable
+					Publications (e.g., how to handle the flow of content in the <a>Viewport</a>). This section defines
+					properties that allow <a>EPUB Creators</a> to control presentation aspects of reflowable
 					content.</p>
 
 				<section id="flow">
 					<h4>The <code>rendition:flow</code> Property</h4>
 
-					<p>The <code>rendition:flow</code> property specifies the EPUB Creator preference for how
-						Reading Systems should handle content overflow. </p>
+					<p>The <code>rendition:flow</code> property specifies the EPUB Creator preference for how Reading
+						Systems should handle content overflow. </p>
 
-					<p id="property-flow-global">When the <a href="#flow"><code>rendition:flow</code> property</a>
-						is specified on a <code>meta</code> element, it indicates the EPUB Creator's global
-						preference for overflow content handling (i.e., for all spine items). EPUB Creators MAY
-						indicate a preference for dynamic pagination or scrolling. For scrolled content, it is also
-						possible to specify whether consecutive <a>EPUB Content Documents</a> are to be rendered as
-						a continuous scrolling view or whether each is to be rendered separately (i.e., with a
-						dynamic page break between each).</p>
+					<p id="property-flow-global">When the <a href="#flow"><code>rendition:flow</code> property</a> is
+						specified on a <code>meta</code> element, it indicates the EPUB Creator's global preference for
+						overflow content handling (i.e., for all spine items). EPUB Creators MAY indicate a preference
+						for dynamic pagination or scrolling. For scrolled content, it is also possible to specify
+						whether consecutive <a>EPUB Content Documents</a> are to be rendered as a continuous scrolling
+						view or whether each is to be rendered separately (i.e., with a dynamic page break between
+						each).</p>
 
 					<p>EPUB Creators MUST use one of the following values with the <code>rendition:flow</code>
 						property:</p>
@@ -5805,11 +5805,11 @@ No Entry</pre>
 						<dt id="scrolled-continuous">scrolled-continuous</dt>
 						<dd id="scrolled-continuous-dd" data-tests="#pkg-flow-scrolled-continuous">
 							<p>Render all Content Documents such that overflow content is scrollable, and the EPUB
-								Publication is presented as one continuous scroll from spine item to spine item
-								(except where <a href="#layout-property-flow-overrides">locally overridden</a>).</p>
-							<p>Note that EPUB Creators SHOULD NOT create publications in which different resources
-								have different block flow directions, as continuous scrolled rendition in EPUB
-								Reading Systems would be problematic.</p>
+								Publication is presented as one continuous scroll from spine item to spine item (except
+								where <a href="#layout-property-flow-overrides">locally overridden</a>).</p>
+							<p>Note that EPUB Creators SHOULD NOT create publications in which different resources have
+								different block flow directions, as continuous scrolled rendition in EPUB Reading
+								Systems would be problematic.</p>
 						</dd>
 
 						<dt id="scrolled-doc">scrolled-doc</dt>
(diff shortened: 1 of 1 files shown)


iherman:
I wait for the reaction of @rdeltour on the epubcheck question; I am fine with the text otherwise.

rdeltour:
> My one concern is calling out the specific allowed html elements. Will this be a source of ongoing headaches for epubcheck, given that the list changes over time?

For EPUBCheck, it certainly adds some complexity. We will have to check that the target is among the list, that it is not empty, etc, etc.

Spec-wise, I'm wondering what is the benefit of being explicit about it being palpable content?

Also, at first sight I doesn't sound totally irrelevant to tie an audio clip to an empty element, for instance to add content that would be present only in the audio version, or to do some kind of hybrid partial-text/full-audio book. But I defer to @marisademeglio's greater expertise here.

iherman:
> Spec-wise, I'm wondering what is the benefit of being explicit about it being palpable content?

Well... it may be a bit too picky. However... the `par` element is defined as

> The `par` element defines the parallel **rendering** of media objects.

(Emphasis is mine.) The definition of fragments is to refine the referencing of some content via, say, a `text` element that is part of a `par` element. Ie, spec-wise, if we do not use the term palpable, we are supposed to define what "rendering" means if the fragment refers to, say, a `script` element.

I agree it is picky, and if it creates major issues for epubcheck then I am fine dropping it; I doubt it would make a big difference in practice.

marisademeglio:
> 
> Spec-wise, I'm wondering what is the benefit of being explicit about it being palpable content?
> 

Agree, is there a problem that this solves? I think that #2096 is fixed by saying the fragment should resolve to part of the HTML content document.

The HTML spec does not treat palpable content as a hard requirement, so I don't think we should put an uppercase "SHOULD" around it. 

By the way, is this palpable content? 
```
<span
          id="pgbreak1"
          role="doc-pagebreak"
          aria-label="14"/>
``` 

> Also, at first sight I doesn't sound totally irrelevant to tie an audio clip to an empty element, for instance to add content that would be present only in the audio version, or to do some kind of hybrid partial-text/full-audio book. But I defer to @marisademeglio's greater expertise here.

We don't have a hybrid book defined (and I think other issues would arise for that use case) but generally speaking I wouldn't want to put a hard restriction on what MO `text src` should refer to unless something breaks without such a restriction.

marisademeglio:
After reading this thread and its related issues and re-reading the MO section of the specs (core and RS), I would make these changes. Let me know if it would be more helpful to have them in separate issues or even just PRs.

1. Change [7.2.2.5](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2105.html#sec-smil-seq-elem) `epub:textref` and  [7.2.2.7](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2105.html#sec-smil-text-elem)`src` definitions to accommodate both fragment and document-level references, as the definition of its value is a URL plus an _optional_ fragment string:

Proposed for both `src` and `epub:textref`:
> Identifies the associated EPUB Content Document or part of it.

2. Change [7.2.2](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2105.html#sec-media-overlays-fragids) to reflect the optional-ness of this URL fragment string:

>The epub:textref attribute and the [text element's](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2105.html#elemdef-smil-text) src attribute both may contain a URL-fragment string that references a specific part (e.g., an element or text string) of the associated EPUB Content Document. 

We could also say something about how if no URL-fragment string is present, the RS may interpret it as a reference to the body element but I don't know if we have to say that or not.

3. Definition of `par`

Current:
> The par element defines the parallel rendering of media objects.

Issue:
As @iherman [mentioned](https://github.com/w3c/epub-specs/pull/2105#issuecomment-1075039924):
>if we do not use the term palpable, we are supposed to define what "rendering" means if the fragment refers to, say, a script element.

So why not remove the word "rendering"? Anyway, rendering is what you do with it, not what it is. 

Proposed:
> The par element is a parallel time container for media objects.

4. Along the same lines as (3): Change [7.3.1](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2105.html#sec-docs-intro) 

Current:
>Since par elements render their children in parallel, Reading Systems play the audio clip and EPUB Content Document fragment at the same time, resulting in a synchronized presentation.

Proposed
>Because par elements' media objects are timed in parallel, Reading Systems play the audio clip and EPUB Content Document fragment at the same time, resulting in a synchronized presentation.

5. Also along the same lines as (3) [7.3.2.1](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2105.html#sec-media-overlays-structure)

Current:
>The par element represents a segment of content to render, such as...

Proposed:
>The par element represents a segment of content, such as...

6. And again also along the same lines as (3)  [7.2.2.5](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2105.html#sec-smil-seq-elem)

Current: 
> The seq element contains a sequential rendering of media objects.

Proposed:
> The seq element is a sequential time container for media objects and/or child time containers.

7. [7.3.2.3](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2105.html#sec-media-overlays-granularity)

Issue: Refers to HTML "target element"

Proposed change in _**emphasis**_:
> The granularity level of the Media Overlay depends on how EPUB Creators mark up the EPUB Content Document and the type of fragment identifier they use in the [text elements'](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2105.html#elemdef-smil-text) src attributes _**and the seq elements' epub:textref attributes. For example, when referencing elements**_, if the finest level of markup is at the paragraph level, then that is the finest possible level for Media Overlay synchronization

8. [RS 8.2.3](https://w3c.github.io/epub-specs/epub33/rs/#sec-rsconf-rendering-text)

Current:
>When presented with a Media Overlay text element [[EPUB-33](https://w3c.github.io/epub-specs/epub33/rs/#bib-epub-33)] whose src attribute references a [target element](https://html.spec.whatwg.org/multipage/#target-element)

Issue: Refers to HTML "target element"

Proposed:
>When presented with a Media Overlay text element [[EPUB-33](https://w3c.github.io/epub-specs/epub33/rs/#bib-epub-33)] whose src attribute references a specific part of the content document

iherman:
@marisademeglio speaking for myself, I think it would be really helpful if you turned https://github.com/w3c/epub-specs/pull/2105#issuecomment-1076825048 into a PR encompassing both the core and the RS specs. Clearly, that PR would make this current PR moot, and would be an alternate solution for #2096. It would also make it easier to discuss and, if needed, adjust the text...

Thanks

PR #2117, opened 2022-03-24
Change in the example on landscape orientation

### LINKED ISSUE #2115: Example in Orientation does not show how rendition:orientation should be used
In [the Example](https://w3c.github.io/epub-specs/epub33/core/#fxl-ex2) of Orientation segment. It has title "Specifying global landscape orientation" but the statement is more related to `rendition:spread`.
```
In this example, the content should also render without synthetic spreads.
```
However I could not find any statement or explanation regarding to "rendition:orientation" in this example.

I also wonder how we should tell reading system about how to deal with `rendition:orientation"landscape"`

iherman:
Fix #2115

The synthetic spread term has been removed from the example. It is defined later, and does not add to the example in terms of the section it is in.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2117.html" title="Last updated on Mar 24, 2022, 2:45 PM UTC (210f2fa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2117/55d5466...210f2fa.html" title="Last updated on Mar 24, 2022, 2:45 PM UTC (210f2fa)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5383,7 +5383,7 @@ No Entry</pre>
 							for setting the property for individual <a>EPUB Content Documents</a>.</p>
 
 						<aside class="example" id="fxl-ex2" title="Specifying global landscape orientation">
-							<p>In this example, the content should also render without synthetic spreads.</p>
+							<p>In this example, items in the spine are to be rendered in landscape mode.</p>
 
 							<pre>&lt;package …>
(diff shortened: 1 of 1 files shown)


Jeffxz:
Looks good. Thanks for fixing it

PR #2136, opened 2022-03-28
Add note about other scripting contexts

### LINKED ISSUE #2098: Scripted SVG
I was just asked a question about embedding scripted svg and couldn't decipher exactly what is valid from the specification.

It's clear that you can have a scripted Top-Level SVG Content Document. And we say that an SVG with script referenced from an HTML iframe element is container-constrained.

But what happens in other cases? Can you embed an SVG with scripting using `object` or is that invalid? Can you embed an SVG with scripting in `img` even though the scripting won't work, or should that be reported as invalid?

mattgarrish:
Best place I could find for the note is right after we first define the top-level and container-constrained scripting contexts.

Fixes #2098

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2136.html" title="Last updated on Mar 30, 2022, 9:25 AM UTC (1709136)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2136/7e81268...1709136.html" title="Last updated on Mar 30, 2022, 9:25 AM UTC (1709136)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3343,8 +3343,8 @@ Package Document:
 								<p>The following example shows a reference to a remote audio file. Because the
 										<code>audio</code> element embeds the audio in its EPUB Content Document, the
 									file is considered a Publication Resource. EPUB Creators therefore must list the
-									audio file in the manifest and indicate that its parent EPUB Content Document
-									contains a remote resource.</p>
+									audio file in the manifest and indicate that its host EPUB Content Document contains
+									a remote resource.</p>
 
 								<pre>XHTML:
 &lt;html …>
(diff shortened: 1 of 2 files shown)


iherman (commented):
Is it the intention that we leave the issue of whether, conceptually, a script in SVG via object is considered to be spine level or not, open? Because this how I read this note.

My higher level impression is that a container level script has some restrictions (no dom or window size modification) whether for spine level scripts anything goes. In this respect, I do not see a problem saying that, conceptually, a script we are talking about may be viewed spine level...

mattgarrish:
> a script in SVG via object is considered to be spine level or not

It's not technically the same as a spine-level script as a spine-level script doesn't have a parent it can manipulate. It's a container-constrained script that isn't constrained by its container, I guess.

That's the only way I can read this RS spec requirement:

> Reading Systems MAY support scripting in other contexts, but this specification does not address such scripting.

So there are no defined constraints on what an SVG referenced from an `object` can do, and no requirement that scripting be supported.

iherman:
> container-constrained script that isn't constrained by its container

:-)

I guess we can just leave it as is then...

iherman (approved):

iherman:
Minor knit, the [section on container-constrained, ie, §3.3.2.2.1](https://w3c.github.io/epub-specs/epub33/core/#sec-scripted-container-constrained) also used the 'parent' terminology that we located elsewhere and replaced by 'host'. Maybe it is a good time to change it...

PR #2143, opened 2022-03-28
Copyedit: fix broken link and clarify that we're talking about XML

dlazin:
(The XML bit is obvious from the heading and the citation, but it's not in the readable part of the sentence; let me know if you'd structure this differently.)


mattgarrish:
Looks fine to me!

I think "all" made sense when we started this revision, but we've continuously whittled down the list.

PR #2156, opened 2022-03-29
Controversial copyedit: add dc namespace to element names

dlazin:
I'm proposing this because "the title element" is ambiguous with the HTML element of that name. I think it's clearer in the core spec, where we use [[DCTERMS]] before each (initial) reference. We could do that too.

Less confusing:
https://w3c.github.io/epub-specs/epub33/core/#sec-opf-dctitle

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -421,25 +421,25 @@
 								data-cite="epub-33#dfn-value">values</a> [[EPUB-33]] before processing.</p>
 					</dd>
 
-					<dt id="dc-identifier">The <code>identifier</code> element</dt>
+					<dt id="dc-identifier">The <code>dc:identifier</code> element</dt>
 					<dd>
-						<p>To determine whether an <code>identifier</code> conforms to an established system or has been
+						<p>To determine whether a <code>dc:identifier</code> conforms to an established system or has been
 							granted by an issuing authority, Reading Systems SHOULD check for an <a
 								data-cite="epub-33#identifier-type"><code>identifier-type</code> property</a>
 							[[EPUB-33]].</p>
 					</dd>
 
-					<dt id="dc-title">The <code>title</code> element</dt>
+					<dt id="dc-title">The <code>dc:title</code> element</dt>
 					<dd>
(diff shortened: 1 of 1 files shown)


mattgarrish:
Even if we add the prefix, there should always be proper citation of where elements/attributes/properties are sourced from (in this case, there should be a [DCTERMS] cite in front).

The problem with adding dc: is we'd probably also then need an authoring shorthand section to explain its mapping.

It looks like we lost a lot of citations because the RS requirements came further down from the initial use back in 3.2.

There are a lot of places where we should be linking across to the core spec, too for epub-specific markup. I don't like that some sections have requirements you can't quickly cross-reference to the other spec.

iherman:
The remark of @mattgarrish notwithstanding, we should make it sure that we do this type of change consistently across the full RS document. @dlazin, have you checked whether that is the only place where such prefix is needed?

dlazin:
> The remark of @mattgarrish notwithstanding, we should make it sure that we do this type of change consistently across the full RS document. @dlazin, have you checked whether that is the only place where such prefix is needed?

This seems to be the only place in the *RS* document where it's needed, and the only place to me where it's confusing. As noted in the PR description, something like https://w3c.github.io/epub-specs/epub33/core/#sec-opf-dctitle is much clearer. But also, I like consistency, and so do we all, and this PR is not consistent with that link.

In short, I don't really expect this PR to be accepted as it is; this is more like an issue report in visual form. How would you (mostly Matt) like to resolve this? Do you want to do it yourself?

(I agree that a [[DCTERMS]] citation would be good, but I don't know whether we want to do that in the headings.)

mattgarrish:
I've integrated the DC changes here into https://github.com/w3c/epub-specs/pull/2163

I didn't want to base off this PR because the main branch is rapidly evolving and I also wanted to fix xrefs and citations. It's not really controversial as much as putting back in the authoring shorthand, but I worked around adding a new section to the RS spec by linking the first instance of the element names back across.

We can close this PR and merge the other unless you spot something amiss in it (warning, it's a lot of added links and citations, so reading the previews might be easier than the diffs if you only want to see the dc changes).

PR #2164, opened 2022-03-30
Minor updates/changes on the exit criteria description

iherman:
Only minor things, but a review of the text in general would be good

- See [preview](https://raw.githack.com/w3c/epub-specs/admin/exit-criteria-march-22/epub33/reports/exit_criteria.html)
- See [diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fepub-specs%2Fepub33%2Freports%2Fexit_criteria.html&doc2=https%3A%2F%2Fraw.githack.com%2Fw3c%2Fepub-specs%2Fadmin%2Fexit-criteria-march-22%2Fepub33%2Freports%2Fexit_criteria.html)

### DIFF
--- a/epub33/reports/exit_criteria.html
+++ b/epub33/reports/exit_criteria.html
@@ -12,11 +12,9 @@
 
     </style>
     <script class="remove">
-        // All config options at https://respec.org/docs/
         var respecConfig = {
-            group: "epub",
-            wgPublicList: "public-epub-wg",
             specStatus: "base",
+            "latestVersion": "https://w3c.github.io/epub-specs/epub33/reports/exit_criteria.html",
             noRecTrack: true,
             editors: [{
                 name: "Ivan Herman",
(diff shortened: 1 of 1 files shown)


mattgarrish:
The only question I have is about WCAG 2.1 AA conformance for the accessibility spec. The specification only requires 2.0 A, but strongly recommends (non-normatively) the latest WCAG version and higher level.

So while I'm not arguing we should seek the lowest possible threshold to show conformance, do we want to say that 2.1 AA is the requirement for passing CR, even if I expect we'll find five publishers with reflowable content that can pass?

Should it maybe say that although 2.0 A is the minimum we need for CR, we will be looking to prove the spec by finding publishers who meet the latest standard?

/cc @avneeshsingh

shiestyle (approved):

iherman:
Thanks @mattgarrish, I did not realize this issue...

> So while I'm not arguing we should seek the lowest possible threshold to show conformance, do we want to say that 2.1 AA is the requirement for passing CR, even if I expect we'll find five publishers with reflowable content that can pass?

I am not sure I understand what you say... 

"even if" -> "because" ?

> Should it maybe say that although 2.0 A is the minimum we need for CR, we will be looking to prove the spec by finding publishers who meet the latest standard?

This would certainly be more consistent with the standard text. If we go down that line, we should emphasize the fact that we have actually 2.1 in the relevant implementation report.

mattgarrish:
> I am not sure I understand what you say...
> "even if" -> "because" ?

I'm just saying that I don't think it'll be hard to find five that meet 2.1 AA, but even though that's my expectation is it really what we want to say is the requirement for CR? I like to err on the side of caution, which in this case is to only require we meet the minimum but do our best to go far beyond that in what we present in our report.

> we should emphasize the fact that we have actually 2.1 in the relevant implementation report.

Yes, this exactly.

iherman:
I have changed the targeted feature to EPUB-A11Y-11_WCAG-20-AA, and changed the reference to WCAG 2.0 AA.

PR #2166, opened 2022-03-30
Less bolding in def boxes

mattgarrish:
I don't know what triggered me about the definition boxes this morning, but the bolded field names are jarring to read, especially when you hit all the ones in the vocabularies.

I looked at the boxes the CSS specs use and their field names are italicized, which I find less conflicting with headings, too.

I also fixed some other inconsistencies, like not always having colons after the names and some names using different cases. (If we don't keep the bolding change, we should still keep these.)

I don't know if the auto-previews will show the style changes, so here's an alternative: https://cdn.statically.io/gh/w3c/epub-specs/style/less-bold/epub33/core/index.html

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2166.html" title="Last updated on Mar 30, 2022, 5:01 PM UTC (3d6425b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2166/e8241a8...3d6425b.html" title="Last updated on Mar 30, 2022, 5:01 PM UTC (3d6425b)">Diff</a>


mattgarrish:
Don't ask me what caused 37 commits to join in the fun, but nothing is changed beyond what I did in my one commit. I might kill this branch and reapply the changes to a new one just to be on the safe side.

PR #2167, opened 2022-03-30
Clean up def boxes (try 2)

mattgarrish:
I don't know what triggered me about the definition boxes this morning, but the bolded field names are jarring to read, especially when you hit all the ones in the vocabularies.

I looked at the boxes the CSS specs use and their field names are italicized, which I find less conflicting with headings, too.

I also fixed some other inconsistencies, like not always having colons after the names and some names using different cases. (If we don't keep the bolding change, we should still keep these.)

I don't know if the auto-previews will show the style changes, so here's an alternative: https://cdn.statically.io/gh/w3c/epub-specs/style/def-box-fixes/epub33/core/index.html

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2167.html" title="Last updated on Mar 30, 2022, 5:10 PM UTC (5d2c2c0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2167/e8241a8...5d2c2c0.html" title="Last updated on Mar 30, 2022, 5:10 PM UTC (5d2c2c0)">Diff</a>

### DIFF
--- a/epub33/common/css/common.css
+++ b/epub33/common/css/common.css
@@ -33,6 +33,15 @@ dl.conformance-list > dd {
     margin-top: 0.7em;
 }
 
+dl.elemdef > dt,
+table.propdef th {
+	font-weight: normal;
+	font-style: italic;
+}
+
+dl.elemdef > dt {
+	margin-bottom: 0.5em;
+}
 
 /* styles for conformance list links */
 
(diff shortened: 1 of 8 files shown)


iherman (approved):
I had to look at the separate preview to understand what you are proposing... and I agree, what you propose is much more readable.

If this is o.k., we will have to update the SSV document, too, but that can be done separately.

mattgarrish:
> If this is o.k., we will have to update the SSV document, too, but that can be done separately.

It doesn't use definition boxes, so I'm not sure there's anything to do with it. I'd have to look at it again, but I think we use italics for the subfields of each term.

mattgarrish:
Let's try this and given its overall insignificance we can always revert back even post CR.

PR #2175, opened 2022-03-30
Copyedit: remove excess spaces in RS spec

dlazin:
If you like this, I can do the same to the core spec (much larger).


mattgarrish:
It would make me happy, but might be worth waiting until all the pull requests quiet down, or even add it as an issue for the post-CR list of things to do.

I have a strong suspicion that the spaces get introduced because we have different authoring tools. I don't start text on a new line inside paragraphs, list items, etc., like this:

```
<p>
	lorem ipsum dolor sit amet ...
</p>
```

Because when oxygen reformats lines it compacts the whitespace after/before the tags to a single space, which as far as I can tell is how we end up with them.

PR #2176, opened 2022-03-31
Added links to the meta-inf section in the RS spec

### LINKED ISSUE #2172: Link RS manifest.xml (etc) to core spec?
https://w3c.github.io/epub-specs/epub33/rs/#sec-container-metainf has a smattering of requirements about various XML files; the first two make sense on their own, but the manifest.xml requirement seems basically to be saying "the RS should not parse a manifest.xml file at all."

a) If it's saying that, can we just say that explicitly?

b) I was only able to interpret this after reading https://w3c.github.io/epub-specs/epub33/core/#sec-container-metainf-manifest.xml, but the RS spec does not link there. Should we link all these files to their longer sections in the RS spec? In particular, I found `This feature exists only for compatibility with ODF` to be helpful.

iherman:
This is to fix #2172.

Note that I did not change the text on the ecryption file, because that paragraph is subject to a separate issue. Any additional reference, if necessary, should be added there...

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1503,14 +1503,16 @@
 						<dd>
 							<p id="container-default-rendition"
 								data-tests="#ocf-package_arbitrary,#ocf-package_multiple">A Reading System MUST, by
-								default, use the Package Document referenced from first <code>rootfile</code> element to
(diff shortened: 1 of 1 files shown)


iherman:
Closing without merge; we duplicated work...

PR #2186, opened 2022-04-01
Editorials on a11y section in RS

### LINKED ISSUE #2182: EPUB Reading Systems 3.3 - Suggestion Edits: 13. Accessibility
Para 3:
"many browser accessibility issues have parallels in EPUB" s.b "many browser accessibility issues have parallels in EPUB-specific user agents"

Bookshelf bullet:
"announce the options" s/b "announce the content"

User Interface Controls bullet: 
"and their actions" should be "and that their actions"

Search bullet:
Does any Reading System allow Search capability to extend to alt attributes, or is this aspirational?

Reading Control:
Question the phrase "physical interaction" here... is manipulating a keyboard in order to open a new document not allowed?

Document Object Model (DOM) bullet:
change "used in the source" to "applied in the source content"

iherman:
Thanks Ben. Some comments below:

> Search bullet:  
> Does any Reading System allow Search capability to extend to alt attributes, or is this aspirational?

I honestly do not know. I am afraid it is more aspirational (@mattgarrish?) but we can leave it in the text anyway (this is not a normative text)

> Reading Control:  
> Question the phrase "physical interaction" here... is manipulating a keyboard in order to open a new document not allowed?

The text says "does not _depend_ on physical interaction" (emphasis is mine). Ie, using the keyboard is fine, but there should be other alternatives.

> Document Object Model (DOM) bullet:  
> change "used in the source" to "applied in the source content"

I am not sure I agree with what you propose. The issue here is that the semantics expressed in the source (e.g., the choice of an HTML element, possible microdata attributes, aria attributes, etc.) should be available. The DOM contains this information by default, and providing access to the DOM is what is needed for this.

I changed "used in the source" to "expressed in the source", this might be clearer.

cc: @BenSchroeter

Fix #2182

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2157,9 +2157,8 @@
 					and buttons to load features such as the table of contents) are exposed to assistive technologies
 					and that their actions do not rely on specific modalities (e.g., they only operate through touch or a
 					mouse).</li>
-				<li>Search &#8212; Search access to the full text content of all EPUB Content Documents (including
-					alternative text attributes, text in SVG images, etc.) is necessary to ensure users can locate
-					information easily.</li>
+				<li>Search &#8212; Search access to the full text content of all EPUB Content Documents is necessary 
+					to ensure users can locate information easily.</li>
 				<li>Display Control &#8212; Provide methods for users to tailor the styling of the content to their
 					preferences (e.g., to change and increase fonts, increase line and word spacing, and apply alternate
 					contrasts).</li>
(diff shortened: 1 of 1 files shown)


mattgarrish:
> I honestly do not know. I am afraid it is more aspirational (@mattgarrish?) but we can leave it in the text anyway

That part is a bit more aspirational in the mainstream. Searching alt text is problematic outside of a screen reader, or other AT that renders the content is a non-visual way, because there isn't anything to display to visual readers.

I don't think a lot is lost by dropping that parenthetical, though.

iherman:
@mattgarrish I have removed the parenthetical part. If you agree with the rest, we can merge I presume...

mattgarrish (approved):
I think being less specific here gives greater flexibility. Otherwise, we'd have to say something like "Depending on the rendering capabilities of the device, the full text may include ..."

I expect someone developing an aural reading system is aware that the full text could include attributes they play back, like alt. We're also not talking about a normative requirement but a consideration to take into account.

PR #2188, opened 2022-04-01
Example 65: DigestValue should not be empty

iherman:
In example 65 the DigestValue element was empty; I do not think it is correct

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2188.html" title="Last updated on Apr 2, 2022, 4:44 AM UTC (a104e39)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2188/b532068...a104e39.html" title="Last updated on Apr 2, 2022, 4:44 AM UTC (a104e39)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7617,9 +7617,7 @@ No Entry</pre>
                &lt;/Transforms>
                &lt;DigestMethod
                    Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
-               &lt;DigestValue>
-	           …
-               &lt;/DigestValue>
+               &lt;DigestValue>…&lt;/DigestValue>
             &lt;/Reference>
             &lt;Reference URI="EPUB/images/cover.jpeg">
                &lt;Transforms>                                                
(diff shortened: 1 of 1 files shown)


epub33/core/index.html
```diff
@@ -7618,6 +7618,7 @@ <h6>The <code>signatures</code> element</h6>
                &lt;DigestMethod
                    Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
                &lt;DigestValue>
+	           …
                &lt;/DigestValue>
```
mattgarrish:
```suggestion
               &lt;DigestValue>…&lt;/DigestValue>
```

epub33/core/index.html
```diff
@@ -7628,6 +7629,7 @@ <h6>The <code>signatures</code> element</h6>
...
```
mattgarrish:
```suggestion
               &lt;DigestValue>…&lt;/DigestValue>
```

mattgarrish (approved):
Yes, the examples it's based on all have some junk data followed by an ellipsis but I don't think we need to fake the value that much.

Only thing I'd do here is compact to a single line. The value would just be a string of characters, not nested tagging, so no need to make the example longer than it is.

PR #2203, opened 2022-04-02
Clean up manifest fallbacks section and add an extra caution

mattgarrish:
As mentioned in [#1464](https://github.com/w3c/epub-specs/issues/1464#issuecomment-1086242145), this PR separates the requirements for spine fallbacks from the requirements for content fallbacks.

I've also taken the technical details about the `fallback` attribute and merged them back into the `item` element definition, as this was the most bothersome part to me in moving this section up.

I've also added another layer of caution to the content fallbacks section that the mechanism was meant for `img` which now has other fallback mechanisms.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2203.html" title="Last updated on Apr 4, 2022, 4:56 PM UTC (2440b0f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2203/5e0b77a...2440b0f.html" title="Last updated on Apr 4, 2022, 4:56 PM UTC (2440b0f)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1381,16 +1381,17 @@
 								the fallback chain MUST contain at least one <a>Core Media Type Resource</a>.</p>
 							<p>EPUB Creators MAY also provide manifest fallbacks for <a>Core Media Type Resources</a>
 								(e.g., to allow Reading Systems to select from more than one image format).</p>
-							<div class="note">
-								<p>As it is not possible to use manifest fallbacks for resources represented in <a
-										href="#sec-data-urls">data URLs</a>, EPUB Creators can only represent Foreign
-									Resources as data URLs where an intrinsic fallback mechanism is available.</p>
-							</div>
 						</dd>
 					</dl>
 
 					<p>Regardless of the type of manifest fallback specified, fallback chains MUST NOT contain
 						self-references or circular references to <code>item</code> elements in the chain.</p>
+
+					<div class="note">
+						<p>As it is not possible to use manifest fallbacks for resources represented in <a
+								href="#sec-data-urls">data URLs</a>, EPUB Creators can only represent Foreign Resources
+							as data URLs where an intrinsic fallback mechanism is available.</p>
+					</div>
 				</section>
 
 				<section id="sec-resource-locations">
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -836,14 +836,14 @@
 						render something for the user to read, as there is no guarantee of support for Foreign Content
 						Documents.</p>
 
-          <p>A mechanism called <a href="#sec-manifest-fallbacks">manifest fallbacks</a> allows EPUB Creators
-            to provide fallbacks for Foreign Content Documents. In this model, the <a>manifest</a> entry for
-            the Foreign Content Document must include a <a href="#attrdef-item-fallback"
-               ><code>fallback</code> attribute</a> that points to the next possible resource for
-            Reading Systems to try when they do not support its format. Although not common, a fallback
-            resource can specify another fallback, thereby making chains many resources deep. The one
-            requirement is that there must be at least one EPUB Content Document in a <a>manifest fallback
-            chain</a>.</p>
+					<p>A mechanism called <a href="#sec-manifest-fallbacks">manifest fallbacks</a> allows EPUB Creators
+						to provide fallbacks for Foreign Content Documents. In this model, the <a>manifest</a> entry for
+						the Foreign Content Document must include a <a href="#attrdef-item-fallback"
+								><code>fallback</code> attribute</a> that points to the next possible resource for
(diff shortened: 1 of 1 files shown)


epub33/core/index.html
```diff
@@ -1340,58 +1337,55 @@ <h5>Manifest Fallbacks</h5>
 						<p>Reading System support for manifest fallbacks is poor and should not be relied upon to create
 							interoperable content. In most cases, if a Reading System does not support a format, it will
 							not use its fallback. This can lead to unexpected failures.</p>
-
-						<p>EPUB Creators should use the intrinsic fallback capabilities of HTML and SVG to provide
-							fallback content whenever possible, and avoid using <a>Foreign Content Documents</a> in the
-							spine.</p>
```
iherman: I understand the reason for the caution but, spec-wise, it looks a bit strange to start a section with essentially saying "ignore this section". This may raise some question like "why do you have this section in the first place?" or "why is it not deprecated instead?". We know the answer but, nevertheless...

I think it looks somewhat less awkward if that caution is moved to the end of the section.

epub33/core/index.html
```diff
@@ -1340,58 +1337,55 @@ <h5>Manifest Fallbacks</h5>
...
-							content</a>).</p>
+						<dt id="content-fallbacks">Content Fallbacks</dt>
+						<dd>
+							<div class="caution">
+								<p>The purpose for keeping content fallbacks in EPUB 3 was primarily to continue to
+									provide a means of specifying fallback images for the [[HTML]] <a
+										data-cite="html#the-img-element"><code>img</code> element</a>. As HTML now has
+									intrinsic fallback mechanism for images, the use of content fallbacks is strongly
+									discouraged. EPUB Creators should always use the intrinsic fallback capabilities of
+									[[HTML]] and [[SVG]] to provide fallback content.</p>
```
iherman: Same comment as above. I think it would be better to move this caution to the end of the `<dd>` part.

epub33/core/index.html
```diff
@@ -1340,58 +1337,55 @@ <h5>Manifest Fallbacks</h5>
...
+								the purpose of each.</p>
+						</dd>
 
-					<p>EPUB Creators MAY provide fallbacks for <a>Top-Level Content Documents</a> that are EPUB Content
-						Documents (e.g., to provide <a href="#confreq-cd-scripted-flbk">fallbacks for scripted
-							content</a>).</p>
+						<dt id="content-fallbacks">Content Fallbacks</dt>
+						<dd>
+							<div class="caution">
+								<p>The purpose for keeping content fallbacks in EPUB 3 was primarily to continue to
```
```suggestion
								<p>The original reason for introducing fallbacks in EPUB 3 was primarily to continue to
```

The term "keeping" makes sense when one is familiar with the history (I presume it came from EPUB 2), but it sounds a bit strange for people who came into this world later.  Hence my proposal for change

iherman (changes requested):

epub33/core/index.html
mattgarrish: But anyone who knows epub 2 knows that fallbacks didn't originate in epub 3, so it just reverses the confusion. 😄 

It also won't work logically to say we introduced fallbacks to continue to provide a means of specifying fallbacks for `img`. The precedent is missing. There was a small overlap between epub 2 and epub 3 when manifest fallbacks were the only way to handle `img`.

But to go off on a tangent, I was thinking of taking the two fallback sections currently under Foreign Resources and merging with manifest fallbacks under a "Resource Fallbacks" section. I'll ponder how to fix this note while I'm doing that.

epub33/core/index.html
mattgarrish: Or should we drop this one for now?

This is essentially pre-empting the "under-implemented" label which is going to tell people exactly the same thing. The caution about not needing to use manifest fallbacks for content anymore is a legitimate separate issue, but perhaps that only needs to be a note given it will also fall into the same general lack of support bucket.

iherman: Yep, I think it is better to drop it for now. 

We will have to think about how we will, editorially, mark up under-implemented features... maybe something more than just a parenthetical word would be needed, We will see.

epub33/core/index.html
mattgarrish: I've modified this in the latest commit to:

> The original purpose for content fallbacks was to specify fallback images for 

Hopefully this works both ways. If you don't know the original purpose comes from EPUB 2, you can think the original purpose was for EPUB 3 and it changes nothing.

("Original purpose" knowingly forgets all about EPUB 2's use for "out-of-line xml islands", which were forgettable, and `object` which became `bindings` in EPUB 3 and then we gave up on in place of intrinsic content in the `object`.)

iherman (approved):

PR #2214, opened 2022-04-03
Fix and explain remote resources

### LINKED ISSUE #2194: Proposal for a new plane: Local or Remote
Some resources are local (i.e., located inside the EPUB Container) while others are remote (i.e., outside of the EPUB Container).  In my understanding, local/remote is orthogonal to the manifest plane, spine plane, and content plane.  I thus propose to add the distinction of local and remote as another plane.

mattgarrish:
I've gone through and cleaned up the definitions and text to make sure "remote resources" is being used consistently for publication resources we've exempted from being in the container.

I've also changed "local resource" to "container resource" to make it more obvious what it is "local" to.

I've also done my best to fix #2194 by adding a couple more paragraphs to break down why some publication resources can be outside the container and why we don't care where linked resources are (plus an explanation that external hyperlinks don't fall into this discussion). If someone still wants to come up with a "plane" for this, I'm open to suggestions.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2214.html" title="Last updated on Apr 6, 2022, 2:14 PM UTC (e88f14e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2214/8f16c4d...e88f14e.html" title="Last updated on Apr 6, 2022, 2:14 PM UTC (e88f14e)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -10860,7 +10860,7 @@ html.my-document-playing * {
 			<h2>Detailed Examples</h2>
 
 			<section id="publication-resources-example">
-				<h3>Publication Resources</h3>
+				<h3>Resources</h3>
 
 				<p>Consider the following extracts of a <a>Package Document</a> and an <a>XHTML Content
 					Document</a>:</p>
@@ -10868,7 +10868,12 @@ html.my-document-playing * {
 				<pre>&lt;package …>
     &lt;metadata …>
         …
-        &lt;link rel="record" href="meta/data.xml" media-type="application/marc"/>
+        &lt;link rel="record" 
+            href="meta/data.xml" 
+            media-type="application/marc"/>
+        &lt;link rel="record" 
+            href="https://www.example.org/meta/data2.xml" 
+            media-type="application/marc"/>
         …
     &lt;/metadata>
     &lt;manifest>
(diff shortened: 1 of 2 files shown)


murata2makoto (commented):
Although I prefer a new plane, I can live with this pull request.  It is certainly a big improvement.

iherman (approved):
Would it be worth changing our example by making the font reference a remote resource? It would not change anything on the example, but would be an example for a Remote Resource while all others are Container Resources.

We could also add an `<a>` element to the html portion of the example linking with an absolute URL, an emphasizing in the example that that resource is not to be added to the manifest.

mattgarrish:
> Would it be worth changing our example by making the font reference a remote resource?

+1

> We could also add an `<a>` element to the html portion of the example linking with an absolute URL, an emphasizing in the example that that resource is not to be added to the manifest.

+1

mattgarrish:
> Although I prefer a new plane

I haven't had any breakthrough on what to call it if we create one.

If we want to keep matching up with epub terminology it feels like it has to be the "container plane". Can we rationalize the plane not with the container itself but with all the resources that the publication contains, sort of like how we rationalize the manifest plane as the manifest proper plus the linked resources?

iherman:
> Can we rationalize the plane not with the container itself but with all the resources that the publication contains, sort of like how we rationalize the manifest plane as the manifest proper plus the linked resources?

Maybe by taking the offline/online angle? Container Resources are available offline for the Reading System, Remote Resources are not?

mattgarrish:
> Maybe by taking the offline/online angle?

The availability plane? Still has a weird ring as you can be "online" and not have access to remote resources.

Let's merge and if we ever can figure out a meaningful concept for representing them separately we can always return to this. It'd just take hiving off a few paragraphs from the manifest plane description.

PR #2218, opened 2022-04-04
Editorial updates to conformance identification

### LINKED ISSUE #2215: Proposed rewrite of "3.5.2 Publication Conformance" in EPUB A11Y 1.1
JDC proposes to replace 

> To indicate that an EPUB Publication conforms to the accessibility requirements of this specification, it MUST include a conformsTo property whose value MUST exactly match (i.e., both in case and spacing) the following pattern:

by 

> Conformance to the accessibility requirements of this specification MUST be represented by a conformTo property exactly matching   (i.e., both in case and spacing) the following pattern.   ...

>Conformance to the accessibility requirements of additional specifications (for example, a specification dedicated to some natural language) MAY be represented by other conformTo property values.

This rewrite is to clearly allow values NOT matching the specified pattern when we would like to indicate that an EPUB Publication conforms to additional specifications (such as those for Japanese EPUB publications).

mattgarrish:
This pull request adds an additional explanation that we are not preventing the use of other dcterms:conformsTo properties to the note that already existed to clarify that you can claim wcag-only conformance.

It also does the following cleanup:

- makes the requirement in first paragraph that the metadata specify a property to avoid mentioning the creator
- removes the "musts" from what each segment of the string specifies, as we already control that by defining the segments. The normative requirement on two of the values is that they not include decimal points, so I've made those normative.
- the example is moved up after the explanation of the segments and the package element added to it
- the note about additional wcag versions is merged with the note about new wcag levels in 3.0 since these are related

Fixes #2215 

- Accessibility [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2215/epub33/a11y/index.html)
- Accessibility [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2215/epub33/a11y/index.html)

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -994,9 +994,12 @@
 				<section id="sec-conf-reporting-pub">
 					<h4>Publication Conformance</h4>
 
-					<p>To indicate that an EPUB Publication conforms to the accessibility requirements of this
-						specification, it MUST include a <code id="dcterms-conformsTo">conformsTo</code> property whose
-						value MUST exactly match (i.e., both in case and spacing) the following pattern:</p>
+					<p>To indicate conformance to the accessibility requirements of this specification, an EPUB
+						Publication's <a data-cite="epub-33#sec-pkg-metadata">metadata section</a> [[EPUB-33]] MUST
+						specify a <a
+							href="https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/conformsTo"
+								><code id="dcterms-conformsTo">conformsTo</code> property</a> [[DCTERMS]] exactly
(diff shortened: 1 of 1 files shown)


GeorgeKerscher:
I reviewed this section and I think it is fine. Thank you for providing a link to the preview, which makes it a lot easier to review.

clapierre (approved):

avneeshsingh (approved):

PR #2225, opened 2022-04-05
Improved the example with the new concepts

iherman:
Added

- a linked element stored remotely
- a font stored remotely
- a hyperlink target stored remotely

I have also made some additional improvements.

If the decision is to define a separate plane for container and remote resources, the relevant text must be completed (but it stands as is now in case we do not define a new plane).

Careful, this is a PR on #2214, not on the main branch!

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2225.html" title="Last updated on Apr 5, 2022, 1:30 PM UTC (736df09)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2225/1144879...736df09.html" title="Last updated on Apr 5, 2022, 1:30 PM UTC (736df09)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -10860,7 +10860,7 @@ html.my-document-playing * {
 			<h2>Detailed Examples</h2>
 
 			<section id="publication-resources-example">
-				<h3>Publication Resources</h3>
+				<h3>Resources</h3>
 
 				<p>Consider the following extracts of a <a>Package Document</a> and an <a>XHTML Content
 					Document</a>:</p>
@@ -10868,7 +10868,12 @@ html.my-document-playing * {
 				<pre>&lt;package …>
     &lt;metadata …>
         …
-        &lt;link rel="record" href="meta/data.xml" media-type="application/marc"/>
+        &lt;link rel="record" 
+            href="meta/data.xml" 
+            media-type="application/marc"/>
+        &lt;link rel="record" 
+            href="https://www.example.org/meta/data2.xml" 
+            media-type="application/marc"/>
         …
     &lt;/metadata>
     &lt;manifest>
@@ -10885,6 +10890,9 @@ html.my-document-playing * {
             media-type="text/css"/>
         &lt;item id="font_otf"
             href="fonts/font-file.otf"
+            media-type="font/otf"/>
+        &lt;item id="font_otf_remote"
+            href="https://www.example.org/fonts/font-file2.otf"
             media-type="font/otf"/>
         &lt;item id="font_cff"
             href="fonts/font-file.cff"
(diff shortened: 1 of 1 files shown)


murata2makoto:
Remote audio or video resources are similar to remote fonts, but they deserve to be mentioned, in my opinion.

iherman:
@murata2makoto I am happy to change the remote font example to a remote video used by a video element in the XHTML file. That may indeed be more "realistic"

mattgarrish:
> I am happy to change the remote font example to a remote video used by a video element in the XHTML file. That may indeed be more "realistic"

This is getting pretty duplicative with the examples in [5.6.2.2](https://www.w3.org/TR/epub-33/#sec-item-elem-examples).

Specifically, there's already a remote audio example there: https://www.w3.org/TR/epub-33/#example-remote-resources-that-are-publication-resources

Perhaps you can delete the duplicates from 5.6.2.2 and add a note to see this example instead.

iherman:
Hm.  Not sure that we should remove that example, @mattgarrish. The point is that the emphasis of the [example in §5.6.2.2](https://www.w3.org/TR/epub-33/#example-remote-resources-that-are-publication-resources) is the usage of the `remote-resources` property, and that is useful/necessary where it is. Ie, I do not think that example should be removed.

On the other hand, indeed, it is better to avoid too much duplication.

What I would prefer to do is to leave the example in the appendix as it is, but would add some text, which would say:

> Similar examples, showing various types of resources, can be found in §5.6.2.2.

As an aside, I wonder whether it is not worth going through those examples and change the explanations by using the "plane" terminology. But I would prefer to do that through a separate PR.

iherman:
> As an aside, I wonder whether it is not worth going through those examples and change the explanations by using the "plane" terminology. But I would prefer to do that through a separate PR.

Although I see that the terminology is up-to-date on the example descriptions, so adding the "on the spine plane" might just become a bit too convoluted...

mattgarrish:
> The point is that the emphasis of the [example in §5.6.2.2](https://www.w3.org/TR/epub-33/#example-remote-resources-that-are-publication-resources) is the usage of the `remote-resources` property

No, 46 and 47 are a pair to show one resource that is a remote resource and another that is not. 5.6.2.1 already has plenty of examples of the properties attribute. That's why I'm finding all these examples repetitive.

murata2makoto:
I did some [analysis ](https://1drv.ms/x/s!An5Z79wj5AZBkponQi9dbqyssnTmCA?e=ZmYiuI) for myself.

iherman:
> > The point is that the emphasis of the [example in §5.6.2.2](https://www.w3.org/TR/epub-33/#example-remote-resources-that-are-publication-resources) is the usage of the `remote-resources` property
> 
> No, 46 and 47 are a pair to show one resource that is a remote resource and another that is not. 5.6.2.1 already has plenty of examples of the properties attribute. That's why I'm finding all these examples repetitive.

Sigh. Is the current text for the example o.k., or should I also add a reference to 5.6.2.1?

mattgarrish:
No one's yet died from too many examples, so I'm going to merge this and look at closing this off.

PR #2235, opened 2022-04-07
Have added export classes to definitions (see issue 2207)

iherman:
Hopefully by merging this, as well as [PR on adding epub to specref](https://github.com/w3c/browser-specs/pull/581) we can use xref better.

In general, I added an export for all terms appearing in the terminology sections, plus some additional ones (like the 'planes', for example). I did find some `<dfn>`-s that I hesitated or did not export:

- For Core:
    - [§4.1.4 Deriving File Paths](https://w3c.github.io/epub-specs/epub33/core/#sec-file-names-to-path-names) has a dfn for “derive the File Path”, which is not even referenced in the document. I think it is a mistake to have that as a dfn
    - [§5.5.2 Metadata Values](https://w3c.github.io/epub-specs/epub33/core/#sec-metadata-values) the term “value” is set as a dfn. It is referenced from elsewhere in the document so, in this respect, it is fine to have it, but I do not think it is to be exported.
- For RS:
    - in [§15 epubReadingSystem Object](https://w3c.github.io/epub-specs/epub33/rs/#app-epubReadingSystem) the HTML navigator object is set as a dfn. We should definitely not export this, but I wonder whether it is correct to use a dfn in the first place. It is a reference to an existing definition defined elsewhere, after all. B.t.w., respec does not behave as if it was a dfn, probably because it is part of a hyperlink… 
    - the values [§15.3 Properties](https://w3c.github.io/epub-specs/epub33/rs/#app-ers-properties) are set as dfn-s. I think that is fine, but we should not export these (they are way too generic). The same holds for the `hasFeature` method in [§15.4.1.1](https://w3c.github.io/epub-specs/epub33/rs/#app-ers-hasFeature-desc)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2235.html" title="Last updated on Apr 7, 2022, 1:24 PM UTC (24e90e1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2235/1287600...24e90e1.html" title="Last updated on Apr 7, 2022, 1:24 PM UTC (24e90e1)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -285,7 +285,7 @@
 
 				<dl class="termlist">
 					<dt>
-						<dfn id="dfn-codec" data-lt="Codecs">Codec</dfn>
+						<dfn class="export" id="dfn-codec" data-lt="Codecs">Codec</dfn>
 					</dt>
 					<dd>
 						<p>Codec refers to content that has intrinsic binary format qualities, such as video and audio
@@ -294,7 +294,7 @@
 					</dd>
 
 					<dt>
-						<dfn id="dfn-local-resource" data-lt="Container Resources">Container Resource</dfn>
+						<dfn class="export" id="dfn-local-resource" data-lt="Container Resources">Container Resource</dfn>
 					</dt>
 					<dd>
 						<p>A <a>Publication Resource</a> that is located within the <a>EPUB Container</a>, as opposed to
@@ -304,7 +304,7 @@
 					</dd>
 
 					<dt>
-						<dfn id="dfn-container-root-url">Container Root URL</dfn>
+						<dfn class="export" id="dfn-container-root-url">Container Root URL</dfn>
 					</dt>
 					<dd>
 						<p>The <a data-cite="url#concept-url">URL</a> [[URL]] of the <a>Root Directory</a> representing
@@ -313,7 +313,7 @@
 					</dd>
 
 					<dt>
-						<dfn id="dfn-content-url">Content URL</dfn>
+						<dfn class="export" id="dfn-content-url">Content URL</dfn>
 					</dt>
 					<dd>
 						<p> The <a data-cite="url#concept-url">URL</a> of a file or directory in the <a>OCF Abstract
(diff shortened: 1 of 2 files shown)


mattgarrish:
> * has a dfn for “derive the File Path”, which is not even referenced in the document. I think it is a mistake to have that as a dfn

Ya, we can probably change that to strong tags for now to highlight the name. If we link to it in the future, it's easy enough to make a dfn at that point.

mattgarrish:
> * but I do not think it is to be exported

No, that's unique to the package document. Because xml elements don't technically have "values", we added that definition rather than come up with alternative wording. It also avoids repeating the whitespace rules everywhere.

mattgarrish:
> * the HTML navigator object is set as a dfn. We should definitely not export this, but I wonder whether it is correct to use a dfn in the first place.

Isn't this required to make the webidl definition valid? If you can remove it without breaking anything, though, then go for it.

mattgarrish:
> * I think that is fine, but we should not export these (they are way too generic).

No, we don't crossreference the properties from anywhere else, so no need to export. Again, I think this was done primarily for webidl validity.

mattgarrish:
Not related to this issue, but should we strip that at-risk feature box from the start of the RSO section?

It doesn't appear to serve any purpose given that we'll be marking features under-implemented before long.

mattgarrish (approved):

iherman:
> > * the HTML navigator object is set as a dfn. We should definitely not export this, but I wonder whether it is correct to use a dfn in the first place.
> 
> Isn't this required to make the webidl definition valid? If you can remove it without breaking anything, though, then go for it.

Well... I found out that, again with xref, you can use `{{Navigator}}` which refers to WebIDL terms (and it links the right way). That seems to work...

iherman:
> Not related to this issue, but should we strip that at-risk feature box from the start of the RSO section?
> 
> It doesn't appear to serve any purpose given that we'll be marking features under-implemented before long.

Indeed. I thought of removing it, but I did not want to do it in this PR. But then again, why not? So I did it now.

PR #2237, opened 2022-04-07
Move details of media overlays to techniques

### LINKED ISSUE #2221: 3.4.2.3.1 Completeness in EPUB A11Y 1.1
https://www.w3.org/TR/epub-a11y-11/#sec-mo-complete

Meeting this Objective
EPUB Creators MUST provide synchronized audio playback via Media Overlays for all textual content.

This statement could be interpreted as saying to provide synchronized audio using the Media Overlays mechanism, and not to use another mechanism.
Since the main purpose of this clause is to say that the complete audio narration should also be provided along with the full text, we believe that "for all textual content" is a weak explanation and should be "for complete textual content".

mattgarrish:
This is my best attempt to make the accessibility objectives less tied to media overlays.

In place of "Media Overlays", I've used "Synchronized Text-Audio Playback" for the main specification.

I've also added a technique to match each objective. These should work as starters, but they need thorough review.

The examples I added for skippability and escapability were largely copied from the ones already in the media overlays section of the core spec. I just tweaked them to exemplify different semantics. I didn't try to make examples for the reading order and complete text techniques, as these would be complex and time-consuming to build. I don't think they're critical to have.

Let me know if you think I have anything wrong, or have thoughts on how to improve what's there.

Fixes #2221 

- [Accessibility preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2221/epub33/a11y/index.html)
- [Accessibility diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2221/epub33/a11y/index.html)
- [Accessibility Techniques preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2221/epub33/a11y-tech/index.html)
- [Accessibility Techniques diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2221/epub33/a11y-tech/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2237.html" title="Last updated on Apr 12, 2022, 7:04 PM UTC (4433c5e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2237/809b2e8...4433c5e.html" title="Last updated on Apr 12, 2022, 7:04 PM UTC (4433c5e)">Diff</a>

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -1894,8 +1894,8 @@
 							[[EPUB-SSV-11]] to identify each.</li>
 						<li>Lists &#8212; using the <a data-cite="epub-ssv-11#list"><code>list</code> semantic</a>
 							[[EPUB-SSV-11]] to identify each.</li>
-						<li>Sidebars &#8212; use the <a data-cite="epub-ssv-11#sidebar"><code>sidebar</code>
-								semantic</a> [[EPUB-SSV-11]] to identify each.</li>
+						<li>Sidebars &#8212; use the <a data-cite="epub-ssv-11#aside"><code>aside</code> semantic</a>
+							[[EPUB-SSV-11]] to identify each.</li>
 						<li>Tables &#8212; use the <a data-cite="epub-ssv-11#table"><code>table</code> semantic</a>
 							[[EPUB-SSV-11]] to identify each.</li>
 					</ul>
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -1881,12 +1881,11 @@
 						before being able to move on. These are called escapable elements, because the user needs to be
 						able to escape from them whenever they choose to simplify the reading experience.</p>
 
-					<p>EPUB 3's <a data-cite="epub-33#sec-media-overlays">Media Overlays feature</a> [[EPUB-33]] does
-						not allow Reading Systems to determine if playback sequences are escapable unless EPUB Creators
-						add additional semantics to the markup, however. EPUB Creators must use the <a
-							data-cite="epub-33#sec-epub-type-attribute"><code>epub:type</code> attribute</a> [[EPUB-33]]
-						to add semantics to <a data-cite="epub-33#sec-smil-seq-elem"><code>seq</code></a> and <a
-							data-cite="epub-33#sec-smil-seq-elem"><code>par</code></a> elements [[EPUB-33]] to allow
+					<p>EPUB 3's <a data-cite="epub-33#sec-media-overlays">Media Overlays feature</a> [[EPUB-33]] only
+						supports escapability if EPUB Creators add structural semantics to the markup. EPUB Creators
+						must use the <a data-cite="epub-33#sec-epub-type-attribute"><code>epub:type</code> attribute</a>
+						[[EPUB-33]] to add semantics to <a data-cite="epub-33#sec-smil-seq-elem"><code>seq</code></a>
+						and <a data-cite="epub-33#sec-smil-seq-elem"><code>par</code></a> elements [[EPUB-33]] to allow
 						Reading Systems to provide users the option to escape their playback sequences.</p>
 
 					<p>The recommended structures to identify for escapability are:</p>
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -1796,7 +1796,7 @@
 									><code>endnote</code> semantic</a> [[EPUB-SSV-11]] for individual endnotes.</li>
 						<li>Footnotes &#8212; use the <a data-cite="epub-ssv-11#footnotes"><code>footnotes</code>
 								semantic</a> for a group of footnotes or the <a data-cite="epub-ssv-11#footnote"
-									><code>footnote</code> semantic</a> [[EPUB-SSV-11]]for individual footnotes.</li>
+									><code>footnote</code> semantic</a> [[EPUB-SSV-11]] for individual footnotes.</li>
 						<li>Page breaks &#8212; use the <a data-cite="epub-ssv-11#pagebreak"><code>pagebreak</code>
 								semantic</a> [[EPUB-SSV-11]] for each page break.</li>
 					</ul>
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -1791,12 +1791,12 @@
 					<p>The recommended structures to identify for skippability are:</p>
 
 					<ul>
-						<li>Endnotes &#8212; use the <a data-cite="epub-ssv-11#endnotes"><code>endnotes</code>
-								semantic</a> for a group of notes or the <a data-cite="epub-ssv-11#endnote"
-									><code>endnote</code> semantic</a> [[EPUB-SSV-11]] for individual endnotes.</li>
-						<li>Footnotes &#8212; use the <a data-cite="epub-ssv-11#footnotes"><code>footnotes</code>
-								semantic</a> for a group of footnotes or the <a data-cite="epub-ssv-11#footnote"
-									><code>footnote</code> semantic</a> [[EPUB-SSV-11]]for individual footnotes.</li>
+						<li>Endnotes &#8212; use the <a data-cite="epub-ssv-11#endnotes"><code>endnotes</code></a> and
+								<a data-cite="epub-ssv-11#endnote"><code>endnote</code></a> semantics [[EPUB-SSV-11]]
+							for groups of notes and individual endnotes, respectively.</li>
+						<li>Footnotes &#8212; use the <a data-cite="epub-ssv-11#footnotes"><code>footnotes</code></a>
+							and <a data-cite="epub-ssv-11#footnote"><code>footnote</code></a> semantics [[EPUB-SSV-11]]
+							for groups of footnotes and individual footnotes, respectively.</li>
 						<li>Page breaks &#8212; use the <a data-cite="epub-ssv-11#pagebreak"><code>pagebreak</code>
 								semantic</a> [[EPUB-SSV-11]] for each page break.</li>
 					</ul>
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -84,10 +84,20 @@
 						"date": "04 September 2010",
 						"publisher": "IDPF"
 					},
+					"SVG": {
+						"title": "SVG",
+						"href": "https://www.w3.org/TR/SVG/",
+						"publisher": "W3C"
+					},
 					"SWF": {
 						"title": "SWF File Format Specification Version 19",
 						"href": "https://www.adobe.com/content/dam/acom/en/devnet/pdf/swf-file-format-spec.pdf",
 						"date": "2012"
+					},
+					"WAI-ARIA": {
+						"title": "Accessible Rich Internet Applications (WAI-ARIA)",
+						"href": "https://www.w3.org/TR/wai-aria/",
+						"publisher": "W3C"
 					},
 					"WCAG2": {
 						"title": "Web Content Accessibility Guidelines (WCAG) 2",
@@ -117,7 +127,7 @@
 					[[EPUB-A11Y-11]] discovery and accessibility requirements for EPUB Publications.</p>
 
 				<p>This document does not cover techniques and best practices already addressed in [[WCAG2]] and
-					[[WAI-ARIA-1.1]] for which no substantive differences in application exist.</p>
+					[[WAI-ARIA]] for which no substantive differences in application exist.</p>
 			</section>
 
 			<section id="sec-terminology">
@@ -176,9 +186,9 @@
 			<p>If EPUB Creators encounter issues that are not covered in these or related techniques, they are
 				encouraged to report the issue to the appropriate community for guidance on how to meet accessibility
 				standards. The <a href="https://www.w3.org/WAI/IG/">W3C Web Accessibility Interest Group</a> has a
-				public mailing list where issues meeting [[WCAG2]] and [[WAI-ARIA-1.1]] requirements can be raised. The
-					<a href="https://github.com/w3c/publ-a11y/issues">W3C Publishing Community Group issue
-				tracker</a> can be used to ask for support implementing EPUB-specific requirements and the <a
+				public mailing list where issues meeting [[WCAG2]] and [[WAI-ARIA]] requirements can be raised. The <a
+					href="https://github.com/w3c/publ-a11y/issues">W3C Publishing Community Group issue tracker</a> can
+				be used to ask for support implementing EPUB-specific requirements and the <a
 					href="https://github.com/w3c/epub-specs/issues">EPUB 3 Working Group's issue tracker</a> to report
 				issues with this document.</p>
 		</section>
@@ -877,7 +887,7 @@
 							<a href="https://idpf.github.io/epub-guides/epub-aria-authoring/">EPUB Type to ARIA Role
 							Authoring Guide</a> guide details notable authoring differences between the two attributes.
 						It also includes a mapping table of semantics in the EPUB Structural Semantics Vocabulary to
-						equivalent ARIA roles in [[DPUB-ARIA-1.0]] and [[WAI-ARIA-1.1]].</p>
+						equivalent ARIA roles in [[DPUB-ARIA-1.0]] and [[WAI-ARIA]].</p>
 				</section>
 
 				<section id="sem-002">
@@ -896,10 +906,10 @@
 						chapter, putting each into a separate document.</p>
 
 					<p>Although visually these restructuring decisions can be hidden from readers, they impact the
-						functionality of Assistive Technologies. In the case of [[WAI-ARIA-1.1]] roles, the result is
-						that only the subset present in the currently-loaded EPUB Content Document are exposed to users.
-						An Assistive Technology cannot provide a list of landmarks for the whole publication, as it
-						cannot see outside the current document.</p>
+						functionality of Assistive Technologies. In the case of [[WAI-ARIA]] roles, the result is that
+						only the subset present in the currently-loaded EPUB Content Document are exposed to users. An
+						Assistive Technology cannot provide a list of landmarks for the whole publication, as it cannot
+						see outside the current document.</p>
 
 					<p>To counteract this destructuring effect, EPUB Creators sometimes think to re-add or re-identify
 						structures in the belief that having this information in every document will be helpful to users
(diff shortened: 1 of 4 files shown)


epub33/a11y-tech/index.html
```diff
@@ -1680,6 +1690,336 @@ <h4>Identify the pagination source</h4>
...
+					<p>EPUB Creators do not need to identify every structure to meet this requirement. The recommended
+						structures to identify for skippability are:</p>
+
+					<ul>
+						<li>Endnotes &#8212; use the <a data-cite="epub-ssv-11#endnotes"><code>endnotes</code>
+								semantic</a> for a group of notes or the <a data-cite="epub-ssv-11#endnote"
+									><code>endnote</code> semantic</a> [[EPUB-SSV-11]] for individual endnotes.</li>
+						<li>Footnotes &#8212; use the <a data-cite="epub-ssv-11#footnotes"><code>footnotes</code>
+								semantic</a> for a group of footnotes or the <a data-cite="epub-ssv-11#footnote"
+									><code>footnote</code> semantic</a> [[EPUB-SSV-11]]for individual footnotes.</li>
```
iherman:
```suggestion
									><code>footnote</code> semantic</a> [[EPUB-SSV-11]] for individual footnotes.</li>
```

epub33/a11y-tech/index.html
```diff
@@ -1680,6 +1690,336 @@ <h4>Identify the pagination source</h4>
...
+							data-cite="epub-33#sec-smil-par-elem"><code>par</code></a> elements [[EPUB-33]], thereby
+						allowing Reading Systems to provide users the option to skip their playback sequences.</p>
+
+					<p>EPUB Creators do not need to identify every structure to meet this requirement. The recommended
+						structures to identify for skippability are:</p>
+
+					<ul>
+						<li>Endnotes &#8212; use the <a data-cite="epub-ssv-11#endnotes"><code>endnotes</code>
+								semantic</a> for a group of notes or the <a data-cite="epub-ssv-11#endnote"
+									><code>endnote</code> semantic</a> [[EPUB-SSV-11]] for individual endnotes.</li>
```
iherman: At first reading I did not realize that both the `endnotes` and the `endnote` values are defined in the same document (ie, SSV). Ie, that the reference to SSV is also valid for `endnotes`. Repeating the reference may be an overkill; maybe making it more explicit saying something like "both defined in [[...]]" might help.

This a repeated pattern here and in the part on skippability.

mattgarrish: I've rephrased to: "use the endnotes and endnote semantics [EPUB-SSV-11] for groups of notes and individual notes, respectively."

Same for footnotes.

epub33/a11y/index.html
```diff
@@ -915,83 +925,96 @@ <h6>Skippability</h6>
...
 								</dd>
 
-								<dt id="sec-mo-skippability-conf">Meeting this Objective</dt>
+								<dt id="sec-sync-skippability-conf">Meeting this Objective</dt>
 								<dd>
-									<p>EPUB Creators SHOULD identify all <a
-											href="https://www.w3.org/TR/epub/#sec-skippability">skippable structures</a>
-										[[EPUB-3]] in Media Overlay Documents.</p>
+									<p>EPUB Creators SHOULD identify all skippable structures if the synchronized
+										text-audio playback format does not handle escaping content by default.</p>
```
marisademeglio: Should it say "does not handle skipping content by default" rather than "escaping", since this is the section on skippability? 

Even so I don't know what it would mean to have a format handle skipping content by default.

epub33/a11y/index.html
```diff
@@ -764,62 +764,63 @@ <h6>Page Breaks</h6>
...
-							sequences, the effect is to deprive users of rich navigation of the content.</p>
+						<p>The provision of synchronized text-audio playback helps address various user needs. It not
+							only enables a seamless visual and auditory reading experience from beginning to end of an
+							EPUB Publication, but is useful to users who only require audio playback (e.g., who cannot
+							see the text or are prevented from reading visually due to motion-sickness) or who only
+							benefit from reading with text highlighting (e.g., readers with dyslexia).</p>
+
+						<p>The most basic forms of synchronized text-audio playback provide only minimal instructions to
+							Reading Systems, however. They indicate the text to highlight and the audio clip that
+							corresponds to the text. The result is that users only have basic start and stop options
```
marisademeglio: Even with basic clip/text alignment, users would also have phrase navigation available to them. I would replace this paragraph with:

"Unlike purely linear listening experiences, EPUB with synchronized text-audio playback preserves the user's ability to navigate around the publication, such as via the TOC, and also introduces audio-centric reading features like phrase navigation, and ways to control which parts of the content are read aloud."

epub33/a11y/index.html
```diff
@@ -764,62 +764,63 @@ <h6>Page Breaks</h6>
...
+							available.</p>
+
+						<p>EPUB Creators need to add additional structure and semantics to allow Reading Systems to
```
marisademeglio: Slight edit to flow better with the previous suggestion: 

"In order to offer users greater control over content presentation, EPUB Creators need to add structure and semantics so that the Reading System has the necessary context to enable this type of user experience. With greater context, a Reading System can provide..."

epub33/a11y/index.html
```diff
@@ -764,62 +764,63 @@ <h6>Page Breaks</h6>
...
+							present more usable experiences. With greater context, a Reading System can provide the
+							ability to skip past secondary content that interferes with the primary narrative, escape
+							users from deeply nested structures like tables, and allow them to navigate through the
+							sections of the publication without having to go to the table of contents.</p>
```
marisademeglio: > allow them to navigate through the sections of the publication without having to go to the table of contents

Is there a text/audio sync feature that does this?

epub33/a11y/index.html
```diff
@@ -915,83 +925,96 @@ <h6>Skippability</h6>
...
 								</dd>
 
-								<dt id="sec-mo-escapability-conf">Meeting this Objective</dt>
+								<dt id="sec-sync-escapability-conf">Meeting this Objective</dt>
 								<dd>
-									<p>EPUB Creators SHOULD identify all <a
-											href="https://www.w3.org/TR/epub/#sec-escapability">escapable structures</a>
-										[[EPUB-3]] in the Media Overlay Documents.</p>
+									<p>EPUB Creators SHOULD identify all escapable structures if the synchronized
+										text-audio playback format does not handle escaping content by default.</p>
```
marisademeglio: Same comment as above: what does it mean for a format to handle escaping content by default?

epub33/a11y/index.html
```diff
@@ -915,83 +925,96 @@ <h6>Skippability</h6>
 										colours or placement so readers can filter this information visually out while
```
marisademeglio: > such as sidebars and footnotes

Suggest change to "print page numbers and footnotes".  Trying to put myself in the reader's position, I would not want to skip a sidebar but I would probably want to escape it.

epub33/a11y/index.html
```diff
@@ -915,83 +925,96 @@ <h6>Skippability</h6>
...
 										reading).</p>
 									<p>Readers who prefer auditory playback, however, cannot skip this information with
-										the same ease by default. Media overlays playback will typically result in
-										Reading Systems rendering secondary content where it occurs.</p>
-									<p>When EPUB Creators add structural semantics to Media Overlay Documents, however,
-										Reading Systems can create reading experiences that allow users to decide which
-										secondary content to skip by default during playback.</p>
+										the same ease by default. Synchronized text-audio playback will typically result
```
marisademeglio: Suggested change:

"Readers who prefer auditory playback, however, cannot skip this information with the same ease in a linear audio-based reading experience. And, without structural semantics, synchronized text-audio playback cannot offer skipping content either."

epub33/a11y/index.html
```diff
@@ -915,83 +925,96 @@ <h6>Skippability</h6>
...
+								<dt id="sec-sync-escapability-obj">Objective</dt>
 								<dd>
 									<p>Enable users to automatically escape from structured content.</p>
 								</dd>
 
-								<dt id="sec-mo-escapability-understand">Understanding this Objective</dt>
+								<dt id="sec-sync-escapability-understand">Understanding this Objective</dt>
 								<dd>
 									<p>When reading visually, users can quickly move through, and escape from, highly
 										structured content such as tables, lists, and figures. Table layouts that use
```
marisademeglio: Suggest change to "sidebars, lists, and figures." And also remove the parts about table layout. I know we've always used tables as an escapability example but I haven't seen any good implementations and I think it's because sync text/audio doesn't really help you here. For example, you can't have a row-based presentation and switch to a column-based presentation. And even to enable basic table reading features like escaping along the axis of cell=>row=>table or scanning the column headers requires really thorough parallel content doc <-> sync text format semantic markup plus a clever reading system.

epub33/a11y/index.html
```diff
@@ -915,83 +925,96 @@ <h6>Skippability</h6>
...
 										rows and columns, for example, allow users to quickly move along either axis to
 										find the information they want, and to easily return to the primary narrative
 										when they finish. Similarly, lists of items allow users to skim their content
 										and escape from them once they locate the desired information.</p>
-									<p>The same ease of escaping from content is not available in Media Overlay
-										Documents by default. Users cannot escape from table cells, rows, or even the
-										table itself, unless EPUB Creators encode the structural semantics of those
-										elements in the document.</p>
+									<p>The same ease of escaping from content is not always available in synchronized
```
marisademeglio: Replace with

"The same ease of escaping from content is only possible if EPUB Creators encode the structural semantics into the synchronized text/audio format."

Because at first glance it looks like we're saying that the features are not available in sync text/audio.

epub33/a11y-tech/index.html
```diff
@@ -1680,6 +1690,338 @@ <h4>Identify the pagination source</h4>
...
+
+					<p>Some content elements are containers for expressing complex information. A table, for example,
+						has data arranged in rows and cells, where each cell may contain several paragraphs of
+						information. Lists similarly may contain many items, and each item can contain complex
+						information. While users may be interested in some of the information in these structures, they
+						also often want to escape from them to keep reading, not have to listen to the entire content
+						before being able to move on. These are called escapable elements, because the user needs to be
+						able to escape from them whenever they choose to simplify the reading experience.</p>
+
+					<p>EPUB 3's <a data-cite="epub-33#sec-media-overlays">Media Overlays feature</a> [[EPUB-33]] does
```
marisademeglio: I think it's more direct if the first sentence says:

"EPUB 3's Media Overlays feature only supports escapability if EPUB Creators add structural semantics to the markup."

Because at first glance it looks like we're saying that the features are not available in sync text/audio.

epub33/a11y-tech/index.html
```diff
@@ -1680,6 +1690,338 @@ <h4>Identify the pagination source</h4>
...
+
+					<p>The recommended structures to identify for escapability are:</p>
+
+					<ul>
+						<li>Figures &#8212; use the <a data-cite="epub-ssv-11#figure"><code>figure</code></a> semantic
+							[[EPUB-SSV-11]] to identify each figure.</li>
+						<li>Lists &#8212; use the <a data-cite="epub-ssv-11#list"><code>list</code></a> and <a
+								data-cite="epub-ssv-11#list-item"><code>list-item</code></a> semantics [[EPUB-SSV-11]]
+							to allow escaping from both items and lists.</li>
+						<li>Tables &#8212; use the <a data-cite="epub-ssv-11#table"><code>table</code></a>, <a
```
marisademeglio: As in the A11y doc, I am a little uneasy about table escapability and in general, nested escapability. I don't want to say that it's not possible, I just don't think nested escapability is implemented anywhere (though I haven't looked into it, it's just a hunch). We don't provide good implementor guidance either, just a basic "this is how you escape something" rule in the [reading systems spec](https://w3c.github.io/epub-specs/epub33/rs/#sec-escapability). Content creators may still want to aim high with their markup, so that's a reason to keep this section as-is; but realistically IMO not worth the trouble to markup MO that extensively.

epub33/a11y/index.html
mattgarrish: Oops, ya, the text was similar so I did some copying... 😄 

It made me wonder if we really need two different objectives, or if adding semantics is one requirement with two useful outcomes. But I don't want to do too much rethinking at this stage.

Saying "by default" is probably not the best phrasing, but I was thinking of the discussion we had earlier about using the content document to provide skippability and escapability. If a technology were to do that, why would we need to bother with this objective? Making sure tables are marked up as tables, lists as lists, etc. is already required by WCAG 1.3.1 Info and Relationships. We'd be duplicating it.

I'm not sure how to say "if the host grammar already has markup that identifies escapable structures you don't need any additional semantics" in a succinct way. We could ignore the problem for now, though, and revisit the objectve whenever an alternative like that becomes a reality.

(Edit: fixed the description I wrote in quotes in the above para as my brain wasn't awake the first time...)

marisademeglio: > or if adding semantics is one requirement with two useful outcomes

To me, this is a slightly better way to present it but the current way is probably just as good for people who will read this document.

> We could ignore the problem for now, though, and revisit the objective whenever an alternative like that becomes a reality

Yeah let's ignore it for now :) I think as things are currently, implementations expect duplicate semantics in the sync text/audio format rather than just going straight to the content doc source. 

I am experimenting a bit with the latter in [a web-based book project](https://daisy.github.io/accessible-books-in-browsers/demos/action-for-heroes/005.html), will report back for EPUB vNextNext 🙃

mattgarrish: The next commit will strip these sentences down to: "EPUB Creators SHOULD identify all (skippable/escapable) structures."

epub33/a11y/index.html
mattgarrish: I'm not sure how it would work. Sounds mixed up with ARIA landmark roles.

We don't even have a way of identifying headings in overlays, so it'd sort of be like escapability at a nebulous "section" level. I'm going to strip that bit in the next commit.

epub33/a11y/index.html
mattgarrish: I'll get rid of that pesky "colours" while I'm at it. (Try as I might to misspell words like Americans... 😉 )

epub33/a11y/index.html
mattgarrish: > And also remove the parts about table layout.

Is this only for the description here or are you suggesting we should not recommend identifying tables in the techniques, too?

I don't know how well escaping works from tables, but if we don't recommend identifying them do we expect users will drop out of MO playback mode to traverse/escape tables?

If so, should we say that in a note in the techniques to explain why it's not there?

Should we also remove the recommendation from the media overlays section on escapability, too?

Does this work for everyone? @avneeshsingh @GeorgeKerscher @clapierre @gregoriopellegrino

epub33/a11y-tech/index.html
mattgarrish: Oh, I guess this partly answers my question above. It would still help to see if we can get consensus on what to do about table markup. I've removed tables from the understanding description for now.

epub33/a11y/index.html
marisademeglio: I think that escaping from a table is fine, as in just going from being in a table to not being in a table. But it's the nested escaping (e.g. Cell => Row => Table) that I'm not sure about.

Like say you're reading a table. You're always in a cell. Say the MO are marked up row-wise. So you're in a cell, in a row. And then you "escape" from the cell out to the next cell. But how do you say you want to escape out to the next row? Or escape out of the table entirely? Of course it's possible if the RS is clever about it and provides a UI for this, but we haven't given any guidance and as far as I know, escapability isn't implemented by anyone, let alone this sort of x-ray-vision level of nested escapability. 

One might say what's the harm of including extra semantics in MO? Well, they could get in the way of a basic escapability implementation, if the RS is just saying "wherever there's a semantic term from this list, I'll let the user escape", in the above example, the user would probably never be able to escape out of the table itself. 

So, I would propose to either dig in and establish an algorithm for complex escapability, or discourage its use.

GeorgeKerscher:
It seems that escaping from the table entirelly is appropriate. Creating a complex user interface and placing authoring requirements is too much.

epub33/a11y/index.html
mattgarrish: Okay, so recommend to identify `table` but not `table-row` and `table cell`?

Does the same apply to a list? Should we only recommend escaping from lists and not list items?

mattgarrish: We can still "encourage" the substructures be identified without recommending, too.

mattgarrish: I've added the following note in the latest commit:

> Sometimes escapable structures may contain escapable structures. For example, tables are composed of many rows and cells that users may want to separately escape from. While identifying the nested escapable structures is encouraged, it is not required at this time as the process of escaping from nested structures is not well defined.

marisademeglio: > Okay, so recommend to identify `table` but not `table-row` and `table cell`?
> 
To me, it comes down to author intent. As in, what is the structure they would like to offer the user the option to escape from?

marisademeglio: > While identifying the nested escapable structures is encouraged, it is not required at this time as the process of escaping from nested structures is not well defined.

I'm afraid that identifying all the nested structures would break the most useful behavior provided by a simple implementation. Eg am RS that uses the rule "if semantic exists, offer escape" and therefore let's you escape cells but isn't clever enough to expose any wrapping escapable containers.

mattgarrish: But if it's purely author intent, then doesn't that make an accessibility requirement/recommendation virtually impossible? You can't verify what someone thinks makes for the best escaping.

And if we try to make people choose only one level when there is nesting (e.g., only identify cells or the table),  it also feels like we're pushing people to only accommodate the least useful escaping possible.

mattgarrish: What if we remove the note and just start with a recommendation to identify the larger structures?

These are only optional requirements, after all, so if we find the guidance is impractical we can always tweak the suggestions later before they become recommendations to follow.

gregoriopellegrino (approved):

clapierre (approved):
Wondering if this example should be coded with role="doc-example"?

> `<aside class="example" title="Identifying a skippable footnote">`

mattgarrish:
> Wondering if this example should be coded with role="doc-example"?

If you're asking about examples in general, that's problematic as the superclass is changing to figure in DPUB 1.1. I don't know yet how that impacts its use on asides. I don't want to muck around with the respec template until I'm sure.

clapierre:
Sorry Matt, the code snippit was stripped out it was an example we were providing I believe.  See above now.

mattgarrish:
> Sorry Matt, the code snippit was stripped out it was an example we were providing I believe.

Yes, this is what I was referring to. The use of `aside` with `class="example"` is part of the template that generates the final specification (triggers reformatting to get the syntax highlighting, example title, etc.).

epub33/a11y/index.html
mattgarrish: The last commit makes a number of changes:

I've taken the note that I put in earlier and moved it to the core specification. I've also modified it to make clear that reading system support is iffy:

> Sometimes escapable structures may contain escapable structures. For example, tables are composed of many rows and cells that users may want to separately escape from. Reading System support for escaping from such structures is complex and not well supported at this time. EPUB Creators should avoid identifying nested escapable structures until better support is available.

I've replaced the note in the techniques document with a shorter explanation that references across to the core spec section.

I've removed table-cell, table-row and list-item from the core spec's list of recommended escapable structures (and added links to the SSV for all the terms we were citing). Alternatively, I added `aside` as a recommended term for identifying sidebars.

Hopefully that'll get us over the finish line! 😄

PR #2238, opened 2022-04-08
Require links to all reproduced pages in page list

mattgarrish:
Fixes #2220 

- [Accessibility preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2220/epub33/a11y/index.html)
- [Accessibility diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2220/epub33/a11y/index.html)

PR #2246, opened 2022-04-12
Fixes to page navigation requirements

mattgarrish:
The concerns raised by @gregoriopellegrino in #2238 made me notice that our requirements on page navigation changed from a requirement to a recommendation because of this paragraph:

> An EPUB Publication SHOULD meet the page navigation objectives defined in this section whenever any of the following cases is true:

EPUB Creators should include page navigation in the listed circumstances, but they still have to meet the requirements. I believe we changed this to try and work around requiring objectives that aren't necessarily required to be met.

I believe we can better address this split by returning the above sentence back to its original form:

> An EPUB Publication SHOULD include page navigation whenever any of the following cases is true:

We can address the need to meet the objectives without making a normative statements about at what conformance level by simply adding the following paragraph to the end of the applicability section:

> When EPUB Creators include page navigation, the objectives defined in this section apply to the EPUB Publication.

This pull request implements both these changes.

To address @gregoriopellegrino's concerns with the previous pull request, I've returned the requirement to link to all page breaks to a SHOULD.

To address the concern about this leading to page lists with only a single entry (the minimum you have to specify to be valid), I've added the following new bullet:

> EPUB Creators MUST include links to all [page break markers](#sec-page-breaks-obj) in the content.

Not only does this make us consistent with the forthcoming [2.4.13 SC to provide page navigation](https://www.w3.org/TR/WCAG22/#page-break-navigation), but it only allows omissions from the page list when there are no markers in the text. That will provide flexibility for books that may not have markers but are reproduced from another source.

Let me know if this works for everyone now.

- [Accessibility preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2220-redux/epub33/a11y/index.html)
- [Accessibility diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2220-redux/epub33/a11y/index.html)

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -1776,6 +1776,11 @@
 					>working group's issue tracker</a>.</p>
 
 			<ul>
+				<li>12-Apr-2022: Restored recommendation to include links to all reproduced pages in the page list and
+					added a requirement to link to all page break markers to address concerns with the previous change.
+					Clarified that meeting the page navigation requirements is required, but adding page navigation is
+					only recommended in some circumstances. See <a href="https://github.com/w3c/epub-specs/pull/2246"
+						>pull request 2246</a>.</li>
 				<li>08-Apr-2022: Changed the recommendation to include links to all reproduced pages in the page list to
 					a requirement. See <a href="https://github.com/w3c/epub-specs/issues/2220">issue 2220</a>.</li>
 				<li>23-Mar-2022: Clarified linking for WCAG conformance versions and levels. See <a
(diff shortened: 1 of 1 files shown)


GeorgeKerscher:
For me, this works.

gregoriopellegrino:
Great for me, thanks!

gregoriopellegrino (approved):

PR #2245, opened 2022-04-12
Using xref

iherman:
Eventually, this will fix #2207

~However, do _NOT_ merge this yet. Currently, this branch includes only the core spec; put it out for review while I will also do a similar change for the RS spec and some notes.~

It can go to the usual comment/review round now, I have gone through most of the notes, too.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2245.html" title="Last updated on Apr 14, 2022, 9:29 AM UTC (7fd88b8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2245/6459c44...7fd88b8.html" title="Last updated on Apr 14, 2022, 9:29 AM UTC (7fd88b8)">Diff</a>

Issue #2241, opened 2022-04-11
Terminology linking


mattgarrish:
The first time a term from the terminology is used in a section, and only the first time, it's supposed to be linked to its definition.

After years of revisions, it's unlikely this practice is consistent anymore (if it ever was), particularly with the splitting off of the RS spec.

Adding a review to the post-CR cleanup chore list, in case we don't have enough things to do... 😄

iherman:
@mattgarrish #2247 was merged; shouldn't this issue be closed?

mattgarrish:
I wish, but no...

The merge only covered the accessibility specification. Even on that relatively tiny spec, it was a headache to do. You lose track of what is linked and not linked in each section when you see the terms over and over.

PR #2247, opened 2022-04-13
Add terminology linking to a11y docs

mattgarrish:
This PR makes similar changes to #2245 to use respec's xref feature to links to the terminology and html elements.

In adding the linking to the terminology, I removed the list of terms from the terminology sections. We added these in the ISO process as a workaround to having to repeat definitions (and not having the ability to link out), but there's no easy way to verify if the lists are complete plus it's a long-term maintenance headache.

I've instead gone through and linked the first instance of each epub-defined term, like we do for the other specifications. (So this also addresses issue #2241 for these two documents.)

Other than that, I've made some very minor editorial corrections (fixing term names, fixing capitalization, etc.).

- [Accessibility preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/add-linking/epub33/a11y/index.html)
- [Accessibility diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/add-linking/epub33/a11y/index.html)
- [Accessibility Techniques preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/add-linking/epub33/a11y-tech/index.html)
- [Accessibility Techniques diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/add-linking/epub33/a11y-tech/index.html)

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -49,6 +49,7 @@
 					branch: "main"
 				},
 				xref: ["epub-33"],
+				profile: "web-platform",
 				localBiblio: {
 					"A11Y-DISCOV-VOCAB": {
 						"title": "Schema.org Accessibility Properties for Discoverability Vocabulary",
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -377,8 +377,8 @@
 					</li>
 					<li>
 						<p>motion simulation — if a resource simulates motion, it can cause a user to become nauseated
-							(e.g., a video game drawn on the [[HTML]] <a data-cite="html#the-canvas-element"
-									><code>canvas</code> element</a> or parallax scrolling with CSS).</p>
+							(e.g., a video game drawn on the [[HTML]] <a data-lt="canvas"
+								><code>canvas</code> element</a> or parallax scrolling with CSS).</p>
 					</li>
 					<li>
 						<p>sound — certain sound patterns, such as ringing and buzzing, can cause seizures, while loud
@@ -863,7 +863,7 @@
 					<p>Since each attribute offers different advantages, it is not necessary that they be used together.
 						Due to the lack of restrictions on where <a>EPUB Creators</a> can use the <code>type</code>
 						attribute, pairing the attributes may cause accessibility issues (e.g., putting roles on the
-						[[HTML]] <code>body</code> element).</p>
+						[[HTML]] <a data-lt="body"><code>body</code> element</a>).</p>
 
 					<p>In particular, the use of the <code>type</code> attribute is not a means of satisfying
 						requirements for ARIA roles in WCAG.</p>
@@ -898,11 +898,11 @@
 
 					<p>To counteract this destructuring effect, EPUB Creators sometimes think to re-add or re-identify
 						structures in the belief that having this information in every document will be helpful to users
-						(e.g., adding an extra [[HTML]] <code>section</code> element around a chapter to indicate it
-						belongs to a part, or putting the part semantic on the <code>body</code> tag). All this practice
-						does, however, is add repetition that is not only disruptive when reading but can make the
-						structure of the publication harder to follow. EPUB Creators are therefore advised not to
-						attempt to rebuild structures in these ways.</p>
+						(e.g., adding an extra [[HTML]] <a data-lt="section"><code>section</code> element</a> around a
+						chapter to indicate it belongs to a part, or putting the part semantic on the
+						<code>body</code> tag). All this practice does, however, is add repetition that is not only
+						disruptive when reading but can make the structure of the publication harder to follow. EPUB
+						Creators are therefore advised not to attempt to rebuild structures in these ways.</p>
 
 					<p>For example, consider a book that has five parts and each part contains five chapters.
 						Structurally, each chapter belongs to its part (i.e., is grouped with it), as in the following
@@ -1148,8 +1148,9 @@
 						to end, even though the content is often split across numerous <a>EPUB Content Documents</a>. As
 						a result, their natural expectation is that the headings reflect their position in the overall
 						hierarchy of the publication, despite the publication not actually being a single document
-						(e.g., if a part heading is expressed in an [[HTML]] <code>h1</code> element, each chapter that
-						belongs to the part will have an <code>h2</code> heading).</p>
+						(e.g., if a part heading is expressed in an [[HTML]] <a data-lt="h1"
+						><code>h1</code> element</a>, each chapter that belongs to the part will have an <a data-lt="h2"
+								><code>h2</code></a> heading).</p>
 
 					<p>Technique <a href="https://www.w3.org/WAI/WCAG21/Techniques/general/G141">G141: Organizing a page
 							using headings</a> instructs <a>EPUB Creators</a> on correctly using numbered headings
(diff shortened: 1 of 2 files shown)


epub33/a11y-tech/index.html
```diff
@@ -48,6 +48,7 @@
 					repoURL: "https://github.com/w3c/epub-specs",
 					branch: "main"
 				},
+				xref: ["epub-33"],
```
iherman, line 51: Are we sure that is enough? In the other specs I added explicitly the "web-platform" profile in case we want to use html, fetch, etc, terms as well via xref. This may be unnecessary for the a11y spec right now, but may be in future...

(Or is the web-platform a default thing for respec? In which case we may want to simplify #2245)

epub33/a11y/index.html
```diff
@@ -50,6 +50,7 @@
 					repoURL: "https://github.com/w3c/epub-specs",
 					branch: "main"
 				},
+				xref: ["epub-33","html"],
```
iherman: See my remark above. If the web platform is default, then we may not need it at all; otherwise using the web platform profile may be more future proof.

epub33/a11y/index.html
```diff
@@ -239,37 +233,17 @@ <h3>Application to Older Versions</h3>
 			<section id="sec-terminology">
 				<h3>Terminology</h3>
 
-				<p>This specification uses the following terminology defined in EPUB 3 [[EPUB-3]]:</p>
+				<p>This specification uses terminology defined in EPUB 3 [[EPUB-3]].</p>
```
iherman: Maybe link it to the terminology section of the epub core?

epub33/a11y-tech/index.html
```diff
@@ -1406,23 +1394,23 @@ <h3>Accessible Alternatives</h3>
 						version</a> is available.</p>
 
 				<p>The [[EPUBMultipleRenditions-10]] specification defines a set of features for creating these types of
```
iherman: Shouldn't that reference be [[multi-rend]] to link to the current note?

epub33/a11y-tech/index.html
```diff
@@ -133,50 +134,28 @@ <h3>Purpose and Scope</h3>
...
-					</li>
-					<li>
-						<a href="https://www.w3.org/TR/epub/#dfn-package-document">Package Document</a>
-					</li>
-					<li>
-						<a href="https://www.w3.org/TR/epub/#dfn-epub-reading-system">Reading System</a>
-					</li>
-				</ul>
+				<p>In addition, it uses the <a href="https://www.w3.org/TR/WCAG2/#dfn-assistive-technologies">WCAG 2
+						definition</a> of <dfn data-lt="assistive technologies">Assistive Technology</dfn>.</p>
```
iherman: Per [xref](https://respec.org/xref/?term=assistive+technologies) the term assistive technologies can be reached via xref as well, directly from WAI-ARIA. Wouldn't it be better to use that rather than redefining the term here? (Same comment goes for the a11y spec)

epub33/a11y-tech/index.html
```diff
@@ -133,50 +134,28 @@ <h3>Purpose and Scope</h3>
...
-						<a href="https://www.w3.org/TR/epub/#dfn-media-overlay-document">Media Overlays Document</a>
-					</li>
-					<li>
-						<a href="https://www.w3.org/TR/epub/#dfn-package-document">Package Document</a>
-					</li>
-					<li>
-						<a href="https://www.w3.org/TR/epub/#dfn-epub-reading-system">Reading System</a>
-					</li>
-				</ul>
+				<p>In addition, it uses the <a href="https://www.w3.org/TR/WCAG2/#dfn-assistive-technologies">WCAG 2
```
iherman: Wouldn't it be cleaner to use data-cite for WCAG2 as well (here and everywhere else)?

iherman (approved):

epub33/a11y-tech/index.html
mattgarrish: We don't reference anything from those specs now and I don't expect we would in the future, at least not in this document. The specific details of how you implement the objectives are in the techniques document. This one's meant to provide higher-level requirements, like WCAG.

I didn't realize web-platform covered them all, though, so we could use that in place of the html reference I put in to future-proof it.

epub33/a11y-tech/index.html
mattgarrish: Does it work for the "latest version" links?

I'll test and see and if it works I'll update.

epub33/a11y/index.html
mattgarrish: Is there a "web-platform" profile? I know the docs say there is, but it doesn't appear to do anything. Respec seems to pick up any specs you have in your informative and normative list to check, but I tried adding dom and url terms and short of putting those specs in the list I just got errors:

> Couldn't find "DOMTokenList" in this document or other cited documents: [epub-3], [epub-33], [epub-a11y-11], [wai-aria], [wcag2], and [web-platform].

> Couldn't find "URL" in this document or other cited documents: [epub-3], [epub-33], [epub-a11y-11], [wai-aria], [wcag2], and [web-platform].

epub33/a11y-tech/index.html
mattgarrish: Ya, but we reference it from WCAG and I'm not sure the definitions are interchangeable. They have similarities, but I suspect the ARIA one has been tailored for the cases it specifically covers.

epub33/a11y/index.html
iherman: This is what I used in the other

```
				xref: {
					profile: "web-platform",
					specs: ["epub-rs-33","epub-33"]
				},
```

and I did get references resolved for `url` for example... that is all I know!

Things are indeed a bit uneven, though, sometimes a bit mysterious...

mattgarrish: Ya, the docs could use some updating in that case. The only place I found the `profile` property is in [example 90](https://respec.org/docs/#example-specify-profile-and-use-additional-specs-for-searching). All the rest show "web-platform" defined with `xref`.

clapierre (approved):

epub33/a11y/index.html
iherman: I think what happens is that respec also tries to do some contextual analysis. That is why you see, eg, wcag2 on the list of the error report, but nothing about the URL spec...

epub33/a11y-tech/index.html
iherman: Ok, I leave this to you... but it is looks as if _we_ defined the term when you click to it elsewhere.

mattgarrish: In a sense we are. We're defining that we use the same meaning as WCAG and also explain that AT may be part of reading systems in the note. That difference makes it unique to our specification.

The way it's phrased right now comes from the rewrite to meet ISO's formatting. It should really be defined as a proper term in a `dl`, and have the note put back closer.

iherman: Do we want to export that definition (which would mean adding the A11y spec to that list of specs, too)?

iherman: It may create lots of confusions, though...

mattgarrish: > It may create lots of confusions, though

Yes, let's not export it. We really only need it in the a11y document and the techniques. I've added a dl for it in the latest commits and the techniques document now references across to both a11y-11 and epub-33 for its terminology, using `data-cite` for AT. That's good enough.

Issue #2135, opened 2022-03-28
Terminology as proper nouns


mattgarrish:
Maybe this is heresy, but do we really need to capitalize _every_ word from the terminology _every_ time it is used?

We've effectively made proper nouns out of all the terminology for reasons that mostly elude me as this practice goes back to at least EPUB 2. Considering that the first instance of a term in a section is already linked to its definition, what does capitalizing all the words add?

From a visual perspective I've always found this jarring to read, and from an editing perspective it's always been a pain to make sure they're all capitalized.

So, painful as it may be to do, should we drop the caps?

mattgarrish:
Searching through various other W3C specs, I haven't yet found another one that follows this practice.

Even the few that capitalize the terms in the terminology don't in the body.

iherman:
Whatever works for improving readability is good, but I'm worried from the work it involves to change at its point:-(

mattgarrish:
> but I'm worried from the work it involves to change at its point:-(

Probably another to leave until after we publish a CR, but it may not be as hard as it seems.

A lot of the most used ones all start with "EPUB", so we can global those out without any worry about getting a lowercase letter in the wrong spot.

There may be a way to speed the rest up with a regex that skips any with a period immediately before, or certain opening tags like p or li since those are also likely the start of sentences.

iherman:
> Probably another to leave until after we publish a CR, but it may not be as hard as it seems.

Let us indeed leave it at that.

Note that, even in the CR phase, we are authorized to re-publish CR drafts via echidna, so this type of change can be done without further ado. We are required to go through a more formal step only if substantial changes (ie, whatever may invalidate implementation reports and/or tests) are done.

mattgarrish:
I made up a "Status-PostCRCleanup" label for this, but if there are other tasks, like sentence-casing the headings or whitespace cleanup, maybe we should start similarly logging them rather than keeping them in the back of our heads.

mattgarrish:
Closing with the merge of #2250 / #2248

PR #2248, opened 2022-04-14
Turned section titles into sentence case

### LINKED ISSUE #2179: Change section headings to use sentence case
This is the new direction for the W3C documents, better adapt to it.

(Enough to do this after CR)

iherman:
This is to fix: #2179

Two notes

- I have not touched the A11y documents to avoid merge conflicts
- I was not sure what to do with section titles that contained an official EPUB term. For the time being I did _not_ turned them into lower case; I guess this is something to settle together with the more general issue #2135

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2248.html" title="Last updated on Apr 15, 2022, 11:28 AM UTC (1655769)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2248/a229169...1655769.html" title="Last updated on Apr 15, 2022, 11:28 AM UTC (1655769)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -80,8 +80,7 @@
 		<style>
 			pre {
 				white-space: break-spaces !important;
-			}
-		</style>
+			}</style>
 	</head>
 	<body>
 		<section id="abstract">
@@ -89,7 +88,7 @@
 				format provides a means of representing, packaging, and encoding structured and semantically enhanced
 				web content — including HTML, CSS, SVG, and other resources — for distribution in a single-file
 				container.</p>
-			<p>This specification defines the authoring requirements for EPUB Publications and represents the third
+			<p>This specification defines the authoring requirements for EPUB publications and represents the third
 				major revision of the standard.</p>
 		</section>
 		<section id="sotd"></section>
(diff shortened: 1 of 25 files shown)


iherman:
@mattgarrish I see that the last a11y PR has been merged, too. I will complete this PR by doing things for the A11y spec, too.

mattgarrish:
> * I was not sure what to do with section titles that contained an official EPUB term. For the time being I did _not_ turned them into lower case

Yes, let's do this as one change so that we don't get mixed cases.

mattgarrish:
> I will complete this PR by doing things for the A11y spec, too.

👍

mattgarrish:
Since you're in the documents, could you change the link in the April 6 change log entry in the core spec so it goes to issue 2200? Right now it's a copy from the previous entry so goes to 2118. I'd push this to main but it'll probably just cause you a merge conflict.

iherman:
> > I will complete this PR by doing things for the A11y spec, too.
> 
> 👍 

Done now, including the change of the apr 6 log entry link.

iherman:
It was easier and quicker than I thought. I found an extension for Virtual Studio Code to turn a text into lower case. The rest is to step through the text via a search on `<h`, select and run that tool and then repeat...

iherman:
@wareid, I have updated the fxl a11y draft, too.

epub33/core/index.html
```diff
@@ -1569,7 +1569,7 @@ <h3>XML Conformance</h3>
 			</section>
 		</section>
 		<section id="sec-ocf">
-			<h2>Open Container Format</h2>
+			<h2>Open container format</h2>
```
mattgarrish: I would keep this in caps even after we change the terminology. It's an acronym. We should also add "(OCF)" as the next heading uses that initialism and there's no place to explain it.

```suggestion
			<h2>Open Container Format (OCF)</h2>
```

mattgarrish:
I think we have to do this with the terminology. There are some inconsistencies in naming -- some terms retain caps in the core but are lowercase in other specs.

I'll try to fix the terminology this afternoon and then I'll give the headings a second pass.

mattgarrish:
I'm generally fine with this and #2250 going in. With the additional lowercasing in that PR, the tocs seemed fine.

epub33/core/index.html
mattgarrish: I fixed this in #2250, so you can ignore or commit, but either way it will get updated.

PR #2250, opened 2022-04-14
Lowercase all terminology

### LINKED ISSUE #2135: Terminology as proper nouns

mattgarrish:
Fixes #2135 

I have a feeling we're going to find weird phrasings of the terms still in caps in the future, but this should fix the vast majority.

(It's a lot easier to spot normative keywords now without the sea of caps to wade through. Already liking this.)

I also fixed some inline cross-references to sections that were still in caps.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2250.html" title="Last updated on Apr 14, 2022, 7:34 PM UTC (df47319)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2250/3feefbd...df47319.html" title="Last updated on Apr 14, 2022, 7:34 PM UTC (df47319)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -80,8 +80,7 @@
 		<style>
 			pre {
 				white-space: break-spaces !important;
-			}
-		</style>
+			}</style>
 	</head>
 	<body>
 		<section id="abstract">
@@ -89,7 +88,7 @@
 				format provides a means of representing, packaging, and encoding structured and semantically enhanced
 				web content — including HTML, CSS, SVG, and other resources — for distribution in a single-file
 				container.</p>
-			<p>This specification defines the authoring requirements for EPUB Publications and represents the third
+			<p>This specification defines the authoring requirements for EPUB publications and represents the third
 				major revision of the standard.</p>
 		</section>
 		<section id="sotd"></section>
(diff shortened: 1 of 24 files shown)


epub33/core/index.html
```diff
@@ -1863,61 +1857,61 @@ <h4>URLs in the OCF Abstract Container</h4>
...
+						</p>
 					</div>
 
-					<p> A string <var>url</var> is a 
-						<dfn class="export" id="dfn-valid-relative-container-url-with-fragment-string">valid-relative-ocf-URL-with-fragment string</dfn> 
-						if it is a <a>path-relative-scheme-less-url string</a>, optionally followed by <code>U+0023 (#)</code> and
-						a <a>url-fragment string</a>, and if the following steps return <var>true</var>: </p>
+					<p> A string <var>url</var> is a <dfn class="export"
+							id="dfn-valid-relative-container-url-with-fragment-string"
+							>valid-relative-ocf-URL-with-fragment string</dfn> if it is a
```
iherman, line 1874: I wonder whether we should not use `url` instead of `URL` in this term definition. All the similar terms of the URL specification use lower case url in their definition (like "path-relative-scheme-less-url string"). The counterargument may be that we use `URL` in all our terms related to urls, like "container root URL".

We could also decide to systematically use lower case url for our other terms as well.

iherman (approved):
It is virtually impossible to do a detailed review for all the files. A huge +1 for these changes overall from me; I would think the best is to merge this (and then #2248) asap, and we may hit edge cases later. I am sure we will have other changes coming up...

epub33/core/index.html
mattgarrish: > All the similar terms of the URL specification use lower case url in their definition

Only in their IDs, which we do, too. The names themselves keep the uppercase form. For example:

- [path-relative-scheme-less-URL string](https://url.spec.whatwg.org/#path-relative-scheme-less-url-string)
- [path-absolute-URL string](https://url.spec.whatwg.org/#path-absolute-url-string)

PR #2252, opened 2022-04-15
Lowercase all normative/informative reference identifiers

### LINKED ISSUE #2251: Use lower case for references?
I really love the effect of #2250 (ie, #2135). I was wondering whether it is worth considering one step forward and use lower case references rather than uppercase ones. Ie, `[[html]]` instead of `[[HTML]]`. I always found that visually more pleasing than the upper case. I do not think there is an obligation to use upper case, see, eg, [the annotation model spec](https://www.w3.org/TR/annotation-model/#normative-references).

mattgarrish:
Chaning case didn't flip any references, so all good.

Fixes #2251

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2252.html" title="Last updated on Apr 15, 2022, 12:40 PM UTC (af5431c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2252/1d6e506...af5431c.html" title="Last updated on Apr 15, 2022, 12:40 PM UTC (af5431c)">Diff</a>

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -51,7 +51,7 @@
 				xref: ["epub-33"],
 				profile: "web-platform",
 				localBiblio: {
-					"A11Y-DISCOV-VOCAB": {
+					"a11y-discov-vocab": {
 						"title": "Schema.org Accessibility Properties for Discoverability Vocabulary",
 						"href": "https://www.w3.org/2021/a11y-discov-vocab/latest/",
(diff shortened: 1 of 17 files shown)


mattgarrish:
Given this is another impossible review, and it was almost all done by regular expression on a pattern that is highly unlikely to have unintended consequences, I'm going to merge this in.

PR #2253, opened 2022-04-17
Various editorial cleanup

mattgarrish:
This PR cleans up a number of things:

- I standardized the text in the terminology sections. We had different wordings and also a couple of documents still mentioned the terms being in caps. I also moved the text about the first term in a section linking to its definition into a note as this isn't anything about the terminology itself.
- I realized I mistakenly used xref in the accessibility specifications when they're supposed to reference the latest version. I had to fix all these to use `href` with TR/epub/ urls.
- I've removed all the plural forms from the terminology definitions as these are generated automatically - we only needed them for the old JS code. I've also taken off the IDs for the same reason
- I've added `dfn` tags to a lot of elements to simplify referencing them, but we have some conflicts (e.g., I skipped the `link` element in the container.xml definition and the SMIL `metadata` element because the package document versions are the more likely to be referenced). I also skipped elements for which there are no references as otherwise we get respec warnings (`smil` element plus most of the elements in the various META-INF files). I've updated the element references.
- To do the previous change, I changed spine/manifest to "EPUB spine (or spine)" and "EPUB manifest (or manifest)" to avoid colliding definitions with the spine/manifest elements. While doing this, I noticed that "EPUB conformance checker" was out of alphabetical order and moved it (probably because it used to be EPUB validator).
- I only export the major package elements (excludes the DC optional and legacy), the MO seq/par/audio/text, and the epub:type attributes. I don't think anything else is referenced outside the core, and doesn't seem likely to be in the future.
- I deleted the dfn-crossref.js file and removed some lingering calls to it in the documents
- "Media Overlays" was in caps in a lot of places talking about the feature generally and also sometimes as a shorthand for media overlay document. I've lowercased these.

There's also the odd really minor fix (link correction, capitalization, etc.).

I don't plan to do any more changes now until we get the CR drafts published.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2253.html" title="Last updated on Apr 19, 2022, 2:17 AM UTC (4c72c3f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2253/692b666...4c72c3f.html" title="Last updated on Apr 19, 2022, 2:17 AM UTC (4c72c3f)">Diff</a>


mattgarrish:
The specs that failed are the ones that reference the newly exported terms.

@iherman I assume we can just ignore the failures and merging will take care of the problem? (Or worst case, they fail again on merge and we have to run the jobs again after the core spec goes through and the database gets updated.)

iherman:
> * I also skipped elements for which there are no references as otherwise we get respec warnings (`smil` element plus most of the elements in the various META-INF files). I've updated the element references.

Isn't there a flag to switch those warnings off? Future versions may need those references...

iherman:
> The specs that failed are the ones that reference the newly exported terms.
> 
> @iherman I assume we can just ignore the failures and merging will take care of the problem? (Or worst case, they fail again on merge and we have to run the jobs again after the core spec goes through and the database gets updated.)

I would expect these will mend themselves, but I would run the jobs manually first. The experience seems to be that changes to the terminology databse take a whike to get updated... (we can also ask François to give a kick)

iherman:
As before, best is to merge this asap, we can always get issues solved later...

iherman (approved):

mattgarrish:
> Isn't there a flag to switch those warnings off?

It depends on what we'd want to do about them. We can shut them all off in the respec config using `no-unused-dfns` or we can individually turn them off using the `lint-ignore` css class.

The other option is to export everything, as exporting automatically silences the warnings, too, assuming because it's expected other specs will reference.

Should we export all the elements regardless of whether anything else references them and regardless of how important they are?

I'm fine either way, but something about exporting all the DC optional elements seemed pointless when we barely define most. Maybe for those we can just export the ones we have our own definitions for? (I still haven't found any way to export two elements with the same name but from different grammars.)

murata2makoto:
@mattgarrish Thanks for lots of work!  I find that "EPUB spine" and "EPUB manifest" are defined but they are never used.  Should we rather always write "spine element" and "manifest element" when we are talking about XML elements?

mattgarrish:
> I find that "EPUB spine" and "EPUB manifest" are defined but they are never used.

There are instances where we talk about the spine and manifest as concepts instead of elements, but we use the same terms "spine" and "manifest" as the element names. We can't have definitions for the concepts and the elements using the same word, so that's why we've put "EPUB" in front of the terminology definitions so that we can have separate definitions.

What it means practically is that when using the concepts, you'll find markup like this (not actual cites from the specs):

```
The <a data-lt="EPUB spine">spine</a> establishes ...
```

whereas when we're talking specifically about the XML element we have

```
... each item in the <a><code>spine</code></a> element references
```

The `data-lt` attribute lets us reference the terminology definition without colliding with the element name.

I don't want to go through and add "EPUB" in front of all the terminology references, as we already overuse that prefix for terms that don't need it.

More an aside, but I'd like to drop the prefix from many of the terms, as about the only "EPUB X" terms that might cause confusion without a prefix are "EPUB publication" and maybe "EPUB Container". (Edit: EPUB content document would be messy to drop EPUB from, given XHTML/SVG/Foreign variations exist.)

Do these really become confusing without "EPUB"?

- conformance checker
- creator
- navigation document
- reading system

mattgarrish:
> Should we export all the elements

FYI, in the last commit I've exported all the elements except the DC optional ones. Since there's only a generic definition for all of them, and we don't have any specific purpose for them, I don't see a point in defining and exporting.

iherman:
> Do these really become confusing without "EPUB"?
> 
> * conformance checker
> * creator
> * navigation document
> * reading system

Not really... the context makes the situation clear.

mattgarrish:
> Not really... the context makes the situation clear.

Ya, the concern I have in changing now is that I got away with not having to check capitalization when I lowercased these, only the solo instances of "reading system" and just those were a major pain in the a-- that we'd have to recheck.

There are easily 1000+ instances of creator and reading system that would need to be reviewed. Not my idea of fun. ☹️ 

This is definitely for post-CR if we really want to tackle it.

PR #2254, opened 2022-04-19
Fix conflicting element citations

mattgarrish:
Have to watch out when our element names match those in html. Need to put a data-cite attribute on a nearby element to indicate which spec has precedence as these errors only popped up once our newly exported elements went live.

It appears to happen most often with the package link and meta elements and MO audio element.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2254.html" title="Last updated on Apr 19, 2022, 8:30 PM UTC (978e1ed)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2254/c70d251...978e1ed.html" title="Last updated on Apr 19, 2022, 8:30 PM UTC (978e1ed)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -143,10 +143,9 @@
 				<p>An EPUB publication is typically represented by a single [=package document=]. This document includes
 					metadata used by [=reading systems=] to present the content to the user, such as the title and
 					author for display in a bookshelf as well as rendering metadata (e.g., whether the content is
-					reflowable or has a fixed layout). It also provides a manifest of resources and includes a <a
-						data-lt="EPUB spine">spine</a> that lists the default sequence in which to render documents as a
-					user progresses through the content. Refer to <a href="#sec-package-doc"></a> for the requirements
-					for the package document.</p>
+					reflowable or has a fixed layout). It also provides a manifest of resources and includes a [=spine=]
+					that lists the default sequence in which to render documents as a user progresses through the
+					content. Refer to <a href="#sec-package-doc"></a> for the requirements for the package document.</p>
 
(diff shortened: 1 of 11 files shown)


iherman:
I have already seen this CI error which seems to refer to the underlying process rather than our code. Did the previous error just disappear, or did you find something @mattgarrish ? (I saw it while I was on my trip on the week-end)

iherman (approved):

mattgarrish:
Everything should be working now except I'm going to wait for the `epub:type` definition to get updated before pushing one final cleanup commit. This gets rid of the ambiguous reference errors which were causing the final crashes.

It's just been a headache waiting for the definitions to get updated while trying to also push fixes.

mattgarrish:
I've been discovering some more secrets of the definition exports this afternoon.

Specifically, we can use `data-dfn-for` to tie definitions together. So, for example, I've added `data-dfn-for="package"` to the package doc section tag and that means we can reference all the elements as `[^package/x^]`. Similarly, `data-dfn-for="smil"` on the media overlays section means we can reference all those elements as `[^smil/x^]`.

That means `[^package/metadata^]` can coexist with `[^smil/metadata^]`, and `[^package/spine^]` can coexist with `[=spine=]`. The display text for these does not include the prefixes.

But this means that I've got to merge, fail a bunch of specs again, and wait around 4 more hours for the database to update

iherman:
> But this means that I've got to merge, fail a bunch of specs again, and wait around 4 more hours for the database to update

Well... we can use the time delay for this. You make the commits that generate build errors on one of the evenings your time, when I am already out, and ping me to restart the actions the following day morning for me. The time difference should work it out if it is indeed a matter of around 4 hours.

PR #2258, opened 2022-04-20
Adding html wrapper around md files for reports

iherman:
I got a bit bothered by having a few impl. report files in markdown when most of the files are in HTML... I used a simple respec trick to provide a more uniform html view of the two markdown files (using an almost empty respec shell that imports the markdown content).

Extra bonus: the generated HTML has anchors that can be used to link to from the specs.

Previews:

- [A11y property report](https://raw.githack.com/w3c/epub-specs/admin/all-reports-html/epub33/reports/a11y-properties-use.html)
- [EPUB property report](https://raw.githack.com/w3c/epub-specs/admin/all-reports-html/epub33/reports/epub-properties-use.html)

### DIFF
--- a/epub33/reports/a11y-properties-use.html
+++ b/epub33/reports/a11y-properties-use.html
@@ -51,7 +51,7 @@
     <script class="remove">
         var respecConfig = {
             specStatus: "base",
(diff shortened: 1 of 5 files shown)


iherman:
(This PR does not affect the spec, so it can be merged in spite of our own moratorium related to the CR publication.)

PR #2277, opened 2022-05-12
Consolidate redundant structural semantics requirements

### LINKED ISSUE #2273: Redundant structural semantics normative statements?
https://w3c.github.io/epub-specs/epub33/rs/#sec-structural-semantics says:

> When processing the [epub:type](https://www.w3.org/TR/epub-33/#dfn-epub-type) attribute, a reading system:

> * ...

> * MUST [ignore structural semantics](https://w3c.github.io/epub-specs/epub33/rs/#confreq-rs-epubtype-prec) that conflict with the carrying element.

> When the Reading system behavior associated with a given epub:type value conflicts with an element's native behavior, the behavior associated with the element MUST be given precedence.

As far as I can tell, these are the same requirement. Can we either combine them or, if they're not the same, clarify how they're distinct?

dlazin:
Fixes #2273.

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2026,10 +2026,6 @@
 				<li>
 					<p id="confreq-rs-epubtype-ign">MUST ignore terms that it does not recognize.</p>
 				</li>
-				<li>
-					<p id="confreq-rs-epubtype-conflict">MUST, if the behavior for a given <code>epub:type</code> value
-						conflicts with the element's native behavior, give precedence to the native behavior.</p>
-				</li>
 			</ul>
 		</section>
 		<section id="sec-vocab-assoc">
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2027,8 +2027,8 @@
 					<p id="confreq-rs-epubtype-ign">MUST ignore terms that it does not recognize.</p>
 				</li>
 				<li>
-					<p id="confreq-rs-epubtype-conflict">MUST ignore the <code>epub:type</code> value if its semantics
-						conflict with the element's native behavior.</p>
+					<p id="confreq-rs-epubtype-conflict">MUST, if the behavior for a given <code>epub:type</code> value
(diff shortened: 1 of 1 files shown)


iherman (approved):

iherman:
@dlazin before you are tempted: do not merge this PR until #2268 is not merged (now that we are in CR the automatic publication procedure changes a tiny bit). I will try to merge that tomorrow when I can also monitor whether things are going o.k.

mattgarrish:
As I mentioned in #2273, the better answer here is to remove the bullet and the paragraph. There are no inherent behaviours associated with the use of epub:type, so if there are conflicts with an implementation (currently only pop-up notes) then the implementation(s) need to account for where and how the semantics are used. That's not in our scope unless we're going to standardize the footnote behaviour.

dlazin:
> As I mentioned in #2273, the better answer here is to remove the bullet and the paragraph. There are no inherent behaviours associated with the use of epub:type, so if there are conflicts with an implementation (currently only pop-up notes) then the implementation(s) need to account for where and how the semantics are used. That's not in our scope unless we're going to standardize the footnote behaviour.

OK, I have done so. (I have no opinion on whether this is better/OK; Ivan, do you agree with Matt on deleting these normative statements entirely?)

iherman (approved):
I am fine leaving out these things altogether. The less we talk about `epub:type` the better is...

While we are at this section, one tiny, though independent, issue before you merge, please? In the item

> MAY associate behaviors with none, some or all of the terms defined

there should be a comma before "or all"; the W3C publishing rules advise using Oxford commas...

dlazin:
> I am fine leaving out these things altogether. The less we talk about `epub:type` the better is...
> 
> While we are at this section, one tiny, though independent, issue before you merge, please? In the item
> 
> > MAY associate behaviors with none, some or all of the terms defined
> 
> there should be a comma before "or all"; the W3C publishing rules advise using Oxford commas...

Done, along with an Oxford comma for the list of font resources too, which I noticed yesterday and meant to fix.

PR #2288, opened 2022-05-13
Add index to core spec

### LINKED ISSUE #2260: Index for spec?
In the EPUB 3.3 Webinar hosted by Booknet Canada, someone asked why the spec didn't have an index. It's a good question. I know that some other W3C specs do have indexes. Can ReSpec help with this?

mattgarrish:
Adds the respec-generated index of terms. I don't think this is useful for any of the other specs as they don't add similar numbers of new terms.

Fixes #2260

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 17, 2022, 11:44 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2F452a1385184e18cc485e6f48ae30addb72d41c0a%2Fepub33%2Fcore%2Findex.html%3FisPreview%3Dtrue)

```

Specification: http://labs.w3.org/spec-generator/uploads/KfgrHj?isPreview=true&publishDate=2022-05-17
ReSpec version: 32.1.6
Error: Error: Evaluation failed: Timeout: document.respec.ready didn't resolve in 27763ms.
    at async file:///u/spec-generator/server.js:244:48

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/epub-specs%232288.)._
</details>

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -1839,6 +1839,7 @@
 				</section>
 			</section>
 		</section>
+		<section id="index"></section>
 		<section id="change-log" class="appendix informative">
 			<h2>Change log</h2>
 
@@ -1852,6 +1853,8 @@
 					>working group's issue tracker</a>.</p>
 
 			<ul>
+				<li>17-May-2022: Added an index of terms. See <a href="https://github.com/w3c/epub-specs/issues/2260"
+						>issue 2260</a>.</li>
 				<li>12-Apr-2022: Restored recommendation to include links to all reproduced pages in the page list and
 					added a requirement to link to all page break markers to address concerns with the previous change.
 					Clarified that meeting the page navigation requirements is required, but adding page navigation is
--- a/epub33/common/css/common.css
+++ b/epub33/common/css/common.css
@@ -376,10 +376,3 @@ details.explanation > summary, details.desc > summary {
     font-weight: normal !important;
     font-style: italic;
 }
-
-code a[data-type~="element"], 
-code a[data-type~="dfn"],
-code a[data-type~="element-attr"]
-{
-    color: #c63501 !important;
-}
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -11510,7 +11510,7 @@ EPUB/images/cover.png</pre>
 					>Working Group's issue tracker</a>.</p>
 
 			<ul>
-				<li>13-May-2022: Added an index of terms. See <a href="https://github.com/w3c/epub-specs/issues/2260"
+				<li>17-May-2022: Added an index of terms. See <a href="https://github.com/w3c/epub-specs/issues/2260"
 						>issue 2260</a>.</li>
 				<li>12-Apr-2022: Added note about complexities of escaping from nested escapable structures and updated
 					the list of semantics to use for escaping to match the guidance. See <a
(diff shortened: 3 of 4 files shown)


mattgarrish:
Can be previewed here: https://raw.githack.com/w3c/epub-specs/feature/add-index/epub33/core/index.html#index

iherman (approved):
Overall, I think it is worth adding it, even if not perfect.

One tiny thing. The generated code seems to be:

```html
<a class="index-term" href="....">
    "container resource"
</a>
```

for elements and then

```html
<a class="index-term" href="....">
    <code>meta</code>
</a>
```

for elements. Would it be better if we added to our css something like

```css
a.index-term code {
   color: #c63501 !important;
}
```

we use that colour all over the place for code, better keep it in the index, too...

mattgarrish:
> Would it be better if we added to our css

Should this be raised against respec?

It's respec that colours the code tags in the body, so if we manually override the index then we have to keep watch that the respec css doesn't change in the future.

iherman:
> > Would it be better if we added to our css
> 
> Should this be raised against respec?
> 
> It's respec that colours the code tags in the body, so if we manually override the index then we have to keep watch that the respec css doesn't change in the future.

Oops, somehow I always thought that we set the color, not respec...

Yes, that should be a respec issue. Not only for this case; I have added the following entries to our common css files to ensure the right colour when using xref:

```css
code a[data-type~="element"], 
code a[data-type~="dfn"],
code a[data-type~="element-attr"]
{
    color: #c63501 !important;
}
```

These should be superfluous as well.

Do we want to start by setting it for ourselves and raise an issue afterwards, or just raise an issue? I would be in favour of the former; I do not know whether such an issue would be taken care of with high priority...

mattgarrish:
> Do we want to start by setting it for ourselves and raise an issue afterwards

That works for me. I'll update the PR when I have a minute.

iherman:
Also: do we want to switch indexing on for RS? It does not define many new concepts, but I also found the generated list of external concept references useful... (I cannot judge the value for the A11y document).

mattgarrish:
> > Do we want to start by setting it for ourselves and raise an issue afterwards
> 
> That works for me.

Although, looking at the result of adding custom CSS, I'm not so sure it's a good idea. Linked code is now hard to distinguish from unlinked code as there may only be a faint dotted underline. We also get links that are part orangey-red and part blue if they include a code-tagged word with other words. It's really quite messy. It might be better to remove the custom CSS and figure out with the respec/spec-prod folks what should be done.

mattgarrish:
> It does not define many new concepts, but I also found the generated list of external concept references useful...

I'm of the opposite view: it's rather clunky and duplicative. It appears to take the text of every `a` element and add an entry. If we link to definitions and sections using the same words, we get different entries, although nothing makes clear which is which until you follow the link into the body and then click on that link to see where it goes.

Here's a random sampling I spotted in the RS spec only looking at a few "p" entries:

```
- package document
- package document
...
- properties
- properties attribute
- properties attribute
...
- publication resources
- publication resources
```

mattgarrish:
I'm opting for a middle road on this and adding indexes to the three rec-track specs but removing the CSS link overrides so we have consistent colouring of our links.

PR #2299, opened 2022-05-19
Add recommendations to use https for remote resources

### LINKED ISSUE #2263: External resources should be loaded securely
External resources should be loaded securely, for example over HTTPS. Otherwise, threats would also include any network attacker when the book is being read (even separate from whether the book contents itself were securely transferred). This is currently a non-normative recommendation, but seems like a good candidate for a normative recommendation.

Network attackers are not currently described in the threat model of either the core or rs specs.

mattgarrish:
Adds authoring recommendation to reference resources via https to limit possibility of network attacks and adds reading system recommendation to only load remote resources referenced via https.
		
This could lead to warnings in existing content, but in this case I think there's a strong security basis for making the change. I don't get the impression that publishers actually use a lot of remote resources, either, but we should discuss this in a WG meeting before merging.

Fixes #2263

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 27, 2022, 1:45 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Ff74eb6516d2662915e58a809eee6591eb2725261%2Fepub33%2Fcore%2Findex.html%3FisPreview%3Dtrue)

```

Specification: http://labs.w3.org/spec-generator/uploads/iYt5U4?isPreview=true&publishDate=2022-05-27
ReSpec version: 32.1.7
Error: Error: Evaluation failed: Timeout: document.respec.ready didn't resolve in 27830ms.
    at async file:///u/spec-generator/server.js:244:48

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/epub-specs%232299.)._
</details>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1270,7 +1270,6 @@
 					automatically becomes exempt from fallbacks.</p>
 			</section>
 
-
 			<section id="sec-resource-fallbacks">
 				<h4>Resource fallbacks</h4>
 
@@ -1385,6 +1384,7 @@
 					</section>
 				</section>
 			</section>
+
 			<section id="sec-resource-locations">
 				<h4>Resource locations</h4>
 
@@ -1410,6 +1410,11 @@
 
 				<p>Storing all resources inside the EPUB container is strongly encouraged whenever possible as it allows
 					users access to the entire presentation regardless of connectivity status.</p>
+
+				<p>When resources have to be located outside the EPUB container, EPUB creators are strongly RECOMMENDED
+					to securely reference them via the <code>https</code> URI scheme [rfc7230] to limit the threat of
+					exposing their publications, and users, to network attacks. Reading systems may not load remote
+					resources referenced using insecure schemes such as <code>http</code>.</p>
 
 				<p>These rules for locating publication resource apply regardless of whether the given resource is a
 					[=core media type resource=] or a [=foreign resource=].</p>
@@ -11503,6 +11508,8 @@ EPUB/images/cover.png</pre>
 					>Working Group's issue tracker</a>.</p>
 
 			<ul>
+				<li>19-May-2022: Added recommendation to only reference remote resources via https. See <a
+						href="https://github.com/w3c/epub-specs/issues/2263">issue 2263</a>.</li>
 				<li>17-May-2022: Added an index of terms. See <a href="https://github.com/w3c/epub-specs/issues/2260"
 						>issue 2260</a>.</li>
 				<li>12-Apr-2022: Added note about complexities of escaping from nested escapable structures and updated
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1411,8 +1411,8 @@
 				<p>Storing all resources inside the EPUB container is strongly encouraged whenever possible as it allows
 					users access to the entire presentation regardless of connectivity status.</p>
 
-				<p>When resources have to be located outside the EPUB container, EPUB creators are strongly RECOMMENDED
-					to reference them via the secure <code>https</code> URI scheme [rfc7230] to limit the threat of
+				<p>When resources have to be located outside the EPUB container, EPUB creators are RECOMMENDED to
+					reference them via the secure <code>https</code> URI scheme [rfc7230] to limit the threat of
 					exposing their publications, and users, to network attacks. Reading systems might not load remote
 					resources referenced using insecure schemes such as <code>http</code>.</p>
 
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1412,7 +1412,7 @@
 					users access to the entire presentation regardless of connectivity status.</p>
 
 				<p>When resources have to be located outside the EPUB container, EPUB creators are strongly RECOMMENDED
-					to securely reference them via the <code>https</code> URI scheme [rfc7230] to limit the threat of
+					to reference them via the secure <code>https</code> URI scheme [rfc7230] to limit the threat of
 					exposing their publications, and users, to network attacks. Reading systems might not load remote
 					resources referenced using insecure schemes such as <code>http</code>.</p>
 
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1413,7 +1413,7 @@
 
 				<p>When resources have to be located outside the EPUB container, EPUB creators are strongly RECOMMENDED
 					to securely reference them via the <code>https</code> URI scheme [rfc7230] to limit the threat of
-					exposing their publications, and users, to network attacks. Reading systems may not load remote
+					exposing their publications, and users, to network attacks. Reading systems might not load remote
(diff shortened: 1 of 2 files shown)


mattgarrish:
* EPUB 3.3:
    * [Preview](https://raw.githack.com/w3c/epub-specs/fix/issue-2263/epub33/core/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/core/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fix/issue-2263/epub33/core/index.html)
* EPUB Reading Systems 3.3:
    * [Preview](https://raw.githack.com/w3c/epub-specs/fix/issue-2263/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fix/issue-2263/epub33/rs/index.html)

mattgarrish:
Also note that I opted to put the recommendations directly in the relevant resources sections as these are very specific to remote resources.

dauwhe (approved):

shiestyle (approved):

epub33/core/index.html
```diff
@@ -1411,6 +1411,11 @@ <h4>Resource locations</h4>
 				<p>Storing all resources inside the EPUB container is strongly encouraged whenever possible as it allows
 					users access to the entire presentation regardless of connectivity status.</p>
 
+				<p>When resources have to be located outside the EPUB container, EPUB creators are strongly RECOMMENDED
+					to securely reference them via the <code>https</code> URI scheme [rfc7230] to limit the threat of
+					exposing their publications, and users, to network attacks. Reading systems may not load remote
```
iherman: Just a minor comment: not all resources on the Web are available through https, and this is not under the control of the author. Maybe it is better to say something like:

>...to securely reference them, whenever possible, via the ...

iherman (approved):

epub33/core/index.html
mattgarrish: Isn't that baked into it being a recommendation instead of a requirement?  If we add "whenever possible" we reduce it to an untestable assertion, so effectively an informative "should".

Remote resources are things the author is explicitly bringing into the publication, not things they're linking out to, so in this case I think the burden should be on the author to only reference secure resources given it's exposing users and RSes to an attack vector.

iherman: > Isn't that baked into it being a recommendation instead of a requirement? I

Oops, you are right. Forget this.

epub33/core/index.html
```diff
@@ -1411,6 +1411,11 @@ <h4>Resource locations</h4>
 				<p>Storing all resources inside the EPUB container is strongly encouraged whenever possible as it allows
 					users access to the entire presentation regardless of connectivity status.</p>
 
+				<p>When resources have to be located outside the EPUB container, EPUB creators are strongly RECOMMENDED
```
dlazin: Is there a practical difference between "are strongly RECOMMENDED" and just "SHOULD"?

If they are equivalent (which I think is what the RFC says), I recommend (ha) saying SHOULD just because we use that term more often and thus it will trip fewer readers up.

epub33/core/index.html
```diff
@@ -1411,6 +1411,11 @@ <h4>Resource locations</h4>
...
+					to securely reference them via the <code>https</code> URI scheme [rfc7230] to limit the threat of
```
dlazin:
```suggestion
					to reference them via the secure <code>https</code> URI scheme [rfc7230] to limit the threat of
```

epub33/core/index.html
dlazin:
```suggestion
					exposing their publications, and users, to network attacks. Reading systems might not load remote
```

dlazin (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-05-26-epub#section1-2) on 2022-05-26

List of resolutions:

- ***[Resolution No. 2](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-05-26-epub#resolution2):  Approve PR 2299, close issue 2265.***

#### 1.2. External Resources should be loaded securely.

_See github issue [epub-specs#2263](https://github.com/w3c/epub-specs/issues/2263)._




_See github pull request [epub-specs#2299](https://github.com/w3c/epub-specs/pull/2299)._




**Dave Cramer:** this issue comes with this PR.  
… this is around remote resources, and recommended that they be served over HTTPS.  
… avoiding person in the middle and other such network attacks.  

**Dan Lazin:** my only concern is the PR itself is not super clear in its language.  
… "RS may not load resources".  
… what we mean is there is a chance RS may not load.  
… I'll do a pass.  

**Dave Cramer:** this is _SHOULD_ level, so would we test whether a remote resource is served via HTTP, and that would result in a warning from epubcheck?.  

**Matt Garrish:** yes, that is what I would expect.  

**Dave Cramer:** this could affect existing content, but I expect it would be rare.  

**Matt Garrish:** generally the direction of the web is HTTPS too.  
… this will cause some warnings, but security outweighs. We should move in the direction of the web.  

**Dave Cramer:** using backwards compatibility as reason not to fix security issue isn't where we want to be.  

> **Proposed resolution: Approve PR 2299, close issue 2265.** *(Wendy Reid)*

> *Brady Duga:* +1.

> *Wendy Reid:* +1.

> *Shinya Takami (高見真也):* +1.

> *Matthew Chan:* +1.

> *Masakazu Kitahara:* +1.

> *Dan Lazin:* +1.

> *Dave Cramer:* +1.

> *Toshiaki Koike:* +1.

> *Matt Garrish:* +1.

> ***Resolution #2: Approve PR 2299, close issue 2265.***

</details>

epub33/core/index.html
mattgarrish: Ya, it's usually used for emphasis when we're speaking non-normatively, like the "strongly encouraged" in the paragraph above.

Since this is now a normative statement, "strongly" should definitely be stripped. There's a similar instance on a security statement in the RS spec I'll erase.

iherman:
@mattgarrish any reason not to merge this?

Issue #2271, opened 2022-05-11
Requirement (MUST vs SHOULD) for Media Overlays is unclear


dlazin:
https://w3c.github.io/epub-specs/epub33/rs/#confreq-rs-epub3-mo says:

> Reading systems with the capability to render prerecorded audio SHOULD support [media overlays](https://www.w3.org/TR/epub-33/#sec-media-overlays) [[epub-33](https://w3c.github.io/epub-specs/epub33/rs/#bib-epub-33)].

But then:

> When a reading system loads a [package document](https://www.w3.org/TR/epub-33/#dfn-package-document), it MUST refer to the [manifest](https://www.w3.org/TR/epub-33/#dfn-epub-manifest) [item](https://www.w3.org/TR/epub-33/#dfn-item) elements' [[epub-33](https://w3c.github.io/epub-specs/epub33/rs/#bib-epub-33)] media-overlay attributes to discover the corresponding media overlays for [EPUB content documents](https://www.w3.org/TR/epub-33/#dfn-epub-content-document).

> Reading systems MUST support playback for [XHTML content documents](https://www.w3.org/TR/epub-33/#dfn-xhtml-content-document), and MAY support [SVG content documents](https://www.w3.org/TR/epub-33/#dfn-svg-content-document).

> Playback MUST start...

This makes it sound like (full/proper) *support* for MOs is optional, but RSes still need to do all the processing. I don't think that's what's intended. I assume we mean "if you are going to support MOs, then here is what is a requirement for support." Elsewhere, we use language like "If a reading system supports scripting...". I think we should sprinkle such language throughout the MO section, assuming my interpretation is correct.

iherman:
I agree with your second interpretation, ie, that "_If_ the reading system support scripting then..."

Instead of sprinkle such language all around (which may lead to fairly convoluted sentences) what about making this clear in the intro section? Eg, by changing the first paragraph right after the yellow box to something like:

> If a reading system does not support media overlays, all normative requirements in this section can be ignored. In particular, such a reading system MUST ignore both:

@mattgarrish ?

mattgarrish:
I'm not sure I see the problem here. If you have to ignore the media-overlay attribute and all SMIL files if you don't support media overlays, which is what the following paragraph says, then by extension none of the processing and playback requirements for the SMIL files apply.

mattgarrish:
[...]

Issue #2181, opened 2022-03-31
EPUB 3.3 (Core Spec) Edits - 1. Introduction


BenSchroeter:
This section is in great shape... only two minor suggested edits: 

Section 1.2, Para 9:
hyphenate "universally named"

Section 1.4, Package Document:
insert Oxford comma after "resources"

mattgarrish:
> hyphenate "universally named"

FYI, I'm trying to keep the documents to the rule of not hyphenating adverbs ending in ly when the are followed by an adjective. See this CMoS faq: https://www.chicagomanualofstyle.org/qanda/data/faq/topics/HyphensEnDashesEmDashes/faq0010.html

(But growing up being told to hyphenate these, it's always my initial instinct to hyphenate them, too.)

PR #2305, opened 2022-05-20
Add reading system conformance section

mattgarrish:
May be stating the obvious by adding this, but probably better to err on the side of clarity:

- states that all required and conditionally required features must be met to be conformant
- for recommended feature, processing can typically be ignored, but not always (e.g., scripting has fallback requirements when not supporting)

Addresses part of #2271 but not going to close that issue automatically.

EPUB Reading Systems 3.3:
- [Preview](https://raw.githack.com/w3c/epub-specs/feature/rs-conformance/epub33/rs/index.html)
- [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/feature/rs-conformance/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -225,12 +225,16 @@
 			<p>Whether a reading system has to support a feature is mentioned at the beginning of its section. To be
 				conformant with this specification, reading systems MUST support all required features as well as all
 				applicable conditionally-required features (e.g., to support image rendering if the reading system has a
-				[=viewport=]).</p>
+				[=viewport=]) as defined in their respective sections.</p>
 
-			<p>When support for a feature is only recommended, reading system developers can typically ignore all the
-				normative statements in its section. In some cases, however, there may be alternative requirements when
-				not implementing a feature (e.g., to <a href="#confreq-rs-scripted-flbk">process fallbacks</a> when
-				scripting is not supported).</p>
+			<p>When supporting recommended and optional features, reading systems MUST meet all normative requirements
+				as defined in their respective sections.</p>
+
+			<p>When reading system developers opt not to support a recommended or optional feature, it does not always
+				mean none of the normative requirements of the section apply. In some cases, there may be alternative
+				requirements when not implementing a feature (e.g., to <a href="#confreq-rs-scripted-flbk">process
+					fallbacks</a> when scripting is not supported). Reading systems MUST meet these alternative
+				requirements when not supporting a feature.</p>
 
 			<div class="note">
 				<p>EPUB publications typically contain more information than there are support requirements for in this
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -224,7 +224,7 @@
 
 			<p>Whether a reading system has to support a feature is mentioned at the beginning of its section. To be
 				conformant with this specification, reading systems MUST support all required features as well as all
-				applicable conditionally-required features (e.g., to support image rendering if the reading system has a
+				applicable conditionally required features (e.g., to support image rendering if the reading system has a
 				[=viewport=]).</p>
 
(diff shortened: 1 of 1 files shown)


epub33/rs/index.html
```diff
@@ -219,6 +219,25 @@ <h4>Relationship to SVG</h4>
 				</section>
 			</section>
 		</section>
+		<section id="sec-rs-conformance">
+			<h3>Reading system conformance</h3>
+
+			<p>Whether a reading system has to support a feature is mentioned at the beginning of its section. To be
+				conformant with this specification, reading systems MUST support all required features as well as all
+				applicable conditionally-required features (e.g., to support image rendering if the reading system has a
```
dlazin, line 227: Nit: no hyphen with adverbs (generally -ly), only compound adjectives.
https://www.merriam-webster.com/words-at-play/6-common-hypercorrections-and-how-to-avoid-them (not the first section; ctrl-f for "-ly" and you'll see it, which quotes Chicago, but Chicago requires a subscription so I can't link there directly).

epub33/rs/index.html
```diff
@@ -219,6 +219,25 @@ <h4>Relationship to SVG</h4>
...
+				[=viewport=]).</p>
+
+			<p>When support for a feature is only recommended, reading system developers can typically ignore all the
+				normative statements in its section. In some cases, however, there may be alternative requirements when
+				not implementing a feature (e.g., to <a href="#confreq-rs-scripted-flbk">process fallbacks</a> when
+				scripting is not supported).</p>
+
+			<div class="note">
+				<p>EPUB publications typically contain more information than there are support requirements for in this
```
```suggestion
				<p>EPUB publications frequently contain information not required by this
```

dlazin (approved):
I like this; thanks.

epub33/rs/index.html
mattgarrish: Lol, ya, I even called that out in review comments here: https://github.com/w3c/epub-specs/issues/2181#issuecomment-1085187900

I always fall into the trap.

mattgarrish, line 227:
```suggestion
				applicable conditionally required features (e.g., to support image rendering if the reading system has a
```

epub33/rs/index.html
```diff
@@ -219,6 +219,25 @@ <h4>Relationship to SVG</h4>
...
+		<section id="sec-rs-conformance">
+			<h3>Reading system conformance</h3>
+
+			<p>Whether a reading system has to support a feature is mentioned at the beginning of its section. To be
+				conformant with this specification, reading systems MUST support all required features as well as all
+				applicable conditionally required features (e.g., to support image rendering if the reading system has a
+				[=viewport=]).</p>
+
+			<p>When support for a feature is only recommended, reading system developers can typically ignore all the
+				normative statements in its section. In some cases, however, there may be alternative requirements when
```
iherman: It may be only a nuance, but I could see Reading Systems interpreting this by claiming implementing, e.g., scripting or MO, but ignoring a bunch of normative statements. We should say that a claim of implementing a feature means implementing _all_ MUST statements.

iherman (approved):
Maybe it is a nuance that is not necessary to mention, but I raise it anyway...

I

epub33/rs/index.html
mattgarrish: Good point. I've committed some changes to try and tighten up the wording. I've added an additional paragraph to separate supporting recommended/optional features from not supporting, and for not supporting made it clearer that there may still be requirements that must be met. Let me know if this works better.

iherman: Thanks, I will mark this as resolved now.

PR #2307, opened 2022-05-20
Remove obsolete editor's note

### LINKED ISSUE #2306: Remove dated ed note
I spotted this editor's note still floating around in the RS scripting section:

> Note that the definition of epubReadingSystem is currently marked as "at risk". If, in the final version of this document, the object becomes non-normative, then each "MUST" statement in the last bullet item would become a "MAY".

https://www.w3.org/TR/epub-rs-33/#issue-container-generatedID-10

Did we ever actually mark it "at risk"? In any case, I assume it's more likely to become under-implemented and the requirements left as they are, or are we going to re-raise the issue of deprecating it?

We don't have to solve that now, but my inclination here is to delete this note and figure out what we do about RSO down the line when we get testing results.

mattgarrish:
Fixes #2306

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1267,10 +1267,6 @@
 					</li>
 				</ul>
 
-				<p class="ednote">Note that the definition of <code>epubReadingSystem</code> is currently marked as "at
-					risk". If, in the final version of this document, the object becomes non-normative, then each "MUST"
-					statement in the last bullet item would become a "MAY".</p>
-
 				<p id="confreq-rs-scripted-flbk">If a reading system does not support scripting, it MUST process
 					fallbacks for scripted content as defined in <a data-cite="epub-33#confreq-cd-scripted-flbk"
 						>Fallbacks for scripted content documents</a> [[epub-33]].</p>


dlazin (commented):
I would listen to Ivan, not me, but my belief is that Ivan expected that we would determine whether this is at risk as a result of running the test reports. Since those haven't been run yet, I don't think we know whether this will be marked as at risk or not (but based on WG discussion, it sounds likely to not be implemented in 2 RSes, so I suspect the result will be marking this at risk).

In summary, I vote for not removing this yet, but whatever Ivan decides is fine.

mattgarrish:
What confuses me about the note is that we haven't defined an at-risk category. I assume that's now synonymous with under-implemented, but then we don't change requirements - authors and developers are just aware from the label that support isn't good. At a minimum, it needs a rewrite.

We also haven't preemptively warned people that a feature may be under-implemented anywhere else.

There's also the problem of marking features under-implemented but potentially also having conflicting support requirements (the RS spec saying you MUST support RSO but the definition of under-implemented saying it's only recommended). We're going to have to solve this whenever we figure out what is getting tagged under-implemented, so I'm not sure this note is very helpful.

But as RSO support is @iherman's personal pet peeve 😉, I'm not going to merge this until he's back -- just wrapping up what I can before the long weekend here...

iherman (approved):
I agree with removing this note. Not only we will use the tests and there is nothing really "at-risk" at this point, but even in practice this is not the case. I have created a series of tests for these and, low and behold, those tests are (mostly) passed by a number of implementations (I presume I tested them on Thorium and Mac's Books). I seem to remember that there is only one, somewhat convoluted test that does not pass.

PR #2309, opened 2022-05-24
Updated accessModeSufficient technique for single values

### LINKED ISSUE #2303: Media overlays in accessMode declarations
As I mentioned in a [mailing list comment](https://lists.w3.org/Archives/Public/public-epub-wg/2022May/0024.html):

> EPUB’s media overlays feature means you could have a sufficient auditory pathway. Media overlays are not how the information is encoded or required to be played by default, so I wouldn’t expect the publisher to list auditory as an access mode based solely on their presence. But they do provide you an auditory playback option provided the complete text is narrated (i.e., they’re more like an affordance).

Full text books with media overlays are not audiobooks as we know them, so saying there is an audio requirement is misleading to me. I would only list auditory as an accessMode when you have to turn on media overlays to get the full information of the book (i.e., when they're more like DAISY audio books with minimal text for chapter navigation).

Whatever we ultimately resolve here should also be noted in the techniques (and other places like the DAISY KB).

/cc @madeleinerothberg

mattgarrish:
I've updated the accessModeSufficient technique in this pull request to emphasize that the single-value AMS sets are the most important to list. I've also noted that you don't just list all the access modes as AMS single values.

Since AMS is only a recommended property, I've strongly recommended the single values and said that you may also list any sets of multiple values (but this will typically only be a rehash of the access modes). Is this acceptable, or should we strongly recommend single values and "regularly" recommend the multi-value sets?

/cc @madeleinerothberg

Fixes #2303 

- [Preview](https://raw.githack.com/w3c/epub-specs/a11y/issue-2302/epub33/a11y-tech/index.html)
- [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/a11y/issue-2302/epub33/a11y-tech/index.html)

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -222,10 +222,10 @@
 				<p>The access modes sufficient to consume an <a href="https://www.w3.org/TR/epub/#dfn-epub-publication"
 						>EPUB publication</a> express a broader picture of the potential usability than do the <a
 						href="#meta-001">basic access modes</a>. Where the basic access modes identify the default
-					nature of the media used in the publication, sufficient access modes identify which modes, or sets
-					of modes, a user requires to read the publication. Sufficient access modes account for the
-					affordances and adaptations that have been provided, allowing a user to determine whether the
-					content will be usable regardless of its default nature.</p>
+					nature of the media used in the publication, sufficient access modes identify all the individual
+					modes, and sets of modes, that allow a user to read a publication. Sufficient access modes account
+					for the affordances and adaptations that the EPUB creators has provided, allowing users to determine
+					whether they can the read content regardless of its default nature.</p>
 
 				<p>Sufficient access modes are identified in the [[schema-org]] <a
 						href="https://schema.org/accessModeSufficient"><code>accessModeSufficient</code> property</a>.
@@ -234,7 +234,7 @@
 				<p>For example, consider an EPUB publication that contains graphics and charts, as well as descriptions
 					for all these images. The publication has both textual and visual content, so the <a
 						href="https://www.w3.org/TR/epub/#dfn-epub-creator">EPUB creator</a> will include the following
-					metadata entries to indicate this:</p>
+						<code>schema:accessMode</code> metadata entries to indicate this:</p>
 
 				<pre>&lt;meta
     property="schema:accessMode">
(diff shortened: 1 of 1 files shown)


clapierre (approved):
Love the explanations Matt, especially the note about adaptations are not required to be included in the accessMode list, that makes a lot of sense.

madeleinerothberg:
This is great -- thank you, Matt. I'd be interested in Chris and Chris' opinions as we were talking through this issue recently. I will email them because I can't seem to at them here.

Two minor wording suggestions:
End of first paragraph, "that the EPUB creators has provided" should be "have"

Above the example before the first note, "For example, the metadata the EPUB creator inputs for the example publication re-establishes that there is a textual and visual requirement:" 
I don't think "requirement" is the right word here -- this example has been described as being fully accessible with text. Perhaps "affordance"? "Option"? "Reading mode"?

mattgarrish:
> I can't seem to at them here.

If they have github accounts you can manually type their IDs in, but github won't help you out with autocomplete (I have to do that for you). That feature seems limited to WG members and W3C staff; probably to save the processing cost of checking all user names.

mattgarrish:
> Two minor wording suggestions:

Thanks, I've correct both now. For the second, I merged a couple of your suggestions to get "reading option".

iherman:
> That feature seems limited to WG members and W3C staff; probably to save the processing cost of checking all user names.

To be more precise: the feature limits to whoever is listed (directly or indirectly via groups) as "collaborator and group" in the repository's settings. @madeleinerothberg I have no problem adding you manually if you want, and any other individuals who you think are active in the A11y work of this WG.

madeleinerothberg:
>  I have no problem adding you manually if you want, and any other individuals who you think are active in the A11y work of this WG.

Thanks, Ivan. It would save Matt some typing :-) I don't know the Chris' github account names.

GeorgeKerscher:
This all looks great!

iherman:
Should work now, @madeleinerothberg

mattgarrish:
Let's merge this so everyone has a chance to see the new fixes. It'll help with other discussions we're having around how to improve understanding.

PR #2310, opened 2022-05-24
Add technique for media overlays and access modes

### LINKED ISSUE #2303: Media overlays in accessMode declarations

mattgarrish:
I've added a new technique to discuss the two cases for media overlays I'm aware of -- full text with full audio and structured audio. For full text with audio, the recommendation is to not set an auditory access mode and instead indicate the synchronizedAudioText feature. For structured audio, you do the opposite and set an auditory access mode and don't set synchronizedAudioText.

Let me know what you all think of this guidance.

Fixes #2303

- [Preview](https://raw.githack.com/w3c/epub-specs/a11y/issue-2303/epub33/a11y-tech/index.html)
- [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/a11y/issue-2303/epub33/a11y-tech/index.html)

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -357,8 +357,7 @@
 						would only declare the following sufficient access modes: "<code>textual</code>",
 							"<code>auditory</code>, and "<code>textual,visual</code>".</p>
 
-					<p>It would <strong>not</strong> declare: "<code>textual</code>", "<code>auditory</code>",
-							"<code>textual,auditory</code>", "<code>textual,visual</code>", and
+					<p>It would <strong>not</strong> declare "<code>textual,auditory</code>" or
 							"<code>textual,visual,auditory</code>".</p>
 				</div>
 
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -317,6 +317,80 @@
 				</div>
 			</section>
 
+			<section id="meta-access-modes-sync-audio">
+				<h3>Setting access modes for synchronized text-audio</h3>
+
+				<p>Setting the correct access modes and sufficient access modes for EPUB 3 publications that contain
+					synchronized text-audio playback requires evaluating whether playback is essential to reading the
+					publication or an additional feature.</p>
+
+				<p>EPUB 3's <a href="https://www.w3.org/TR/epub/#sec-media-overlays">media overlays</a> [[EPUB-3]]
+					allows EPUB creators to synchronize the full text of a publication with full audio narration. These
+					types of publications are commonly referred to as "read aloud" books, as the user chooses whether or
+					not to turn on the narration (unlike traditional audiobooks where only the audio is available).</p>
+
+				<p>In this case, because the audio playback is an extra feature, EPUB creators would
+						<strong>not</strong> list "<code>auditory</code>" as an <a href="#meta-001">access mode</a>.
+					Rather, they would indicate the presence of text and audio synchronization as an <a href="#meta-003"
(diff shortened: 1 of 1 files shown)


clapierre (commented):
These two statements seem to conflict with each other.

				<p>For example, a publication with media overlays that has text and images (with text alternatives)
						would only declare the following sufficient access modes: "<code>textual</code>",
							"<code>auditory</code>, and "<code>textual,visual</code>".</p>

					<p>It would <strong>not</strong> declare: "<code>textual</code>", "<code>auditory</code>",
							"<code>textual,auditory</code>", "<code>textual,visual</code>", and
							"<code>textual,visual,auditory</code>".</p>

Shouldn't it only have "NOT" "<code>textual,visual,auditory</code>" ?

mattgarrish:
It would also omit "textual,auditory" since that implies switching back and forth between the two.

But I agree it might be simpler to list the two that are not declared than show both complete sets.

clapierre (approved):

madeleinerothberg:
Agree it would be clearer to only show the omitted items in the NOT list.

>It would also omit "textual,auditory" since that implies switching back and forth between the two.

I don't fully understand media overlays. Is it not possible to listen while looking at the text? In that case, yes, that combo would also be omitted.

A separate comment: the text says
>Likewise, since the text in the publication only provides heading navigation, EPUB creators will not list a textual access mode. The user does not need to, and is not expected to, visually read this text.

In this case, it isn't only that the user doesn't need to or isn't expected to read the text (assuming they can listen straight through with pause/play, if they can't read the TOC structure). They actually **can't** visually read the text because it isn't there.

mattgarrish:
> Is it not possible to listen while looking at the text? In that case, yes, that combo would also be omitted.

It's up to the user. If they want to read along with the narration, the text will be highlighted. But you can also use media overlays to switch back and forth between visual and auditory reading. The typically example is listening to audio while commuting but reading visually elsewhere. So in that sense there is a pathway that is both textual and auditory, but I don't think it's a case we need to highlight.

> They actually **can't** visually read the text because it isn't there.

There always has to be some text to synchronize with for media overlays. In the structured navigation case, you'd just get a list of all the headings in the viewport, with each one being highlighted in turn (assuming your device has a viewport). So there is technically _some_ of the textual content of the book there, but it's entirely useless as far as visual reading goes.

mattgarrish:
This one's also been open for a couple of weeks now without any new change requests. I'm going to merge it since it's only a technique and we can easily change it if new issues come up.

PR #2326, opened 2022-06-06
Split change logs to identify cr changes

mattgarrish:
I'll integrate this into the privacy+security PR after, so this way all new branches have the split sections.

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jun 7, 2022, 12:02 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Fca35a4598745b73bb297c6a293edd6442a56aa5b%2Fepub33%2Fcore%2Findex.html%3FisPreview%3Dtrue)

```

Specification: http://labs.w3.org/spec-generator/uploads/Gnyvmf?isPreview=true&publishDate=2022-06-07
ReSpec version: 32.1.8
Error: Error: Evaluation failed: Timeout: document.respec.ready didn't resolve in 27713ms.

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/epub-specs%232326.)._
</details>

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -1853,8 +1853,8 @@
 					>working group's issue tracker</a>.</p>
 
 			<section id="change-latest">
-				<h3>Substantive changes since the <a href="https://www.w3.org/TR/2022/CR-epub-a11y-11-20220512/"
-						>previous Candidate Recommendation Snapshot</a></h3>
+				<h3>Substantive changes since <a href="https://www.w3.org/TR/2022/CR-epub-a11y-11-20220512/">Candidate
+						Recommendation</a></h3>
 
 				<ul>
 					<li>17-May-2022: Added an index of terms. See <a
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -1852,74 +1852,88 @@
 					href="https://github.com/w3c/epub-specs/issues?q=is%3Aissue+is%3Aclosed+sort%3Aupdated-desc+label%3ASpec-Accessibility+label%3AAccessibility11"
 					>working group's issue tracker</a>.</p>
 
-			<ul>
-				<li>17-May-2022: Added an index of terms. See <a href="https://github.com/w3c/epub-specs/issues/2260"
-						>issue 2260</a>.</li>
-				<li>12-Apr-2022: Restored recommendation to include links to all reproduced pages in the page list and
-					added a requirement to link to all page break markers to address concerns with the previous change.
-					Clarified that meeting the page navigation requirements is required, but adding page navigation is
-					only recommended in some circumstances. See <a href="https://github.com/w3c/epub-specs/pull/2246"
-						>pull request 2246</a>.</li>
-				<li>12-Apr-2022: Generalized discussion of synchronized text-audio playback objectives and moved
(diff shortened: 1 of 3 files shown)


epub33/a11y/index.html
```diff
@@ -1852,74 +1852,88 @@ <h2>Change log</h2>
...
-				<li>16-Nov-2020: References to the optional accessibility control and API metadata have been removed
-					from the discoverability section. These metadata properties are more applicable to reading systems.
-					See <a href="https://github.com/w3c/epub-specs/issues/1327">issue 1327</a>.</li>
-				<li>26-Sept-2020: The revisions made to the Accessibility 1.0 specification as part of publishing it as
-					an ISO standard have been incorporated into the initial draft text.</li>
-			</ul>
+			<section id="change-latest">
+				<h3>Substantive changes since the <a href="https://www.w3.org/TR/2022/CR-epub-a11y-11-20220512/"
+						>previous Candidate Recommendation Snapshot</a></h3>
+
```
iherman, line 1858: Is the idea that if I we have a new snapshot published, then these items are manually moved down to the general list? Or that we'd have a subsection for every CR snapshot? The latter might be an overkill, the former is a bit of a pain at every snapshot publication...

iherman (approved):

epub33/a11y/index.html
mattgarrish: The former. We migrate the old changes down to the complete list after the snapshot. Same process we were using before we went to publishing automatic working drafts.

The only headache with the new publishing system is we can't prep the change log immediately after publishing the snapshot because it would be pushed out right away. It has to happen with the first substantive change post-snapshot.

mattgarrish: Actually, I was just looking at the github documentation and you can temporarily disable the workflows without having to remove them: https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow

We could turn off publication and push the updated change log after the snapshot, which would make life a lot simpler.

iherman: We can try... next time. For the time being, per the comment of PlH, we may not have a snapshot to worry about...

mattgarrish: Well, ya, I think this is useful to get in now regardless so we have a clear record of changes since we went to CR. Even if we don't publish another CR snapshot, it's nice to see what's happened since the transition occurred. If we do publish another snapshot, we can fix the lists then. It's not that complicated.

mattgarrish: One thing we could do, too, is change the title of the section to "Substantive changes since Candidate Recommendation" (with a link to the first snapshot) and perhaps not worry about every individual snapshot.

iherman: > Even if we don't publish another CR snapshot, it's nice to see what's happened since the transition occurred.

It may also become handy when we move to PR...

PR #2328, opened 2022-06-06
Explain bypass blocks are not necessary

### LINKED ISSUE #2319: Mention bypass blocks don't apply to epub
I thought we already mentioned this in the techniques, but it appears not. In any case, EPUBs don't have the trappings of web pages, so there aren't repeated blocks to bypass when reading documents. Adding a "main" element/role is consequently superfluous, and as far as I remember Ace already suppresses any warnings omitting one would generate.

mattgarrish:
Clarify that this success criterion is for bypassing repetitive web site content. In theory if you had an epub that reproduced a set of pages from a web site, wrapper and all, then you'd need bypass blocks, but I can't say I've seen that before. Closest is when we reproduced the set of EPUB specifications as an EPUB, but even then there weren't any repetitive IDPF site blocks to bypass. At any rate, I noted it's a possibility, if a tiny one.

Fixes #2319 

- [Preview](https://raw.githack.com/w3c/epub-specs/a11y/issue-2319/epub33/a11y-tech/index.html)
- [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/a11y/issue-2319/epub33/a11y-tech/index.html)

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -839,6 +839,33 @@
 						or in the spine). EPUB creators can include links to these additional navigation lists in the
 						table of contents.</p>
 				</section>
+
+				<section id="access-bypass-blocks">
+					<h4>Bypass blocks not necessary</h4>
+
+					<p>Web sites are constructed very differently from EPUB publications. A typical web site wraps the
+						content of each page within a repeating template, for example. This template gives each page a
+						consistent look and feel, but users are rarely interested in the wrapper content after visiting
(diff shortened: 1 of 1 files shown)


clapierre (approved):
Looks great!

avneeshsingh (approved):

PR #2297, opened 2022-05-18
Privacy and security changes

mattgarrish:
This pull request comes from a response from @iherman to the email from @npdoty recorded in the chairs mailing list. His suggestion was to potentially make the privacy and security sections normative but ensure that they are only recommendations (almost all were).

I've gone through the three REC track documents and fixed them so that all applicable recommendations are now normative.

We will probably want to discuss this with the WG on a call before making this change, though.

The PR also fixes #2265 by noting under the malicious content heading that because we don't have a standard means of signing publications there isn't always a way to verify the integrity. I've also added a pargraph to the end of the recommendations to say that reading systems should treat unsigned content loaded by the user as untrustworthy. Improvements to the text welcome.

(If we decide not to incorporate the first part, I'll split the second out, but they're tightly bound to each other until we know which direction we're taking.)

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jun 7, 2022, 12:46 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2F31b6584ae7edce3d941a9d88e019bcf757ea9db5%2Fepub33%2Fcore%2Findex.html%3FisPreview%3Dtrue)

```

😭  Sorry, there was an error generating the HTML. Please report this issue!
Specification: http://labs.w3.org/spec-generator/uploads/a51ive?isPreview=true&publishDate=2022-06-07
ReSpec version: 32.1.8
File a bug: https://github.com/w3c/respec/
Error: Error: Evaluation failed: Timeout: document.respec.ready didn't resolve in 27539ms.
    at ExecutionContext._evaluateInternal (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:221:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:110:16)
[...]

PR #2329, opened 2022-06-07
Disallow file URLs

### LINKED ISSUE #2324: Disallowing file URLs
(This is a spin-off of a discussion in #2266, also raised during the discussion with @GJFR at the [EPUB meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-06-03-epub); raising it as a separate issue for a better tracking)

At the moment, the only reference to file: URLs in the spec is a _SHOULD NOT_ for the `href` attribute in the package file. These urls are obvious security issues for content documents that are supposed to be 'installed' in various places, and there is no reason to use them. The proposal is to explicitly disallow their usage in EPUB.

iherman:
This is to fix #2324

I was wondering, however, whether disallowing File URLs in general is implementable by an RS; it may require the RS to patch the fetch library or something similar if a WebView is used for implementing the system. I would welcome the reaction of @bduga or @danielweck on this. (Note that a similar question may arise for Data URLs.)

See:

* For EPUB 3.3:
    * [Preview](https://raw.githack.com/w3c/epub-specs/disallow-file-scheme/epub33/core/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/core/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/disallow-file-scheme/epub33/core/index.html)
    * [Preview](https://raw.githack.com/w3c/epub-specs/disallow-file-scheme/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/disallow-file-scheme/epub33/rs/index.html)

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jun 8, 2022, 9:45 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2F64670ea5fbf01ce5109b7d44862e609b193c5c26%2Fepub33%2Fcore%2Findex.html%3FisPreview%3Dtrue)

```

Specification: http://labs.w3.org/spec-generator/uploads/rIFmii?isPreview=true&publishDate=2022-06-08
ReSpec version: 32.1.8
Error: Error: Evaluation failed: Timeout: document.respec.ready didn't resolve in 27558ms.

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/epub-specs%232329.)._
</details>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3206,8 +3206,7 @@
 				<section id="attrdef-href">
 					<h4>The <code>href</code> attribute</h4>
 
-					<p>A [=valid URL string=] [[url]] that references a resource. If the value is an <a>absolute-URL
-							string</a>, it MUST NOT use the "file" URI scheme [[rfc8089]].</p>
+					<p>A [=valid URL string=] [[url]] that references a resource.</p>
 
 					<aside class="example" title="Linking a metadata record">
 						<pre>&lt;package …>
@@ -11536,6 +11535,8 @@ EPUB/images/cover.png</pre>
 						Recommendation</a></h3>
 
 				<ul>
+					<li>07-06-2022: The usage of File URLs has been disallowed. See <a
+						href="https://github.com/w3c/epub-specs/issues/2263">issue 2324</a>.</li>
 					<li>27-May-2022: Added recommendation to only reference remote resources via https. See <a
 							href="https://github.com/w3c/epub-specs/issues/2263">issue 2263</a>.</li>
 					<li>20-May-2022: Add recommendation not to store sensitive user data in persistent storage, and to
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1502,6 +1502,19 @@
 						restrictions</a>).</p>
 			</section>
 
+			<section id="sec-file-urls">
+				<h3>File URLs</h3>
+				<p>
+					The <a data-cite="rfc8089#"><code>file:</code> URL scheme</a> is defined in [[rfc8089]] as "identifying an 
+					object (a 'file') stored in a structured object naming and accessing environment on a host (a 'file system')."
+					It is typically used to retrieve files from within one's own computer.
+				</p>
+				<p>
+					Using a file URL in an [=EPUB publication=], which can be transferred among different hosts, represents a security risk.
+					As a consequence,  [=EPUB creators=] MUST NOT use file URLs in EPUB publications.
+				</p>
+			</section>
+
 			<section id="sec-xml-constraints">
 				<h3>XML conformance</h3>
 
@@ -3194,7 +3207,7 @@
 					<h4>The <code>href</code> attribute</h4>
 
 					<p>A [=valid URL string=] [[url]] that references a resource. If the value is an <a>absolute-URL
-							string</a>, it SHOULD NOT use the "file" URI scheme [[rfc8089]].</p>
+							string</a>, it MUST NOT use the "file" URI scheme [[rfc8089]].</p>
 
 					<aside class="example" title="Linking a metadata record">
 						<pre>&lt;package …>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1510,7 +1510,7 @@
 					It is typically used to retrieve files from within one's own computer.
 				</p>
 				<p>
-					Using a file URL in an [=EPUB publication=], which can be transferred among different hosts, represents a security risk.
+					Using a file URL in an [=EPUB publication=], which can be transferred among different hosts, represents a security risk and is also non-interoperable.
 					As a consequence,  [=EPUB creators=] MUST NOT use file URLs in EPUB publications.
 				</p>
 			</section>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -302,7 +302,7 @@
 			<section id="sec-epub-rs-conf-file-urls">
 				<h4>File URLs</h4>
 				<p id="confreq-rs-file-urls">
-					Reading systems MUST prevent file URLs [[rfc8089]] from opening.
+					Reading Systems MUST prevent access to resources referenced via file URLs [[rfc8089]], both from elements that refer to resources and from script API calls. 
 				</p>
 			</section>
 
@@ -2575,6 +2575,8 @@ alert("Feature " + feature + " supported?: " + conformTest);</pre>
 						Recommendation</a></h3>
 
 				<ul>
+					<li>07-06-2022: The usage of File URLs has been disallowed. See <a
+						href="https://github.com/w3c/epub-specs/issues/2263">issue 2324</a>.</li>
 					<li>27-May-2022: Added reading system conformance section. See <a
 							href="https://github.com/w3c/epub-specs/issues/2271">issue 2271</a>.</li>
 					<li>27-May-2022: Added recommendation to only load remote resources referenced via https. See <a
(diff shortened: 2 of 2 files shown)


iherman:
cc @GJFR

OriKuttner:
‏Yes, it is implementable as the RS decides on each file to load. most RS's
do not simply hand files to WebView or other loaders, they first get the
file name from the OPF file and then loads it.

-- 
Ori Idan CEO Helicon Books
http://www.heliconbooks.com

iherman:
> Yes, it is implementable as the RS decides on each file to load. most RS's do not simply hand files to WebView or other loaders, they first get the file name from the OPF file and then loads it.

And what happens if a runtime script tries to use a file URL?

OriKuttner:
This is a little more complicated, this requires the RS to make a hook on
file opening, but it is still possible.

iherman:
@mattgarrish I have not added the entry to the list of changes yet. I would prefer to do that before merge but after #2326 will have been merged (and maybe also #2297), to avoid nasty merge conflicts...

mattgarrish:
I'll merge main into this to get the new change log structure added and then if you add an entry it shouldn't be any worse than integrating any others.

epub33/core/index.html
```diff
@@ -3194,7 +3207,7 @@ <h4>The <code>dir</code> attribute</h4>
 					<h4>The <code>href</code> attribute</h4>
 
 					<p>A [=valid URL string=] [[url]] that references a resource. If the value is an <a>absolute-URL
-							string</a>, it SHOULD NOT use the "file" URI scheme [[rfc8089]].</p>
+							string</a>, it MUST NOT use the "file" URI scheme [[rfc8089]].</p>
```
mattgarrish: I'd delete this requirement since it's now redundant, or perhaps make it an informative reference back to the new section.

epub33/rs/index.html
```diff
@@ -299,6 +299,13 @@ <h4>Data URLs</h4>
 					content documents.</p>
 			</section>
 
+			<section id="sec-epub-rs-conf-file-urls">
+				<h4>File URLs</h4>
+				<p id="confreq-rs-file-urls">
+					Reading systems MUST prevent file URLs [[rfc8089]] from opening.
+				</p>
```
mattgarrish, line 306: I wonder if "opening" is the right term. That seems to cover a scenario like trying to load a file reference into an iframe, but is a script that uses fetch or xhr to get the data of the file "opening" the file?

What about something like: "Reading Systems MUST prevent access to resources referenced from file URLs [[rfc8089]], both from elements that embed resources and from script API calls." ?

epub33/core/index.html
iherman: You're right. I have simply removed it.

epub33/rs/index.html
iherman: The term "embed" does not sound right to my ears, it seems to say that something like `<a href="file:...">` is allowed. 

I have used:

> Reading Systems MUST prevent access to resources referenced via file URLs [[rfc8089]], both from elements that refer to resources and from script API calls.

epub33/core/index.html
```diff
@@ -1502,6 +1502,19 @@ <h3>Data URLs</h3>
...
 
+			<section id="sec-file-urls">
+				<h3>File URLs</h3>
+				<p>
+					The <a data-cite="rfc8089#"><code>file:</code> URL scheme</a> is defined in [[rfc8089]] as "identifying an 
+					object (a 'file') stored in a structured object naming and accessing environment on a host (a 'file system')."
+					It is typically used to retrieve files from within one's own computer.
+				</p>
+				<p>
+					Using a file URL in an [=EPUB publication=], which can be transferred among different hosts, represents a security risk.
```
bduga: It's also just plain non-interoperable. What file could possibly exist on all reading systems with the exact same path and name?

epub33/rs/index.html
bduga: I don't like using both here, since it implies those are the only two cases. I would drop the list, which seems fine. If we think we need to clarify when and how file URLs are used, we should list this as potential examples. But it feels like "Reading Systems MUST prevent access to resources referenced via file URLs [[rfc8089]]" full stop is exactly enough, and tacking on more just introduces more surfaces for confusion.

bduga (commented):
I am fine with the restriction, it seems implementable to me.

epub33/rs/index.html
mattgarrish: That's fine with me, or even just putting the examples in a parenthetical "e.g." to make clear its not a complete list.

It was more a reaction to the limits that "opening" resources has, but is not strictly necessary.

iherman: Nuked the list :-)

epub33/core/index.html
iherman: Added that, too. I leave the remark on security, it does not harm...

mattgarrish (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-06-09-epub#section1) on 2022-06-09

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-06-09-epub#resolution1):  Remove file URLs, merge PR 2329 and close issue 2324.***

### 1. Disallowing File URLs.

_See github issue [epub-specs#2324](https://github.com/w3c/epub-specs/issues/2324)._




_See github pull request [epub-specs#2329](https://github.com/w3c/epub-specs/pull/2329)._




**Dave Cramer:** I can't think of good reason to have these, but a lot of good reasons to prohibit them.  

**Brady Duga:** yes, sounds good. File URLs seem to be interoperable. What file would you load from an epub?.  

**Ben Schroeter:** most common thing i've seen is youtube videos, but those aren't file URLs, right?.  

**Brady Duga:** depends on how the epub is created. Could be a link to youtube, or link to external resource that plays in your epub, but neither of those are file URLs.  

**Dave Cramer:** file URL goes against idea that epub should be self contained, you don't want epub author to look at your files in your local machine.  

**Ben Schroeter:** when Play gets file URL what happens?.  

**Brady Duga:** probably gets stripped on the server, but probably gets intercepted and rejected. We might try to open it in the browser, but then the browser would probably reject it.  
… there are also a lot of origin issues with file URLs.  

**Dave Cramer:** I propose we forbid file URLs.  

**Brady Duga:** there's already a PR open for that.  


> **Proposed resolution: Remove file URLs, merge PR 2329 and close issue 2324.** *(Wendy Reid)*

> *Ben Schroeter:* +1.

> *Dave Cramer:* +1.

> *Brady Duga:* +1.

> *Masakazu Kitahara:* +1.

> *Wendy Reid:* +1.

> *Matthew Chan:* +1.

> *Matt Garrish:* +1.

> *Ben Schroeter:* +1.

> *Toshiaki Koike:* +1.

> ***Resolution #1: Remove file URLs, merge PR 2329 and close issue 2324.***


</details>

PR #2332, opened 2022-06-10
Clarify that data blocks are exempt from fallbacks

### LINKED ISSUE #2331: Script data blocks and foreign resources
If we allow data files to travel in the epub container for use by scripts without having to meet CMT requirements, shouldn't we also allow foreign data in `script` tags?

It's not completely clear to me if it's allowed, as it depends on how you interpret "embedding" here:

> it is not embedded directly in EPUB content documents (e.g., via [html] [embedded content](https://html.spec.whatwg.org/multipage/dom.html#embedded-content-category) and [svg] [image](https://www.w3.org/TR/SVG/embedded.html#ImageElement) and [foreignObject](https://www.w3.org/TR/SVG/embedded.html#ForeignObjectElement) elements).

Since it's not a definitive list of cases, and the data is embedded in a content document when in `script` tags, that seems to make it illegal. But it doesn't make a lot of sense to allow in external files but not in `script` tags.

mattgarrish:
I've added a new section to the intrinsic fallbacks section to clarify that they're not needed for data blocks.

There are about three places this exemption could go, but this seemed the most logical. It could also go in a couple of places in  the exempt resources section, but even if we consider them resources they aren't for use by reading systems which breaks the definition.

Fixes #2331

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jun 13, 2022, 12:27 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2F613e6e2b905a265e805ffc64d0af31b46518cfba%2Fepub33%2Fcore%2Findex.html%3FisPreview%3Dtrue)

```

Specification: http://labs.w3.org/spec-generator/uploads/fOB39g?isPreview=true&publishDate=2022-06-13
ReSpec version: 32.1.8
Error: Error: Evaluation failed: Timeout: document.respec.ready didn't resolve in 27316ms.

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/epub-specs%232332.)._
</details>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1268,6 +1268,10 @@
 					resources (i.e., the requirement for a fallback for the foreign content document covers any
 					rendering issues within it). As the resource is not referenced from an EPUB content document, it
 					automatically becomes exempt from fallbacks.</p>
+
+				<p>Although technically not separate resources from the XHTML and SVG content documents that embed them,
+						<a data-cite="html#data-block">data blocks</a> [[HTML]] expressed in [=script=] elements are
+					also considered exempt resources.</p>
 			</section>
 
 			<section id="sec-resource-fallbacks">
@@ -1504,15 +1508,12 @@
 
 			<section id="sec-file-urls">
 				<h3>File URLs</h3>
-				<p>
-					The <a data-cite="rfc8089#"><code>file:</code> URL scheme</a> is defined in [[rfc8089]] as "identifying an 
-					object (a 'file') stored in a structured object naming and accessing environment on a host (a 'file system')."
-					It is typically used to retrieve files from within one's own computer.
-				</p>
-				<p>
-					Using a file URL in an [=EPUB publication=], which can be transferred among different hosts, represents a security risk and is also non-interoperable.
-					As a consequence,  [=EPUB creators=] MUST NOT use file URLs in EPUB publications.
-				</p>
+				<p> The <a data-cite="rfc8089#"><code>file:</code> URL scheme</a> is defined in [[rfc8089]] as
+					"identifying an object (a 'file') stored in a structured object naming and accessing environment on
+					a host (a 'file system')." It is typically used to retrieve files from within one's own computer. </p>
+				<p> Using a file URL in an [=EPUB publication=], which can be transferred among different hosts,
+					represents a security risk and is also non-interoperable. As a consequence, [=EPUB creators=] MUST
+					NOT use file URLs in EPUB publications. </p>
 			</section>
 
 			<section id="sec-xml-constraints">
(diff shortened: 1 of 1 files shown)


iherman:
See:

* For EPUB 3.3:
    * [Preview](https://raw.githack.com/w3c/epub-specs/fix/issue-2331/epub33/core/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/core/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fix/issue-2331/epub33/core/index.html)

epub33/core/index.html
```diff
@@ -1382,6 +1382,25 @@ <h5>HTML <code>img</code> fallbacks</h5>
...
+						<div class="note">
+							<p>The exemption from fallbacks aligns data blocks with the exemption made to <a
+									href="confreq-foreign-no-fallback">allow data files</a> to travel in the [=EPUB
+								container=]. As data blocks are not separate resources from their host [=EPUB content
+								document=], they do not fall under the same exemption.</p>
+						</div>
+						<div class="note">
+							<p>Since the [[SVG]] <span data-cite="svg">[^script^] element</span> is an implementation of
+								the HTML element, the same exemption applies to both formats.</p>
+						</div>
```
iherman: Tiny comment. In HTML, the `href` attribute in a `script` element used for datablock is ignored. Ie, a non-JS content is _always_ embedded. This explicit statement is missing in the SVG version of the `script` element's definition; it is therefore, in theory, possible to say something like

```xml
<script href="http://example.org/a.ttl" type="text/turtle"/>
```

I regard that as a bug in the SVG spec...

I think we can safely ignore this dark corner in the spec, I just want to do note this to ourselves...

iherman (approved):

epub33/core/index.html
mattgarrish: Ya, this is why I mention SVG `script` in a note rather than cite both elements as distinct. SVG doesn't mention data blocks at all, and seems to assume that all `script` elements contain some kind of scripting and only to not execute ones with unknown media types.

It just jogged my memory that we ran into this earlier while defining scripted content documents. In that case, we put in this note:

> [[svg](https://w3c.github.io/epub-specs/epub33/core/#bib-svg)] does not define data blocks as of publication, but the same exclusion would apply if a future update adds the concept.

https://w3c.github.io/epub-specs/epub33/core/#sec-scripted-support

Maybe for consistency we should use the same note here.

iherman: Your memory is bottomless... 😀

Yes, the same remark should apply here.

shiestyle (approved):

PR #2335, opened 2022-06-17
Clarify that vocabulary association mechanisms support is optional

### LINKED ISSUE #2317: Reading system requirement to resolve property datatype values
We've been having a [discussion in the test repository](https://github.com/w3c/epub-tests/issues/158#issuecomment-1143888901) about whether reading systems are required to process the compact URIs in attributes that accept a [property datatype value](https://w3c.github.io/epub-specs/epub33/core/#sec-property-datatype).

My understanding since we created this framework in 3.0 is that it was optional and the rules are only defining how to resolve them for reading systems that want full URIs for verification. All the metadata that use these values are either unprefixed or have reserved prefixes you're not allowed to redefine so that there's never any ambiguity that needs resolving. The need to resolve prefixes only affects custom metadata introduced by the author.

The specification is a bit ambiguous about the requirement to process these values, however. Nowhere do we say outright that you must resolve all property datatypes, but there is a [processing section](https://w3c.github.io/epub-specs/epub33/rs/#sec-vocab-assoc) that explains how you must process them. It says:

> Reading systems MUST expand [property values](https://www.w3.org/TR/epub-33/#sec-property-datatype) [[epub-33](https://w3c.github.io/epub-specs/epub33/rs/#bib-epub-33)] as follows:

My take on this is that it's more akin to: "If a reading system requires the full URI from a propery value, it MUST expand ..."

The requirement at the top of the processing section, however, says reading systems must support the vocabulary mechanisms. It is new to 3.3, though. If you go back in time, there isn't a requirement like that anywhere else.

But as I don't develop reading systems for a living, my understanding of the intent here may not match those who do. Do any reading systems support all the vocabulary mechanisms and expand all the property values, or is it more common to treat the values as tokens?

For example, do you take `properties="nav"` in a manifest declaration and expand the value to `http://idpf.org/epub/vocab/package/#nav` to determine the resource is the nav document?

If no one is expanding the values, we might want to consider lowering the support requirement to a "should" and clarifying that expanding property values is only required when a full URI is needed.

mattgarrish:
The primary change is to make section 11 optional.

There were a number of statements about the URLs to use to expand default vocab values sprinkled around the package document section, but this was a recent change as part of splitting out the reading system requirements. In the past, we had all this info in a default vocabularies section. See https://www.w3.org/publishing/epub3/epub-packages.html#sec-metadata-default-vocab

To fix these, I've taken all the statements and recombined them under section 11. There's no need for them to be with their attribute definitions when all of this processing is optional. Better to have all the information in one place for anyone who wants to implement the processing.

Fixes #2317 

EPUB Reading Systems 3.3:
- [Preview](https://raw.githack.com/w3c/epub-specs/fix/issue-2317/epub33/rs/index.html)
- [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fix/issue-2317/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2046,7 +2046,8 @@
 			<h2>Vocabulary association mechanisms</h2>
 
 			<p id="confreq-rs-vocab-assoc" class="support">Reading systems MAY support <a
-					data-cite="epub-33#sec-vocab-assoc">vocabulary association mechanisms</a> [[epub-33]] in [=EPUB content documents=].</p>
+					data-cite="epub-33#sec-vocab-assoc">vocabulary association mechanisms</a> [[epub-33]] in [=EPUB
+				content documents=].</p>
 
 			<dl class="conformance-list">
 				<dt id="sec-metadata-reserved-prefixes">Reserved Prefixes</dt>
@@ -2068,8 +2069,8 @@
 				<dd>
 					<p id="confreq-rs-vocab-prefix-attr">If the <code>prefix</code> attribute includes a declaration for
 						a <a href="#sec-metadata-reserved-prefixes">predefined prefix</a>, reading systems MUST use the
-						URL mapping defined in the <code>prefix</code> attribute, regardless of whether of it maps to
-						the same URL as the predefined prefix.</p>
+						URL mapping defined in the <code>prefix</code> attribute, regardless of whether it maps to the
+						same URL as the predefined prefix.</p>
 				</dd>
 
 				<dt id="sec-default-vocab">Default vocabularies</dt>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2045,9 +2045,8 @@
 		<section id="sec-vocab-assoc">
 			<h2>Vocabulary association mechanisms</h2>
 
-			<p id="confreq-rs-vocab-assoc" class="support">Reading systems support for <a
-					data-cite="epub-33#sec-vocab-assoc">vocabulary association mechanisms</a> [[epub-33]] is
-				OPTIONAL.</p>
+			<p id="confreq-rs-vocab-assoc" class="support">Reading systems MAY support <a
+					data-cite="epub-33#sec-vocab-assoc">vocabulary association mechanisms</a> [[epub-33]] in [=EPUB content documents=].</p>
 
 			<dl class="conformance-list">
 				<dt id="sec-metadata-reserved-prefixes">Reserved Prefixes</dt>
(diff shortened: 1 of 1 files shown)


epub33/rs/index.html
```diff
@@ -2066,8 +2045,9 @@ <h2>Processing structural semantics</h2>
 		<section id="sec-vocab-assoc">
 			<h2>Vocabulary association mechanisms</h2>
 
-			<p id="confreq-rs-vocab-assoc" class="support">Reading systems MUST support <a
-					data-cite="epub-33#sec-vocab-assoc">vocabulary association mechanisms</a> [[epub-33]].</p>
+			<p id="confreq-rs-vocab-assoc" class="support">Reading systems support for <a
+					data-cite="epub-33#sec-vocab-assoc">vocabulary association mechanisms</a> [[epub-33]] is
+				OPTIONAL.</p>
```
```suggestion
			<p id="confreq-rs-vocab-assoc" class="support">Reading systems MAY support <a
					data-cite="epub-33#sec-vocab-assoc">vocabulary association mechanisms</a> [[epub-33]] in [=EPUB content documents=].</p>
```

(Just to be consistent with similar notices elsewhere. Besides, I believe the current language was not correct.)

epub33/rs/index.html
```diff
@@ -2093,15 +2073,38 @@ <h2>Vocabulary association mechanisms</h2>
 						the same URL as the predefined prefix.</p>
 				</dd>
 
-				<dt id="sec-property-datatype">Expanding <code>property</code> Data Types</dt>
```
iherman, line 2096: The github review does not allow me to put comments to unchanged parts of the text; in the `<dd>` for prefixes:

whether of it maps to -> whether it maps to

iherman (approved):
Note that these changes, while fine, do not remove our obligation of adding some tests on the association mechanism, in line with similar situation (like MO). 

Cc @dlazin @wareid

mattgarrish:
> do not remove our obligation of adding some tests on the association mechanism

Responding to vocabulary values in the attributes, you mean? Processing the values as tokens shouldn't require invoking any of the vocabulary association mechanisms, as you're only responding to the value.

iherman:
> > do not remove our obligation of adding some tests on the association mechanism
> 
> Responding to vocabulary values in the attributes, you mean? Processing the values as tokens shouldn't require invoking any of the vocabulary association mechanisms, as you're only responding to the value.

*If* (and that is a big if) the RS is working based on URL-s (ie, sort of RDF way) then one can expect to behave similarly if a full URL is used and/or a different prefix is used with the same URL. From an RDF graph point of view this is the required behaviour. Of course, these tests would be relevant only if the RS does implement this voc. association mechanism, ie, for most of them the expected entry in the test suite will be "n/a". But we have to provide those tests nevertheless.

dauwhe (approved):

shiestyle (approved):

wareid (approved):

PR #2336, opened 2022-06-17
Update external link handling recommendations

### LINKED ISSUE #2320: Consider security implications of URI schemes
Placeholder to discuss this, as mentioned by Gertjen.

mattgarrish:
When I read over the external links section again, I didn't like what was there so I've made more changes than just adding a warning about opening external links.

Specifically, the section was talking about opening all external links in a browser, but I'm not sure how much sense that makes for mailto and tel links. Our original focus was only on ensuring the browser security model was available when launching web content. I'm also not sure "links that resolve outside the epub publication" is an accurate description of mailto and tel links and the like. They require a specialized application, but what are they really resolving to?

To try and fix these ambiguities, I've reworded the recommendation to "hyperlinks that declare a scheme". That should cover everything that is not internal without having to be specific about what is being linked.

For the original recommendation about opening in a browser, I've generalized it to opening an appropriate application.

Otherwise, the only other change I made was to add a note to the end of the security recommendations to point to the security sections we have in the body. We talk about external links, network access and scripting in the threats, but then kind of fall silent in the recommendations because the specifics are already in the body.

Anyway, let me know if the new wording makes sense or if it still needs improvement.

Fixes #2320 

EPUB Reading Systems 3.3:
- [Preview](https://raw.githack.com/w3c/epub-specs/fix/issue-2320/epub33/rs/index.html)
- [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fix/issue-2320/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -387,9 +387,9 @@
 				<p>When a link has an <code>http</code> or <code>https</code>
 					<a data-cite="url#concept-url-scheme">scheme</a> [[url]], reading systems:</p>
 
-				<ul>
-					<li id="sec-external-links-consent">SHOULD obtain the user's consent before opening the link.</li>
-					<li data-tests="#pub-external-links" id="sec-external-links-open">SHOULD open the link in a new browser instance to ensure that the browser's security and privacy
+				<ul data-tests="#pub-external-links" id="sec-external-links-open">
+					<li>SHOULD obtain the user's consent before opening the link.</li>
+					<li>SHOULD open the link in a new browser instance to ensure that the browser's security and privacy
 						controls are available to users.</li>
 				</ul>
 
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -388,7 +388,7 @@
 							data-cite="url#concept-url-scheme">scheme</a> [[url]] (i.e., that resolves to web content
 						outside the [=EPUB publication=] or requires a system application to open).</span>
 					<span data-tests="#pub-external-links">If the user consents to open the link, reading systems SHOULD
-						open the link using the default application registered for the scheme (i.e., to help ensure
+						open it using the default application registered for the scheme (i.e., to help ensure
 						appropriate security and privacy controls are available).</span></p>
 
 				<div class="note">
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -393,7 +393,7 @@
 						controls are available to users.</li>
 				</ul>
 
-				<p>For links that use schemes that require a helper application to load, reading systems SHOULD obtain
+				<p>For links that have schemes that require a helper application to load, reading systems SHOULD obtain
 					the user's consent and follow platform guidance for opening the helper application.</p>
 
 				<div class="note">
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -384,7 +384,7 @@
 			<section id="sec-epub-rs-external-links">
 				<h3>External links</h3>
 
-				<p data-tests="#pub-external-links">When a link has an <code>http</code> or <code>https</code>
+				<p>When a link has an <code>http</code> or <code>https</code>
 					<a data-cite="url#concept-url-scheme">scheme</a> [[url]], reading systems:</p>
 
 				<ul>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -384,18 +384,18 @@
 			<section id="sec-epub-rs-external-links">
 				<h3>External links</h3>
 
-				<p>Reading systems SHOULD alert users before opening hyperlinks that declare a <a
-						data-cite="url#concept-url-scheme">scheme</a> [[url]] (i.e., that resolve outside the [=EPUB
-					publication=] or require a specialized application to open).</p>
+				<p><span>Reading systems SHOULD alert users before opening a link that declares a <a
+							data-cite="url#concept-url-scheme">scheme</a> [[url]] (i.e., that resolves to web content
+						outside the [=EPUB publication=] or requires a system application to open).</span>
+					<span data-tests="#pub-external-links">If the user consents to open the link, reading systems SHOULD
+						open the link using the default application registered for the scheme (i.e., to help ensure
+						appropriate security and privacy controls are available).</span></p>
 
-				<p data-tests="#pub-external-links">Reading systems SHOULD open links that declare a scheme in an
-					appropriate application to ensure security and privacy controls are available.</p>
-
-				<p>Although links to external web sites and resources are commonly found in <a>EPUB content
-						documents</a>, these are not the only sources of hyperlinks with schemes. For example, if a
-					reading system provides access to external <a data-cite="epub-33#sec-link-elem">linked records</a>
-					[[epub-33]] in the <a>package document</a> metadata, it should open the links in a new browser
-					instance.</p>
+				<div class="note">
+					<p>External links are not only found in [=EPUB content documents=]. A reading system might provides
+						access to external <a data-cite="epub-33#sec-link-elem">linked records</a> [[epub-33]] in the
+							<a>package document</a> metadata, for example.</p>
+				</div>
 
 				<div class="note">
 					<p>For more information about security and privacy issues with external links, see <a
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -393,7 +393,7 @@
 						controls are available to users.</li>
 				</ul>
 
-				<p>For links that have schemes that require a helper application to load, reading systems SHOULD obtain
+				<p id="sec-external-links-consent" data-tests="#pub-external-links_consent">For links that have schemes that require a helper application to load, reading systems SHOULD obtain
 					the user's consent and follow platform guidance for opening the helper application.</p>
 
 				<div class="note">
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -384,15 +384,20 @@
 			<section id="sec-epub-rs-external-links">
 				<h3>External links</h3>
 
-				<p><span>Reading systems SHOULD alert users before opening a link that declares a <a
-							data-cite="url#concept-url-scheme">scheme</a> [[url]] (i.e., that resolves to web content
-						outside the [=EPUB publication=] or requires a system application to open).</span>
-					<span data-tests="#pub-external-links">If the user consents to open the link, reading systems SHOULD
-						open it using the default application registered for the scheme (i.e., to help ensure
-						appropriate security and privacy controls are available).</span></p>
+				<p data-tests="#pub-external-links">When a link has an <code>http</code> or <code>https</code>
+					<a data-cite="url#concept-url-scheme">scheme</a> [[url]], reading systems:</p>
+
+				<ul>
+					<li>SHOULD obtain the user's consent before opening the link.</li>
+					<li>SHOULD open the link in a new browser instance to ensure that the browser's security and privacy
+						controls are available to users</li>
+				</ul>
+
+				<p>For links that use schemes that require a helper application to load, reading systems SHOULD obtain
+					the user's consent and follow platform guidance for opening the helper application.</p>
 
 				<div class="note">
-					<p>External links are not only found in [=EPUB content documents=]. A reading system might provides
+					<p>External links are not only found in [=EPUB content documents=]. A reading system might provide
 						access to external <a data-cite="epub-33#sec-link-elem">linked records</a> [[epub-33]] in the
 							<a>package document</a> metadata, for example.</p>
 				</div>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -388,7 +388,7 @@
 					<a data-cite="url#concept-url-scheme">scheme</a> [[url]], reading systems:</p>
 
 				<ul>
-					<li>SHOULD obtain the user's consent before opening the link.</li>
+					<li id="sec-external-links-consent">SHOULD obtain the user's consent before opening the link.</li>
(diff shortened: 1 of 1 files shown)


OriKuttner:
Wouldn't it be better to say "Links should be open in an appropriate user
agent"?

-- 
Ori Idan CEO Helicon Books
http://www.heliconbooks.com

mattgarrish:
> an appropriate user agent

I'm not sure if that works for applications like Outlook for email or Skype for telephone numbers. User agent tends to be a very webby word in my experience -- for retrieving / rendering some type of web content -- but it's easy to change if others disagree.

epub33/rs/index.html
```diff
@@ -384,14 +384,18 @@ <h3>Network access</h3>
 			<section id="sec-epub-rs-external-links">
 				<h3>External links</h3>
 
-				<p data-tests="#pub-external-links">Reading systems SHOULD open links that resolve outside the <a>EPUB
-						publication</a> in a new browser instance to ensure that the browser's security and privacy
-					controls are available to users.</p>
+				<p>Reading systems SHOULD alert users before opening hyperlinks that declare a <a
+						data-cite="url#concept-url-scheme">scheme</a> [[url]] (i.e., that resolve outside the [=EPUB
+					publication=] or require a specialized application to open).</p>
```
iherman: I wonder about the terminology "declare a scheme"; it looks a bit strange to me. I looked at the URL spec and this terminology is not used either. What about "includes a scheme"?

I am a bit neutral about the usage of "application" vs. "user agent" (referring to the remark of @OriIdan). I have a mild preferene for user agent, which seems to be more in line with what other W3C specs are using (like UAAG).

epub33/rs/index.html
```diff
@@ -384,14 +384,18 @@ <h3>Network access</h3>
...
+
+				<p data-tests="#pub-external-links">Reading systems SHOULD open links that declare a scheme in an
+					appropriate application to ensure security and privacy controls are available.</p>
```
iherman: (This is a matter of taste:) I wonder if it does not read better to merge the two paragraphs into one.

epub33/rs/index.html
```diff
@@ -384,14 +384,18 @@ <h3>Network access</h3>
...
 
 				<p>Although links to external web sites and resources are commonly found in <a>EPUB content
-						documents</a>, these are not the only sources. For example, if a reading system provides access
-					to <a data-cite="epub-33#sec-link-elem">linked records</a> [[epub-33]] in the <a>package
-						document</a> metadata, it should similarly open the links in a new browser instance.</p>
+						documents</a>, these are not the only sources of hyperlinks with schemes. For example, if a
+					reading system provides access to external <a data-cite="epub-33#sec-link-elem">linked records</a>
+					[[epub-33]] in the <a>package document</a> metadata, it should open the links in a new browser
+					instance.</p>
```
iherman: For the sake of consistency, we should use the term "specialized external application" (or "user agent") here as well.

iherman (approved):

epub33/rs/index.html
mattgarrish: "has a scheme" is probably simplest.

mattgarrish: > user agent ... seems to be more in line with what other W3C specs are using (like UAAG)

Even UAAG defines a user agent as:

> any software that retrieves, renders and facilitates end-user interaction with web content.

A telephone number really isn't web content. That's where my disconnect comes in, as say your phone dialer gets launched, not a program that dials over the web.

epub33/rs/index.html
mattgarrish: I didn't want to interfere with the existing test, but I guess that could be shuffled to a span, right?

epub33/rs/index.html
mattgarrish: Actually, I'm not sure there's any point to this paragraph anymore. It was added because the first one was so focused on opening web content. At most we might mention that links aren't only found in content documents, but I'm not sure that's really a major clarification.

epub33/rs/index.html
mattgarrish: How about a hybrid statement:

> To ensure security and privacy controls are available, Reading systems SHOULD open the link in the user agent or system application registered to handle the scheme.

mattgarrish: Here's what's in the latest commit:

> Reading systems SHOULD alert users before opening a link that declares a [scheme](https://url.spec.whatwg.org/#concept-url-scheme) [url] (i.e., that resolves to web content outside the [EPUB publication](https://www.w3.org/TR/epub-33/#dfn-epub-publication) or requires a system application to open). If the user consents to open the link, reading systems SHOULD open it using the default application registered for the scheme (i.e., to help ensure appropriate security and privacy controls are available).

epub33/rs/index.html
```diff
@@ -384,14 +384,18 @@ <h3>Network access</h3>
...
+					<span data-tests="#pub-external-links">If the user consents to open the link, reading systems SHOULD
+						open it using the default application registered for the scheme (i.e., to help ensure
+						appropriate security and privacy controls are available).</span></p>
 
-				<p>Although links to external web sites and resources are commonly found in <a>EPUB content
-						documents</a>, these are not the only sources. For example, if a reading system provides access
-					to <a data-cite="epub-33#sec-link-elem">linked records</a> [[epub-33]] in the <a>package
-						document</a> metadata, it should similarly open the links in a new browser instance.</p>
+				<div class="note">
+					<p>External links are not only found in [=EPUB content documents=]. A reading system might provides
```
```suggestion
					<p>External links are not only found in [=EPUB content documents=]. A reading system might provide
```

epub33/rs/index.html
```diff
@@ -384,14 +384,18 @@ <h3>Network access</h3>
 			<section id="sec-epub-rs-external-links">
 				<h3>External links</h3>
 
-				<p data-tests="#pub-external-links">Reading systems SHOULD open links that resolve outside the <a>EPUB
-						publication</a> in a new browser instance to ensure that the browser's security and privacy
-					controls are available to users.</p>
+				<p><span>Reading systems SHOULD alert users before opening a link that declares a <a
+							data-cite="url#concept-url-scheme">scheme</a> [[url]] (i.e., that resolves to web content
```
bduga: This seems vague, and I can't say I have done an exhaustive review to say if this is entirely correct. For instance, data:, blob:, and chrome: are all url schemes that are local or even part of the publication. I am not sure if you can "open a link" to data: or blob:, but I am also not sure that you can't.

epub33/rs/index.html
```diff
@@ -384,14 +384,18 @@ <h3>Network access</h3>
 			<section id="sec-epub-rs-external-links">
 				<h3>External links</h3>
 
-				<p data-tests="#pub-external-links">Reading systems SHOULD open links that resolve outside the <a>EPUB
-						publication</a> in a new browser instance to ensure that the browser's security and privacy
```
bduga, line 388: There are kind of two issues here - links that open in the same app as the reading system, and links that open externally. If we change this statement, we lose the recommendation that a browser based client open links in a new browser context (eg open a new window/tab, don't replace the contents of the current window/tab, or worse open the content in a subframe of the reading system). But maybe that is fine, since we do still recommend a warning in that case.

epub33/rs/index.html
```diff
@@ -384,14 +384,18 @@ <h3>Network access</h3>
...
-					controls are available to users.</p>
+				<p><span>Reading systems SHOULD alert users before opening a link that declares a <a
+							data-cite="url#concept-url-scheme">scheme</a> [[url]] (i.e., that resolves to web content
+						outside the [=EPUB publication=] or requires a system application to open).</span>
+					<span data-tests="#pub-external-links">If the user consents to open the link, reading systems SHOULD
+						open it using the default application registered for the scheme (i.e., to help ensure
```
bduga: I am not sure this makes sense. For instance, Chrome may open PDFs in Chrome, even if there is another default viewer for PDF. In fact, it would be very odd to recommend we open system default apps since browsers typical do *not* do that. And not all operating systems even have the concept of a default. For instance, Android will let you pick an app from list. The user can decide one is app is the one they always want to use, but that isn't required.

bduga (commented):
An alternative would be:

* Leave the requirement for new browser context
* Add a should for opening links that open a new browser context
* Add a general caution for non http(s) based schemes when following links (ask for permission, follow platform guidance for opening helper applications)

mattgarrish:
The latest commit is my best attempt to implement @bduga's comment. "helper application" is much better than user agent or system application. Didn't realize that one existed.

Let me know if the revised text needs further clarification, though.

epub33/rs/index.html
```diff
@@ -384,14 +384,23 @@ <h3>Network access</h3>
 			<section id="sec-epub-rs-external-links">
 				<h3>External links</h3>
 
-				<p data-tests="#pub-external-links">Reading systems SHOULD open links that resolve outside the <a>EPUB
-						publication</a> in a new browser instance to ensure that the browser's security and privacy
-					controls are available to users.</p>
+				<p data-tests="#pub-external-links">When a link has an <code>http</code> or <code>https</code>
```
```suggestion
				<p>When a link has an <code>http</code> or <code>https</code>
```

epub33/rs/index.html
```diff
@@ -384,14 +384,23 @@ <h3>Network access</h3>
...
+					<a data-cite="url#concept-url-scheme">scheme</a> [[url]], reading systems:</p>
 
-				<p>Although links to external web sites and resources are commonly found in <a>EPUB content
-						documents</a>, these are not the only sources. For example, if a reading system provides access
-					to <a data-cite="epub-33#sec-link-elem">linked records</a> [[epub-33]] in the <a>package
-						document</a> metadata, it should similarly open the links in a new browser instance.</p>
+				<ul>
+					<li>SHOULD obtain the user's consent before opening the link.</li>
+					<li>SHOULD open the link in a new browser instance to ensure that the browser's security and privacy
+						controls are available to users</li>
```
```suggestion
					<li data-tests="#pub-external-links" id="sec-external-links-open">SHOULD open the link in a new browser instance to ensure that the browser's security and privacy
						controls are available to users.</li>
```

epub33/rs/index.html
```diff
@@ -384,14 +384,23 @@ <h3>Network access</h3>
...
-					controls are available to users.</p>
+				<p data-tests="#pub-external-links">When a link has an <code>http</code> or <code>https</code>
+					<a data-cite="url#concept-url-scheme">scheme</a> [[url]], reading systems:</p>
 
-				<p>Although links to external web sites and resources are commonly found in <a>EPUB content
-						documents</a>, these are not the only sources. For example, if a reading system provides access
-					to <a data-cite="epub-33#sec-link-elem">linked records</a> [[epub-33]] in the <a>package
-						document</a> metadata, it should similarly open the links in a new browser instance.</p>
+				<ul>
+					<li>SHOULD obtain the user's consent before opening the link.</li>
```
iherman, line 391:
```suggestion
					<li id="sec-external-links-consent">SHOULD obtain the user's consent before opening the link.</li>
```

iherman (approved):
I have proposed changes for the testing; both list items may (should...) get a test and better add the right ID-s now. Also the existing test should be used for the second bullet item not the paragraph preceding them.

Finally, note to ourselves: if this goes ahead, a separate PR should be raised for a change in  https://github.com/w3c/epub-tests/blob/main/tests/pub-external-links/EPUB/package.opf (the reference to the spec must be updated).

epub33/rs/index.html
```diff
@@ -384,14 +384,23 @@ <h3>Network access</h3>
...
-						documents</a>, these are not the only sources. For example, if a reading system provides access
-					to <a data-cite="epub-33#sec-link-elem">linked records</a> [[epub-33]] in the <a>package
-						document</a> metadata, it should similarly open the links in a new browser instance.</p>
+				<ul>
+					<li id="sec-external-links-consent">SHOULD obtain the user's consent before opening the link.</li>
+					<li data-tests="#pub-external-links" id="sec-external-links-open">SHOULD open the link in a new browser instance to ensure that the browser's security and privacy
+						controls are available to users.</li>
+				</ul>
+
+				<p>For links that use schemes that require a helper application to load, reading systems SHOULD obtain
```
```suggestion
				<p>For links that have schemes that require a helper application to load, reading systems SHOULD obtain
```

epub33/rs/index.html
```diff
@@ -384,14 +384,23 @@ <h3>Network access</h3>
...
-					to <a data-cite="epub-33#sec-link-elem">linked records</a> [[epub-33]] in the <a>package
-						document</a> metadata, it should similarly open the links in a new browser instance.</p>
+				<ul>
+					<li id="sec-external-links-consent">SHOULD obtain the user's consent before opening the link.</li>
+					<li data-tests="#pub-external-links" id="sec-external-links-open">SHOULD open the link in a new browser instance to ensure that the browser's security and privacy
+						controls are available to users.</li>
+				</ul>
+
+				<p>For links that have schemes that require a helper application to load, reading systems SHOULD obtain
+					the user's consent and follow platform guidance for opening the helper application.</p>
```
iherman: While creating the relevant epub test modification, I have added a test for the user consent requirement.

```suggestion
				<p id="sec-external-links-consent" data-tests="#pub-external-links_consent">For links that have schemes that require a helper application to load, reading systems SHOULD obtain
					the user's consent and follow platform guidance for opening the helper application.</p>
```

iherman (approved):

shiestyle (approved):

wareid (approved):

PR #2337, opened 2022-06-21
Adding reference to symbolic links to the threat models (issue 2322)

### LINKED ISSUE #2322: Adding malicious symbolic links to an EPUB publication
As discussed in the meeting with @GJFR (see [EPUB meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-06-03-epub)) a malicious EPUB publication may include a symbolic link in its content pointing 'outside' the container to the local file system. This threat should be documented in the [§15.2 Thread model](https://www.w3.org/TR/epub-rs-33/#epub-threat-model) of the reading system.

We may also consider explicitly disallowing such files in the content document, but only if epubcheck is able to find those in the content. @rdeltour @mattgarrish ?

iherman:
The PR adds a reference to symbolic links to the threat models of both the core and the RS specs, essentially along the lines of what had been proposed in https://github.com/w3c/epub-specs/issues/2322#issuecomment-1156635255. Both threat models already had an item on "malicious content", and that seemed like the good place to add this.

Because the threat models' section implicitly says to RS systems that "you ought to deal with these", adding a separate warning along these lines on symbolic links (and friends), as also mentioned in https://github.com/w3c/epub-specs/issues/2322#issuecomment-1156635255, seemed superfluous.

I did not find a proper place to add a note on _not_ using the `-y` option in the `zip` command; we do not refer to that command in the spec in the first place. So I dropped this, too.

Fix #2322

See:

* For EPUB 3.3:
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/symb-link-warning-issue-2322/epub33/core/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/core/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/symb-link-warning-issue-2322/epub33/core/index.html)
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/symb-link-warning-issue-2322/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/symb-link-warning-issue-2322/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2337.html" title="Last updated on Jun 21, 2022, 1:22 PM UTC (2544a68)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2337/c4ced5a...2544a68.html" title="Last updated on Jun 21, 2022, 1:22 PM UTC (2544a68)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -9341,7 +9341,10 @@ html.my-document-playing * {
 					<dd>
 						<p>Resources embedded in the EPUB container are not immune to malicious actors, especially when
 							EPUB publications are obtained from untrusted sources. Resources may contain exploits or
-							forms may submit sensitive information to unintended parties.</p>
+							forms that may submit sensitive information to unintended parties.
+							Such actors may also try to gain access to [=remote resources=] using file indirection techniques,
+							such as symbolic links or file aliases.
+						</p>
 						<p>The use of third-party content, such as games and quizzes, may also lead to security and
 							privacy issues if the EPUB creator is not able to fully vet the content.</p>
 					</dd>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2245,8 +2245,12 @@
 
 					<dt>Malicious content</dt>
 					<dd>
-						<p>EPUB publications may contain resources designed to exploit security flaws in reading systems
-							or the operating systems they run on.</p>
+						<p>
+							EPUB publications may contain resources designed to exploit security flaws in reading systems
+							or the operating systems they run on. Attackers may also try to gain access to
+							[=remote resources=] using file indirection techniques, such as symbolic links or 
+							file aliases.
+						</p>
 						<p>The lack of a standard method of signing EPUB publications means that reading systems cannot
 							always verify whether the content has been tampered with between authoring and loading in
 							the device.</p>


mattgarrish:
> I did not find a proper place to add a note on _not_ using the `-y` option in the `zip` command

It's not universal to all zip tools, either, so not sure it would make sense to call it out specifically.

Should we say anything in the recommendations about perhaps being less trusting of resources that are not declared in the package document? Something along the lines of although these are allowed, reading systems may want to attempt to verify their format before unpacking/allowing access to them?

iherman:
> > I did not find a proper place to add a note on _not_ using the `-y` option in the `zip` command
> 
> It's not universal to all zip tools, either, so not sure it would make sense to call it out specifically.
> 

+1

> Should we say anything in the recommendations about perhaps being less trusting of resources that are not declared in the package document? Something along the lines of although these are allowed, reading systems may want to attempt to verify their format before unpacking/allowing access to them?

Isn't that a separate issue? If I want to be nasty, I can create a symbolic link, and I then I declare it in the package document. I guess epubcheck may catch this, but that is simply because they are that good :-)

mattgarrish:
It's true for all resources since you can fake out epubcheck by declaring it as a foreign resource. Won't catch sideloaded content, either. This issue just highlights the problem.

wareid (approved):

PR #2342, opened 2022-06-24
Update text on self-containment and add reference to iso 22424

### LINKED ISSUE #1876: Self-contained Packages
From the PING review: 

> Self-contained packages have potential huge privacy advantages, but it's not clear that the EPUB spec or current implementations fulfill these opportunities. Is that a goal that the community could work towards?

> The current spec anticipates and requires (at least a SHOULD) loading of remote resources from arbitrary origins. This introduces risks of additional data collection about who is reading what book (and from where), and what part of the book is being read at a particular moment (depending on the implementation or requirements on how remote resources are loaded). And remote resource loading should also make explicit that the author/publisher of the book may effectively be collecting data on the reader's habits, in addition to the reading system. Different levels of scripting access are defined, but it's not clear whether any such level would indicate that user reading behavior would not be disclosed.

> It would be useful to specify a privacy threat model specific to EPUB, to the extent that it varies from the Web. Can we guarantee that reading habits will not be surveilled, by the publisher, the retailer, the reading system, or other parties? Or if that data is revealed, then we should clarify to whom or under what conditions. Book-like privacy could be achieved, but would require significant changes from the current spec and current popular implementations.

> The spec suggests that the manifest is an exhaustive pre-stated list of resources, including remote resources, but it's not clear how that's intended to be handled by reading systems. Should a reading system refuse to fetch any remote resource not included in the manifest? Are remote resources intended to be the same for all readers of a book, or might they be personalized?

mattgarrish:
This PR reworks the last paragrah about self containment and adds a reference to OAIS preservation document that was produced in ISO as an example.

Fixes #1876

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2342.html" title="Last updated on Jun 24, 2022, 9:54 AM UTC (76b381f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2342/947d70f...76b381f.html" title="Last updated on Jun 24, 2022, 9:54 AM UTC (76b381f)">Diff</a>

### DIFF
--- a/epub33/core/biblio.js
+++ b/epub33/core/biblio.js
@@ -53,6 +53,11 @@ var biblio = {
 		"href": "https://www.itu.int/ITU-T/recommendations/rec.aspx?rec=13189",
 		"date": "2017-04-13"
 	},
+	"iso22424": {
+		"title": "ISO/IEC TS 22424-1:2020 — Digital publishing — EPUB3 preservation",
+		"href": "https://www.iso.org/standard/73163.html",
+		"date": "2020-01"
+	},
 	"isoschematron": {
 		"title": "ISO/IEC 19757-3: Rule-based validation — Schematron",
 		"href": "http://standards.iso.org/ittf/PubliclyAvailableStandards/c040833_ISO_IEC_19757-3_2006(E).zip",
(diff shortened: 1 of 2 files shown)


iherman (approved):

iherman:
@mattgarrish merging this?

PR #2343, opened 2022-06-24
Remove redundant processing requirement for hreflang attribute

mattgarrish:
As discussed in https://github.com/w3c/epub-tests/pull/173, we have a redundant requirement on language processing for the `hreflang` attribute.

The internationalization section already defines that language and direction are obtained from resources, not information in the package document, and specifically calls out `hreflang` as one case.

This pull request just removes the redundant must for now, replacing it with a pointer to the overarching requirement, but I'm easily convinced we can drop the entire paragraph. It only seems to serve the purpose of reminding a developer of an existing requirement, but we don't repeat this for every element that has a language or direction property, so why here?

EPUB Reading Systems 3.3:
- [Preview](https://raw.githack.com/w3c/epub-specs/fix/hreflang-req/epub33/rs/index.html)
- [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fix/hreflang-req/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -393,8 +393,9 @@
 						controls are available to users.</li>
 				</ul>
 
-				<p id="sec-external-links-consent" data-tests="#pub-external-links_consent">For links that have schemes that require a helper application to load, reading systems SHOULD obtain
-					the user's consent and follow platform guidance for opening the helper application.</p>
+				<p id="sec-external-links-consent" data-tests="#pub-external-links_consent">For links that have schemes
+					that require a helper application to load, reading systems SHOULD obtain the user's consent and
+					follow platform guidance for opening the helper application.</p>
 
 				<div class="note">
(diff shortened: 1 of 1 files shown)


iherman (approved):
You asked, in https://github.com/w3c/epub-tests/pull/173#issuecomment-1165577423

> I'll open a pull request with an additional reference to this paragraph, but I wonder if we can delete this bit of duplication completely?

I would be in favour of keeping it. It does not harm, it does not add any normative text, but may be helpful for RS implementers.

iherman:
I would think this could be merged, @mattgarrish. If you do, could you also close https://github.com/w3c/epub-tests/pull/173 without merge?

PR #2349, opened 2022-07-05
Allow other properties in viewport meta

mattgarrish:
This PR adds a simple "compat" definition that allows other properties to optionally be used around and between the width and height definitions just to show how this might alternatively be done. 

I used a prose description for it, as I don't believe we need to expend much effort on these other properties when they're ignored. We could always list off the four properties from the apple def and use a prose descriptor to avoid time on defining their values, but I'm not sure why we even need to care if someone makes a typo or adds a completely unknown property. It's all the same to getting the ICB.

(I also allow for other syntactic weirdness that doesn't harm anything by changing the separator definition as I mentioned in the original PR.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2349.html" title="Last updated on Jul 5, 2022, 1:06 PM UTC (0594c00)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2349/f008cee...0594c00.html" title="Last updated on Jul 5, 2022, 1:06 PM UTC (0594c00)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7625,7 +7625,7 @@ No Entry</pre>
 								<p>The <code>device-width</code> and <code>device-height</code> values refer to the 100%
 									of the width and height, respectively, of the reading system's [=viewport=].</p>
 								<p class="note">The value of the <code>viewport</code> [^meta^] element's
-										<code>content</code> attribute allows the inclusion of the additional properties
+										<code>content</code> attribute allows the inclusion of additional properties
 									only for compatibility with the Apple definition this metadata tag was originally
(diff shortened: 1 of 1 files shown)


mattgarrish:
I've got a simpler idea in mind for this, but I'll have to open another PR. In essence, we move all the EBNF to an appendix. Instead of defining the properties of the tag, though, we only describe how to define/separate any property in a generic manner. The FXL section then requires height and width properties and describes their values.

mattgarrish:
Closing this as https://github.com/w3c/epub-specs/pull/2350 is the simpler option if we want to allow other properties.

PR #2350, opened 2022-07-05
Simpler viewport meta definition 

mattgarrish:
This PR moves the EBNF to an appendix, leaving only prose requirements for the values of the required height and width definitions in the ICB section.

I've also stripped the ordering requirements from the meta definition. The ICB section already requires the two necessary properties, so the definition should be fine just stating that the tag contains one or more properties.

I've listed the six possible properties defined in the apple doc, but we could even make the name field a generic "? ascii characters ?" definition and not bother ourselves at all about what properties are allowed. (Keeping to the known set is probably best, though.)

I also added a note explaining what whitespace normalization entails, as I'm not sure the average author trying to understand the spec will get that this is a post-processed syntax. Then again, they probably won't be able to read the EBNF, either...

    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-2292/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-2292/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2350.html" title="Last updated on Jul 19, 2022, 3:57 AM UTC (1def8b3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2350/265c3cf...1def8b3.html" title="Last updated on Jul 19, 2022, 3:57 AM UTC (1def8b3)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7505,7 +7505,7 @@ No Entry</pre>
 									<dt>name</dt>
 									<dd>The attribute value MUST be <code>viewport</code>.</dd>
 									<dt>content</dt>
-									<dd> <p>The attribute value MUST be of the following form:</p>
+									<dd> <p>The <a href="https://www.w3.org/TR/2008/REC-xml-20081126/#AVNormalize">normalized</a> [[XML]] attribute value MUST be of the following form:</p>
 										<table class="productionset">
 											<caption>(EBNF productions <a href="https://www.iso.org/standard/26153.html">ISO/IEC
 												14977</a>)<br /> All terminal symbols are in the Unicode Block 'Basic Latin' (U+0000 to
@@ -7553,18 +7553,15 @@ No Entry</pre>
 												</td>
 												<td>=</td>
 												<td>
-													{ <a href="#viewport.ebnf.whitespace">whitespace</a> }, 
-													(";" | "," | #x20), 
-													{ <a href="#viewport.ebnf.whitespace">whitespace</a> } ;</td>
+													[space], (";" | "," | space), [space]
+												</td>
 											</tr>
 											<tr>
 												<td id="viewport.ebnf.assign">
 													<a href="#viewport.ebnf.assign">assign</a>
 												</td>
 												<td>=</td>
-												<td>{ <a href="#viewport.ebnf.whitespace">whitespace</a> }, 
-													"=", 
-													{ <a href="#viewport.ebnf.whitespace">whitespace</a> } ;</td>
+												<td>[space], "=", [space] ;</td>
 											</tr>
 											<tr>
 												<td id="viewport.ebnf.dimension">
@@ -7574,11 +7571,11 @@ No Entry</pre>
 												<td>? positive number ? ;</td>
 											</tr>
 											<tr>
-												<td id="viewport.ebnf.whitespace">
-													<a href="#viewport.ebnf.whitespace">whitespace</a>
+												<td id="viewport.ebnf.space">
+													<a href="#viewport.ebnf.space">space</a>
 												</td>
 												<td>=</td>
-												<td>(#x20 | #x9 | #xD | #xA) ;</td>
+												<td>#x20 ;</td>
 											</tr>
 										</table>
 									</dd>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7498,136 +7498,18 @@ No Entry</pre>
 								<p>For XHTML [=fixed-layout documents=] the <a
 										href="https://www.w3.org/TR/CSS2/visudet.html#containing-block-details">initial
 										containing block</a> [[css2]] is obtained from the <code>height</code> and
-										<code>width</code> definitions in a <code>viewport</code> [^meta^] [[html]]
-									tag.</p>
-								<p>The <code>viewport</code>
-									<code>meta</code> tag MUST have <code>name</code> and <code>content</code>
-									attributes that conform to the following definition:</p>
-								<dl>
-									<dt>name</dt>
-									<dd>The value of the attribute after <a
-											href="https://www.w3.org/TR/2008/REC-xml-20081126/#AVNormalize">whitespace
-											normalization</a> [[xml]] MUST be <code>viewport</code>.</dd>
-									<dt>content</dt>
-									<dd>
-										<p>The value of the attribute after <a
-												href="https://www.w3.org/TR/2008/REC-xml-20081126/#AVNormalize"
-												>whitespace normalization</a> [[xml]] MUST be of the following form:</p>
-										<table class="productionset">
-											<caption>(EBNF productions <a href="https://www.iso.org/standard/26153.html"
-													>ISO/IEC 14977</a>)<br /> All terminal symbols are in the Unicode
-												Block 'Basic Latin' (U+0000 to U+007F).</caption>
-
-											<tr>
-												<td id="viewport.ebnf.def">
-													<a href="#viewport.ebnf.def">viewport</a>
-												</td>
-												<td>=</td>
-												<td>( <a href="#viewport.ebnf.width-and-height">width-and-height</a> |
-														<a href="#viewport.ebnf.height-and-width">height-and-width</a> )
-													;</td>
-											</tr>
-											<tr>
-												<td id="viewport.ebnf.width-and-height">
-													<a href="#viewport.ebnf.width-and-height">width-and-height</a>
-												</td>
-												<td>=</td>
-												<td>{ <a href="#viewport.ebnf.compat">compat</a>, <a
-														href="#viewport.ebnf.sep">sep</a> }, <a
-														href="#viewport.ebnf.width">width</a>, { <a
-														href="#viewport.ebnf.sep">sep</a>, <a
-														href="#viewport.ebnf.compat">compat</a> }, <a
-														href="#viewport.ebnf.sep">sep</a>, <a
-														href="#viewport.ebnf.height">height</a>, { <a
-														href="#viewport.ebnf.sep">sep</a>, <a
-														href="#viewport.ebnf.compat">compat</a> } ;</td>
-											</tr>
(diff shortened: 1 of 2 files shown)


epub33/rs/index.html
```diff
@@ -1594,9 +1595,16 @@ <h4>Initial containing block dimensions</h4>
...
-								<span id="confreq-fxl-rs-xhtml-icb" data-tests="#fxl-xhtml-icb">They MUST clip content
-									positioned outside of the ICB.</span>
+								<span id="confreq-fxl-rs-xhtml-icb" data-tests="#fxl-xhtml-icb,fxl-xhtml-icb_multi">They
+									MUST clip content positioned outside of the ICB.</span>
 							</p>
+							<p>If the <code>viewport</code>
+								<code>meta</code> does not contain a width or a height value, or if these values are
+								invalid, reading systems SHOULD consider these as having the values
+									<code>device-width</code> and <code>device-height</code>, respectively. Reading
+								systems MAY also use best effort in interpreting values that follow a deprecated syntax
```
mattgarrish: This will have to change as we can't normatively ask for best efforts... 😉 

It's not clear what it refers to, either, as the CSS module doesn't define a syntax, let alone a deprecated one.

iherman:
This is just after a quick glance, will do a more thorough review when I am back from vacations... 

If I was a first time reader of this spec I would be really messed upon why on earth are these guys bothering with the introduction of those terms in the section that seem to be unnecessary (e.g., initial-scale). In this respect, I preferred the extended EBNF syntax that you proposed in another PR that left open the other possible terms (and we can put a reference to the apple spec). That PR also had a reference to history as for why we bother in the first place (because we try to be backward compatible, that is). I also think that we should try to be a bit more forthcoming in saying that authors should not use those other property values and reading systems should ignore them. 

(I like the idea of moving the formal grammar to the appendix, b.t.w.)

mattgarrish:
> I preferred the extended EBNF syntax that you proposed in another PR that left open the other possible terms

Ya, I was thinking this morning that we don't want to validate for names, but that's probably where it leaves us if we're too specific. I've changed the definition to:

> name = "height" | "width" | ? character data ? ;

It just feels like we should list the two properties that we're expecting to be set. That's what initially led me to list all the ones from the apple def, but I think the above is a good compromise.

> I also think that we should try to be a bit more forthcoming in saying that authors should not use those other property values and reading systems should ignore them.

Other than discouraging their use we fall into the trap of potentially invalidating existing content.

But if reading systems have to ignore everything but height and width in fxl documents, then does it become required that they completely ignore viewport meta declarations in reflowable? It would be weird to disallow it in one rendering and not in the other.

Another question is what is a reading system supposed to do if there is more than one viewport declaration? If it uses the first, what if it's invalid? It's not like viewBox where you can only have one declaration.

mattgarrish:
The key changes in the last commit are:

- the ebnf definition is more generic than I mentioned in my last comment. It's just "? character data ?" now, as we define the height and width requirements in the fxl section. The default definition can be as simple as possible.
- it adds a clarification that height and width are required properties as changing the definition meant we only had requirements on the values
- it adds a requirement for RS to strip non-numeric character data and use resulting number as the dimension.
- it adds a requirement to use first viewport declaration if more than one (if this ends up being invalid, then the result would be that device-width and device-height are used as the defaults)
- it adds a new section to ignore viewport meta declarations in both reflow and fxl (except for getting icb dimensions)

The latter has me the most concerned, as I'm not sure what effect setting things like initial-scale has in an epub. It's necessary for mobile, so if the document gets read in a mobile RS is it good practice to set these properties?

Do we need to tell reading systems anything about other properties when they aren't in scope of the ICB calculation?

epub33/core/index.html
```diff
@@ -10592,6 +10607,140 @@ <h6>The <code>-epub-text-underline-position</code> property</h6>
...
+		<section id="app-viewport-meta" class="appendix">
+			<h2>The <code>viewport</code>
+				<code>meta</code> tag</h2>
+
+			<section id="app-viewport-meta-intro" class="informative">
+				<h3>Introduction</h3>
+
+				<p>As the <a
+						href="https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html#//apple_ref/doc/uid/TP40008193-SW6"
+						>Safari definition</a> of the <code>viewport</code>
```
```suggestion
						>Safari HTML definition</a> of the <code>viewport</code>
```

epub33/core/index.html
```diff
@@ -10592,6 +10607,140 @@ <h6>The <code>-epub-text-underline-position</code> property</h6>
...
+					<code>meta</code> tag is not an officially recognized standard, this specification defines a basic
```
```suggestion
					<code>meta</code> tag, that was used in earlier versions of EPUB 3, is not an officially recognized standard, this specification defines a basic
```

epub33/core/index.html
```diff
@@ -10592,6 +10607,140 @@ <h6>The <code>-epub-text-underline-position</code> property</h6>
...
+
+			<section id="app-viewport-meta-syntax">
+				<h3>Syntax</h3>
+
+				<p>A <code>viewport</code> [^meta^] tag [[html]] MUST have <code>name</code> and <code>content</code>
+					attributes that conform to the following definition:</p>
+
+				<dl>
+					<dt id="viewport-name-attr">name</dt>
+					<dd>The value of the [^meta/name^] attribute [{html]] after <a data-cite="xml#AVNormalize"
```
```suggestion
					<dd>The value of the [^meta/name^] attribute [[html]] after <a data-cite="xml#AVNormalize"
```

epub33/core/index.html
```diff
@@ -10592,6 +10607,140 @@ <h6>The <code>-epub-text-underline-position</code> property</h6>
...
+								<td>#x20 ;</td>
+							</tr>
+						</table>
+					</dd>
+				</dl>
+
+				<p>This specification only defines <a href="#viewport.ebnf.value">value</a> requirements for the
+						<code>height</code> and <code>width</code> properties (see <a href="#sec-fxl-content-dimensions"
+					></a>). It imposes no restrictions on other property names and values except that they MUST NOT
+					contain <a href="#viewport.ebnf.sep_char">separator characters</a> or the <a
```
```suggestion
					contain <a href="#viewport.ebnf.sep-char">separator characters</a> or the <a
```

epub33/core/index.html
```diff
@@ -10592,6 +10607,140 @@ <h6>The <code>-epub-text-underline-position</code> property</h6>
...
+
+				<div class="note">
+					<p>Although the <code>viewport</code>
+						<code>meta</code> tag allows EPUB creators to use properties other than <code>height</code> and
+							<code>width</code>, such use is strongly discouraged. Setting other properties may have
+						unintended consequences on the rendering of [=fixed-layout documents=].</p>
+				</div>
+
+				<div class="note">
+					<p class="note">The authoring requirements in this section apply <em>after</em>
```
```suggestion
					<p>The authoring requirements in this section apply <em>after</em>
```

epub33/rs/index.html
```diff
@@ -1588,25 +1588,44 @@ <h4>Initial containing block dimensions</h4>
...
 							<p>
-								<span id="confreq-fxl-rs-xhtml-icb-def"> Reading systems MUST create the initial
+								<span id="confreq-fxl-rs-xhtml-icb-def">Reading systems MUST create the initial
 									containing block (ICB) using the width and height expressions declared in the
 										<code>viewport</code>
 									<code>meta</code> tag for [=XHTML content documents=], as defined in <a
 										data-cite="epub-33#sec-fxl-icb-html">Expressing in HTML</a> [[epub-33]].</span>
-								<span id="confreq-fxl-rs-xhtml-icb" data-tests="#fxl-xhtml-icb">They MUST clip content
-									positioned outside of the ICB.</span>
+								<span id="confreq-fxl-rs-xhtml-icb" data-tests="#fxl-xhtml-icb,fxl-xhtml-icb_multi">They
```
```suggestion
								<span id="confreq-fxl-rs-xhtml-icb" data-tests="#fxl-xhtml-icb,#fxl-xhtml-icb_multi">They
```

epub33/rs/index.html
```diff
@@ -1588,25 +1588,44 @@ <h4>Initial containing block dimensions</h4>
...
+									MUST clip content positioned outside of the ICB.</span>
 							</p>
-							<p id="confreq-fxl-rs-xhtml-icb-ratio"> When the ICB aspect ratio does not match the aspect
+							<p>If the width or height values in the <code>viewport</code>
+								<code>meta</code> tag contain a non-numeric character but start with a number (e.g., the
+								value includes a unit of length declaration such as "500px"), the number prefix is used
+								as the pixel value.</p>
```
iherman: Is it "number prefix SHOULD be used as the pixel value" or "number prefix MUST be used as the pixel value"? We should say this. (I would opt for the former.)

iherman (changes requested):
I think all these comments are fairly simple. I would propose to handle these and the merge this. We have beaten this issue to death, we have a WG agreement to do this, and if there are minor issues we can handle them separately.

iherman:
Just note for archival purposes: this PR grew out of the extensive discussions in #2341; it was simpler to put everything into a new PR rather than carry on over there. But the technical discussions in #2341 are certainly "part" of this PR, too.

epub33/rs/index.html
mattgarrish: I'm fine either way. If we say "should", does that leave the potential to use non-pixel values, though? Do we need to say something like:

> the number prefix SHOULD be used as the pixel value, otherwise the value MUST be treated as invalid

iherman: Agreed; I have made the change.

PR #2352, opened 2022-07-05
Reports/add gca

clapierre:
Here are all the Publishers, Vendors and Bookstores that are adding in this accessibility metadata based on Benetech's GCA program.  These publishers, vendors and bookstores were also also asked if they had any objections to their name being added to this report.


mattgarrish:
Thanks as always @clapierre ! You've singlehandedly passed all our accessibility metadata except that pesky certifierReport...

PR #2358, opened 2022-07-12
Clarify svg title phrasing content restrictions

### LINKED ISSUE #2355: What is "XHTML content document Phrasing content"?
The [SVG content document definition](https://www.w3.org/TR/epub-33/#sec-svg-restrictions) has this requirement:

> The [[svg](https://www.w3.org/TR/epub-33/#bib-svg)] [title](https://www.w3.org/TR/SVG/struct.html#TitleElement) element MUST contain only valid [XHTML content document Phrasing content](https://www.w3.org/TR/epub-33/#sec-xhtml-req).

The link just redirects back to the general xhtml requirements, however; there are no references to phrasing content.

I assume this is supposed to mean HTML phrasing content that conforms to the XHTML content document restrictions, but it's not clear at all from the description.

It also makes it sound like you have to have an xhtml namespaced element even if the title is text only. I know the phrasing content definition includes text, but the description makes it sound like all content needs to be in the xhtml namespace.

Could we perhaps rewrite this more simply like:

> The [[svg](https://www.w3.org/TR/epub-33/#bib-svg)] [title](https://www.w3.org/TR/SVG/struct.html#TitleElement) element MAY contain [[html]] [phrasing content](https://www.w3.org/TR/epub-33/#sec-xhtml-req)<strike> that conforms to [6.1.4 XHTML requirements](https://www.w3.org/TR/epub-33/#sec-xhtml-req)</strike>. It MUST NOT contain elements from other namespaces.

mattgarrish:
This PR implements the text in https://github.com/w3c/epub-specs/issues/2355#issue-1301109618

I'm still not sure I see the point of the restriction, though.

Fixes #2355

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2358.html" title="Last updated on Jul 14, 2022, 10:22 AM UTC (1cfd608)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2358/df86d5a...1cfd608.html" title="Last updated on Jul 14, 2022, 10:22 AM UTC (1cfd608)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5981,10 +5981,9 @@ No Entry</pre>
 							</ul>
 						</li>
 						<li>
-							<p id="confreq-svg-title">The [[svg]] <a
-									href="https://www.w3.org/TR/SVG/struct.html#TitleElement"><code>title</code></a>
-								element MUST contain only valid <a href="#sec-xhtml-req">XHTML content document Phrasing
-									content</a>.</p>
+							<p id="confreq-svg-title" data-cite="svg">The [[svg]] [^title^] element MAY contain [[html]]
+								[=phrasing content=] that conforms to <a href="#sec-xhtml-req"></a>. It MUST NOT contain
+								elements from other namespaces.</p>
 						</li>
 					</ul>
 				</section>
@@ -11440,9 +11439,9 @@ EPUB/images/cover.png</pre>
 				<p>This appendix registers the media type <code>application/epub+zip</code> for the EPUB Open Container
 					Format (OCF).</p>
 
-				<p>An OCF ZIP container, or EPUB container, file is a container technology based on the [[zip]]
-					archive format. It is used to encapsulate the EPUB publication. OCF and its related standards are
-					maintained and defined by the <a href="https://www.w3.org">World Wide Web Consortium</a> (W3C).</p>
+				<p>An OCF ZIP container, or EPUB container, file is a container technology based on the [[zip]] archive
+					format. It is used to encapsulate the EPUB publication. OCF and its related standards are maintained
+					and defined by the <a href="https://www.w3.org">World Wide Web Consortium</a> (W3C).</p>
 
 				<dl class="variablelist">
 					<dt>MIME media type name:</dt>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5982,8 +5982,7 @@ No Entry</pre>
 						</li>
 						<li>
 							<p id="confreq-svg-title" data-cite="svg">The [[svg]] [^title^] element MAY contain [[html]]
-								[=phrasing content=] that conforms to <a href="#sec-xhtml-req"></a>. It MUST NOT contain
-								elements from other namespaces.</p>
+								[=phrasing content=]. It MUST NOT contain elements from other namespaces.</p>
 						</li>
 					</ul>
 				</section>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5981,8 +5981,9 @@ No Entry</pre>
 							</ul>
 						</li>
 						<li>
-							<p id="confreq-svg-title" data-cite="svg">The [[svg]] [^title^] element MAY contain [[html]]
-								[=phrasing content=]. It MUST NOT contain elements from other namespaces.</p>
+							<p id="confreq-svg-title" data-cite="svg">If the [[svg]] <code>title</code> element contains
+								marked-up text, the markup MUST contain only elements declared in the <a
+									data-cite="dom#html-namespace">HTML namespace</a> [[dom]].</p>
 						</li>
 					</ul>
 				</section>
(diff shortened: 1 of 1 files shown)


iherman (approved):
To be honest, I think the first sentence is unnecessary, because it repeats what the SVG spec already says. Per SVG spec a `title` element may include just about anything using an XML namespace, and the first sentence is just repeating that. The only important restriction, consistent with the restriction on foreign objects, is that only the HTML namespace is acceptable.

The first sentence is, per se, not wrong, so if we are worried about backward compatibility, we may just as well keep it, although it seems that epubcheck did not do anything with the content of the title element so far...

mattgarrish:
The spec sort of alludes to elements defaulting into the html namespace in a note, although I'm not sure why this would be true except maybe when embedded in html:

> The HTML parser treats all markup within ‘[title](https://www.w3.org/TR/SVG2/struct.html#TitleElement)’ and ‘[desc](https://www.w3.org/TR/SVG2/struct.html#DescElement)’ the same way it treats markkup in an HTML fragment; most elements will be assigned to the HTML namespace.

I find these one-type requirements awkward to phrase, especially since it's not required to use any HTML markup. I guess we could try:

> If the [svg] `title` element contains marked-up text, the markup MUST contain only [HTML elements](https://html.spec.whatwg.org/multipage/infrastructure.html#html-elements) [html].

iherman:
> The spec sort of alludes to elements defaulting into the html namespace in a note, although I'm not sure why this would be true except maybe when embedded in html:
> 
> > The HTML parser treats all markup within ‘[title](https://www.w3.org/TR/SVG2/struct.html#TitleElement)’ and ‘[desc](https://www.w3.org/TR/SVG2/struct.html#DescElement)’ the same way it treats markkup in an HTML fragment; most elements will be assigned to the HTML namespace.

Yep, I found the SVG spec a bit awkward in this respect (let alone the fact that I doubt SVG implementation really do too much with titles...)

> 
> I find these one-type requirements awkward to phrase, especially since it's not required to use any HTML markup. I guess we could try:
> 
> > If the [svg] `title` element contains marked-up text, the markup MUST contain only [HTML elements](https://html.spec.whatwg.org/multipage/infrastructure.html#html-elements) [html].

Yeah, better. But isn't it correct, per SVG spec, that it must be in the HTML namespace, ie, with an explicit xmlns statement? That is the way I understand the SVG spec... It may be better to make that explicit in our case, too...

mattgarrish:
> But isn't it correct, per SVG spec, that it must be in the HTML namespace

Yes, the HTML elements definition says its all elements in the http://www.w3.org/1999/xhtml namespace. We could make it even more explicit with:

> If the [svg] title element contains marked-up text, the markup MUST contain only elements declared in the [HTML namespace](https://infra.spec.whatwg.org/#html-namespace) [dom].

I don't think we have to be that explicit, since we're using the xml flavour of svg, but no harm, either. You could write non-namespaced elements when embedded in HTML (for the HTML parser), but xml validation should flag them as non-conforming to the schema since they wouldn't be from a foreign namespace.

But I can't find an example of anyone using html in a title anywhere. We could probably restrict the title to text if this is really to make reading system developer life simpler, since that's probably what almost all SVGs use anyway.

Testing some more with epubcheck, though, it enforces the weird all-or-nothing approach the original prose made me think of. If you use anything but text in a title, all the content has to be in an html element to be valid:

```<title><span  xmlns="http://www.w3.org/1999/xhtml">My <em>big scary</em> SVG!</span></title>```

You can't mix text and html elements like this or you get an error:

```<title>My <em xmlns="http://www.w3.org/1999/xhtml">big scary</em> SVG!</title>```

Given it took me a while to finally stumble on how to make epubcheck not complain, I'd be even more amazed if there are any EPUBs that have html markup in their titles.

iherman:
> > But isn't it correct, per SVG spec, that it must be in the HTML namespace
> 
> Yes, the HTML elements definition says its all elements in the http://www.w3.org/1999/xhtml namespace. We could make it even more explicit with:
> 
> > If the [svg] title element contains marked-up text, the markup MUST contain only elements declared in the [HTML namespace](https://infra.spec.whatwg.org/#html-namespace) [dom].
> 
> I don't think we have to be that explicit, since we're using the xml flavour of svg, but no harm, either. You could write non-namespaced elements when embedded in HTML (for the HTML parser), but xml validation should flag them as non-conforming to the schema since they wouldn't be from a foreign namespace.

I would opt for using the text above, and call it a day...

> 
> But I can't find an example of anyone using html in a title anywhere. We could probably restrict the title to text if this is really to make reading system developer life simpler, since that's probably what almost all SVGs use anyway.
> 
> Testing some more with epubcheck, though, it enforces the weird all-or-nothing approach the original prose made me think of. If you use anything but text in a title, all the content has to be in an html element to be valid:
> 
> `<title><span xmlns="http://www.w3.org/1999/xhtml">My \<em\>big scary\</em\> SVG!</span></title>`
> 
> You can't mix text and html elements like this or you get an error:
> 
> `<title>My <em xmlns="http://www.w3.org/1999/xhtml">big scary\</em\> SVG!</title>`
> 
> Given it took me a while to finally stumble on how to make epubcheck not complain, I'd be even more amazed if there are any EPUBs that have html markup in their titles.

So would I. I would go even further: I would be amazed if any HTML+SVG implementation would ever use the styling in the svg title element in displaying the tooltip, regardless of EPUB, and that anybody around would ever use this feature. Taking into account that the title element in HTML accepts only text as its content, I think that the richer structure of the SVG title element is an aspirational definition for now...

But that is not our problem after all. Our reduced text just puts the title element in line with the rest of the SVG restrictions, and that is all we should do.

iherman:
I forgot to react on this:

> But I can't find an example of anyone using html in a title anywhere. We could probably restrict the title to text if this is really to make reading system developer life simpler, since that's probably what almost all SVGs use anyway.

I would not bother. It puts an extra load of checking on the RS Implementers, whereas currently RS implementations can simply push the corpse into WebViews' courtyard without any further ado. Let the WebViews implementers deal with this.

We could put a note into our spec warning implementers that if they care about interoperability they should stick with pure text.

PR #2363, opened 2022-07-20
Update access mode sufficient technique to emphasize single values

### LINKED ISSUE #2302: Prioritize single accessModeSufficient values
As discussed on the mailing list starting with @madeleinerothberg's [comment](https://lists.w3.org/Archives/Public/public-epub-wg/2022May/0030.html):

> The thing about accessModeSufficient…. It’s the entries with single modes that are most useful to users, probably, because it is the users with restrictions on their perceptual modes who most need to know if they can use a particular resource. We can list out multiple sets of accessModeSufficient and there is no reason not to, but the most impactful entries will be those with single modes or the LACK of them.

Should we recommend the single values be listed and make other sets optional in the techniques document, or should we only emphasize the importance of single values?

mattgarrish:
I took another look at the access mode sufficient technique and made a couple of minor changes to emphasize single-value declarations.

First, I've moved the paragraph that explains these are the most important up higher in the technique. You had to read through the example for a while before finding it, which was suboptimal.

Second, I've added a list of the two single-value modes that are most common for epubs: textual and auditory. I've also explained when these are set.

The only question I have @GeorgeKerscher is if I've got the right WCAG level for stating that a publication is screenreader friendly. If you meet WCAG Level A, you should be able to set "textual" as a single sufficient access mode, correct? I don't believe AA adds anything more for TTS playback, but if you can confirm that would be a big help.

Also, if anyone spots anything else that needs improving, please let me know now. Issue #2302 has been open for a long time now, so I'd like to close it off with this pull request.

Fixes #2302

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -244,8 +244,8 @@
 					<li><code>textual</code> &#8212; setting this value indicates that the publication is screen-reader
 						friendly (i.e., all the content is available for text-to-speech rendering by an assistive
 						technology). EPUB creators can set this value for EPUB publications that conform to [[WCAG2]] <a
-							href="https://www.w3.org/TR/WCAG2/#cc1">Level A</a> as there have to be textual alternatives
-						for non-text content to meet this threshold.</li>
+							href="https://www.w3.org/TR/WCAG2/#cc1">Level A</a>, for example, as there have to be
+						textual alternatives for non-text content to meet this threshold.</li>
 					<li><code>auditory</code> &#8212; setting this value indicates that pre-recorded audio is available
 						for the publication. EPUB creators can set this value for EPUB publications that provide media
 						overlays for all the content.</li>
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -244,8 +244,8 @@
 					<li><code>textual</code> &#8212; setting this value indicates that the publication is screen-reader
 						friendly (i.e., all the content is available for text-to-speech rendering by an assistive
 						technology). EPUB creators can set this value for EPUB publications that conform to [[WCAG2]] <a
-							href="https://www.w3.org/TR/WCAG2/#cc1">Level A</a>, for example, as there have to be
-						textual alternatives for non-text content to meet this threshold.</li>
+							href="https://www.w3.org/TR/WCAG2/#cc1">Level A</a> or higher, for example, as there must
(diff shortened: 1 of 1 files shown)


mattgarrish:
- [Preview](https://raw.githack.com/w3c/epub-specs/a11y/issue-2302/epub33/a11y-tech/index.html)
- [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/a11y/issue-2302/epub33/a11y-tech/index.html)

clapierre (approved):
really like these changes Matt.

GeorgeKerscher:
Yes, if the publication meets WCAG at the A level, it would be screen reader friendly. I wonder if a publication may meet accessModeSufficient="textual" but not meet WCAG at the A level for other reasons. Is this something we should consider?

mattgarrish:
> I wonder if a publication may meet accessModeSufficient="textual" but not meet WCAG at the A level for other reasons. Is this something we should consider?

I think that will make the guidance tougher to understand as we'd have to get into the details of which success criteria have to pass. It's also hopefully not a common occurrence where someone fails level A and still follows the discovery reporting.

We're not saying you have to meet level A to set the value, only if a publication meets that level then you can set it. I can emphasize this by adding "for example" to the sentence:

> EPUB creators can set this value for EPUB publications that conform to [[WCAG2](https://raw.githack.com/w3c/epub-specs/a11y/issue-2302/epub33/a11y-tech/index.html#bib-wcag2)] [Level A](https://www.w3.org/TR/WCAG2/#cc1), for example, as there have to be textual alternatives for non-text content to meet this threshold.

madeleinerothberg:
These changes look good to me. It is certainly possible that a resource could be screen-reader friendly without passing WCAG A, but as Matt said it would make our guidance more complex and address a fairly limited group of resources. Given that many WCAG criteria don't apply directly to the EPUB file but are reading system requirements, I think the draft text is fine.

GeorgeKerscher:
Looks good.

mattgarrish:
I'm going to go ahead and merge this now. If anyone wants to suggest any other edits to clarify the AMS property, please open a new issue.

PR #2366, opened 2022-07-22
Add all former eds and move list to the acknowledgements

mattgarrish:
As the title says.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2366.html" title="Last updated on Jul 23, 2022, 3:29 PM UTC (340db0c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2366/319d4cb...340db0c.html" title="Last updated on Jul 23, 2022, 3:29 PM UTC (340db0c)">Diff</a>

### DIFF
--- a/epub33/common/acknowledgements-dedication.html
+++ b/epub33/common/acknowledgements-dedication.html
@@ -4,8 +4,7 @@
 		been there every step of the way, from the very first OEB 1.0 in 1999 to today's EPUB 3.3. None of this
 		would have happened without Garth's vision, knowledge, and preternatural good nature. We dedicate
 		EPUB 3.3 to his memory. We are forever in your debt, Garth.</p>
-	<p>The following members of the EPUB 3 Working Group contributed to the development of this
-		specification:</p>
+	<p>The following members of the EPUB 3 Working Group contributed to the development of this document:</p>
 	<ul class="ack">
 		<li>Will AWAD (Newgen Knowledgeworks)</li>
 		<li>Sofia Bautista (Legible Media Inc.)</li>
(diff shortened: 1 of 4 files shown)


murata2makoto:
I am a bit sad not to see Markus's name here.

dauwhe (approved):

shiestyle (approved):

iherman (approved):

iherman:
> I am a bit sad not to see Markus's name here.

You mean in the group? I agree... But his name does appear in the list of former editors (as it should).

murata2makoto:
> > I am a bit sad not to see Markus's name here.
> 
> You mean in the group? I agree... But his name does appear in the list of former editors (as it should).

I overlooked it.

I had a look at EPUB 3.0 specifications.  Even if we eliminate CFIs (which are dropped from EPUB 3.3), two former editors (Elika J. Etemad and I) are missing.

P.S.  I overlooked Elika's name.

mattgarrish:
> Even if we eliminate CFIs (which are dropped from EPUB 3.3)

This acknowledgements list is only for the core and reading systems specs, as these are brand new and combine the various releases of Publications/Packages, Content Docs, OCF, and Media Overlays plus the EPUB 3.2 spec. There are a lot of people to acknowledge, so having them all listed at the top is a bit of clutter, especially if it keeps growing with future releases.

You're still acknowledged in the former editor's list at the top of the Overview, @murata2makoto, along with the others that worked on it, as has not been absorbed. The same is true for the former editors of CFI and Multiple Renditions.

Sorry if that wasn't clear.

I can add some additional prose that it's for the combined specs if that'll help. Something like:

> the former editors of the EPUB 3 specifications that were merged to make this document:

mattgarrish:
Or we can drop the former editors from the overview and use this acknowledgements for it, too, plus @murata2makoto

iherman:
> Or we can drop the former editors from the overview and use this acknowledgements for it, too, plus @murata2makoto

Let us do this for the overview, too.

murata2makoto:
Sorry.  I did not understand that we also have "EPUB 3.3 Overview".  I will raise a separate issue about that.

My name was listed as an editor in the original "EPUB 3 Overview", not because I contribute to editing the document, but because I contributed to the global language support of EPUB 3.  

>Or we can drop the former editors from the overview and use this acknowledgements for it, too, plus @murata2makoto

I would be honored.

PR #2369, opened 2022-07-23
Added the epubcfi reference to the media type registrations

### LINKED ISSUE #2367: Canonical Fragment Identifiers in media type registrations
In the registration template of EPUB media types, canonical fragment identifiers are mentioned, but no references are provided.

iherman:
Fix #2367

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2369.html" title="Last updated on Jul 24, 2022, 4:03 AM UTC (0c611b8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2369/a959518...0c611b8.html" title="Last updated on Jul 24, 2022, 4:03 AM UTC (0c611b8)">Diff</a>

### DIFF
--- a/epub33/core/biblio.js
+++ b/epub33/core/biblio.js
@@ -19,6 +19,22 @@ var biblio = {
 		"date": "26 June 2014",
 		"publisher": "IDPF"
 	},
+	"epubcfi" : {
+		"authors": [
+			"Peter Sorotokin",
+			"Garth Conboy",
+			"Brady Duga",
+			"John Rivlin",
+			"Don Beaver",
+			"Kevin Ballard",
+			"Alastair Fettes",
+			"Daniel Weck"
+		],
+		"title": "EPUB Canonical Fragment Identifier (epubcfi) Specification",
+		"href": "https://idpf.org/epub/linking/cfi/epub-cfi-20111011.html",
+		"publisher": "IDPF",
+		"date": "11 October 2011"
+	},
 	"epubpackages-32": {
 		"authors":[
 		"Matt Garrish",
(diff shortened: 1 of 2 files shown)


murata2makoto (approved):

mattgarrish:
Can we use a respec reference for this? It works in the appendixes, but aren't the actual IANA template registrations all text documents, so you'll get nothing but "[epubcfi]" with no further information? (See the current [epub+zip](https://www.iana.org/assignments/media-types/application/epub+zip) template where we had a URL to the linking registry.)

In other words, do we needed to include a reference URL like:

> ```<p>EPUB Canonical Fragment Identifiers are custom fragment identifiers that can resolve to <code>application/epub+zip</code> and <code>application/oebps-package+xml</code> documents. These identifiers are defined at https://idpf.org/epub/linking/cfi/</p>```

I'm more concerned that using respec references leads to a normative reference to the IDPF CFI spec, which can't be a good thing.

(If we do change, there are a couple of other instances of respec-formatted references that it might also be good to be rid of: rfc4389 is a bracketed reference in one and a hyperlink in the other, and there's also a [zip] reference.)

iherman:
> (If we do change, there are a couple of other instances of respec-formatted references that it might also be good to be rid of: rfc4389 is a bracketed reference in one and a hyperlink in the other, and there's also a [zip] reference.)

Yeah, I was wondering about this issue myself, and it was the presence of those references that made me think that using [[...]] is all right. But you are right that making a proper reference may be a better idea. That should include rfc as well.

iherman:
I have made the reference explicit, and did the same for the rfc and zip references.

I left epubcfi in the biblio.js file. It does not do any harm, and may come handy if we ever touch that thing again (which is an option raised in the locator task force).

mattgarrish (approved):

PR #2375, opened 2022-07-27
Clarify not to set access modes for non-essential content

mattgarrish:
This pull request adds a paragraph clarifying that you do not set access modes for content that does not contain information necessary to read a publication. For example, a book with only a cover image or publisher logo does not have a visual access mode, as these are not important to reading the book itself.

This change just reinforces what the schema.org [vocabulary definitions](https://w3c.github.io/a11y-discov-vocab/#auditory) already say.

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -206,6 +206,14 @@
 &lt;/meta></pre>
 				</aside>
 
+				<p><strong>Do not</strong> list acess modes for content that does not contain information necessary to
+					understand to read a publication. Most EPUB publications contain cover images, for example, but it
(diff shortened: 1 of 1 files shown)


clapierre (approved):
Great addition. one typo "access" in "<strong>Do not</strong> list acess modes"

madeleinerothberg:
Agreed that this a great addition -- one other typo is "to understand to read". Matt is pumping out documents at an unprecedented rate!

madeleinerothberg (approved):

mattgarrish:
Ack, ya, too much hopping between documents today. I thought I could get away without a spellcheck...

The sentence now reads:

> Do not list access modes for content that does not contain information necessary to understand a publication.

PR #2379, opened 2022-07-28
Re-implement vocabulary association mechanisms as an algorithm

### LINKED ISSUE #2378: Normative statements for the vocabulary association mechanism?
The [§11 Vocabulary Association Mechanism](https://w3c.github.io/epub-specs/epub33/rs/#sec-vocab-assoc) section is altogether a _MAY_; it starts with:

> [Reading systems](https://www.w3.org/TR/epub-33/#dfn-epub-reading-system) MAY support [vocabulary association mechanisms](https://www.w3.org/TR/epub-33/#sec-vocab-assoc) [[epub-33](https://w3c.github.io/epub-specs/epub33/rs/#bib-epub-33)] in [EPUB content documents](https://www.w3.org/TR/epub-33/#dfn-epub-content-document).

We actually know that RS-s almost never implement this. It is therefore a bit disturbing to see the section full of _MUST_ statements (which all would also require tests). It all sounds very "aspirational" to me. Wouldn't it be better to rephrase this section? Maybe a simple solution to replace all the _MUST_ statements by plain, lower case "must"-s?

We had a similar situation with, e.g., MOL, which is also an optional feature, but the section contains lots of _MUST_-s. But this section is not optional in terms of _SHOULD_, but in terms of _MAY_. Also, we know that MOL is implemented by RS-s (although not by all).

### LINKED ISSUE #2382: Clarify that the scheme attribute does not have a default vocabulary.
It's the only attribute that requires a prefix for its values.

mattgarrish:
Not looking for approval at this time. Just curious for early feedback on whether this makes sense as a replacement.

<strike>I kept the one requirement not to blow up when encountering unknown prefixes.</strike>

Fixes #2378 
Fixes #2382 

* [Preview](https://cdn.statically.io/gh/w3c/epub-specs/experiment/issue-2378/epub33/rs/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/experiment/issue-2378/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2379.html" title="Last updated on Aug 6, 2022, 12:11 PM UTC (30ed27d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2379/706c5a7...30ed27d.html" title="Last updated on Aug 6, 2022, 12:11 PM UTC (30ed27d)">Diff</a>

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2177,10 +2177,6 @@
 
 			<p>Reading systems do not have to [=url parser | parse the resulting URL=] [[url]] or attempt to dereference
 				the resource.</p>
-
-			<p>As changes to the reserved prefixes and updates to reading systems are not always going happen in
-				synchrony, reading systems MUST NOT fail when encountering unrecognized prefixes (i.e., not reserved and
-				not declared using the <code>prefix</code> attribute).</p>
 		</section>
 		<section id="sec-epub-rs-conf-backward">
 			<h3>Backward compatibility</h3>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2108,7 +2108,7 @@
 			<ol class="algorithm">
 				<li>
 					<p>Let <var>baseURL</var>, <var>expandedURL</var>, <var>propertyPrefix</var>, and
-							<var>propertyName</var> be empty [=strings=].</p>
+							<var>propertyReference</var> be empty [=strings=].</p>
 					<details class="explanation">
 						<summary>Explanation</summary>
 						<p>In this algorithm:</p>
@@ -2117,25 +2117,28 @@
 								assigned by the EPUB creator, corresponds to a default vocabulary, or comes from a
 								reserved prefix.</li>
 							<li><var>expandedURL</var> will hold the final URL that results from the concatenation of
-								the base URL and property name.</li>
-							<li><var>propertyPrefix</var> will hold the prefix for the property (i.e., the value before
-								the colon). Properties from a default vocabulary will not have a prefix, but all others
-								will.</li>
-							<li><var>propertyName</var> will hold the name of the property (i.e., the value after the
-								prefix). All valid properties require a name.</li>
+								the base URL and property reference.</li>
+							<li><var>propertyPrefix</var> will hold the property's <a
+									data-cite="epub-33#property.ebnf.prefix">prefix</a> [[epub-33]] (i.e., the value
+								before the colon). Properties from a default vocabulary will not have a prefix, but all
+								others will.</li>
+							<li><var>propertyReference</var> will hold the property's <a
+									data-cite="epub-33#property.ebnf.reference">reference</a> [[epub-33]] (i.e., the
+								value after the prefix). All valid properties require a reference.</li>
 						</ul>
 					</details>
 				</li>
 				<li>
-					<p>Obtain the values of <var>propertyPrefix</var> and <var>propertyName</var> as follows:</p>
+					<p>Obtain the values of <var>propertyPrefix</var> and <var>propertyReference</var> as follows:</p>
 					<ol>
 						<li>
-							<p>If <var>value</var> does not contain a colon (U+003A), set <var>propertyName</var> to
-									<var>value</var> and continue processing at the next step.</p>
+							<p>If <var>value</var> does not contain a colon (U+003A), set <var>propertyReference</var>
+								to <var>value</var>.</p>
 							<details class="explanation">
 								<summary>Explanation</summary>
 								<p>If a <code>property</code> value does not have a colon, it does not have a prefix. In
-									this case, the value is drawn from the default vocabulary as described in the next step of the algorithm.</p>
+									this case, the value is drawn from the default vocabulary as described in the next
+									step of the algorithm.</p>
 							</details>
 						</li>
 						<li>
@@ -2147,38 +2150,26 @@
 									set.</p>
 							</details>
 						</li>
-						<li id="property-exp-colon-no-slash">
-							<p>Otherwise, if <var>value</var> does not contain a slash (U+002F) and <var>value</var>
-								ends with a colon or has more than one colon, the value is invalid. Return <a
-									data-cite="infra#nulls">null</a>.</p>
-							<details class="explanation">
-								<summary>Explanation</summary>
-								<p>If a <code>property</code> does not contain any slashes, then the colon represents
-									the prefix delimiter. It is not valid for the value to end with a colon, as that
-									would indicate there is only a prefix and no name.</p>
-								<p>Likewise, there cannot be more than one colon, as the name must be a
-									[=path-relative-scheme-less-URL string=] (i.e., any colons must be after the first
-									[=URL-path-segment string=]).</p>
-								<p>For example, the following is a valid <code>property</code> value as the second colon
-									follows a slash: <code>foo:bar/ba:z</code>. This example shows an invalid value, as
-									the second colon could represent a scheme: <code>foo:bar:baz/qux</code>.</p>
-							</details>
-						</li>
 						<li>
 							<p>Otherwise, split <var>value</var> on the first colon and set <var>propertyPrefix</var> to
-								the string before the colon and set <var>propertyName</var> to the string after the
+								the string before the colon and set <var>propertyReference</var> to the string after the
 								colon.</p>
 						</li>
 					</ol>
-					<p>If <var>propertyName</var> contains a colon, and that colon occurs before a slash character, the
-						value is invalid. Return <a data-cite="infra#nulls">null</a>.</p>
+					<p>If <var>propertyReference</var> is an empty string or is not a valid
+						[=path-relative-scheme-less-URL string=], as required by the <a
+							data-cite="epub-33#property.ebnf.reference"><code>property</code> data type definition</a>
+						[[epub-33]], the value is invalid. Return <a data-cite="infra#nulls">null</a>.</p>
 					<details class="explanation">
 						<summary>Explanation</summary>
-						<p>If a <var>propertyName</var> has a colon at this stage, then there were multiple colons in
-							the original value. The first delimited the prefix. Any subsequent ones left in the name
-							must occur after a slash, otherwise the name is not a valid [=path-relative-scheme-less-URL
-							string=]. (Refer to the explanation for <a href="#property-exp-colon-no-slash">substep 3</a>
-							for additional explanation about valid path-relative-scheme-less-URL strings.)</p>
+						<p>If the reference is an empty string, the colon separating the prefix from the reference must
+							have been at the end of <var>value</var>. A prefix without a reference is invalid.</p>
+						<p>The path-relative-scheme-less-URL string definition has a number of restrictions that apply
+							to the reference whether it has a prefix or not.</p>
+						<p>For example, the reference must not begin with a [=URL-scheme string=] followed by a colon.
+							This restriction means that the following is not a valid <code>property</code> value as the
+							second colon could represent a scheme: <code>foo:bar:baz/qux</code>.</p>
+						<p>There are also restrictions on what characters must be percent encoded.</p>
 					</details>
 				</li>
 				<li>
@@ -2251,7 +2242,7 @@
 					</details>
 				</li>
 				<li>Set <var>expandedURL</var> to the concatenated values of <var>baseURL</var> and
-						<var>propertyName</var>. Do not add a separator when concatenating the values.</li>
+						<var>propertyReference</var>. Do not add a separator when concatenating the values.</li>
 				<li>
 					<p>If <var>expandedURL</var> is not a [=valid URL string=], the value is invalid. Return <a
 							data-cite="infra#nulls">null</a>.</p>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2178,6 +2178,9 @@
 							<ol data-cite="epub-33">
 								<li>If <var>attr</var> is [^/epub:type^] [[epub-33]], set <var>baseURL</var> to
 										<br /><code>http://idpf.org/epub/vocab/structure/#</code></li>
+								<li>Otherwise, if <var>elem</var> is the package document [^meta^] element [[epub-33]]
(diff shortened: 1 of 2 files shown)


iherman:
I think that is great. It would get around the issue of the awkward _MUST_ statements and it works well.

Thanks!

mattgarrish:
Okay, I think this is good to go now. The last commit mostly added explanations for all the steps. I also made a few tweaks to infra-ize things a bit more.

The two changes of note are:

- I've added an optional item to the epub:type processing list to get an expanded url. It was a bit strange to describe processing the attribute without mentioning that you could get an expanded value to process.
- I also added a paragraph to the top of the algorithm clarifying that it's not required to expand property values to use them. Reading systems that don't support the vocabulary association mechanisms can just treat them like strings.

mattgarrish:
And as reading the diff is a bit of a nightmare, here's a link to the current section if you want to manually compare against the preview: https://www.w3.org/TR/epub-rs-33/#sec-vocab-assoc

epub33/rs/index.html
```diff
@@ -2067,92 +2073,193 @@ <h2>Processing structural semantics</h2>
...
+					data-cite="infra#nulls">null</a> value is returned for invalid properties.</p>
+
+			<p>This algorithm takes the following arguments:</p>
+
+			<ul>
+				<li><var>value</var>: the <code>property</code> value to process.</li>
+				<li><var>attr</var>: the name of the attribute that contains <var>value</var>.</li>
+				<li><var>elem</var>: the name of the element that <var>attr</var> is attached to.</li>
+				<li><var>doc</var>: the document being processed (e.g., [=package document=], [=media overlay
+					document=], [=epub content document=]).</li>
```
iherman, line 2097: I am not sure why, but on my cloned version of the document these references do not appear as references, just plain texts. This may be some respec/github black art... :-( If it works on your side, ignore this.

epub33/rs/index.html
```diff
@@ -2067,92 +2073,193 @@ <h2>Processing structural semantics</h2>
...
 						<li>
-							<p id="confreq-rs-vocab-property-reference">If the property consists only of a reference,
-								concatenate the prefix URL associated with the <a href="#sec-default-vocab">default
-									vocabulary</a> to the reference.</p>
+							<p>If <var>value</var> does not contain a colon, set <var>propertyName</var> to
+									<var>value</var> and continue processing at the next step.</p>
+							<details class="explanation">
+								<summary>Explanation</summary>
+								<p>If a <code>property</code> value does not have a colon, it does not have a prefix. In
+									this case, the value is drawn from the default vocabulary.</p>
```
```suggestion
									this case, the value is drawn from the default vocabulary as described in the next step of the algorithm.</p>
```

epub33/rs/index.html
```diff
@@ -2067,92 +2073,193 @@ <h2>Processing structural semantics</h2>
...
-							</p>
+							<p>Otherwise, if <var>value</var> contains a single colon:</p>
+							<ol>
+								<li>
+									<p>If <var>value</var> begins or ends with a colon, the value is invalid. Return <a
+											data-cite="infra#nulls">null</a>.</p>
+								</li>
+								<li>
+									<p>Otherwise, let <var>tokens</var> be the result of [=strictly split a
+										string|splitting=] <var>value</var> using a colon (U+003A) as the delimiter. Set
```
iherman: Here again I do not see the infra reference in my local copy... :-(

epub33/rs/index.html
```diff
@@ -2067,92 +2073,193 @@ <h2>Processing structural semantics</h2>
...
+							</details>
 						</li>
 						<li>
-							<p id="confreq-rs-vocab-property-only-prefix">If the property consists only of a prefix
-								(i.e., there is no reference after the colon), the property is invalid and reading
-								systems MUST ignore it.</p>
+							<p>Otherwise, the value is invalid. Return <a data-cite="infra#nulls">null</a>.</p>
+							<details class="explanation">
+								<summary>Explanation</summary>
+								<p>A <code>property</code> value cannot have more than one colon.</p>
```
iherman: Is that specified somewhere? I thought a colon is an acceptable character in value... In which case the previous step should be modified to say that we split at the first occurrence of a colon...

iherman (approved):

iherman:
I have sketched out an implementation in typescript, just to check. Emphasis is on "sketch", but it helped me to follow the infra stuff:-)

epub33/rs/index.html
mattgarrish: I'm getting proper links in my local version. Are all references broken for you?

epub33/rs/index.html
mattgarrish: Must be a glitch either locally for you or that's been resolved now.

epub33/rs/index.html
mattgarrish: Ya, I over-interpreted the relative path requirement a bit, but it looks to be more complicated than either of us has right now.

According to the [EBNF for the data type](https://w3c.github.io/epub-specs/epub33/core/#sec-property-datatype), the value has to be a [path-relative-scheme-less-URL string](https://url.spec.whatwg.org/#path-relative-scheme-less-url-string).

It says in that definition that the value "does not start with: a [URL-scheme string](https://url.spec.whatwg.org/#url-scheme-string), followed by U+003A (:)".

So if you circle down through the many definitions you get:

- A path-relative-scheme-less URL must be a path-relative-URL string
- A path-relative-URL string must be zero or more URL-path-segment strings, separated from each other by U+002F (/), and not start with U+002F (/)
- A URL-path-segment string must be zero or more URL units excluding U+002F (/) and U+003F (?)

So wouldn't this make a colon illegal in the first URL-path-segment string? If you have one, it has to be after the first slash.

Do we want to try and spec that out, though, or can we leave the validity to the end? The ultimate arbiter is that you get a valid URL string. Multiple colons is a pretty niche case.

mattgarrish: Maybe we could test this after splitting. So, if there are one or more colons, split on the first. Then before moving to the next step say if `propertyName` contains a colon before a first slash, the value is invalid?

iherman: If I follow the definition for the property:

- path-relative-scheme-less-URL string -> cannot start with a URL-scheme string followed by a (:) but otherwise is a path-relative-URL string. 
- a URL-scheme is ASCII alpha, followed by ASCII alphanumeric or '+','-','.'
- path-relative-URL string allows for the ':' character.

I.e., the value indeed cannot be "abcd:efgh". But I believe it can be something like "abcd/efg:hij". E.g., I believe this is valid:

```
… prefix="test: http://www.example.org/"
… property="test:abcd/#efg:hij"
```

yielding `http://www.example.org/abcd/#efg:hij`.

This means, I believe, that we should be prepared for the case where the property contains more than one ':' character; meaning that the split must happen at the first occurrence of ':' and then we are done...

iherman: Just realized a practical example, which we all love: epubcfi. If a property includes an epubcfi fragment, it may have a ':' character... :-)

mattgarrish: The latest commit rewrites step 2 to account for multiple colons.

I couldn't use the infra splitting algorithm anymore, as that could lead to multiple chunks which would then have to be recombined. Seemed excessive. I have it now to split on the first colon and assign the strings before and after to the appropriate variables. I'm assuming splitting is widely enough known that we don't have to cite anything.

Anyway, if there are multiple colons, these changes should only allow valid cases to pass through - where the colon is not in the first segment of the name. I added a couple of examples to the definition to help explain this: "foo:bar/ba:z" is valid while "foo:bar:baz/qux" is not.

If you don't think I've got this right, though, let me know.

iherman: I am not sure the restriction is right: I believe `foo:bar#ba:z` is also a valid case. Ie, (3) would create a false negative, wouldn't it? There are URI schemes that do not rely on slash character (DID, ISBN, mailto)

I would rather go down the line of

- remove current (3)
- split at the first colon i.e., what is now (4)
- before setting the value of _propertyName_ check whether the value begins with a [URL Scheme string](https://url.spec.whatwg.org/#url-scheme-string) followed by a colon. If true, it is an error and return null, otherwise we set the _propertyName_ and we are done.

Would that work?

mattgarrish: > check whether the value begins with a [URL Scheme string](https://url.spec.whatwg.org/#url-scheme-string) followed by a colon

Maybe simplest of all would be to say that `propertyName` must be a valid path-relative-scheme-less-URL string, since that's where that exception comes from and will match up with the EBNF. I agree these details of what is valid are better left to the explanation.

mattgarrish: I've redone step 2 to make a few changes.

- step one no longer says to continue processing at the next step. Even if the value doesn't have a colon, that doesn't mean it's a valid PRSLU string.
- the former third step is gone
- after processing, `propertyReference` is checked that it is not an empty string or an invalid PRSLU

I also renamed `propertyName` to `propertyReference` to match the EBNF naming of the components.

mattgarrish: Hm, this still may not be right. Looking at the PRSLU definition, an empty string is valid. That means that a prefix-only property value is valid, which also matches the CURIE definition. For example, you could do: `<meta ... scheme="myScheme:"

I know we considered this invalid in the past, as made some weird hacks to create schemes that would have both a prefix and reference (marc:relators comes to mind). But I guess it's valid, so I should take that test out.

mattgarrish: See also https://github.com/w3c/epub-specs/issues/2382. I've fixed this, too.

iherman: Yes indeed, I know that in the RDFa/Turtle practice using a CURIE without reference (ie, re-using the prefix as a URL on its own right) is actually fairly frequent. But I thought you have some good reasons to disallow the trailing ':' case...

Otherwise, it looks good to me. Personally, I do not see any reason not to merge this and consider it done...

mattgarrish: > But I thought you have some good reasons to disallow the trailing ':' case...

Nope... 😄 

The unfortunate truth is that we lacked the requisite expertise to be implementing RDFa in the package document when we tried to build this framework for 3.0. We probably should have abandoned it completely when it was pointed out we were absuing the `about` attribute, but rewriting the whole framework at the last minute likely wouldn't have had much better results.

In this case, I think the definition of an optional prefix followed by a reference is what confused people. Whoever made the first claim that it meant the reference couldn't be an empty string was just accepted as gospel. I know this is the deepest I've ever dug into the fine print, so for that alone this issue has been insightful.

But, yes, I think this captures everything we need now. Do you think we need a group resolution before merging tihs, since we haven't discussed doing this kind of a rewrite? It's pretty substantive, at least in terms of removing normative requirements.

mattgarrish: I've added a final clarification to the data type definition that a prefix and colon with an empty reference is a valid value. I added this after the statement that an empty string is technically valid but non-conforming, as that's likely where the confusion arose from.

iherman: Yes, it would be good to have a WG approval. Except that... @wareid will be out for the next two weeks on vacations, and she should not be worrying about the WG. I do not know when @dauwhe is back, not sure that he will be around next week. Also, there is no real reason these days to hold WG calls, it is all quiet with some people making tests...

Let us wait if @dauwhe is back next week. What we could do is that you send a mail to the WG explaining what has been done, and leave a week to respond/object to the changes. I would hate to leave this PR open for several weeks...

mattgarrish: >  Also, there is no real reason these days to hold WG calls

Hm, ya, I wonder if I'm overthinking this. We're not adding new requirements, only loosening existing ones. And it's for an optional feature that we haven't heard has been implemented.

It's noted in the change log, too, so anyone keeping track of what's going on will find it. Maybe we should just go ahead an merge...

iherman: I would like to get at least @dauwhe 's reaction (@wareid is out on vacations)

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-08-04-epub#section2) on 2022-08-04

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-08-04-epub#resolution1):  The WG supports merging #2379 pending approval from the assigned reviewers..***

### 2. Re-implement vocabulary association mechanisms as an algorithm (pr epub-specs#2379)

_See github pull request [epub-specs#2379](https://github.com/w3c/epub-specs/pull/2379)._




_See github issue [epub-specs#2378](https://github.com/w3c/epub-specs/issues/2378)._




_See github issue [epub-specs#2382](https://github.com/w3c/epub-specs/issues/2382)._




**Dave Cramer:** there are a couple issues associated with this (#2378, #2382).  
… this is very technical, I wish mgarrish were here to explain.  
… doesn't really seems to change how the spec works, more an editorial change.  
… seems like they were just looking for WG approval before merging, even though most of us probably won't have strong opinions on this.  
… i'm generally in favor of explaining things in an algorithmic fashion rather than hard to understand paragraphs.  
… maybe we just say that we don't have objection to merging this pending github reviews?.  

> **Proposed resolution: The WG supports merging #2379 pending approval from the assigned reviewers..** *(Dave Cramer)*

> *Dave Cramer:* +1.

> *Matthew Chan:* +1.

> *Brady Duga:* +1.

> *Toshiaki Koike:* +1.

> *Shinya Takami (高見真也):* +1.

> ***Resolution #1: The WG supports merging #2379 pending approval from the assigned reviewers..***

</details>

iherman:
Per WG resolution, we need an approval of at least @shiestyle and @dauwhe (@wareid and, I presume, @rdeltour being on vacations...). Shinya, Dave, could you do this, please?

mattgarrish:
I think they gave us the green light to merge it once we are happy with it. I think the "reviewers" in this case means me and you, as it sound like the group is fine with whatever we resolve on for this given no one's hugely passionate about a feature that's only for internal reading system use... :smile:

PR #2380, opened 2022-07-28
Made the epub:type restrictions explicit and adapted the RS

### LINKED ISSUE #2377: "head" elements semantics?
The RS specification says, in [§6.1.1.2](https://w3c.github.io/epub-specs/epub33/rs/#sec-xhtml-structural-semantics)

> In addition to the requirements in [10. Processing structural semantics](https://w3c.github.io/epub-specs/epub33/rs/#sec-structural-semantics), a reading system MUST ignore semantics expressed on the [[html](https://w3c.github.io/epub-specs/epub33/rs/#bib-html)] [head](https://html.spec.whatwg.org/multipage/semantics.html#the-head-element) element or its descendants.

I am not sure what this requirement aims to achieve. The Structural Semantics section in the RS contains _MAY_ statements, it is a bit strange to restrict the HTML standard with such an optional set of statements. Also, this seems to suggest that, for example, the `<title>` element in the `<head>` must be ignored, which becomes very convoluted for an RS running in a browser. Finally, an alternative to semantics described by RDFa is to add a datablock (using a `<script>` element in the head) containing RDF metadata in JSON-LD (this is one of the proposed approaches to add schema.org metadata to an HTML file), and this restriction seems to disallow this for EPUB.

My proposal is to remove that section altogether.

Cc @dauwhe @mattgarrish @wareid

iherman:
Went roughly along the line of https://github.com/w3c/epub-specs/issues/2377#issuecomment-1197985804.

In the §10 of the RS I was tempted to turn the two _MUST_-s into _SHOULD_-s; ignoring those terms is untestable in my view. But I did not do it for now

Fix: #2377

    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/issue-2377-head-element-semantics/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/issue-2377-head-element-semantics/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2380.html" title="Last updated on Jul 29, 2022, 1:08 PM UTC (730bd2d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2380/8b0feb6...730bd2d.html" title="Last updated on Jul 29, 2022, 1:08 PM UTC (730bd2d)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -9721,8 +9721,16 @@ html.my-document-playing * {
 
 					<dt>Usage:</dt>
 					<dd>
-						<p><a data-cite="html#global-attributes">Global attribute</a>. [=EPUB creators=] MAY specify on
-							all elements.</p>
+						<p><a data-cite="html#global-attributes">Global attribute</a>. It MAY be used
+							in [[html]], [[svg]], and on [=Media overlay documents=] with the following restrictions:</p>
+						<ul>
+							<li>For [[html]] content: it MUST be used on [=palpable content=].</li>
+							<li>For [[svg]] content: it MUST be used on 
+								<a href="https://www.w3.org/TR/SVG11/intro.html#TermStructuralElement">structural</a>,
+								<a href="https://www.w3.org/TR/SVG11/intro.html#TermShapeElement">shape</a>, or
+								<a href="https://www.w3.org/TR/SVG11/text.html">text</a> elements.</li>
+							<li>For [=media overlay document=]: it MUST be used on the [^body^] element or its descendants.</li>
+						</ul>
 					</dd>
 
 					<dt>Value:</dt>
@@ -11710,6 +11718,8 @@ EPUB/images/cover.png</pre>
 						Recommendation</a></h3>
 
 				<ul>
+					<li>28-July-2022: The restrictions on the elements where <code>epub:type</code> may be used 
+						has been made explicit. See <a href="https://github.com/w3c/epub-specs/issues/2377">issue 2377</a>.</li>
 					<li>18-July-2022: The viewport <code>meta</code> element for specifying the initial containing
 						boundary in fixed-layout documents is now formally defined. See <a
 							href="https://github.com/w3c/epub-specs/issues/2292">issue 2292</a>. </li>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -9722,15 +9722,15 @@ html.my-document-playing * {
 					<dt>Usage:</dt>
 					<dd>
 						<p><a data-cite="html#global-attributes">Global attribute</a>. It MAY be used
-							in [[html]], [[svg]], and on [=Media overlay documents=] with the following restrictions:</p>
+							in [[html]] and [[svg]] with the following restrictions:</p>
 						<ul>
 							<li>For [[html]] content: it MUST be used on [=palpable content=].</li>
 							<li>For [[svg]] content: it MUST be used on 
 								<a href="https://www.w3.org/TR/SVG11/intro.html#TermStructuralElement">structural</a>,
 								<a href="https://www.w3.org/TR/SVG11/intro.html#TermShapeElement">shape</a>, or
 								<a href="https://www.w3.org/TR/SVG11/text.html">text</a> elements.</li>
-							<li>For [=media overlay document=]: it MUST be used on the [^body^] element or its descendants.</li>
 						</ul>
+						<p>For use in [=media overlay documents=] refer to <a href="#sec-overlays-def"></a>.</p>
 					</dd>
 
 					<dt>Value:</dt>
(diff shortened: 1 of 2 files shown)


epub33/core/index.html
```diff
@@ -9721,8 +9721,16 @@ <h3>The <code>epub:type</code> attribute</h3>
...
-							all elements.</p>
+						<p><a data-cite="html#global-attributes">Global attribute</a>. It MAY be used
+							in [[html]], [[svg]], and on [=Media overlay documents=] with the following restrictions:</p>
+						<ul>
+							<li>For [[html]] content: it MUST be used on [=palpable content=].</li>
+							<li>For [[svg]] content: it MUST be used on 
+								<a href="https://www.w3.org/TR/SVG11/intro.html#TermStructuralElement">structural</a>,
+								<a href="https://www.w3.org/TR/SVG11/intro.html#TermShapeElement">shape</a>, or
+								<a href="https://www.w3.org/TR/SVG11/text.html">text</a> elements.</li>
+							<li>For [=media overlay document=]: it MUST be used on the [^body^] element or its descendants.</li>
```
mattgarrish: The MO content model defines where epub:type can be used, so I don't think we need a bullet.

Maybe this list could just be for HTML and SVG and then have a paragraph or note after that says "For use in [=media overlay documents=] refer to `<a href="#sec-overlays-def"/>.`"

mattgarrish:
> I was tempted to turn the two MUST-s into SHOULD-s; ignoring those terms is untestable in my view

I don't understand the point of this one, to be honest:

> MUST ignore terms that it does not recognize.

This looks like a relic of when we thought epub:type might influence assistive technologies. What is it going to do with terms it doesn't recognize and why do we care?

I'd just erase it.

mattgarrish:
It should be possible to write a test for the "must" you're adding in this PR, though. Since we know pop-up footnotes are implemented on at least a couple of RSes, make a test document that tries to make a meta tag a footnote and see if it works.

It's not a perfect example of the requirement's restriction, but we only have to prove it can be done, right?

iherman:
> It should be possible to write a test for the "must" you're adding in this PR, though. Since we know pop-up footnotes are implemented on at least a couple of RSes, make a test document that tries to make a meta tag a footnote and see if it works.
> 
> It's not a perfect example of the requirement's restriction, but we only have to prove it can be done, right?

Actually, I did not know about the footnotes being implemented...

Let us see if this sketch for a test would work, containing two content files, p1 and p2

- On p1 have a proper footnote and see if the system implements it in the first place
- On p2, have a footnote term added to, say, the `<title>` element and show that that footnote is ignored

Is this what you mean? Which RS does implement a footnote with `epub:type`, b.t.w.? I mean, how would I test to see if the test itself is o.k.?

epub33/core/index.html
iherman: Agreed. Will modify this accordingly.

iherman:
> > I was tempted to turn the two MUST-s into SHOULD-s; ignoring those terms is untestable in my view
> 
> I don't understand the point of this one, to be honest:
> 
> > MUST ignore terms that it does not recognize.
> 
> This looks like a relic of when we thought epub:type might influence assistive technologies. What is it going to do with terms it doesn't recognize and why do we care?
> 
> I'd just erase it.

Happy to do it :-)

mattgarrish:
> Which RS does implement a footnote with epub:type, b.t.w.?

iBooks was the first to implement them. You can read how to define them here: https://help.apple.com/itc/booksassetguide/en.lproj/itccf8ecf5c8.html

Thorium also has the same.

And just on a quick search there are some others like Calibre, Moon+, and even Kindle. I expect there are more, too.

iherman:
> > Which RS does implement a footnote with epub:type, b.t.w.?
> 
> iBooks was the first to implement them. You can read how to define them here: https://help.apple.com/itc/booksassetguide/en.lproj/itccf8ecf5c8.html
> 
> Thorium also has the same.
> 
> And just on a quick search there are some others like Calibre, Moon+, and even Kindle. I expect there are more, too.

Brilliant! I have created two tests:

1. Essentially implementing what is in that apple note with an `aside` element
2. The same but the `epub:type="footnote" id="the_note"` is put on the `title` element.

According to our spec, but also according to its earlier incarnations, the `title` element text should _not_ appear in a footnote popup. Guess what? It does in both Thorium and Apple Books...

Is it o.k. to merge this PR? I will add the anchor to these two tests, too, it will save another PR.

mattgarrish:
> It does in both Thorium and Apple Books...

At least it's testable. I thought for sure they'd both ignore such a case, but I guess they aren't paying attention to where the semantic is set.

> Is it o.k. to merge this PR?

Looks fine to me now.

PR #2383, opened 2022-08-02
IANA comments on the registrations.

iherman:
This PR is a proposed answer to https://lists.w3.org/Archives/Public/public-epub-wg/2022Aug/0000.html

To directly answer to the comments

> It was wrong before and it is still a bit wrong, because this field can
> have only 1 of 4 values: 7bit/8bit/binary/framed.
> I think "binary" is suitable here, as UTF-16 always use "binary" encoding.

However, the package document may be in UTF-8 or UTF-16, i.e., if my understanding is correct, it should either in 8bit or binary encoding. I have therefore changed the text to make this explicit.

(I got inspired by some other registrations, like https://datatracker.ietf.org/doc/html/rfc4627)

> The newly added "Fragment Identifiers" says:
> EPUB Canonical Fragment Identifiers are custom fragment identifiers
> that can resolve to application/oebps-package+xml documents.
>
> I don't know what this means and the above doesn't give me enough
> information about where more information can be found.
> I found Section 9.3.2.2 in <https://www.w3.org/TR/epub-33/> and it seems
> to contain the text I expected here. I suggest the above text is
> replaced with the reference to this section.

It is correct that the current description of the EPUB Canonical Fragment Identifiers (epubcfi) is a by cryptic and is to be improved. Unfortunately, the section you refer to is not the right one, it refers to fragment identifiers for HTML, SVG, etc, files in a specific context. This Pull Request contains an alternative formulation; hopefully that would work. 

~Furthermore, the comment made me think that it was wrong to list epubcfi for the application/oebps-package+xml media type; it is a fragment that is to be used with the URL of the publication and not with the package document. Ie, it is valid for the application/epub+zip media type only.~

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2383.html" title="Last updated on Aug 5, 2022, 10:56 AM UTC (95746ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2383/c90ecf5...95746ae.html" title="Last updated on Aug 5, 2022, 10:56 AM UTC (95746ae)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -11497,7 +11497,13 @@ EPUB/images/cover.png</pre>
 
 					<dt>Encoding considerations:</dt>
 					<dd>
-						<p>Package documents are UTF-8 or UTF-16 encoded XML.</p>
+						<p>8bit if UTF-8; binary if UTF-16.</p>
+						<p>
+							Package documents are in XML, represented either in UTF-8 or UTF-16. When the 
+							package document is written in UTF-8, the file is 8bit compatible.  
+							When it is written in UTF-16, the binary content-transfer-encoding
+							must be used.
+						</p>
 					</dd>
 
 					<dt>Security considerations:</dt>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -11559,10 +11559,6 @@ EPUB/images/cover.png</pre>
 								<p>TEXT</p>
 							</dd>
 
-							<dt>Fragment Identifiers:</dt>
-							<dd>
-								<p>EPUB Canonical Fragment Identifiers are custom fragment identifiers that can resolve to <code>application/epub+zip</code> and <code>application/oebps-package+xml</code> documents. These identifiers are defined at <a href="https://idpf.org/epub/linking/cfi/">https://idpf.org/epub/linking/cfi/</a>.</p>
-							</dd>
 						</dl>
 					</dd>
 
@@ -11685,9 +11681,14 @@ EPUB/images/cover.png</pre>
 
 							<dt>Fragment identifiers:</dt>
 							<dd>
-								<p>EPUB Canonical Fragment Identifiers are custom fragment identifiers that can resolve to <code>application/epub+zip</code> and <code>application/oebps-package+xml</code> documents. These identifiers are defined at <a href="https://idpf.org/epub/linking/cfi/">https://idpf.org/epub/linking/cfi/</a>.</p>
+								<p>
+									EPUB Canonical Fragment Identifiers are custom fragment identifiers defined for [= EPUB Publications =]. 
+									In combination with the URL of the publication, they may be used to refer to an arbitrary content within 
+									any [=publication resource=] defined for the publication.
+									These identifiers are defined at <a href="https://idpf.org/epub/linking/cfi/">https://idpf.org/epub/linking/cfi/</a>.
+								</p>	
 							</dd>
-						</dl>
+					</dl>
 					</dd>
 
 					<dt>Person &amp; email address to contact for further information:</dt>
@@ -11724,6 +11725,11 @@ EPUB/images/cover.png</pre>
 						Recommendation</a></h3>
 
 				<ul>
+					<li>02-August-2022: Updated the media type registrations, following the 
+						<a href="https://lists.w3.org/Archives/Public/public-epub-wg/2022Aug/0000.html">official IANA review comments</a>
+						on updating the previous registrations. 
+						See <a href="https://github.com/w3c/epub-specs/issues/1398">issue 1398</a>.
+					</li>
 					<li>28-July-2022: The restrictions on the elements where <code>epub:type</code> may be used 
 						has been made explicit. See <a href="https://github.com/w3c/epub-specs/issues/2377">issue 2377</a>.</li>
 					<li>18-July-2022: The viewport <code>meta</code> element for specifying the initial containing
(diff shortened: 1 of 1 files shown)


epub33/core/index.html
```diff
@@ -11679,9 +11681,14 @@ <h3>The <code>application/epub+zip</code> media type</h3>
 
 							<dt>Fragment identifiers:</dt>
 							<dd>
-								<p>EPUB Canonical Fragment Identifiers are custom fragment identifiers that can resolve to <code>application/epub+zip</code> and <code>application/oebps-package+xml</code> documents. These identifiers are defined at <a href="https://idpf.org/epub/linking/cfi/">https://idpf.org/epub/linking/cfi/</a>.</p>
+								<p>
+									EPUB Canonical Fragment Identifiers are custom fragment identifiers defined for [= EPUB Publications =]. 
+									In combination with the URL of the publication, they may be used to refer to an arbitrary content within
```
mattgarrish: This doesn't sound quite right, as [intra-publication CFIs](https://idpf.org/epub/linking/cfi/#sec-intra-cfis) are attached to a relative path to the package document. You could put the URL to the publication in front of that, but I'm not sure it would make sense anymore. Do we need to be this specific, or can we go without everything to the comma?

epub33/core/index.html
```diff
@@ -11497,7 +11497,13 @@ <h3>The <code>application/oebps-package+xml</code> media type</h3>
 
 					<dt>Encoding considerations:</dt>
 					<dd>
-						<p>Package documents are UTF-8 or UTF-16 encoded XML.</p>
+						<p>8bit if UTF-8; binary if UTF-16.</p>
+						<p>
+							Package documents are in XML, represented either in UTF-8 or UTF-16. When the 
+							package document is written in UTF-8, the file is 8bit compatible.  
+							When it is written in UTF-16, the binary content-transfer-encoding
```
mattgarrish, line 11504: RFC7303 says it must be base64 in s. 9.1, so while that is a form of binary, is "binary" technically correct here?

I don't know what to suggest for this given the four values they seem to want. All the other +xml registrations I've looked at either defer to rfc 3023 or 7303, depending on their age, without getting into these kinds of details.

epub33/core/index.html
mattgarrish: Sorry, missed that you deleted the package registration reference to CFIs in the github diff. This change would highlight the CFIs that don't work at the expense of the ones that some reading systems do support internally. Is this really what we want?

iherman: I missed the intra-publication CFI term and section, my bad. Which may also mean that I would have to put the section back for the publication documents, too. (Although this definition does sound a bit strange to me spec-wise, to be honest, but we are not here to get into the details of the epubcfi spec...).

I am fine removing that part of the sentence, and it certainly should be removed in this case. I hope it still remains understandable...

epub33/core/index.html
iherman: Isn't it nice to see this consistency across rfc-s? 

7303 says

> ...but binary data, for example, data with a UTF-16 encoding, MUST be content- transfer-encoded in base64.

In my reading it says that UTF-16 _is_ binary, which then require some additional things. So, technically, what is in the text (and which is based on the JSON registration, quite a bit simpler) is correct, and 7303 adds some other things. Maybe we could say something like

> For further details on encoding, rfc7303 also apply.

The only thing I know is that what the commenter proposed, ie, to simply use "binary", is probably wrong, because using UTF-8 is perfectly fine, too.

epub33/core/index.html
iherman: > Sorry, missed that you deleted the package registration reference to CFIs in the github diff. This change would highlight the CFIs that don't work at the expense of the ones that some reading systems do support internally. Is this really what we want?

Yeah, as I said, the intro-publication thingy means that I would have to put it back to package documents, too.

iherman:
@mattgarrish is it o.k. to merge this? I would then answer IANA and, hopefully, get the issue closed...

mattgarrish:
Ya, I would go ahead and merge it. It's ultimately going to be IANA's call if we've got this to their liking.

mattgarrish:
Sorry, didn't mean to close it. Accidental button click using the mouse pad...

epub33/core/index.html
murata2makoto: I am puzzled by what the reviewer said about "Encoding considerations".  [This page](https://www.iana.org/assignments/media-types/media-types.xhtml#text) shows lots of media types derived from XML.  I think that we should follow what they do.  Here is my proposal.

Encoding considerations: Identical to those of  "application/xml" as
described in [Section 3 of [RFC7303]](https://www.rfc-editor.org/rfc/rfc7303#section-3) .

iherman: I am sorry @murata2makoto, but we merged the PR yesterday, and I have already responded to IANA referring, and quoting from, the new text. So I'd prefer not to change the text now (we will see what they respond).

But I believe we fundamentally agree: just putting 'binary' would have been a mistake. What we have in the text now is:

> 8bit if UTF-8; binary if UTF-16.
>
> Package documents are in XML, represented either in UTF-8 or UTF-16. When the package document is written in UTF-8, the file is 8bit compatible. When it is written in UTF-16, the binary content-transfer-encoding must be used. For further details, see [[rfc7303](https://www.w3.org/TR/epub-33/#bib-rfc7303)].

Which is, I believe, essentially the same as what you propose, just a bit more verbose.

PR #2391, opened 2022-08-05
Added a reference to MathML and vertical writing for EPUB 3.0

iherman:
Fix #2371

See:

* For EPUB 3 Overview:
    * [Preview](https://raw.githack.com/w3c/epub-specs/additional-epub3-mentions-issue-2010/epub33/overview/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/overview/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/additional-epub3-mentions-issue-2010/epub33/overview/index.html)

PR #2392, opened 2022-08-05
Revised overview

### LINKED ISSUE #2368: Status of "EPUB 3 Overview"
I am not sure if this document deserves to exist.  First, EPUB 3.3 already has a section "Overview".  Second, some descriptions (most notably, the accessibility section) in "EPUB 3 Overview" are outdated.

We might want to simply cancel this document.  Or, we might want to eliminate outdated sections and change the title of this document to "EPUB revision history".

### LINKED ISSUE #2370: 2.10  Text-to-speech in "EPUB 3 Overview"
[This subsection](https://www.w3.org/TR/epub-overview-33/#sec-tts) is very outdated.  SSML and PLS were removed from EPUB 3.3.   I am not sure if CSS Speech has been implemented.  

I propose to remove everything but the last paragraph in this subsection.

### LINKED ISSUE #2371: "A.2 EPUB 3.0: 2010" in "EPUB 3 Overview"
I propose to mention "vertical writing" and MathML, both of which are mentioned in [an IDPF announcement](https://idpf.org/news/epub-3-proposed-specification-released) of EPUB 3 as proposed specifications.

mattgarrish:
In the interest of having competing PRs, here's my rewrite of the overview... ;)

I've actually incorporated the change in #2391, but I don't see the point in calling out TTS even for media overlays. It's a underspecified and underimplemented feature of them, so let's just not talk about technologies that are only defined in a note.

This PR is a lot of shuffling of the existing content. Rather than one big section of "features", I've tried to break the body up so that related topics are grouped. I also de-structured the global language support and accessibility sections so that they don't seem to repeat the same sections we've already discussed (I used `dl` lists instead of subsections).

The one big new piece of content is the introduction, as the old document just dropped right into features without really explaining its purpose.

I've largely kept the prose that we had before, but I did do some clean up of it and also added more context where I found the discussion a bit stunted.

Fixes #2368, fixes #2370, and fixes #2371 

* [Preview](https://raw.githack.com/w3c/epub-specs/review/overview/epub33/overview/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/overview/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/review/overview/epub33/overview/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2392.html" title="Last updated on Aug 9, 2022, 11:37 AM UTC (21c9617)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2392/706c5a7...21c9617.html" title="Last updated on Aug 9, 2022, 11:37 AM UTC (21c9617)">Diff</a>

### DIFF
--- a/epub33/common/css/common.css
+++ b/epub33/common/css/common.css
@@ -321,8 +321,3 @@ details.explanation > summary, details.desc > summary {
     font-weight: normal !important;
     font-style: italic;
 }
-
-a code
-{
-    color: rgb(3,69,117) !important;
-}
--- a/epub33/common/js/remove-empty-index.js
+++ b/epub33/common/js/remove-empty-index.js
@@ -1,0 +1,24 @@
+
+// this function removes an empty "Terms defined by this specification" section (for indexes in notes)
+
+function removeEmptyIndex() {
+	// update the body
+	var emptyIndex = document.getElementById('index-defined-here');
+	if (emptyIndex !== null) {
+		emptyIndex.parentNode.removeChild(emptyIndex);
+	}
+	var renumIndex = document.querySelector('section#index-defined-elsewhere bdi.secno');
+	if (renumIndex !== null) {
+		renumIndex.innerHTML = renumIndex.innerHTML.replace('.2','.1');
+	}
+	// update the toc
+	var emptyLink = document.querySelector('a[href="#index-defined-here"]')
+	if (emptyLink !== null) {
+		var li = emptyLink.parentNode;
+		li.parentNode.removeChild(li);
+	}
+	var renumLink = document.querySelector('a[href="#index-defined-elsewhere"] bdi.secno')
+	if (renumLink !== null) {
+		renumLink.innerHTML = renumLink.innerHTML.replace('.2','.1');
+	}
+}
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -119,29 +119,31 @@
 			<section id="sec-intro-spec-org" class="informative">
 				<h3>Organization</h3>
 
-				<p>This section reviews the organization of the EPUB specifications through the central product they
-					define: the [=EPUB publication=].</p>
+				<p>This section reviews the organization of this specification through the central product it defines:
+					the [=EPUB publication=].</p>
 
-				<p>An EPUB publication is typically represented by a single [=package document=]. This document includes
+				<p>An EPUB publication is, in its most basic sense, a bundle of resources with instructions on how to
+					render those resources to present the content in a logical order. The types of resources that are
+					allowed in EPUB publication, as well as restrictions on their use, are defined in <a
+						href="#sec-publication-resources"></a>.</p>
+
+				<p>A ZIP-based archive with the file extension <code>.epub</code> bundles the EPUB publication's
+					resources for distribution. As conformant ZIP archives, EPUB publications can be unzipped by many
+					software programs, simplifying both their production and consumption.</p>
+
+				<p>The container format not only provides a means of determining that the zipped content represents an
+					EPUB publication (the <code>mimetype</code> file), but also provides a universally named directory
+					of non-normative resources (<code>/META-INF</code>). Key among these resources is the
+						<code>container.xml</code> file, which directs reading systems to the available [=package
+					documents=]. Refer to <a href="#sec-ocf"></a> for more information about the container format.</p>
+
+				<p>An EPUB publication is typically represented by a single package document. This document includes
 					metadata used by [=reading systems=] to present the content to the user, such as the title and
 					author for display in a bookshelf as well as rendering metadata (e.g., whether the content is
 					reflowable or has a fixed layout). It also provides a manifest of resources and includes a [=EPUB
 					spine | spine=] that lists the default sequence in which to render documents as a user progresses
 					through the content. Refer to <a href="#sec-package-doc"></a> for the requirements for the package
 					document.</p>
-
-				<p>An EPUB publication also includes another key file called the [=EPUB navigation document=]. This
-					document provides critical navigation capabilities, such as the table of contents, that allow users
-					to navigate the content quickly and easily. Refer to <a href="#sec-nav"></a> for more information
-					about this document.</p>
-				<!--
-				<figure>
-					<figcaption> The following example visually represents the structure of the EPUB format. </figcaption>
-					<object data="images/epub.svg" width="350">
-						<img src="images/epub.png" width="350" alt="" />
-					</object>
-				</figure>
-				-->
 
 				<p>The actual content of an EPUB publication &#8212; what users are presented with when they begin
 					reading &#8212; is built on the Open Web Platform and comes in two flavors: XHTML and SVG. Called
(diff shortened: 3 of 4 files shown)


mattgarrish:
I forgot to mention that I also did some work on the "Organization" section of the core spec. I reordered some of it to match the ordering of the specification and also edited out the inference that there are still multiple specifications.

murata2makoto:
"For a list of all the recommendation-track EPUB 3 specifications and non-normative notes, refer to ."  in "Reading the specifications" appears to lack the last word.  Should we create a separate web page for enumerating all our documents and reference it here?

P.S.  It appears that the section "EPUB 3 documents" is referenced here.  But I still prefer an external web page for enumerating all our documents including some new notes about accessibility among others.

murata2makoto:
Overall, I think that this is a big improvement.  In particular, I like "Version numbering" and "Layouts" containing "Reflowable layouts" and "Fixed layouts".  I also think that "Global language support" no longer looks outdated.

I will have a more careful look but I support this pull request strongly.

iherman:
The source makes a reference to a `removeEmptyIndex` script and a `./common/js/remove-empty-index.js` file; the latter is not in the repository. As a consequence, respec does not complete (which is not a very graceful way of handling things for respec, but that is another issue...)

iherman:
Overall, I am fine with moving along with this PR to handle #2370 and #2371 (I will close the other PR-s). I think the reorganization is helpful (but I still need to make a closer review).

I am not (yet?) convinced by the complete removal of TTS. I realize it is under-implemented, but I have to see whether this isn't the case with other features that are also listed in the text (eg, is setting the landscape/portrait mode by the user implemented?). But this is not an issue for which I would lie down the road…

murata2makoto:
@iherman 

>I am not (yet?) convinced by the complete removal of TTS. I realize it is under-implemented, but I have to see whether this isn't the case with other features that are also listed in the text (eg, is setting the landscape/portrait mode by the user implemented?). But this is not an issue for which I would lie down the road…

SSML is implemented and used by the biggest textbook publisher in Japan, while PLS is implemented and used by another Japanese company.   They are not used for trade books but used for textbooks.   I did not lie down the road to block the removal of SSML and PLS from EPUB 3.3, because I was late and also because W3C has [another attempt ](https://www.w3.org/TR/spoken-html/).  But I continue to miss SSML and PLS.

iherman:
> @iherman
> 
> > I am not (yet?) convinced by the complete removal of TTS. I realize it is under-implemented, but I have to see whether this isn't the case with other features that are also listed in the text (eg, is setting the landscape/portrait mode by the user implemented?). But this is not an issue for which I would lie down the road…
> 
> SSML is implemented and used by the biggest textbook publisher in Japan, while PLS is implemented and used by another Japanese company. They are not used for trade books but used for textbooks. I did not lie down the road to block the removal of SSML and PLS from EPUB 3.3, because I was late and also because W3C has [another attempt ](https://www.w3.org/TR/spoken-html/). But I continue to miss SSML and PLS.

But then... why would be remove any reference to them from the Overview? The fact of being in a Note "only" means that we did not expect to pass the CR bar for a Recommendation, but nothing prevents us to turn it into a Rec later, if the implementation landscape changes...

mattgarrish:
> the latter is not in the repository

Ah, I didn't dig deep enough in the failure message. I just saw that it was failing to build and I didn't have any local issues, so I thought it was the same bug you ran into. I'll commit the script. I had to delete the "references defined in this document" (or whatever it's called) section as the overview doesn't define any terms so respec outputs an empty heading and I couldn't find any way to stop it.

mattgarrish:
> But then... why would be remove any reference to them from the Overview?

Are features without support in reading systems worth highlighting to people as something they can use? I'm fully expecting the media overlays feature will be marked underimplemented as there's no evidence any reading systems support it.

Look at how terrible the navigation section reads with its excuse that the html formatting features we know reading systems don't support in their auto-generated tables of contents can still sort of work if you put the navigation document in the spine. Do we really want to go even further and add notes after every feature we know is not supported. Essentially saying to people, hey, look at this cool feature, by the way don't expect it to work at all!

It's not about whether technologies are notes, but about how long we want to continue claim we have features that we know have limited to no support. And, yes, if we discover from testing that there are additional features that aren't fully supported, I would also argue that they don't belong in the overview either. It was fine to claim all kinds of wonderful features when we wrote it, and before any reading systems existed, but more than a decade on I'd like to know what EPUB really supports.

The TTS note is included in the list of notes and explains what it defines, so it's not like there are no references to it. I'd also be fine with mentioning it and multiple renditions in an "Experimental features" section at the end of the body, but I don't want to give people the wrong impression about these technologies.

mattgarrish:
> It appears that the section "EPUB 3 documents" is referenced here. But I still prefer an external web page for enumerating all our documents including some new notes about accessibility among others.

The preview is broken right now, probably from the script I missed. Respec hasn't run on it, so all the adjustments it makes are missing.

And, yes, I agree it would be helpful if we had a document that covered all the documents the WG, CG, and BG produce for EPUB. It's already a headache trying to figure out what is published where. But for now all we have to reference is the appendix.

mattgarrish:
I've added a section to explain our use of evolving standards and also mention multi-rend and tts and why they're published as notes.

I also noticed that the accessibility list items still read like guidance on how to author content rather than explanations of the features that are available, so I did some minor tweaking to improve them.

murata2makoto:
> Are features without support in reading systems worth highlighting to people as something they can use? 

SSML and PLS are supported by reading systems for Japanese textbooks.  They are *not* underimplemented.  TTS for the Japanese language is unlikely to be perfect without using them.

mattgarrish:
> > Are features without support in reading systems worth highlighting to people as something they can use?
> 
> SSML and PLS are supported by reading systems for Japanese textbooks. They are _not_ underimplemented.

The discussion was only around why I don't think it's appropriate to list them for media overlays. There's no support of dropping into TTS mode when only a text element is specified that I've yet heard of, let alone supporting PLS or SSML when that's done.

That's why I've added the new and experimental features section to mention the note, as that's where the TTS technologies currently fall. You can use them, but we're not guaranteeing support for them. It will be up to publishers to determine if it is beneficial to them.

murata2makoto:
> And, yes, I agree it would be helpful if we had a document that covered all the documents the WG, CG, and BG produce for EPUB. It's already a headache trying to figure out what is published where.

This deserves to be discussed in the WG, IMHO.

iherman:
> > And, yes, I agree it would be helpful if we had a document that covered all the documents the WG, CG, and BG produce for EPUB. It's already a headache trying to figure out what is published where.
> 
> This deserves to be discussed in the WG, IMHO.

I do not know whether this requires a general WG discussion, but I am not sure whether we should mix it into the Overview document. The problem is that the Overview is a publication on /TR, whose update, though relatively easy, is still a separate process. I would rather create a separate page on the WG's Web site, as an extension to the [Publication page](https://www.w3.org/publishing/groups/epub-wg/PublStatus) whose maintenance is much easier. Such lists have the tendency of becoming out of data quickly.

iherman:
> That's why I've added the new and experimental features section to mention the note, as that's where the TTS technologies currently fall. You can use them, but we're not guaranteeing support for them. It will be up to publishers to determine if it is beneficial to them.

I am fine with that approach, at least for now.

There is a more general issue, though, that may have to be discussed, probably at the F2F meeting: how do we plan to reflect an under-implemented feature in our documents when we get to that point? Do we just flag them as such, or do we re-engineer the spec by separating those more visibly? In some cases, that may become a major surgery (e.g., fallbacks). That discussion may also affect how we would treat those in the Overview document (if applicable).

I would propose to merge this PR (which otherwise looks good to me) as of now, knowing that we may have to reopen a new issue/PR once those discussions have taken place and, more importantly, we have some testing results in our hands.

iherman (approved):

murata2makoto (approved):
We might later create smaller issues, but this pull request is overall a great reengineering.

mattgarrish:
> In some cases, that may become a major surgery (e.g., fallbacks).

I'm a little sceptical about this, but it's a topic for a another day... 😄 

(Some parts of fallbacks may not be fully supported, but even back around 2016 when we were maintaining epubtest.org there were more than enough reading systems that supported manifest fallbacks for the `img` element. I'd be surprised if that's changed)

mattgarrish:
I'll merge this tomorrow if there's no other feedback.

dauwhe (approved):

shiestyle (approved):

iherman:
With all reviews positive, I just merge it to get it out of the way...

PR #2400, opened 2022-08-12
Update conformance identifiers to new expanded pattern

mattgarrish:
This pull request implements the new pattern for identifiers we discussed to make the metadata field more human readable.

Because the value now has whitespace in it, I added that the value must match our pattern string after xml whitespace normalization. This will allow the value to be brokan by line endings and indents but still be valid.

Other than that, it's a pretty basic expanding of a few abbreviations and removing the hyphenation and underscores that made it one contiguous string of characters.

Fixes #2398 

- Accessibility [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2398/epub33/a11y/index.html)
- Accessibility [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2398/epub33/a11y/index.html)

Issue #2398, opened 2022-08-12
Change text pattern of conformance identifiers


mattgarrish:
Per the discussions that went on in #2116, we want to make the `conformsTo` value both machine and human readable.

The new pattern will simply expand the truncated names and remove the hyphens and underscores.

For example, "EPUB-A11Y-11_WCAG-21-AA" will become "EPUB Accessibility 1.1 - WCAG 2.1 Level AA".

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-08-26-epub#section2) on 2022-08-26

List of resolutions:

- ***[Resolution No. 2](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-08-26-epub#resolution2):  Merge PRs 2400 and 2401.***

<details><summary><i>View the transcript</i></summary>

### 2. Changes to accessibility conformance identifiers and `accessibilitySummary`.

_See github pull request [epub-specs#2400](https://github.com/w3c/epub-specs/pull/2400)._




_See github pull request [epub-specs#2401](https://github.com/w3c/epub-specs/pull/2401)._




_See github issue [epub-specs#2399](https://github.com/w3c/epub-specs/issues/2399)._




_See github issue [epub-specs#2398](https://github.com/w3c/epub-specs/issues/2398)._




**Avneesh Singh:** the main issue here is how to handle `accessibilitySummary`.  
… should it be mandatory, or only recommended?.  
… we settled on recommended, and highlight the cases where you should definitely provide this data.  
… this also requires a change in something else. People were using this field to explain a11y conformance.  
… mgarrish suggested that we should make the conformance field more human-friendly.  
… which means you don't need to explain this in `accessibilitySummary`.  

> *Dave Cramer:* ak ch.

**Avneesh Singh:** so these are the two PRs.  

**Charles LaPierre:** agreed. In addition, mgarrish also clarified that there should only be one `accessibilitySummary`, in the language of the epub.  
… the `accessibilitySummary` is now recommended. It should include anything that is not captured in the other (machine readable) meta fields.  
… not duplication.  
… we will rewrite a11y best practices around how to display epub metadata to users.  

**George Kerscher:** the EUAA will require the presentation of a11y data, so we feel that if the other fields that are currently machine readable can be communicated to users via best practices, then accessibilitySummary can just be for anything else.  

[...]

PR #2401, opened 2022-08-12
Change accessibilitySummary to a recommended field

### LINKED ISSUE #2399: Make summaries recommended and update guidance on use
Per the discussions that went on in #2116, the new proposal is to both change `schema:accessibilitySummary` from a required to recommended field and provide additional guidance on using it to supplement the other discovery metadata so it is not repeating information.

mattgarrish:
This pull request moves accssibilitySummary to the recommended metadata list with accessModeSufficient. The definition is also slightly tweaked to emphasize that it is not for duplicating information expressed elsewhere.

The techniques document section has also been rewritten to emphasize avoiding duplication.

Have a good look at these changes and let me know if there's anything else we should say.

Fixes #2399 

- Accessibility [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2399/epub33/a11y/index.html)
- Accessibility [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2399/epub33/a11y/index.html)

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -2336,9 +2336,12 @@
 					>working group's issue tracker</a>.</p>
 
 			<ul>
+				<li>12-Aug-2022: Corrected the invalid guidance than an <code>xml:lang</code> attribute on an
+						<code>accessibilitySummary</code> property represents a translation. See <a
+						href="https://github.com/w3c/epub-specs/pulls/2401">pull request 2401</a>.</li>
 				<li>12-Aug-2022: Updated the <code>accessibilitySummary</code> explanation to avoid repeating
 					information available in the other metadata. See <a
-						href="https://github.com/w3c/epub-specs/issues/#2399">issue 2399</a>.</li>
+						href="https://github.com/w3c/epub-specs/issues/2399">issue 2399</a>.</li>
 				<li>20-July-2022: Added additional emphasis on setting single value access mode sufficient declarations
 					and also listed and explained when to apply the two most common for EPUBs. See <a
 						href="https://github.com/w3c/epub-specs/issues/2302">issue 2302</a>.</li>
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -543,10 +543,34 @@
 
 				<p>An accessibility summary provides a brief, human-readable description of the accessibility
 					characteristics of an <a href="https://www.w3.org/TR/epub/#dfn-epub-publication">EPUB
-						publication</a>, or lack thereof.</p>
+						publication</a> that cannot be expressed through the other discovery metadata.</p>
+
+				<p>The accessibility summary should not simply repeat the conformance information provided in the
+						<code>dcterms:conformsTo</code> property, for example, or the features listed in the
+						<code>schema:accessibilityFeature</code> properties. As this metadata is required by the EPUB
+					Accessibility specification, systems that process EPUB publications can already present it to users.
+					Repeating it in the summary only makes them hear the information again.</p>
+
+				<aside class="example" title="Accessibility summary that complements the conformance statement">
+					<pre>&lt;meta
+    property="dcterms:conformsTo">
+   EPUB Accessibility 1.1 - WCAG 2.1 Level AA
+&lt;/meta>
+&lt;meta
+    property="schema:accessibilitySummary">
+   In addition to the requirements of its conformance claim,
+   this publication includes sign language interpretation
+   for all audio content.&#8230;
+&lt;/meta>
+					</pre>
+				</aside>
+
+				<p>EPUB creators should not include an accessibility summary when they have nothing more to add to the
+					conformance claim and other discovery metadata.</p>
 
 				<p>If an EPUB publication does not meet the requirements for content accessibility in [[epub-a11y-11]],
-					the reason(s) it fails should be noted in the summary.</p>
+					the reason(s) it fails should be noted in the summary. Similarly, if an EPUB creator is hesitant to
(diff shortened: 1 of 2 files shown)


clapierre (approved):
Wondering Matt if we should include in the Highlights of what changed the fact that we no longer recommend having multiple summaries for different languages?  Maybe we don't need to do that as it wasn't stated in 1.0.

mattgarrish:
I was surprised to see that text in the techniques, as we already had a conversation about translations earlier in the revision and how putting `xml:lang` on an element only indicates what language the content is in, not that the element represents a translation of another. It was never a formal practice we mentioned in the specification.

The problem is that we discussed this in the publ-a11y repository (in https://github.com/w3c/publ-a11y/issues/94). We got to the right answer there, but it doesn't look like anyone spotted that we were saying the opposite here in the techniques.

It wasn't a requirement or recommendation, so we technically aren't reversing anything substantive, but I can add a change log entry to highlight that the incorrect advice has been fixed easily enough.

mattgarrish:
For anyone interested, the original wording didn't use the word translation, but the text suggested that was the use case:

> Do not repeat this property unless it contains the summary in another language. In the case of multiple summaries, use the xml:lang attribute to differentiate the language.

I don't think we want to try and explain what multiple summaries with different `xml:lang` attributes represents. Better to just say that they are not translations.

gregoriopellegrino (approved):

iherman (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-08-26-epub#section2) on 2022-08-26

List of resolutions:

- ***[Resolution No. 2](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-08-26-epub#resolution2):  Merge PRs 2400 and 2401.***

### 2. Changes to accessibility conformance identifiers and `accessibilitySummary`.

_See github pull request [epub-specs#2400](https://github.com/w3c/epub-specs/pull/2400)._




_See github pull request [epub-specs#2401](https://github.com/w3c/epub-specs/pull/2401)._




_See github issue [epub-specs#2399](https://github.com/w3c/epub-specs/issues/2399)._




_See github issue [epub-specs#2398](https://github.com/w3c/epub-specs/issues/2398)._




**Avneesh Singh:** the main issue here is how to handle `accessibilitySummary`.  
… should it be mandatory, or only recommended?.  
… we settled on recommended, and highlight the cases where you should definitely provide this data.  
… this also requires a change in something else. People were using this field to explain a11y conformance.  
… mgarrish suggested that we should make the conformance field more human-friendly.  
… which means you don't need to explain this in `accessibilitySummary`.  

> *Dave Cramer:* ak ch.

**Avneesh Singh:** so these are the two PRs.  

**Charles LaPierre:** agreed. In addition, mgarrish also clarified that there should only be one `accessibilitySummary`, in the language of the epub.  
… the `accessibilitySummary` is now recommended. It should include anything that is not captured in the other (machine readable) meta fields.  
… not duplication.  
… we will rewrite a11y best practices around how to display epub metadata to users.  

**George Kerscher:** the EUAA will require the presentation of a11y data, so we feel that if the other fields that are currently machine readable can be communicated to users via best practices, then accessibilitySummary can just be for anything else.  

**John Foliot:** what happens if the epub is in more than one language? Then what lang should `accessibilitySummary` be if there should only be one? Multiple languages in one?.  

**George Kerscher:** there's always a base language for the publication, so the accessibilitySummary would use that language..  

**John Foliot:** the `accessibilitySummary`, as explained, is intended to be human readable. So screen readers should be able to read it. I should be able to use tools like language tagging, pronunciation markup, etc..  
… so maybe that human readable summary should support multiple languages.  

**George Kerscher:** the name summary is inaccurate. It's not a summary of the whole document. It uses the term summary as a holdover from Schema.org.  
… we would propose to reference the Schema.org description in the a11y best practices guide, to make this clear.  

**Avneesh Singh:** a lot of the details which have been shared about a11y guidelines have been working in Publishing CG.  

**Ivan Herman:** i'm looking at preview of the spec where description of `accessibilitySummary` does not make explicit reference to languages.  

_See github issue [publ-a11y#94](https://github.com/w3c/publ-a11y/issues/94)._




**Ivan Herman:** also `accessibilitySummary` is part of the package document, so you can repeat it in a different language?.  

**Wendy Reid:** there's a note in the PR about not repeating the `accessibilitySummary`.  
… "do not repeat this property to declare translations...".  
… since there is no hierarchy, the RS might pick the wrong language.  

**Ivan Herman:** it's an xml document, so adding a language is technically possible.  

**Avneesh Singh:** we've been discussing `accessibilitySummary` for a couple months now, JF has made good points but we are in CR now. We will defer these things to the next version rather than expanding our work on this currently.  

**Wendy Reid:** any other questions re. `accessibilitySummary`? Changes to conformance identifiers?.  

> **Proposed resolution: Merge PRs 2400 and 2401.** *(Wendy Reid)*

> *Ivan Herman:* +1.

> *Charles LaPierre:* +1.

> *Wendy Reid:* +1.

> *Matthew Chan:* +1.

> *Avneesh Singh:* +1.

> *Brady Duga:* +1.

> *Masakazu Kitahara:* +1.

> *Dave Cramer:* +1.

> ***Resolution #2: Merge PRs 2400 and 2401.***

</details>

PR #2402, opened 2022-08-16
Deprecate Media Overlays embedded media

### LINKED ISSUE #2397: Deprecate embedded audio or video as targets for Media Overlay <text> reference
My apologies for the verbose background below; I believe it is necessary to make an informed decision. This is not a widely known area of the spec…

This is the usual way of using Media Overlays (MO):

```xml
<html>
    …
    <p id="txt">...</p>
</html>

<smil>
    …
    <par>
        <text src="...#txt">
        <audio src="link_to_audio_file" clipBegin="…" clipEnd="…">
    </par>
    …
<smil>
```

On Reading Systems implementing MO, if read-aloud is switched on, the content of the audio file is read while the text in the XHTML content is rendered. 

However, what happens if we have, instead, the following:

```xml
<html>
    …
    <video id="movie">...</video>
</html>

<smil>
    …
    <par>
        <text src="...#movie">
        <audio src="link_to_audio_file" clipBegin="…" clipEnd="…">
    </par>
    …
<smil>
```

The answer is… it is complicated. There are indeed two time-dependent media rendering jobs in parallel (the movie and the audio clip), and the spec should be clear about which one has the upper hand in timing control. The problem is that it isn't ☹️: 

- [§9.2.2.4.1](https://www.w3.org/TR/epub-33/#sec-emb-audio-video) in the EPUB 3.3 content spec does _not_ talk about this case, only the case when there is _no_ audio element in the MO file. Specifying what is and isn't allowed would require further work.
- [§9.3.2](https://www.w3.org/TR/epub-rs-33/#sec-embedded-media) in the Reading System is a bit clearer; what it specifies is that the audio file's timing is in control: the video must be started by the RS when the audio starts, it must be stopped when the audio ends, the reader has no control over the video (only the RS has). But of course there are some other cases, see that section for details, which makes this specification _much_ more complicated. In my view that section requires further editorial work to make it clearer.

Some tests have been created for [§9.3.2](https://www.w3.org/TR/epub-rs-33/#sec-embedded-media) (see [`mol-embed`](https://github.com/w3c/epub-tests/tree/main/tests/mol-embed) or  [`mol-embed_deactivate_playback`](https://github.com/w3c/epub-tests/tree/main/tests/mol-embed_fxl)) but, at the moment, there are no known implementations around, and we may be looking at a problem at the end of the CR phase. Also: it is not clear what the use case may be for such a structure in the EPUB context. (If the goal is to add an audio track to a video as, say, background music, there are widely available video editing programs to do that more easily these days.)

(Note that this introduction only referred to video but, in fact, it could also be an audio track within the XHTML content document. The spec used the term "embedded media" as a general term, although that also includes still images.)

---

Sorry for the long introduction; here is the real issue I indent to raise. I think we should consider deprecating the feature of "embedded media" (maybe with the exception of still images). It is an overly complex part of the spec (mainly the RS spec) without known use cases and implementations; we can preempt the problems we will have at the end of the CR phase by doing so.

I can see several ways we could proceed:

1. declare the whole of [Content spec §9.2.2.4.1](https://www.w3.org/TR/epub-33/#sec-emb-audio-video) as well as [Reading System §9.3.2](https://www.w3.org/TR/epub-rs-33/#sec-embedded-media) a _SHOULD_ or _MAY_ (and avoid using _MUST_ statements in the text, much like the [§9.3.3 TTS](https://www.w3.org/TR/epub-rs-33/#sec-text-to-speech) section); or
2. we extract the use case whereby video or audio in a `<par>` without an accompanying `<audio>` means playing that media under the implementation control (without user control, that is), and we declare all other cases as deprecated; or
3. declare the whole feature as deprecated: the content document MO file disallows a reference to a video/audio from the `<text>` element (either directly or indirectly with a, say, video in the middle of a text flow with read loud) and the RS specification just says that handling those cases is implementation dependent.

I do not think we would create problems with existing EPUB deployment; indeed, I do not believe there is any out there... And it would make our spec cleaner.

marisademeglio:
Fixes #2397 

In the core spec:

Removed 9.3.2.4 entirely, added note to MO text element about how referring to embedded timed media with the MO text element is possible but unspecified.

In the RS spec:

Removed 9.3.2 and added a note about deprecation to the top of 9.3.

See:

    * [Preview](https://raw.githack.com/marisademeglio/epub-specs/fix-issue-2397-mo-embed/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/marisademeglio/epub-specs/fix-issue-2397-mo-embed/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/marisademeglio/epub-specs/pull/2402.html" title="Last updated on Aug 29, 2022, 3:42 PM UTC (76eb127)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2402/d9c0385...marisademeglio:76eb127.html" title="Last updated on Aug 29, 2022, 3:42 PM UTC (76eb127)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -8257,7 +8257,7 @@ No Entry</pre>
 								href="https://www.w3.org/TR/SVG11/shapes.html">basic shapes</a>, or <a
 								href="https://www.w3.org/TR/SVG11/text.html">text</a> elements in SVG). </p>
 						
-						<p class="note">While the <code>src</code> attribute of a <code>text</code> element may refer to embedded timed media, such as an [[html]] [^video^] element, this is atypical usage and, as Reading System behavior is unspecified, may produce unpredictable results.</p>
+						<p class="note">[[epub-rs-33]] no longer provides guidance for reading systems on the playback of timed media (i.e., the automatic starting of the referenced media). Although the <code>src</code> attribute of a <code>text</code> element may refer to embedded timed media (e.g., via an [[html]] [^video^] element), referencing such media may have unpredictable results.</p>
 					</section>
 
 					<section id="sec-smil-audio-elem">
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -11707,7 +11707,8 @@ EPUB/images/cover.png</pre>
 						Recommendation</a></h3>
 
 				<ul>
-					<li>28-August-2022: Added a note to Media Overlays <code>text</code> element definition regarding using embedded timed media.</li>
+					<li>28-August-2022: Added a note to Media Overlays <code>text</code> element definition regarding using embedded timed media.
+						See <a href="https://github.com/w3c/epub-specs/issues/2397">issue 2397</a>.</li>
 					<li>02-August-2022: Updated the media type registrations, following the 
 						<a href="https://lists.w3.org/Archives/Public/public-epub-wg/2022Aug/0000.html">official IANA review comments</a>
 						on updating the previous registrations. 
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -8200,7 +8200,7 @@ No Entry</pre>
 
 						<p>The <code>text</code> element references an element in an [=EPUB content document=]. A
 								<code>text</code> element typically refers to a textual element but can also refer to
-							other [=EPUB content document=] media elements (see <a href="#sec-embedded-media"></a>). In
+							other [=EPUB content document=] media elements. In
 							the absence of a sibling [^audio^] element, textual content referred to by this element may
 							be rendered via <a href="#sec-tts">text-to-speech</a>.</p>
 
@@ -8256,6 +8256,8 @@ No Entry</pre>
 								href="https://www.w3.org/TR/SVG11/paths.html">paths</a>, <a
 								href="https://www.w3.org/TR/SVG11/shapes.html">basic shapes</a>, or <a
 								href="https://www.w3.org/TR/SVG11/text.html">text</a> elements in SVG). </p>
+						
+						<p class="note">While the <code>src</code> attribute of a <code>text</code> element may refer to embedded timed media, such as an [[html]] [^video^] element, this is atypical usage and, as Reading System behavior is unspecified, may produce unpredictable results.</p>
 					</section>
 
 					<section id="sec-smil-audio-elem">
@@ -8643,43 +8645,6 @@ No Entry</pre>
 							navigating by word or phrase and when searching the text but increases the file size of the
 							[=media overlay documents=]. Fragment identifier schemes that do not rely on the presence of
 							elements could provide even finer granularity, where supported.</p>
-					</section>
-
-					<section id="sec-embedded-media">
-						<h5>Embedded media</h5>
-
-						<p>Any [=EPUB content document=] associated with a [=media overlay document=] MAY contain
-							embedded media such as video, audio, and images. [=EPUB creators=] MAY use the media overlay
-							[^text^] element in such instances to reference the embedded media by its element's
-								<code>id</code> attribute value.</p>
-
-						<section id="sec-emb-audio-video">
-							<h6>Embedded audio and video</h6>
-
-							<p> When a [^text^] element references embedded audio or video, [=reading systems=] will
-								initiate playback of the media in the absence of an [^audio^] element sibling. </p>
-
-							<p>[=EPUB creators=] SHOULD avoid using scripts to control playback of referenced embedded
-								[=EPUB content document=] media, as this might conflict with media overlays playback
-								behavior.</p>
-
-							<p>EPUB creators should carefully examine any overlapping audio situations and deal with
-								them at the production stage, as reading systems handling of simultaneous volume levels
-								is optional.</p>
-						</section>
-
-						<section id="sec-emb-img">
-							<h6>Embedded images</h6>
-
-							<p>When a <code>text</code> element references an embedded image, the [^audio^] sibling
-								element is OPTIONAL. In the absence of an <code>audio</code> element, [=reading
-								systems=] will voice the image using <a href="#sec-tts">text-to-speech
-								rendering</a>.</p>
-
-							<p>[=EPUB creators=] MUST ensure they provide fallback text for an image when an omitting an
-									<code>audio</code> element (e.g., using the [[html]] <code>alt</code>
-								attribute).</p>
-						</section>
 					</section>
 
 					<section id="sec-tts">
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -11707,6 +11707,7 @@ EPUB/images/cover.png</pre>
 						Recommendation</a></h3>
 
 				<ul>
+					<li>28-August-2022: Added a note to Media Overlays <code>text</code> element definition regarding using embedded timed media.</li>
 					<li>02-August-2022: Updated the media type registrations, following the 
 						<a href="https://lists.w3.org/Archives/Public/public-epub-wg/2022Aug/0000.html">official IANA review comments</a>
 						on updating the previous registrations. 
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1866,6 +1866,9 @@
 			<section id="sec-behaviors-interaction">
 				<h4>Interacting with the EPUB content document</h4>
 
+				<p class="note">Earlier versions of this specification included some information about embedded audio and 
+					video [epubmediaoverlays-32]. This feature has been deprecated.</p>
+
 				<section id="sec-rsconf-navigation">
 					<h5>Navigation</h5>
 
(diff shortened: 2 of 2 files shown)


mattgarrish:
I wonder if we should give this the TTS treatment and instead of making it like it never existed, move the authoring and reading system requirements to a note? (Maybe including the TTS section.)

It might never get implemented, but publishing as a note leaves it as an option.

marisademeglio:
> I wonder if we should give this the TTS treatment and instead of making it like it never existed, move the authoring and reading system requirements to a note? (Maybe including the TTS section.)
> 
> It might never get implemented, but publishing as a note leaves it as an option.

I honestly don't think it's worth the effort. I don't think it will ever get implemented and to specify it properly, even as a note, would be a not-insignificant time investment.

epub33/rs/index.html
```diff
@@ -1866,6 +1866,9 @@ <h5>Rendering EPUB content document elements</h5>
 			<section id="sec-behaviors-interaction">
 				<h4>Interacting with the EPUB content document</h4>
 
+				<p class="note">Earlier versions of this specification included some information about embedded audio and 
+					video [epubmediaoverlays-32]. This feature has been deprecated.</p>
```
```suggestion
					video [[epubmediaoverlays-32]]. This feature has been deprecated.</p>
```

and a reference to epubmediaoverlays-32 must be added to the bibliography part on the respec structure of the header.

(The `[[...]]` forces respec to use the bibliography entry in the references.)

iherman (changes requested):
@marisademeglio  (also answering your question) the entries in the change sections are added by hand. Copy-paste-modify is your friend...

iherman:
> > I wonder if we should give this the TTS treatment and instead of making it like it never existed, move the authoring and reading system requirements to a note? (Maybe including the TTS section.)
> > It might never get implemented, but publishing as a note leaves it as an option.
> 
> I honestly don't think it's worth the effort. I don't think it will ever get implemented and to specify it properly, even as a note, would be a not-insignificant time investment.

I agree. The TTS seems to be on much more stable foundations, using external and, mainly, more widely known standards. Although MO is based on SMIL, the latter is, alas!, a completely unknown standard and probably the only reason it has not been deprecated is EPUB's usage of (a small subset of) SMIL. A WG note, to be made worthy of publication, would require quite some work.

mattgarrish:
> The TTS seems to be on much more stable foundations

Look out for multiple renditions, in that case...

At any rate, if we don't, then this introduces ambiguity onto the authoring side, as it doesn't say what's supposed to happen now if you do reference an audio or video element. It might be helpful to add a note where media elements are mentioned that previous version of the specification defined auto-playback of these elements, but it's no longer the case (if audio is not provided, then TTS kicks in). Authors should consider alerting users that these elements exist and they need to go back and find them after listening to the work.

epub33/rs/index.html
mattgarrish: We can't deprecate in a note, otherwise the deprecation doesn't have normative standing. I'd recreate 9.3.2 rather than completely deleting the section so it matches other deprecated features. I can't suggest a fix through the github interface, but put back something like:

```
<section id="sec-embedded-media">
	<h4>Embedded audio and video (deprecated)</h4>
	<p>Automatic playback of embedded audio and video is now <a data-cite="epub-33#deprecated">deprecated</a>.</p>
	<p>Refer to its definition in [[epubmediaoverlays-32]] for more information.</p>
</section>
```

mattgarrish:
> mentioned that previous version of the specification defined auto-playback of these elements

Actually, re-reading the note that's there I can probably live with it as-is. I was trying to read the line in the diff and it looked like it was only referring to video.

iherman:
> > The TTS seems to be on much more stable foundations
> 
> Look out for multiple renditions, in that case...
> 
> At any rate, if we don't, then this introduces ambiguity onto the authoring side, as it doesn't say what's supposed to happen now if you do reference an audio or video element. It might be helpful to add a note where media elements are mentioned that previous version of the specification defined auto-playback of these elements, but it's no longer the case (if audio is not provided, then TTS kicks in). Authors should consider alerting users that these elements exist and they need to go back and find them after listening to the work.

There is such a mention in the RS spec. Do you mean to add something similar in the core spec as well?

epub33/core/index.html
```diff
@@ -8256,6 +8256,8 @@ <h5>The <code>text</code> element</h5>
 								href="https://www.w3.org/TR/SVG11/paths.html">paths</a>, <a
 								href="https://www.w3.org/TR/SVG11/shapes.html">basic shapes</a>, or <a
 								href="https://www.w3.org/TR/SVG11/text.html">text</a> elements in SVG). </p>
+						
+						<p class="note">While the <code>src</code> attribute of a <code>text</code> element may refer to embedded timed media, such as an [[html]] [^video^] element, this is atypical usage and, as Reading System behavior is unspecified, may produce unpredictable results.</p>
```
mattgarrish: This is still kind of bugging me this morning. I wonder if we should be specific that we don't expect playback anymore, and that authors should consider informing users that audio/video material is available.

```suggestion
						<p class="note">[[epub-rs-33]] no longer provides guidance for reading systems on the playback of timed media (i.e., the automatic starting of the referenced media). Although the <code>src</code> attribute of a <code>text</code> element may refer to embedded timed media (e.g., via an [[html]] [^video^] element), referencing such media may have unpredictable results.</p>
```

marisademeglio: I don't know, it feels out of place to me for the spec to be giving best practice advice for what a book's contents should say.

iherman: "should consider" in an informal note is a fairly weak statement, ie, a weak form of best practice advice. I am not bothered by this.

I am mildly bothered by another point, though. I believe it is formally o.k. to say

```xml
<text src="link_to_a_section_in_a_content"/>
<text src="link_to_a_video"/>
```

Ie, MO is not, strictly speaking, for audio playback only; I could use a video playback, too. Otherwise, the very term "Media Overlay" would be a misnomer. I realize audio is _the_ major use case, and maybe no RS would be able to do anything with the structure above (is it worth a test?) and I am not sure what the a11y use case would be, but our spec is allowing it... Hence, the "for synchronized audio playback" seems to be out of place.

iherman: (B.t.w, if this usage is valid and has a minimum usability, we should probably take over the SMIL `<video>` element... in a future release.

marisademeglio: @iherman well you could have a `par` containing either of your `text` element examples but you cannot have both of those `text`s in parallel with each other.

iherman: Hm. What this means is that I cannot use a video for playback. Ie, if there is a usage for it, we need a `<video>` element (yes, that is for a future version, if any...). But this means is that "Media Overlay" is really a misnomer. Is it too late or too awkward to rename the full functionality to "Audio Overlay" or "Audio playback". This is really what we are talking about, isn't?

marisademeglio: You could still have this sort of thing:

```
<par>
  <text src="link_to_a_video"/>
</par>
<par>
  <text src="link_to_html_para"/>
  <audio src="audio_file".../>
</par>
...
```

The changes in my PR have not in any way affected an author's ability to use MO in this way. I think what you were asking about (two texts in parallel, one of them an embedded video) has never been possible as a `par` must have exactly one `text` child. 

The PR changes do affect what a RS may do when it would encounter the above markup as we are now leaving it entirely up to implementations.

Regarding renaming MO ... I don't think we should do that. Not only would it be confusing for consumers of the spec, to have a major feature take on a new name but do exactly the same thing; it closes the door that we have been intentionally leaving open to potential future expansion of media types (that was the original reason for "media" and not "audio" overlays). 

Although, between you and me and anyone reading this thread, I am hoping that, in EPUB's future, we can adopt some of the VTT-based work in the [Sync Media CG](https://w3c.github.io/sync-media-pub/#with-webvtt) rather than make an increasingly hard to implement SMIL flavor.

iherman: What this whole thread and discussion tells me is that there is a need to have a separate reflection and plan for the future evolution of MO or anything replacing it. But that is definitely not for the current PR nor the Working Group as it stands today. Plan to participate (remotely) at the Publishing Salon at TPAC, @marisademeglio ?

marisademeglio: I may be able to join for a bit -- do you think there will be time and interest to discuss the future of MO at the Publishing Salon?

iherman: Something to check with @wareid and @TzviyaSiegman !

iherman: However, this very issue/PR may be on the F2F meeting's agenda, so it would be important for us that you would be there!

iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-08-26-epub#section1) on 2022-08-26

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-08-26-epub#resolution1):  Once the edits are finalized, merge PR 2402 and close issue 2397.***

### 1. Deprecate embedded audio or video as targets for Media Overlay `<text>` references.

_See github issue [epub-specs#2397](https://github.com/w3c/epub-specs/issues/2397)._




_See github pull request [epub-specs#2402](https://github.com/w3c/epub-specs/pull/2402)._




> *Ivan Herman:* See [Description of the issue background](https://github.com/w3c/epub-specs/pull/2402#issue-1340871554).

**Ivan Herman:** on a high level, what you can do with MO is associate a piece of time-stamped audio with a piece of text.  
… it is used in a bunch of a11y situations.  
… it is pretty well specified.  
… based on the tests I have made so far it is well implemented too.  
… however the specifications also try to allow association of an audio clip with a video (or several videos).  
… now you have two time related media side-by-side.  
… for that to work, the spec needs to define what is happening, considering that videos themselves often have UI controls themselves.  
… the spec tries to make some sense of this use-case, but this is really underspecified.  
… also, even for the cases which are relatively clear, we haven't found any implementations.  
… so we have a complicated case, underspecified in spec, and not implemented.  
… in side conversation with Avneesh, we did not find any important proper use case in a11y area.  
… the tests I created were trying to take a video and add a piece of background audio. But this is an artificial use case, as adding background audio to a video can be done much more easily in editing software rather than via MO.  
… so this lead to suggestion that we should just deprecate this type of video/audio MO from spec.  
… Marisa, mgarrish, and I opened #2397 to discuss.  
… based on this Marisa created PR #2402.  
… but we wanted to discuss with WG first.  

**Dave Cramer:** the fact that we don't have implementations of this weighs heavily to me.  
… reluctant to keep this in the spec for that reason.  

> *John Foliot:* is anyone aware of popcorn.js?.

**Ivan Herman:** is there any use-cases for this in the a11y area which requires this?.  
… MO has been included in the spec primarily for a11y reasons, although it is useful for other things.  

**George Kerscher:** i agree generally. Audio descriptions for the blind is implemented in TV as a second audio channel that can be turned on.  
… don't know the technology behind it (mixer behind the scenes that combines audio channels?).  
… don't know that we can break new ground here by adding that feature, but its something people would want.  
… what happens with descriptive audio is that the on-screen action is described for the blind person.  

> *John Foliot:* [http://ncamftp.wgbh.org/rd/hgp/hgp.html](http://ncamftp.wgbh.org/rd/hgp/hgp.html).

> *John Foliot:* watch the example at: [http://ncamftp.wgbh.org/rd/hgp/hgp.html](http://ncamftp.wgbh.org/rd/hgp/hgp.html).

**John Foliot:** there is a package called popcorn.js. Work done a while back by WGBH, where they were using this js library to use time-stamp markers to activate functions.  
… this is a demonstrated use-case, but i'm not aware of anyone who has taken this and run with it.  
… not sure that we need this in our spec, as long as we're not frustrating the use of it in epub.  

**Avneesh Singh:** this is removal of 9.3.2 from the RS spec. No one is blocking the creation of such MO files.  
… say you have a video going on, and you need description for it, MO doesn't seem like a good solution for this.  

**Ivan Herman:** the largest part of the MO concentrates on associating audio track with text.  
… this is not being touched.  
… the section that is affected is "embedded-media overlays", i.e. where HTML includes a video track.  
… to what John said, if there are js libraries that can do things in a website, they can do it in an epub as well. No problem..  
… what you described is a nice tool that can be used to get additional feature while a video is running. This can work in an epub as well.  

**Wendy Reid:** agreed. I think we're just removing something that is creating confusion/opening ourselves up to adding another unimplemented feature.  
… i haven't seen this in an epub or in any RS.  
… description of video of visually impaired is not done via MO.  
… RS would have to implement accessible media-player or epub would have to come with one, larger problem.  

> *John Foliot:* More info about accessible media requirements: [https://www.w3.org/TR/media-accessibility-reqs/](https://www.w3.org/TR/media-accessibility-reqs/).

**Ivan Herman:** can we pass a resolution that once editors are ready, that we can merge PR #2402?.  

> **Proposed resolution: Once the edits are finalized, merge PR 2402 and close issue 2397.** *(Wendy Reid)*

> *Matthew Chan:* +1.

> *Wendy Reid:* +1.

> *Dave Cramer:* +1.

> *GeorgeK:* +1.

> *Ivan Herman:* +1.

> *Brady Duga:* +1.

> *Masakazu Kitahara:* +1.

> *Avneesh Singh:* +1.

> *John Foliot:* +1.

> *Charles LaPierre:* +1.

> ***Resolution #1: Once the edits are finalized, merge PR 2402 and close issue 2397.***

</details>

iherman:
@marisademeglio @mattgarrish we have now a WG resolution to merge this after we all agree on the textual details. From my point of view, it is a go; I do not have any strong feeling about https://github.com/w3c/epub-specs/pull/2402#discussion_r949025461

marisademeglio:
> @marisademeglio @mattgarrish we have now a WG resolution to merge this after we all agree on the textual details. From my point of view, it is a go; I do not have any strong feeling about [#2402 (comment)](https://github.com/w3c/epub-specs/pull/2402#discussion_r949025461)

Great, no more changes needed then ?

iherman:
> > @marisademeglio @mattgarrish we have now a WG resolution to merge this after we all agree on the textual details. From my point of view, it is a go; I do not have any strong feeling about [#2402 (comment)](https://github.com/w3c/epub-specs/pull/2402#discussion_r949025461)
> 
> Great, no more changes needed then ?

Well, you should wait for @mattgarrish : he seems to have proposed some changes and those seem to be open still. Once that is settled it is a go...

epub33/core/index.html
mattgarrish: We don't need the bit about announcing that audio/video is there. What bothered me more was that we are saying the behaviour is unspecified when we're deprecating playback. We should be more direct about the change so authors know that it's not something that will get specified later.

I've trimmed the suggested text above to focus on that part only.

marisademeglio: I like it! At the start, where it says:

> Reading system playback of timed media is now deprecated.

Could it instead be:

"Guidance for reading system playback of timed media is now deprecated."

?

mattgarrish: I've switched the suggestion to: "[[epub-rs-33]] no longer provides guidance for reading systems on the playback of timed media ..."

Does that work better?

epub33/rs/index.html
mattgarrish: This change should also be integrated before merging.

epub33/core/index.html
marisademeglio: Ok final adjustment, I promise! How about 

"<p class="note">[[epub-rs-33]] no longer provides guidance for reading systems on the playback of embedded timed media. This scenario arises when a Media Overlays <code>text</code> element refers to an [[html]] [^video^] element. While this usage is permitted in Media Overlays, reading systems behavior is implementation-specific.</p>"

epub33/rs/index.html
marisademeglio: Done in https://github.com/w3c/epub-specs/pull/2402/commits/9d5bb342b48b788443f95e0cbda2721bd1795e52

mattgarrish:
This looks fine to me now but looks like we need to add a biblio entry for MO 3.2 to the RS file. I'm going to merge and manually correct that now.

PR #2404, opened 2022-08-25
Adding the value of "hidden" to the example

### LINKED ISSUE #2403: EPUB 3.3 specs - Error in Example 53?
[EXAMPLE 53](https://www.w3.org/TR/epub-33/#example-hiding-a-nav-element-in-spine): Hiding a nav element in spine
`<nav epub:type="page-list"  hidden="">`
Shouldn't 'hidden' be added? 
`<nav epub:type="page-list" hidden="hidden">`

iherman:
Although, per the [HTML spec](https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute), the empty string is valid and is equivalent to the value of `hidden`, it is probably a better practice to make it explicit (as for a number of attributes in the XHTML), so this PR does that.

Fix #2403

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2404.html" title="Last updated on Aug 25, 2022, 9:38 AM UTC (89a4c97)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2404/d9c0385...89a4c97.html" title="Last updated on Aug 25, 2022, 9:38 AM UTC (89a4c97)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -6722,7 +6722,7 @@ No Entry</pre>
 
 					<pre>&lt;nav
     epub:type="page-list"
-    hidden="">
+    hidden="hidden">
    &lt;h2>Pagebreaks of the print version, third edition&lt;/h2>
    &lt;ol>
       &lt;li>


mattgarrish:
I believe it's technically the other way around. [Boolean attributes](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes) don't really have a value; its the presence or absence of the attribute that matters.

You can add the name of the attribute as its value, but using the name often leads people to believe they can use other values, like true and false, with unintended consequences. The HTML spec even notes that.

iherman:
Yes, but, per [the html spec](https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute), the `hidden` attribute is not boolean but "enumerated". The empty string and the value of `hidden` (as opposed to `until-found`) are equivalent.

mattgarrish:
Ah, it's changed recently... or because we switched off the W3C spec.

mattgarrish:
Looks like a very recent change, to answer my own question. Even MDN is still referring to it as a boolean attribute without reference to the new `until-found` value.

It doesn't appear to change anything as far as nav doc hiding goes, though, since the attribute in its entirety has to be ignored by reading systems, not based on any specific value it has.

mattgarrish (approved):

PR #2411, opened 2022-09-06
Add note for summaries

mattgarrish:
This is an update for #2401. Part of the resolution for the issue was to add a note explaining that summaries will still be useful for market segments where automated processing and display of metadata are not expected - library systems, in particular, as we've discussed. This PR adds the missing note.

- Accessibility [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/summary-use-note/epub33/a11y/index.html)
- Accessibility [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/summary-use-note/epub33/a11y/index.html)

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -352,6 +352,14 @@
 					[[schema-org]] accessibility metadata not specified in this section.</p>
 
 				<div class="note">
+					<p>This specification assumes that conformance and discoverability metadata will be processed and
+						presented in a human-readable manner, removing the need for a summary unless there is additional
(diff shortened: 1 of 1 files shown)


clapierre (approved):
works for me.

gregoriopellegrino (approved):

PR #2425, opened 2022-09-14
Rearrange unsupported features

### LINKED ISSUE #2423: Explain legacy elements
There still seems to be confusion about the purpose of the legacy elements. They exist only for allowing EPUB 3s to render, as much as possible, in EPUB 2 reading systems.

The spec doesn't introduce this purpose in the package document's legacy content section, so it requires users to follow the link on the word legacy in almost all cases to get to the definition (oddly, we explain the purpose for the meta element in a note).

I wonder if we should even bother having legacy elements described in the unsupported features section, since they only apply to those elements (plus the toc attribute for referencing the NCX). It would be clearer to move the prose up to the section.

### LINKED ISSUE #2424: Reading system requirements in the authoring spec
The requirements for how reading systems handle unsupported features would seemingly belong in the reading system specification now. It's the only case where we still have crossover requirements.

mattgarrish:
This PR makes the following changes:

- the explanation of "legacy features" is integrated into the existing package document section as these are the only legacy features
- adds a new intro to try and explain that the legacy features only exist to allow epub 2 reading systems to open epub 3 content, and there probably isn't much use for them anymore
- makes some minor tweaks to the legacy feature descriptions to better highlight that they're only for epub 2 reading systems
- moves the reading system support requirement for legacy features to a new corresponding section in the reading system spec
- also moves the under-implemented and deprecated support requirements to the RS spec

Fixes #2423 
Fixes #2424 

    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-2424/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fix/issue-2424/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2425.html" title="Last updated on Sep 14, 2022, 11:45 PM UTC (5bedee4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2425/0792d0d...5bedee4.html" title="Last updated on Sep 14, 2022, 11:45 PM UTC (5bedee4)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -11707,9 +11707,9 @@ EPUB/images/cover.png</pre>
 						Recommendation</a></h3>
 
 				<ul>
-					<li>14-Sep-2022: Combined the legacy feature section into the package document definition. See <a
+					<li>14-Sept-2022: Combined the legacy feature section into the package document definition. See <a
 							href="https://github.com/w3c/epub-specs/issues/2423">issue 2423</a>.</li>
-					<li>07-Sep-2022: Added clarifying requirement that the manifest only list publication resources. See
+					<li>07-Sept-2022: Added clarifying requirement that the manifest only list publication resources. See
 							<a href="https://github.com/w3c/epub-specs/issues/2413">issue 2413</a>.</li>
 					<li>28-Aug-2022: Added a note to Media Overlays <code>text</code> element definition regarding using
 						embedded timed media. See <a href="https://github.com/w3c/epub-specs/issues/2397">issue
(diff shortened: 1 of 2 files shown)


iherman (approved):
A note to @dlazin : this PR introduce a new _MUST_ statement that will require a test. The test structure could be to have both a nav and an ncx document in the publication, with the ncx being slightly different; the test passes if the ncx version is not visible.

wareid (approved):

PR #2445, opened 2022-10-03
Update epub-properties-use.md

nekennedy:
Adding PRH usage


mattgarrish:
Thanks so much for putting this together @nekennedy!

Can you do the IPR magic on this and the a11y PR, @iherman ?

PR #2449, opened 2022-10-05
Extra note on RS conformance

### LINKED ISSUE #2433: Can a conforming RS reject EPUBs that reference external entities?
In the [RS spec](https://www.w3.org/TR/epub-rs-33/#confreq-rs-xml-extid) we say:

```
In addition, when processing XML documents, a reading system MUST NOT resolve 
[external identifiers](https://www.w3.org/TR/xml/#NT-ExternalID) in DOCTYPE, 
ENTITY and NOTATION declarations [[xml](https://www.w3.org/TR/epub-rs-33/#bib-xml)].
``` 

However, EPUB check will report an error for such content. For instance, when run against [the test for this](https://github.com/w3c/epub-tests/tree/main/tests/pub-xml-external-id) it will produce this error:

```
EPUB/content_001.xhtml:4, 36: External entities are not allowed in EPUB v3 documents. 
   External entity declaration found: xxe.
EPUB/foo.xhtml:1, 7: Error while parsing file: element "span" not allowed here; 
   expected element "html" (with xmlns="http://www.w3.org/1999/xhtml")
```
It is unclear if a RS that rejects this content is conformant. Must it process the document and display the &xxe; in text? There may be other cases where EPUB check reports an error but the spec gives guidance on how to process the document. For content that comes from a publisher, it seems like failing is the best thing to do, as it forces a fix to the content. Should we add some sort of blanket statement that allows an RS to reject such content? And do we need to specify that that is an acceptable outcome for the tests?

iherman:
This PR offers a solution for the core issue raised in #2433.

Fix #2433.

For preview see:

    * [Preview](https://raw.githack.com/w3c/epub-specs/note-on-rs-conformance/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/note-on-rs-conformance/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -224,16 +224,18 @@
 			<h3>Reading system conformance</h3>
 
 			<p>Whether a [=reading system=] has to support a feature is mentioned at the beginning of its section. To be
-				conformant with this specification, reading systems MUST, with one exception, support all required features as well as all
+				conformant with this specification, reading systems MUST support all required features as well as all
 				applicable conditionally-required features (e.g., to support image rendering if the reading system has a
 				[=viewport=]) as defined in their respective sections.</p>
 
-			<p>
-				The one exception to the conformance requirement stated above is for reading systems that do not allow
-				users to load their own EPUB publications (e.g., all content is controlled and delivered from a bookstore or
-				library system). In this case, if a reading system developer can demonstrate that a requirement is handled
-				in an upstream process (e.g., EPUB publications with duplicate [^itemref^] entries [[epub-33]] cannot enter
-				the system), the reading system is still considered conforming.
+			<p class="note">
+				As a reading system is not necessarily a single application, but may exist as a distributed system,
+				it is not always the case that reading system requirements will be met in the application that 
+				renders EPUB publications to users. An example is a reading system that solely interacts with a
+				controlled content repository (e.g., a bookstore or library system). In this case, if a reading
+				system developer can demonstrate that requirements are not applicable to the application
+				(e.g., EPUB publications with duplicate [^itemref^] entries [[epub-33]] cannot enter the system),
+				the reading system as a whole is still considered conforming.
 			</p>
 
 			<p>When supporting recommended and optional features, reading systems MUST meet all normative requirements
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -224,9 +224,18 @@
 			<h3>Reading system conformance</h3>
 
 			<p>Whether a [=reading system=] has to support a feature is mentioned at the beginning of its section. To be
-				conformant with this specification, reading systems MUST support all required features as well as all
+				conformant with this specification, reading systems MUST, in general, support all required features as well as all
 				applicable conditionally-required features (e.g., to support image rendering if the reading system has a
 				[=viewport=]) as defined in their respective sections.</p>
+
+			<p>
+				There is one exception to the conformance requirement as stated above. Not all reading systems offer the possibility
+				to load an EPUB publication directly (a feature often referred to as "side-loading"); instead, the reading 
+				system is part of a larger publishing workflow which automatically filters out non-conformant EPUB publications 
+				(e.g., whether the publication contains duplicate <a data-cite="epub-33#itemref"><code>itemref</code></a> entries in 
+				its <a data-cite="epub-33#spine">spine</a> [[epub-33]]). It is reasonable for such reading systems not to implement
+				strong checks for those cases. This decision does <em>not</em> affect the conformance of such reading systems.
+			</p>
 
 			<p>When supporting recommended and optional features, reading systems MUST meet all normative requirements
 				as defined in their respective sections.</p>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -224,17 +224,16 @@
 			<h3>Reading system conformance</h3>
 
 			<p>Whether a [=reading system=] has to support a feature is mentioned at the beginning of its section. To be
-				conformant with this specification, reading systems MUST, in general, support all required features as well as all
+				conformant with this specification, reading systems MUST, with one exception, support all required features as well as all
 				applicable conditionally-required features (e.g., to support image rendering if the reading system has a
 				[=viewport=]) as defined in their respective sections.</p>
 
 			<p>
-				There is one exception to the conformance requirement as stated above. Not all reading systems offer the possibility
-				to load an EPUB publication directly (a feature often referred to as "side-loading"); instead, the reading 
-				system is part of a larger publishing workflow which automatically filters out non-conformant EPUB publications 
-				(e.g., whether the publication contains duplicate <a data-cite="epub-33#itemref"><code>itemref</code></a> entries in 
-				its <a data-cite="epub-33#spine">spine</a> [[epub-33]]). It is reasonable for such reading systems not to implement
-				strong checks for those cases. This decision does <em>not</em> affect the conformance of such reading systems.
+				The one exception to the conformance requirement stated above is for reading systems that do not allow
+				users to load their own EPUB publications (e.g., all content is controlled and delivered from a bookstore or
+				library system). In this case, if a reading system developer can demonstrate that a requirement is handled
+				in an upstream process (e.g., EPUB publications with duplicate [^itemref^] entries [[epub-33]] cannot enter
+				the system), the reading system is still considered conforming.
 			</p>
 
 			<p>When supporting recommended and optional features, reading systems MUST meet all normative requirements


iherman:
Note that this also would abide to the [WG resolution](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-09-29-epub#resolution1) of 2022-09-29.

murata2makoto:
I think that the proposed additional wording says too much without obvious advantages.  In general, I do not believe that conformance requirements can ensure usefulness or safety.  I prefer shorter wording.

iherman:
>In general, I do not believe that conformance requirements can ensure usefulness or safety. 

The spec has to say what a conformant EPUB RS is; it is an important point for a RS manufacturer to be able to rightfully claim to be conformant to a standard. I am happy to be more succinct, if possible, but we need something like that imho.

mattgarrish:
I'm a bit concerned that we're blurring the lines of conformance with this change. I understand where you're trying to go with the paragraph, but couldn't we do it in a note without having to add an "in general" to supporting requirements?

I'd maybe adjust it more like this:

```
<div class="note">
	<p>Not all reading systems offer the possibility to load an EPUB publication directly &#8212; a feature
	often referred to as "side-loading". Instead, the reading system is part of a system that automatically
	filters out non-conforming EPUB publications before they can reach users. In these cases, some
	conformance checks defined in this specification are performed upstream from the reading system
	(e.g., rejecting publications with duplicate <a data-cite="epub-33#itemref"><code>itemref</code></a>
	entries [[?epub-33]]). This model does <em>not</em> affect the conformance of such reading systems;
	it is only required that the outcome be the same.</p>
</div>
```

What we're saying here doesn't feel all that different from the note about OCF processors being able to ignore RS requirements: https://w3c.github.io/epub-specs/epub33/rs/#sec-ocf

iherman:
> I'm a bit concerned that we're blurring the lines of conformance with this change. I understand where you're trying to go with the paragraph, but couldn't we do it in a note without having to add an "in general" to supporting requirements?
> 
> I'd maybe adjust it more like this:
> 

That was my first thought, too. What made me pause and then change is the fact that we define, "normatively", so to say, the conformance of a RS in the first paragraph, and then we would slightly weaken it in the note. Do you think that is still o.k.?

mattgarrish:
> we define, "normatively", so to say, the conformance of a RS in the first paragraph, and then we would slightly weaken it in the note

I don't know that it weakens the requirement so much as expands what the definition of a "reading system" can be. It's not just the app the user reads the content from but possibly the whole chain that gets the content from publisher to user when that's a closed loop.

What bothered me about the original phrasing is that "in general" weakens the requirement for all reading systems. The next paragraph then seemed to suggest there's another exception to only generally having to meet the requirements that applies for this case of a reading system with upstream control over the content.

It's always nasty business to carve out exceptions on requirements, which is why I was hoping we could do it via a note that explains that upstream filtering is considered equivalent to meeting the requirement.

If you're more comfortable with the prose being normative, though, I'd suggest switching "in general" to "with one exception" to more clearly tie that into the next paragraph, which might only need to say:

```
<p>The one exception to the conformance requirement stated above is for reading systems that do not allow
	users to load their own EPUB publications (e.g., all content is controlled and delivered from a bookstore or
	library system). In this case, if a reading system developer can demonstrate that a requirement is handled
	in an upstream process (e.g., EPUB publications with duplicate [^itemref^] entries [[epub-33]] cannot enter
	the system), the reading system is still considered conforming.</p>
```

bduga:
Having read this, now I am not sure I like having it at all. For instance, does this mean if you support side loading you can't reject due to epub check (or other) errors? Another option is to just leave things as they are and allow reading systems to reject for arbitrary reasons without mentioning it in the spec (yes, I know I raised the issue to begin with). We do already reject perfectly valid EPUBs for other reasons. O maybe a note that says reading systems are free to impose additional restrictions on the EPUBs they will accept?

iherman:
I like the alternative of @mattgarrish, so I have made those changes to have a common ground for discussion.

iherman:
@bduga wrote:

> Having read this, now I am not sure I like having it at all. For instance, does this mean if you support side loading you can't reject due to epub check (or other) errors? 

I am not sure what you mean. While the spec does not say what "reject" mean, the RS system must react somehow on invalid EPUB for conformance. (E.g., by a pop up window, or something similar. If I submit an invalid XML to the RS on my machine, it may display an XML error page instead of the content, following the way a browser would react on an invalid XHTML document.)

> Another option is to just leave things as they are and allow reading systems to reject for arbitrary reasons without mentioning it in the spec (yes, I know I raised the issue to begin with).

I.e., put back the genie into the bottle that you opened?😀 Sorry about that, but you _did_ ask the right question!

> We do already reject perfectly valid EPUBs for other reasons. O maybe a note that says reading systems are free to impose additional restrictions on the EPUBs they will accept?

It is not only a matter of additional restrictions; I believe it is more often the case of RS-s being more permissive. In the aforementioned case of invalid XML, a RS may decide to reparse the content as tag-soup HTML and display it without comment. Or it may decide to accept an EPUB with duplicate itemrefs in the spine and do something sensible (like dropping the second occurrence). How would you characterize this? I am a bit worried that we would open the flood-gates by specifying this. What would conformance mean in the first place?

mattgarrish:
> For instance, does this mean if you support side loading you can't reject due to epub check (or other) errors?

This is where I thought the original issue was going. I'm fine, either with the above or in place of it, stating (or noting) that we aren't defining how a reading system has to behave when it encounters content that violates content/processing requirements.

So maybe something like:

> This specification does not require reading systems to load EPUB publications, or resources within them, when they violate content authoring or processing requirements (e.g., they define entities in an external DTD subset).

It is a different issue from whether requirements are handled upstream.

(Edit: cleaned up the suggested text a bit.)

iherman:
> > For instance, does this mean if you support side loading you can't reject due to epub check (or other) errors?
> 
> This is where I thought the original issue was going. I'm fine, either with the above or in place of it, stating (or noting) that we aren't defining how a reading system has to behave when it encounters content that violates content/processing requirements.
> 
> So maybe something like:
> 
> > This specification does not require reading systems to load EPUB publications, or resources within them, when they violate content authoring or processing requirements (e.g., they define entities in an external DTD subset).

I do not have any problem adding such a generic statement into the spec which _is_ good to have in any case. But I do not see it as a replacement to the text which is in the PR right now, because it does not say anything about conformance. In a way, it says that the spec does not specify a behaviour if the RS is _not_ conformant... 

So the question really is (and that is something for @bduga, but also @wareid or @davemanhall to say): is it important for a RS vendor to be able to say that their RS is conformant per the W3C specification? If not, we can indeed _replace_ the proposed text with this; if it is, then better keep both...

wareid:
It depends on who is asking. Publishers may want to know if we're conformant, but users won't know/care/understand. I am in favour of being as explicit as possible though, so I like Matt's proposals.

wareid (approved):

mattgarrish:
I think we're worrying too much about making reading systems in closed loops conformant. I'd be fine leaving that part unstated and let those vendors it affects explain why they don't feel requirements apply to them (assuming anyone cares). We're sort of endorsing potential security risks by formally okaying any claim that content can never slip past an upstream check.

But it's not a critical issue to me, so I'm fine with whatever people decide we need to say here.

davemanhall:
The definition of a 'reading system' already contains the idea that the 'system' may be distributed:
https://www.w3.org/TR/epub-rs-33/#sec-intro-overview.

Given that definition, I'm not sure any change really needs to be made to this section?

iherman:
I am not sure where to go. We got one approval from @wareid, otherwise we seem to got stalled...

iherman:
> The definition of a 'reading system' already contains the idea that the 'system' may be distributed: https://www.w3.org/TR/epub-rs-33/#sec-intro-overview.
> 
> Given that definition, I'm not sure any change really needs to be made to this section?

@davemanhall there are two issues with that. (1) that section is not normative and (2) the original issue, ie, making statements about the conformance of a RS, is not addressed in that text

mattgarrish:
I'd still like to drop the formal exception and only make a note if we can. Could we link in "distributed system" here to explain that requirements can be satisfied in other ways? For example:

> Note
> As a reading system is not necessarily a single application, but may exist as a distributed system, it is not always the case that reading system requirements will be met in the application that renders EPUB publications to users. An example is a reading system that solely interacts with a controlled content repository (e.g., a bookstore or library system). In this case, if a reading system developer can demonstrate that requirements are not applicable to the application (e.g., EPUB publications with duplicate [^itemref^] entries [[epub-33]] cannot enter the system), the reading system as a whole is still considered conforming.

I don't see how this weakens the requirements as it's not like we're excusing them from being considered and handled in some way. It just doesn't have to happen in the app.

iherman:
I have taken over your proposal, @mattgarrish, it is indeed better that way.

I would think this is ready to merge.

davemanhall (approved):
This feels like an appropriate addition.

PR #2450, opened 2022-10-05
Succinct wording for conformance requirements on XML processing

murata2makoto:
Here is my proposed rewrite of 3.6 XML processing in EPUB 3.3 RS.  It is intended to address #2433 at least partially.  But more wording about XML security issues is needed somewhere else.

See:

* For EPUB 3.3 Reading Systems:
    * [Preview](https://raw.githack.com/murata2makoto/epub-specs/xml-processing-conformance/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/murata2makoto/epub-specs/xml-processing-conformance/epub33/rs/index.html)

PR #2451, opened 2022-10-06
Made the XML conformance section more succinct

### LINKED ISSUE #2447: XML security risk: expansion of internal parsed entities
RFC 7303 (XML Media Types) has a section ["Security considerations"](https://datatracker.ietf.org/doc/html/rfc7303#section-10),  The last paragraph is quoted below:

> Note that it is also possible to construct [XML](https://datatracker.ietf.org/doc/html/rfc7303#ref-XML) documents that make
>    use of what XML terms "[XML-]entity references" to construct repeated
>    expansions of text.  Recursive expansions are prohibited by [XML] and
>    XML processors are required to detect them.  However, even non-
>    recursive expansions may cause problems with the finite computing
>    resources of computers, if they are performed many times.  For
>    example, consider the case where XML-entity A consists of 100 copies
>    of XML-entity B, which in turn consists of 100 copies of XML-entity
>    C, and so on.

Note that such repetitive expansion of entities does not need external parsed entities (see #2433).   Internal entities can easily cause such expansion.  [entityExpansion.zip](https://github.com/w3c/epub-specs/files/9706493/entityExpansion.zip) contains two EPUB publications containing such internal parsed entities.  

One of them (IPAttackConformant.epub) conforms to EPUB 3.3.   This is because [3.9 XML conformance](https://www.w3.org/TR/epub-33/#sec-xml-constraints) in EPUB 3.3 does not prohibit internal parsed entities within internal DTD subsets.  Meanwhile, [3.6 XML processing](https://www.w3.org/TR/epub-rs-33/#sec-epub-rs-conf-xml) in EPUB 3.3 RS is silent about expansion of internal XML parsed entities.  Moreover, [5.1 Validating and Non-Validating Processors ](https://www.w3.org/TR/2008/REC-xml-20081126/#proc-types) in the XML recommendation requires that all conformant XML processors expand internal parsed entities declared in internal DTD subsets.  Thus,  EPUB 3.3 RS cannot prohibit the expansion of internal parsed entities.

The other EPUB publication in the ZIP file is derived from the first one, but its XHTML content documents contain DOCTYPE statements not allowed by 1[3.1.1 The DOCTYPE](https://html.spec.whatwg.org/multipage/syntax.html#the-doctype) in WHATWG HTML.

If the repetitive expansion of entities is a problem, an easy solution is to prohibit internal DTD subsets in [3.9 XML conformance](https://www.w3.org/TR/epub-33/#sec-xml-constraints) in EPUB 3.3.  But this change might invalidate some existing EPUB publications.

It is worth noting that some RSes such as Thorium do not expand references to internal parsed entities.  But this means that XML processing in Thorium does not conform to XML.
  

P.S.  I should have raised this issue long time ago.  I learned this issue when XML was being designed. My apology.

iherman:
This PR is a replacement for #2450, which was mistakenly branched off on an earlier state of the spec.

The PR also adds some text to the security section (also anticipating some WG discussion on #2447 to happen on 2022-10-06).

See:

    * [Preview](https://raw.githack.com/w3c/epub-specs/makoto-xml-conformance-change/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/makoto-xml-conformance-change/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -309,24 +309,14 @@
 			<section id="sec-epub-rs-conf-xml">
 				<h4>XML processing</h4>
 
-				<p>A reading system MUST be both of the following:</p>
+				<p  id="confreq-rs-xml-nval"
+				data-tests="#pub-xml-non-validating_invalid,#pub-xml-non-validating_unclosed">A [= reading system =] MUST use a 
+				non-validating XML processor [[xml]] that:</p>
 
 				<ul class="conformance-list">
-					<li>
-						<p id="confreq-rs-xml-nval"
-							data-tests="#pub-xml-non-validating_invalid,#pub-xml-non-validating_unclosed">a <a
-								data-cite="xml#proc-types">conformant non-validating XML processor</a> [[xml]].</p>
-					</li>
-					<li>
-						<p id="confreq-rs-xml-ns" data-tests="#pub-xml-names">a <a
-								data-cite="xml-names#ProcessorConformance">conformant processor</a> as defined in
-							[[xml-names]].</p>
-					</li>
+					<li id="confreq-rs-xml-extid" data-tests="#pub-xml-external-id">does not read external DTD subsets [[xml]];</li>
+					<li id="confreq-rs-xml-ns" data-tests="#pub-xml-names">conforms to [[xml-names]].</li>
 				</ul>
-
-				<p id="confreq-rs-xml-extid" data-tests="#pub-xml-external-id">In addition, when processing XML
-					documents, a reading system MUST NOT resolve <a data-cite="xml#NT-ExternalID">external
-						identifiers</a> in DOCTYPE, ENTITY and NOTATION declarations [[xml]].</p>
 			</section>
 
 			<section id="sec-epub-rs-i18n">
@@ -2426,6 +2416,14 @@
 					publications through the process of "sideloading") SHOULD treat such content as insecure (e.g.,
 					prompt users to allow scripting and network access).</p>
 
+				<p >When processing XML documents, a reading system SHOULD NOT resolve 
+					<a data-cite="xml#NT-ExternalID">external identifiers</a> in DOCTYPE, ENTITY and NOTATION declarations [[xml]].
+					Reading systems SHOULD also consider security risks related to <a data-cite="xml#sec-internal-ent">internal</a> or 
+					<a data-cite="xml#sec-external-ent">external</a> XML entities like, for example, DoS attacks 
+					also known as <a href="https://en.wikipedia.org/wiki/Billion_laughs_attack">"Billion laughs attacks"</a>
+					or <a href="https://en.wikipedia.org/wiki/XML_external_entity_attack">"XML external entity attacks"</a>.
+				</p>
+	
 				<div class="note">
 					<p>Additional security recommendations for external links, network access and scripting are
 						available in <a href="#sec-epub-rs-external-links"></a>, <a href="#sec-epub-rs-network-access"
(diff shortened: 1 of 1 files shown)


iherman:
The issue was discussed in a [meeting](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-10-07-epub#section2) on 2022-10-07

- no resolutions were taken

### 2. XML security risk: expansion of internal parsed entities (issue epub-specs#2447)

_See github issue [epub-specs#2447](https://github.com/w3c/epub-specs/issues/2447)._




_See github issue [epub-specs#2433](https://github.com/w3c/epub-specs/issues/2433)._




**Murata Makoto:** old issue; more than 20 years ago.  
… related to external parsed entities issue 2433.  
… this issue is about INTERNAL parsed entities.  
… doesn't use external identifiers or URLs.  
… these are defined in internal DTD subsets.  

> *Dave Cramer:* See [description of the attack](https://en.wikipedia.org/wiki/Billion_laughs_attack).

**Murata Makoto:** if an internal reference references a different internal reference, things can get out of hand quickly.  
… some EPUB reading systems ignore rec of XML and ignore internal parsed entities. This is non-conformant but avoids security issue.  
… what should we do? Prohibit internal DTD subsets?.  
… which gets rid of internal parsed entities..  
… can exissting RSes be destryeed by malicious content?.  

_See github pull request [epub-specs#2451](https://github.com/w3c/epub-specs/pull/2451)._




**Ivan Herman:** thanks Makoto for raising the issue.  
… the problem is that removing the possibility to use internal entities is something we can't do because of existing epub docs that might use this.  
… we cannot invalidate existing content per our charter.  
… as part of a pull request that makoto and I did together is a separate section.  
… in the security section of RS spec.  

> *Ivan Herman:* See [proposed text in the PR's security section](https://raw.githack.com/w3c/epub-specs/makoto-xml-conformance-change/epub33/rs/index.html#security-privacy-recommendations).

**Ivan Herman:** which say that RS should be aware of this problem and should deal with it..  
… so we should draw attention to the issue.  

**Dave Cramer:** I don't think we can or should forbid internal entities.  
… the problem has been around 20 years and EPUB is still functional.  
… it's a security vulnerability, we may need to warn people in our security section.  
… ordering RSs to follow a specific algorithm might be overkill.  
… let's warn people.  

**Murata Makoto:** Basically, not our problem.  

**Brady Duga:** this is a known problem in XML. Are there known solutions?.  
… do they have mitigations? Like in libxml?.  

> *Dave Cramer:* "Defenses against this kind of attack include capping the memory allocated in an individual parser if loss of the document is acceptable, or treating entities symbolically and expanding them lazily only when (and to the extent) their content is to be used.".

**Ivan Herman:** makoto may know about libraries.  
… makoto made some sample epubs, and we tested them.  
… I tested on Thorium and Apple Books.  
… both reacted by rejecting the EPUB.  
… without identifying the problem.  
… we suspect they don't accept internal entities.  

**Murata Makoto:** I'm not aware of general solutions.  
… XML WG was aware of this issue, but had no good solutions.  
… thorium doesn't reject publication, it ignores the entity but not the entire publication.  
… this behavior is non-conformant.  

**Dave Cramer:** I don't think it's entirely fair to say that this implementation is non-conformant because it's not deploying my vulnerability.  
… not fair to RS.  

**Wendy Reid:** it's like the viewport discussion.  
… we want to avoid putting burdens on readers.  
… does the fail quietly solution work?.  
… is that a bad thing?.  
… I think because we are running out of time.  
… people should look at the PR.  
… there are other things in the PR.  
… then we move on.  

**Dave Cramer:** I think we add something to the security section.  
… it hasn't been a problem in the past, internal entities aren't usually a problem unless malicious.  
… what are the consequences of a malicious use.  
… the consequences seem small.  
… good enough to say "XML has a known vulnerability:.  

**Brady Duga:** I agree putting this in the security section, maybe mentioning other xml vulnerabilities.  
… any attack like this can jeopardize personal information.  
… although no one has bothered to do this.  

**Dave Cramer:** Some of this is also outside the scope of the WG, we can't ask for people to patch the OS they are running dev machines on.  

**Wendy Reid:** consensus: let's look people look at Ivan's PR.  
… comment on it.  
… ten minutes remaining.  
… two more topics.  

</details>

iherman:
@murata2makoto @mattgarrish we got stalled... are we ready to merge this?

Note that the WG has briefly discussed the addition on the security section on its [last call](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-10-07-epub#section2) and we may get input from there, too. But at least the three of us should be aligned...

mattgarrish:
Seems fine to me now.

iherman:
@murata2makoto I intend to merge this on the week end unless you have some further comments.

murata2makoto:
Sorry for my belated response.  This PR looks fine to me.

PR #2453, opened 2022-10-11
Handling problematic viewport/ICB settings 

### LINKED ISSUE #2442: Some comments on the viewport meta tag
First, sorry about missing the discussion in #2292, #2341, and #2350 last summer. It went totally under my radar! 😅

A couple (late) comments:

- there has been recent developments around the concept of viewport:
  - viewport is one of the topics of the [interop 2022](https://github.com/web-platform-tests/interop-2022-viewport) project
  - the [CSS WG resolved](https://github.com/w3c/csswg-drafts/issues/4766#issuecomment-585314599) a while ago to retire the CSS Device Adaptation spec, remove the definition of `@viewport`, and move the `viewport` `meta` tag definition to a new spec. A [CSS Viewport draft](https://drafts.csswg.org/css-viewport/) was just created.
- the new EBNF syntax of the `viewport` `meta` tag in EPUB 3.3 is somewhat restrictive when compared to the parsing algorithm defined in CSS Viewport (formerly CSS Device Adaptation).
  - the parsing algorithm in css-viewport does not produce errors, but always returns a (possibly empty) se tof name/value pairs.
  - in contrast, our definition defines a static syntax, but does not completely tell how RS should deal with an invalid value (except for number-prefixed height/width values).
  - for example, for a viewport value of `",invalid,,height=200,width=,width=100"` the algorithm returns the name/value pair list `("height","200"),("width","100")`, when our spec just says the value is invalid.
  - that said, the CSS algorithm itself has room for improvement: it could define parsing errors, error handling, and how to deal with duplicate properties. (also, it would be worth being rewritten based on Infra).

Some concerns:

- is it ok to reference `css-device-adapt` (even informatively) when we know the spec is doomed?  
  At least, I would rephrase the following:
  > The syntax of this grammar is also influenced by the processing model for transforming `viewport` `meta` tags to `@viewport` CSS rules, as defined in [css-device-adapt](https://www.w3.org/TR/epub-33/#bib-css-device-adapt).  

  to something along the lines of:
  > The syntax of this grammar is also influenced by the parsing algorithm for the `viewport` `meta` element, as defined in [css-device-adapt](https://www.w3.org/TR/epub-33/#bib-css-device-adapt).

  (so that we no longer reference the for-now-defunct `@viewport` at-rule.
- I have a (slightly) nagging feeling with having our own normative definition of the `viewport` `meta` element in EPUB:
  - what if this definition conflicts with whatever will (hopefully) be defined in CSS Viewport? It already kinda does. Although our definition seems to be a subset: something valid to the EPUB definition will be parsed as expected by the CSS algorithm, but not all values that are parsed into something by the CSS algorithm are conforming to our definition.
  - how confident are we that this definition matches the reality of implementations? (the existing EPUB tests not extensively evaluate the syntax variations)
  - can our syntax be informative only, somehow?
  - these types of structured values are better defined algorithmically (in my opinion), so that the algorithm can both define syntax (i.e. parsing errors) and error handling (i.e. recoverable errors that still produce a returned value).

### LINKED ISSUE #2456: Editorial: `meta` points to the package doc element instead of HTML element in F.2
In [F.2 Syntax](https://www.w3.org/TR/2022/CRD-epub-33-20221003/#app-viewport-meta-syntax):

> A viewport [meta](https://www.w3.org/TR/2022/CRD-epub-33-20221003/#dfn-meta) tag [[html](https://www.w3.org/TR/2022/CRD-epub-33-20221003/#bib-html)] MUST have name and content attributes that conform to the following definition

`meta` refers to the package document element rather than the HTML element. I don't know ReSpec enough to submit a PR 😅.

iherman:
As agreed on the [WG call](https://www.w3.org/publishing/groups/epub-wg/Meetings/Minutes/2022-10-07-epub#resolution1) this is an attempt to fix #2442. Some notes:

- The case of repeated `width`/`height` setting within the `viewport` tag is now part of the `viewport` definition in the core spec. Having it there means that epubcheck can consider such a `viewport` setting as valid (although it may choose to issue a warning).
- Clearly, the reaction of various reading systems on the problematic cases is different, contrast, e.g., https://github.com/w3c/epub-specs/issues/2442#issuecomment-1271820337 and https://github.com/w3c/epub-specs/issues/2442#issuecomment-1268691239. My conclusion from this, but also of the discussion, is that the RS can only include non-binding solutions, in other words, _MAY_ statements. This is what is there. Maybe in the discussion we can agree to put a _SHOULD_ to one of these.
- There may have to be a more general statement in the RS spec on what happens if erronous content is met, e.g., an invalid XHTML content file. I believe a note or some sentence in §2 (Reading system conformance) would be in order which says something like 'the RS _MAY_ issue and error message'. I would let @mattgarrrish think about it before doing it. Note that, if we decide to put something there, the last (new) paragraph in the PR may become moot.

See:

* For EPUB 3.3:
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/viewport-spec-issues/epub33/core/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/core/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/viewport-spec-issues/epub33/core/index.html)
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/viewport-spec-issues/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/viewport-spec-issues/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2453.html" title="Last updated on Oct 15, 2022, 4:10 AM UTC (f8b1398)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2453/2fb7a66...f8b1398.html" title="Last updated on Oct 15, 2022, 4:10 AM UTC (f8b1398)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -10733,7 +10733,7 @@ html.my-document-playing * {
 						href="#viewport.ebnf.assign">assignment character</a>.</p>
 
 				<p>
-					The [^meta/content^] attribute SHOULD NOT include several <code>height</code>, respectively <code>width</code>,  assignments.
+					The [^meta/content^] attribute MUST NOT include several <code>height</code>, respectively <code>width</code>,  assignments.
 				</p>
 
 				<p>There are no restrictions on any other attributes allowed on the [^meta^] element by the [[html]]
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -10732,6 +10732,12 @@ html.my-document-playing * {
 					contain <a href="#viewport.ebnf.sep-char">separator characters</a> or the <a
 						href="#viewport.ebnf.assign">assignment character</a>.</p>
 
+				<p>
+					If the <code>height</code> or the <code>width</code> assignments have multiple occurences whithin an otherwise
+					correct value of the [^meta/content^] attribute, only the first occurence of the <code>height</code>, respectively
+					<code>width</code>, definition counts, and all others MUST be ignored.
+				</p>
+
 				<p>There are no restrictions on any other attributes allowed on the [^meta^] element by the [[html]]
 					grammar.</p>
 
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -10733,9 +10733,7 @@ html.my-document-playing * {
 						href="#viewport.ebnf.assign">assignment character</a>.</p>
 
 				<p>
-					If the <code>height</code> or the <code>width</code> assignments have multiple occurences whithin an otherwise
-					correct value of the [^meta/content^] attribute, only the first occurence of the <code>height</code>, respectively
-					<code>width</code>, definition counts, and all others MUST be ignored.
+					The [^meta/content^] attribute SHOULD NOT include several <code>height</code>, respectively <code>width</code>,  assignments.
 				</p>
 
 				<p>There are no restrictions on any other attributes allowed on the [^meta^] element by the [[html]]
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -10644,7 +10644,7 @@ html.my-document-playing * {
 			<section id="app-viewport-meta-syntax">
 				<h3>Syntax</h3>
 
-				<p>A <code>viewport</code> [^meta^] tag [[html]] MUST have <code>name</code> and <code>content</code>
+				<p>A <code>viewport</code> <a data-cite="html#meta"><code>meta</code></a> tag [[html]] MUST have <code>name</code> and <code>content</code>
 					attributes that conform to the following definition:</p>
 
 				<dl>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -11783,6 +11783,9 @@ EPUB/images/cover.png</pre>
 				<h3>Substantive changes since <a href="https://www.w3.org/publishing/epub32/">EPUB 3.2</a></h3>
 
 				<ul>
+					<li>11-Oct-2023: Additional text on ICB setting when dimensions are set multiple times.
+						See <a href="https://github.com/w3c/epub-specs/issues/2442">issue 2442</a>.
+					</li>
 					<li>12-Apr-2022: Added note about complexities of escaping from nested escapable structures and
 						updated the list of semantics to use for escaping to match the guidance. See <a
 							href="https://github.com/w3c/epub-specs/issues/2221">issue 2221</a>.</li>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2669,6 +2669,9 @@ alert("Feature " + feature + " supported?: " + conformTest);</pre>
 						Recommendation</a></h3>
 
 				<ul>
+					<li>11-Oct-2023: Additional text sketching the RS behavior in the case of erroneous ICB setting.
+						See <a href="https://github.com/w3c/epub-specs/issues/2442">issue 2442</a>.
+					</li>
 					<li>23-Sep-2023: Added a note on a possible future feature, whereby the value of
 							<code>rendition:flow</code> would control the publication of webtoon-like publications. See
 							<a href="https://github.com/w3c/epub-specs/issues/2412">issue 2412</a>. </li>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1611,17 +1611,19 @@
 								"500px"), the number prefix SHOULD be used as the pixel value, otherwise the value MUST
 								be treated as invalid.</p>
 
-							<p>
+							<p id="confreq-fxl-rs-xhtml-icb_invalid_meta"
+							   data-tests="#lay-fxl-xhtml-icb_invalid_meta">
 								Reading systems SHOULD attempt to extract <code>width</code> and <code>height</code> values from a <code>viewport meta</code> tag even if its syntax is invalid.
 							</p>
 
-							<p>
+							<p id="confreq-fxl-rs-xhtml-icb_repeated-in-meta"
+							   data-tests="#lay-fxl-xhtml-icb_repeated-in-meta">
 								Reading systems MUST use the first declaration of a <code>width</code> and <code>height</code> property in a <code>viewport meta</code> tag (i.e., ignore repeated declarations).
 							</p>
 							
 							<p>
 								If the <code>viewport meta</code> does not contain a width or a height value, or if these
-								values are invalid, reading systems MAY supply the values. For example, a reading system might:
+								values are invalid, reading systems MAY supply the values. For example, a reading system may:
 							</p>
 
 							<ul>
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1604,23 +1604,46 @@
 									data-tests="#lay-fxl-xhtml-icb,#lay-fxl-xhtml-icb_multi">They MUST clip content
 									positioned outside of the ICB.</span>
 							</p>
+
 							<p id="confreq-fxl-rs-xhtml-icb-units" data-tests="#lay-fxl-xhtml-icb_units">If the width or
 								height values in the <code>viewport meta</code> tag contain a non-numeric character but
 								start with a number (e.g., the value includes a unit of length declaration such as
 								"500px"), the number prefix SHOULD be used as the pixel value, otherwise the value MUST
 								be treated as invalid.</p>
-							<p>If the <code>viewport meta</code> does not contain a width or a height value, or if these
-								values are invalid, reading systems SHOULD consider these as having the values
-									<code>device-width</code> and <code>device-height</code>, respectively.</p>
+
+							<p>
+								If the <code>viewport meta</code> tag is invalid, the reading system MAY attempt to
+								extract valid <code>width</code> and <code>height</code> values and, if successful, use those to establish
+								the initial containing block.
+							</p>
+							
+							<p>
+								If the <code>viewport meta</code> does not contain a width or a height value, or if these
+								values are invalid, reading systems MAY choose to:
+							</p>
+
+							<ul>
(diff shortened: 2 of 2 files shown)


rickj (approved):

epub33/rs/index.html
```diff
@@ -1604,23 +1604,46 @@ <h4>Initial containing block dimensions</h4>
...
 								"500px"), the number prefix SHOULD be used as the pixel value, otherwise the value MUST
 								be treated as invalid.</p>
-							<p>If the <code>viewport meta</code> does not contain a width or a height value, or if these
-								values are invalid, reading systems SHOULD consider these as having the values
-									<code>device-width</code> and <code>device-height</code>, respectively.</p>
+
+							<p>
+								If the <code>viewport meta</code> tag is invalid, the reading system MAY attempt to
+								extract valid <code>width</code> and <code>height</code> values and, if successful, use those to establish
+								the initial containing block.
```
mattgarrish: Can we retain this as a should? For example:

> Reading systems SHOULD attempt to extract `width` and `height` values from a `viewport meta` tag even if its syntax is invalid.

epub33/rs/index.html
```diff
@@ -1604,23 +1604,46 @@ <h4>Initial containing block dimensions</h4>
...
+							</p>
+							
+							<p>
+								If the <code>viewport meta</code> does not contain a width or a height value, or if these
+								values are invalid, reading systems MAY choose to:
```
mattgarrish: Do we want to place the MAY here on the list of suggestions, which then makes them the normative list of what you're allowed to do?

```suggestion
								values are invalid, reading systems MAY supply the values. For example, a reading system might:
```

epub33/rs/index.html
```diff
@@ -1604,23 +1604,46 @@ <h4>Initial containing block dimensions</h4>
...
 								document order to obtain the height and width dimensions. Subsequent declarations MUST
 								be ignored.</p>
+
 							<p id="confreq-fxl-rs-xhtml-icb-ratio">When the ICB aspect ratio does not match the aspect
 								ratio of the reading system [=content display area=], reading systems MAY position the
 								ICB inside the area to accommodate the user interface; in other words, added
 								letter-boxing space MAY appear on either side (or both) of the content.</p>
+
+							<p>
+								In all cases when the <code>viewport meta</code> tag is invalid, the reading system MAY issue
```
mattgarrish: Isn't this true for all errors? Feels like it could be stated somewhere more general for error handling.

epub33/core/index.html
```diff
@@ -10732,6 +10732,12 @@ <h3>Syntax</h3>
 					contain <a href="#viewport.ebnf.sep-char">separator characters</a> or the <a
 						href="#viewport.ebnf.assign">assignment character</a>.</p>
 
+				<p>
+					If the <code>height</code> or the <code>width</code> assignments have multiple occurences whithin an otherwise
+					correct value of the [^meta/content^] attribute, only the first occurence of the <code>height</code>, respectively
+					<code>width</code>, definition counts, and all others MUST be ignored.
```
mattgarrish: Isn't this statement in the wrong spec? Sounds like the author is not allowed to delete the duplicates? 😉 

At most you might note that reading systems will only process the first value, but it seems kind of pointless to let authors know how invalid syntax is handled. The publication is gone and broken at that point.

mattgarrish: To finish the thought, I'd expected this to be a reading system requirement like:

> Reading systems MUST use the first declaration of a `width` and `height` property in a `viewport meta` tag (i.e., ignore repeated declarations).

epub33/rs/index.html
iherman: Personally, I would be o.k. with this, nay, I would prefer this. The question is whether RS implementers would be fine with this, so we should get some specific feedbacks for this.

epub33/rs/index.html
iherman: Yes, this is my last remark in the issue description. There are a number of other places in the spec that refer to errors explicitly, although all are referring to package errors. There seems to be no mention of a generic error handling, though, for example, what happens if the XHTML is invalid?

Maybe, at this point, the best would be to raise a separate issue/PR for error handling statement and then remove this from the current PR.

epub33/core/index.html
iherman: The question is: is it syntactically ok to have multiple `height` clause in the viewport setting? At this moment, per the BNF, it actually is. The reason I put the statement where it is now is to make this explicit.

Another possibility is to make a statement in the core spec whereby there _SHOULD NOT_ be repeated values in the viewport specification, and expand the RS spec with the statement you propose. That may be a better choice indeed.

epub33/rs/index.html
iherman: I will change this text accordingly. We will see if the SHOULD is o.k. or we should change it to MAY, but the language is better in any case.

epub33/core/index.html
iherman: I will change the text in both specs along these lines.

epub33/rs/index.html
wareid: Something more general on error handling is likely a good idea. I know in practice, when RS run into XHTML or parsing errors, we just throw the usual browser error page.

wareid (approved):

epub33/core/index.html
mattgarrish: >Another possibility is to make a statement in the core spec whereby there SHOULD NOT be repeated values

I'm fine with a statement that says properties MUST NOT be repeated for authoring. There's no valid need for allowing repetition that I can think of. RSes need to account for this possibility, but it shouldn't be an allowed authoring pattern.

iherman: CC-ing to @rdeltour and @wareid. If we put a MUST NOT, this becomes an epubcheck issue and, theoretically, a backward compatibility issue (that is why I shied away from a MUST NOT, actually). Is it o.k. to put a MUST NOT? Do we think that we would invalidate content out there? Do we care because that would be invalid content already?

danielweck: From a reading system perspective (more specifically, Thorium / Readium), we use ad-hoc / laser-focused regular expressions to parse the `content` attribute value, based on "common sense" syntactical expectations (i.e. unique `width` and `height` key/value pairs). I have personally never encountered problematic multiple `width` or `height` declarations. I haven't formally tested the edge case either, but I suspect that my parsing code would break somewhere :)

danielweck: UPDATE: our parsing code works with multiple `width` / `height` key/value pairs, we just pick the first regular expression match. Phew! ;)

iherman: Indeed, I have prepared some tests yesterday (https://github.com/w3c/epub-tests/pull/235) and I ran it through Thorium and the test worked 😀

epub33/rs/index.html
iherman: Considering #2454 and #2455 I will pull this paragraph from this PR.

epub33/core/index.html
wareid: I'll have to ask around about what we're doing but I don't doubt it's entirely possible it's something similar to what @danielweck has described, since we favour getting content to work whenever possible. I think a MUST NOT will be good for publishers though, to catch any accidents in the production process. Even with our internal EPUBCheck scans, not every error is disqualifying.

iherman: At the moment, epubcheck does not catch a duplicate w/h setting. If we put a _MUST NOT_, it eventually will. Won't this violate our charter restriction? The [Safari guide](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html#//apple_ref/doc/uid/TP40008193-SW6) that was used in previous incarnations of EPUB, is also silent on the issue.

At this moment, to stay in line with the charter, I think I would prefer to keep it a _SHOULD NOT_. To my spite, to be honest, but violation of the charter could not be justified by some possible security problems, which was the case with some other cases.

davemanhall (approved):

epub33/core/index.html
rdeltour: Whatever we decide can be implemented in EPUBCheck. In fact I changed the way the viewport is parsed, using dedicated parsing code instead of the naïve string matching used in v4.x.

@iherman wrt the concern about backward compatibility:
- a SHOULD will make EPUBCheck raise warnings which we know often means rejection by distributors.
- EPUBCheck v5 will still be backward incompatible due to the EBNF which is more restrictive to what we had before (EPUBCheck did not complain about `"anything, width=100, height=100` but will raise a syntax error in v5.0.0).

mattgarrish: I don't see how putting a "should not" in helps. That suggests there are valid cases for duplicating the dimensions.

We're not supposed to invalidate valid content, but in this case all we're doing is better defining what was always the only valid representation. That's not breaking our charter goal. I can't imagine someone would be upset that they're now being alerted to a potential incompatibility across reading systems.

I don't care if the ebnf allows properties without values, if that's really a concern. If it's to capture a blown height/width declaration, that's going to happen anyway. If it's to allow other garbage through, so be it.

iherman: Ok. I changed it to _MUST NOT_.

iherman:
@mattgarrish @wareid  I guess this could be merged now, right?

davemanhall (approved):

PR #2455, opened 2022-10-13
Add error handling and reporting subsections

### LINKED ISSUE #2454: Add error handling and reporting sections
https://github.com/w3c/epub-specs/pull/2449#issuecomment-1269809452 and https://github.com/w3c/epub-specs/pull/2453#discussion_r992543790 raise the question of whether we should include some information about error handling and reporting.

The most logical place for these would probably be as subsections of "2. Reading system conformance".

The error handling section could use the text proposed in the referenced discussion:

> This specification does not require reading systems to load EPUB publications, or resources within them, when they violate content authoring or processing requirements (e.g., they define entities in an external DTD subset).

If we add an error reporting section, should we say that reading systems "SHOULD" provide a way for users to access any issues the RS encountered loading a publication? I know not every reading systems will implement an EPUB version of the browser dev tools, or will report every issue, but it would certainly make life easier for users if they did. Saying that they may provide reporting doesn't really move the needle much.

mattgarrish:
I've made the error reporting section informative since we're not recommending or requiring it.

Fixes #2454 

* Reading Systems:
    * [Preview](https://raw.githack.com/w3c/epub-specs/fix/issue-2454/epub33/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fix/issue-2454/epub33/rs/index.html)

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -221,7 +221,7 @@
 			</section>
 		</section>
 		<section id="sec-rs-conformance">
-			<h3>Reading system conformance</h3>
+			<h2>Reading system conformance</h2>
 
 			<p>Whether a [=reading system=] has to support a feature is mentioned at the beginning of its section. To be
 				conformant with this specification, reading systems MUST support all required features as well as all
@@ -242,6 +242,24 @@
 					document=] metadata). Reading systems may use this additional information for any purposes (e.g., to
 					improve the user interface).</p>
 			</div>
+
+			<section id="rs-conf-error-handling">
+				<h3>Error handling</h3>
+
+				<p>Reading systems are not required to load [=EPUB publications=], or resources within them, when they
+					violate content authoring or processing requirements.</p>
+			</section>
+
+			<section id="rs-conf-error-reporting" class="informative">
+				<h3>Error reporting</h3>
+
+				<p>Although reading systems are not required to report errors encountered while processing and rendering
+					[=EPUB publications=] (e.g., if the dimensions of a [=fixed-layout document=] have been inferred),
+					they are strongly encouraged to provide users a means of accessing this information.</p>
+
+				<p>Allowing access to processing errors can help users determine whether problems with their EPUB
+					publications should be reported to the reading system developer or to the EPUB creators.</p>
+			</section>
 		</section>
 		<section id="sec-pub-resources">
 			<h3>Publication resource processing</h3>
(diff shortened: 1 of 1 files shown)


iherman (approved):
Just food for thought, and I can go either way. 

During the discussion on the call, one argument against error reporting was that the final end-user (ie, the reader) does not really want to know about error handling, it will only mess her up. On the other hand, such errors may indeed be very important for EPUB creators. Hence the idea of providing some extra flags, a bit like the developers' flag in a browser, to see those error messages. Ie, a browser-like behaviour.

So the question is whether we should go into somewhat more suggested details in the section on error reporting, eg., by referring to the error reporting strategies of browsers that could be used as an example for reading systems. But maybe what is there is enough; not sure, I trust your decision, @mattgarrish .

wareid (approved):
I wonder, since this section is informative, would this be an appropriate place to mention errors in EPUBCheck. I don't know how I'd word it, but something about validating content before it reaches the user?

mattgarrish:
I've done some rewriting of the section to focus it more on debugging. I brought in web dev tools and mention providing access to issues from, for example, the HTML, CSS and JS rendering, but I'm not sure if we want to get any deeper into it than that. I also mention epubcheck as a source of validation info.

If you have any more ideas on what to say here, though, feel free to offer them up!

mattgarrish:
I also added a subsection for all the text that precedes the error sections (called "requirements" for lack of a better name). It looked strange in the table to have a conformance section with only two error reporting subsections (the stray paragraph problem that means you can't cite those requirements without also referring to all of section 2).

iherman:
> I also added a subsection for all the text that precedes the error sections (called "requirements" for lack of a better name). It looked strange in the table to have a conformance section with only two error reporting subsections (the stray paragraph problem that means you can't cite those requirements without also referring to all of section 2).

I like it as is. As far as I am concerned, it is a go!

Thanks @mattgarrish

davemanhall (approved):
This feels like a great addition and will hopefully improve the development environment for authors.  👍

wareid (approved):

iherman:
@mattgarrish @wareid I think this can be merged now.

PR #2457, opened 2022-10-15
Allow properties without values in viewport meta tags

mattgarrish:
@rdeltour would this solve your concerns about not blocking EPUB publications with (potentially) invalid property definitions because epubcheck will complain?

Really, what other data is in the tag isn't our concern since it isn't even supposed to be used, so by making the assignment operator and value optional, that should allow any weird combinations through (or future valid declarations that don't require a value, if such things ever come to pass).

Our requirement for a valid height and width declaration will still capture what we need even with this change.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2457.html" title="Last updated on Oct 17, 2022, 12:34 PM UTC (131a4c3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2457/032fcce...131a4c3.html" title="Last updated on Oct 17, 2022, 12:34 PM UTC (131a4c3)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7575,6 +7575,9 @@ No Entry</pre>
 								</ul>
 								<p>The <code>device-width</code> and <code>device-height</code> values refer to the 100%
 									of the width and height, respectively, of the reading system's [=viewport=].</p>
+								<p>EPUB creators MUST NOT specify duplicate <code>height</code> or <code>width</code>
+									definitions either within a single <code>viewport meta</code> tag or by specifying
+									multiple <code>viewport meta</code> tags.</p>
 								<aside class="example"
 									title="Specifying the initial containing block in a viewport meta tag">
 									<pre>&lt;html …>
@@ -10732,9 +10735,6 @@ html.my-document-playing * {
 						href="#viewport.ebnf.sep-char">separator characters</a> or the <a href="#viewport.ebnf.assign"
 						>assignment character</a>.</p>
 
-				<p>The [^meta/content^] attribute MUST NOT include duplicate property assignments (e.g., two or more
-          <code>height</code> or <code>width</code> definitions).</p>
-
 				<p>The authoring requirements in this section apply <em>after</em>
 					<a data-cite="xml#AVNormalize">whitespace normalization</a> [[xml]] (i.e., after a reading system
 					strips leading and trailing whitespace and compacts all instances of multiple whitespace within the
@@ -11783,9 +11783,8 @@ EPUB/images/cover.png</pre>
 				<h3>Substantive changes since <a href="https://www.w3.org/publishing/epub32/">EPUB 3.2</a></h3>
 
 				<ul>
-					<li>11-Oct-2023: Additional text on ICB setting when dimensions are set multiple times.
-						See <a href="https://github.com/w3c/epub-specs/issues/2442">issue 2442</a>.
-					</li>
+					<li>11-Oct-2023: Additional text on ICB setting when dimensions are set multiple times. See <a
+							href="https://github.com/w3c/epub-specs/issues/2442">issue 2442</a>. </li>
 					<li>12-Apr-2022: Added note about complexities of escaping from nested escapable structures and
 						updated the list of semantics to use for escaping to match the guidance. See <a
 							href="https://github.com/w3c/epub-specs/issues/2221">issue 2221</a>.</li>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -10674,8 +10674,8 @@ html.my-document-playing * {
 									<a href="#viewport.ebnf.property">property</a>
 								</td>
 								<td>=</td>
-								<td><a href="viewport.ebnf.name">name</a>, <a href="viewport.ebnf.assign">assign</a>, <a
-										href="viewport.ebnf.value">value</a> ;</td>
+								<td><a href="viewport.ebnf.name">name</a>, [ <a href="viewport.ebnf.assign">assign</a>,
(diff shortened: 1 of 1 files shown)


iherman (approved):
What you propose is fine and spec-wise correct. But I wonder whether in the following paragraph would not make it a bit more readable and more explicit to the casual reader by saying something like

> Only the  <code> height</code> and <code>width</code> properties are required, by this specification, to have a value assignment,  with the corresponding value requirement specified in <a href="#sec-fxl-content-dimensions"></a>. It imposes…

But I am fine if you prefer to leave it as is.

mattgarrish:
Ya, I was thinking of putting a note in saying it allows invalid properties, but didn't like saying it out loud. Reinforcing we have name/value requirements for height/width isn't a bad thing

Looking over the full section, though, I found some further inconsistencies to clean up:

- I think it's better to say right at the beginning that this syntax is only for FXL.
- I fixed a few hrefs in the ebnf that were missing hashes at the start.
- I added space around all the ebnf special characters, as sometimes we had whitespace and sometimes not.
- I took the text about whitespace handling out of a note, as it is saying normatively when to validate the syntax.
- I changed the vague reference to HTML whitespace to a specific reference to ASCII whitespace in the infra spec (since that's what HTML references).

Finally, I put the text about the required properties and values into the note that discourages other uses, plus added some text that we don't even require values for other properties. That leaves all the processing normative, rather than interspersing some text about why we have this syntax. It's not as immediate as saying why we have the syntax right after the ebnf, but feels a little clearer for reading (we also mention it's for fxl height and width in the intro, so it's plenty covered now).

iherman:
Looks fine to me, and there is a clearer separation between the Syntax of the tag and the definition of what w/h are for. 

One minor thing. In the meantime, I merged the PR on viewport setting, which included this:

> The [content](https://html.spec.whatwg.org/multipage/semantics.html#attr-meta-content) attribute MUST NOT include several height, respectively width, assignments.

I presume that should still stay and github's merge magic will take care of that...

mattgarrish:
> I presume that should still stay and github's merge magic will take care of that...

Ya, it's still there, but does it make sense to have that specific a restriction in the general syntax definition? If we want to make it exclusive to height and width, then shouldn't that be part of section on setting the FXL dimensions?

For now, I've tweaked the paragraph to say that all duplicate definitions are illegal and put width and height into a parenthetical.

iherman:
> > I presume that should still stay and github's merge magic will take care of that...
> 
> Ya, it's still there, but does it make sense to have that specific a restriction in the general syntax definition? If we want to make it exclusive to height and width, then shouldn't that be part of section on setting the FXL dimensions?
> 
> For now, I've tweaked the paragraph to say that all duplicate definitions are illegal and put width and height into a parenthetical.

As we do not put any restrictions on other parts of the definitions, making any statement (even duplication) on a term other than w/h may not be appropriate. And you are right that, in this new setting, the statement should probably go to the FXL definition.

mattgarrish:
> making any statement (even duplication) on a term other than w/h may not be appropriate

Right, and I woke up this morning thinking that we should also be saying something about not including duplicate definitions by having multiple viewport tags, since reading systems are supposed to ignore subsequent ones. I've moved the restriction and also added that clarification into it.

iherman:
It looks good to me now. B.t.w., the RS spec already says, in the section on ICB §8.1.2. how to handle multiple viewport declarations, which is good; all loose ends are now tied.

It's a go for me.

epub33/core/index.html
```diff
@@ -7575,6 +7575,9 @@ <h4>Content document dimensions</h4>
 								</ul>
 								<p>The <code>device-width</code> and <code>device-height</code> values refer to the 100%
 									of the width and height, respectively, of the reading system's [=viewport=].</p>
+								<p>EPUB creators MUST NOT specify duplicate <code>height</code> or <code>width</code>
+									definitions either within a single <code>viewport meta</code> tag or by specifying
+									multiple <code>viewport meta</code> tags.</p>
```
rdeltour, line 7580: This creates a backward incompatibility.
Not that I'm objecting, just wanted it to be clear 😊

rdeltour (commented):
With my EPUBCheck hat on, this PR clarifies the duplication issue, which is good. Note that it does introduce a backward incompatibility, since we did allow that before.

Outside of EPUBCheck's consideration, my own stance, as you know, is that ideally this would be better handled algorithmically, which would allow some well-defined error handling and fully specify the RS interpretation.  
That said, given the complexity of coming up with an algo and the lateness of the issue, I understand it's not optimal here 😊.

A careful reader might wonder why the EBNF explicitly allows some invalid forms (value-less keys) while forbidding others (e.g. leading or trailing separators). But practically I don't think it will happen often with real-world content, so whatever the EBNF allows/restricts is probably ok!

epub33/core/index.html
mattgarrish: Ya, but this falls into ensuring invalid/incompatible content is better captured.

We want to avoid invalidating content that is actually valid.

mattgarrish:
> A careful reader might wonder why the EBNF explicitly allows some invalid forms (value-less keys) while forbidding others (e.g. leading or trailing separators). But practically I don't think it will happen often with real-world content, so whatever the EBNF allows/restricts is probably ok!

Right, I don't think anyone is going to care given that the viewport meta tag has only one purpose in fixed layouts. We could add a note clarifying that the syntax is meant to be as loose as possible to ensure that (invalid) properties we don't care about don't invalidate the publication, but this is all a technical detail related to W3C process. I doubt authors will notice we even have our own syntax, let alone try to decipher the EBNF! They'll just copy the examples.

PR #2459, opened 2022-10-17
Add recommendation against using spaces in file paths and names

### LINKED ISSUE #2458: Spaces in file names
While going through epubcheck's tests, I noticed that it throws an error if file names have spaces in them. We previously discussed whether these should be allowed in https://github.com/w3c/epub-specs/issues/808

But in this revision we've changed to referencing the URL standard, so should we revist this issue again? Otherwise, it's strange for epubcheck to ban them when that specification doesn't.

I'm not suggesting we have to allow spaces now, as suddenly allowing spaces may expose interoperability issues hidden by the previous ban, but perhaps we should add a recommendation/restriction against their use to the OCF file and path naming section? It would help to have epubcheck aligned with spec requirements so it's not seen as adding arbitrary restrictions.

mattgarrish:
Fixes #2458

Also makes some minor editorial tweaks:
- all bullets are capitalized for consistentcy
- s/file names and paths/file paths and names/ in one bullet to be consistent with section name and ordering in another bullet

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2459.html" title="Last updated on Oct 17, 2022, 7:07 PM UTC (b7b6a17)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2459/f101c88...b7b6a17.html" title="Last updated on Oct 17, 2022, 7:07 PM UTC (b7b6a17)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1761,7 +1761,7 @@
 							</ul>
 						</li>
 						<li>
-							<p id="ocf-fn-space">For compatibility with older reading systems, file names SHOULD NOT
+							<p id="ocf-fn-space">For compatibility with older [=reading systems=], file names SHOULD NOT
 								contain SPACE (U+0020) characters.</p>
 						</li>
 						<li>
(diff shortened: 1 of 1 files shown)


iherman (approved):

rdeltour (approved):
LGTM.

Although it is not related to this PR, I just noticed while reviewing that I do not understand the following paragraph in the final note:

> If EPUB creators need to ensure compatibility with EPUB 2 [reading systems](https://www.w3.org/TR/epub-33/#dfn-epub-reading-system) that only accept URIs [[rfc3986](https://www.w3.org/TR/epub-33/#bib-rfc3986)], they should further consider restricting resource names to the ASCII character set [[us-ascii](https://www.w3.org/TR/epub-33/#bib-us-ascii)].

What is an RS that only accept URIs? How is that related to restricting to ASCII only? (given non-ASCII characters can always be percent-encoded).

(I'm not opening an issue since this is only a note and it may be a clear to others).

mattgarrish:
> What is an RS that only accept URIs? How is that related to restricting to ASCII only?

Ya, I suggested throwing it out when we updated the text above it in https://github.com/w3c/epub-specs/pull/1935. I'm still good with that.

Life is simpler when only URIs are supported by using ASCII, but as you say it's not critical.

The primary issue, I believe, has always been zip tool handling (plus any other program under the hood that might not handle unicode in file names).

Issue #2461, opened 2022-10-17
Encoding in the abstract container


rdeltour:
It's mostly a detail, but something struck me as maybe out of place in the file paths and names restrictions.

The spec says:
> file names and paths MUST be UTF-8 [unicode] encoded

What does it mean to talk about encoding in the abstract container? Encoding is a concrete thing that happens when actually storing the info.

The ZIP section already requires:
> OCF ZIP containers MUST encode File System Names using UTF-8 [unicode].

rdeltour:
I'm suggesting maybe referring to the Infra string type (or scalar value string?) in the definition of file path and file name.

mattgarrish:
> OCF ZIP containers MUST encode File System Names

Is "file system names" in caps for a bit of imperiousness?

It seems to have been that way forever for no particular reason. The zip appnote makes no reference to it as a term.

mattgarrish:
> maybe referring to the Infra string type (or scalar value string?) in the definition of file path and file name

What if the first sentence of the section incorporated it:

> In the context of the OCF abstract container, file paths and file names are case sensitive scalar value strings [infra].

rdeltour:
> What if the first sentence of the section incorporated it:

>> In the context of the OCF abstract container, file paths and file names are case sensitive scalar value strings [infra].

Looks better to me!
To nitpick even more: case sensitiveness is implied (upper/lower case characters are two different code points), it could be moved to a note.

mattgarrish:
> case sensitiveness is implied (upper/lower case characters are two different code points)

Sure, if all we're considering is code points, but isn't the issue here the virtual file system of the abstract container? If we remove this as a normative statement, doesn't that mean they can be handled case insensitively and it doesn't matter?

PR #2466, opened 2022-10-20
Remove utf-8 encoding requirement for abstract container file names

mattgarrish:
I've only removed the encoding bullet and lowercased "File System Names" per #2461 with this pull request, as those are reasonably straightforward.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2466.html" title="Last updated on Oct 20, 2022, 1:24 PM UTC (be81cdd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2466/ea362bb...be81cdd.html" title="Last updated on Oct 20, 2022, 1:24 PM UTC (be81cdd)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1699,9 +1699,6 @@
 
 					<ul class="conformance-list">
 						<li>
-							<p id="ocf-fn-encoding">File paths and names MUST be UTF-8 [[unicode]] encoded.</p>
-						</li>
-						<li>
 							<p id="ocf-fn-length">File names MUST NOT exceed 255 bytes.</p>
 						</li>
 						<li>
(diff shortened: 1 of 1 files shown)


iherman (approved):
(Tiny bit: this PR should close #2461, better add it as a fix, or mark it explicitly in the 'development' list on the right.)

mattgarrish:
> this PR should close #2461

Well, there's also the request that we treat file names and paths as strings and remove the case sensitivity requirement. I'm not sure those changes fit with describing a virtual file system. If we opt to leave them alone, then, yes, this fixes the issue. Otherwise, we still have to figure out another part before closing.

iherman:
> remove the case sensitivity requiremen

It may be an overspecification to mention that, but I think it is not bad to leave it there to emphasize for casual readers. It does not harm, and may be helpful to some. I would leave it as is.

PR #2470, opened 2022-10-27
Spine vocabulary improvement on the usage of page-spread-*

### LINKED ISSUE #2468: Spine properties vocabulary should mention mutual exclusivity of page spread properties
`itemref` + D.7 state that the default vocabulary for `itemref` elements consists of `page-spread-left` and `page-spread-right`. These are probably meant to be mutually exclusive, and I therefore propose the wording for these to be clarified, adding something like (for the page-spread-left property):

> This property is mutually exclusive with the `page-spread-right` property. EPUB creators MUST NOT include the `page-spread-right` property on elements carrying the `page-spread-left` property.

and vice versa on the page-spread-right property.

iherman:
This is to fix #2468.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2470.html" title="Last updated on Nov 1, 2022, 11:27 AM UTC (67fa715)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2470/3340110...67fa715.html" title="Last updated on Nov 1, 2022, 11:27 AM UTC (67fa715)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -7445,8 +7445,8 @@ No Entry</pre>
 							and omit the properties on spine items where one-up or two-up presentation is equally
 							acceptable.</p>
 
-						<p>EPUB creators MUST NOT declare more than one <code>page-spread-*</code> property on any given
-							spine item.</p>
+						<p>EPUB creators MUST NOT declare more than one <code>rendition:page-spread-*</code> property,
+							or their unprefixed equivalents, on any given spine item.</p>
 
 						<div class="note" id="note-page-spread-aliases">
 							<p>The <code>rendition:page-spread-left</code> and <code>rendition:page-spread-right</code>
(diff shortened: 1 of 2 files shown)


mattgarrish:
This looks pretty confusing in the rendered spec. It comes out as a paragraph of the page-spread-right property as we can't really have floating paras between sections.

Since all these kinds of property restrictions are defined in their respective FXL sections, not the vocabulary, I'd prefer to incorporate the statement with the existing restriction. As it stands, we have the same statement repeated twice, instead of one mentioning the prefixed and one the unprefixed versions (but we only need one statement to cover all).

I think it'd be better to incorporate this into the existing requirement, perhaps like this:

> <p>EPUB creators MUST NOT declare more than one <code>rendition:page-spread-*</code> property, or their unprefixed equivalents, on any given spine item.</p>

I'd also favour linking the page-spread-* vocabulary entries to 8.2.2.4, as requested in the issue, and making clear that they have newer aliases. Putting aside the restriction, it's not easy to work your way back from the vocabulary to how to use these properties.

I'll propose some edits in another PR.

mattgarrish (approved):

PR #2471, opened 2022-10-27
Changed the alternate-script definition on language tags.

### LINKED ISSUE #2465: Add recommendation or requirement about the existence of xml:lang on alternate-script meta-element
An `alternate-script` meta element without an `xml:lang` attribute is not very useful and inheriting it from the `metadata` or `package` element should IMO be considered an anti-pattern (the refined metadata (`dc:title`, `dc:creator`, etc) overrides the base `xml:lang` only to be refined by the base `xml:lang` again).

A recommendation would look like:

When specifying an `alternate-script` property, creators SHOULD include an `xml:lang` attribute identifying the language and script of the associated property value.

iherman:
This is to fix #2465, adopting the proposed text in https://github.com/w3c/epub-specs/issues/2465#issuecomment-1284257358.

Takes also care of the remark of @r12a in https://github.com/w3c/epub-specs/issues/2465#issuecomment-1291769787.

### DIFF
--- a/epub33/core/vocab/meta-property.html
+++ b/epub33/core/vocab/meta-property.html
@@ -27,8 +27,9 @@
 					<th>Description:</th>
 					<td>
 						<p>The <code>alternate-script</code> property provides an alternate expression of the
-							associated property value in a language and script identified by the
-								<code>xml:lang</code> attribute.</p>
+							associated property value in a different language and/or script. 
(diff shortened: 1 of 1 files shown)


iherman:
cc @Jesse-Baker

mattgarrish (approved):

PR #2479, opened 2022-11-11
Updates to nav processing

### LINKED ISSUE #2477: Example of non-textual descendant concatenated label in nav elements
[Section 7.3 The nav element: restrictions](https://www.w3.org/TR/epub-33/#sec-nav-def-model) states:

> The child a or span element MUST provide a non-zero-length text label after concatenation of all child content and application of white space normalization rules. When determining compliance with this requirement, the concatenated label MUST include text content contained in title or alt attributes for non-textual descendant elements.

> If an a or span element contains instances of [HTML embedded content](https://html.spec.whatwg.org/multipage/dom.html#embedded-content-category) [[html](https://www.w3.org/TR/epub-33/#bib-html)] that do not provide intrinsic text alternatives, the element MUST also contain a title attribute with an alternate text rendering of the link label.

These rules could use an example to clarify what constitutes a valid text label in the presence of nested child elements and embedded content, like

```xml
<nav epub:type="…">
   <h1>…</h1>
   <ol>
      <li>
         <a href="chap1.xhtml">
            <audio src="..."/>
            <audio src="..." title="Chapter 1"/>
         </a>
      </li>
      <li>
         <span>
             <img src="..." alt="Chapter 2"/>
         </span>
         <ol>...</ol>
      </li>
   </ol>
</nav>
```

mattgarrish:
This pull request fixes #2477 as follows:

- adds a requirement that reading systems use the alternative text for non-text elements that are not supported in a non-html rendering of a nav component. Also adds precedence rule for alt over title and lets RS do what it wants when neither are specified.
- adds a caution to the authoring spec that reading systems that build their own navigation widgets may not be able to retain element- and attribute-based rendering instructions, so authors need to be aware of possible usability issues
- adds an image with alt text to the nav patterns example

I don't expect that requiring the use of the alt text will be controversial, as when it's not used you generally get a really bad user experience. Open to suggestions on the authoring caution, though. I don't want to tell people not to use html formatting, as it could be fine in a browser-based reading system, or when used in spine, so I tried to tie it into being a usability issue.

Reading Systems:
* [Preview](https://raw.githack.com/w3c/epub-specs/fix/issue-2477/epub33/rs/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fix/issue-2477/epub33/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2479.html" title="Last updated on Nov 16, 2022, 12:47 PM UTC (da8e5cc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2479/5894e8d...da8e5cc.html" title="Last updated on Nov 16, 2022, 12:47 PM UTC (da8e5cc)">Diff</a>

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1420,6 +1420,14 @@
 							data-cite="epub-33#sec-nav-toc"><code>toc nav</code> element</a> [[epub-33]].</p>
 				</li>
 				<li>
+					<p id="confreq-nav-alt-text">MUST, when generating non-HTML based navigation widgets, replace
+						unsupported non-text elements in headings and labels with their <a
+							data-cite="epub-33#confreq-nav-a-cnt">alternative text</a> [[epub-33]]. If both
+							<code>alt</code> and <code>title</code> attributes are present on an element, preference
+						SHOULD be given to the <code>alt</code> attribute. If neither attribute is present, reading
+						systems MAY provide their own text or ignore the element.</p>
+				</li>
+				<li>
 					<p id="confreq-nav-pagelist-access">SHOULD provide a method to navigate to the listed page breaks
 						when a <a data-cite="epub-33#sec-nav-pagelist"><code>page-list nav</code> element</a>
 						[[epub-33]] is present.</p>
@@ -2702,6 +2710,9 @@ alert("Feature " + feature + " supported?: " + conformTest);</pre>
 						Recommendation</a></h3>
 
 				<ul>
+					<li>11-Nov-2022: Added requirement to replace unsupported non-text elements with their text
+						alternatives when generating non-HTML navigation widgets. See <a
+							href="https://github.com/w3c/epub-specs/issues/2477">issue 2477</a>.</li>
 					<li>13-Oct-2022: Added sections to explain expectations for error handling and reporting. See <a
 							href="https://github.com/w3c/epub-specs/issues/2454">issue 2454</a>.</li>
 					<li>11-Oct-2023: Additional text sketching the RS behavior in the case of erroneous ICB setting. See
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -1420,7 +1420,7 @@
 							data-cite="epub-33#sec-nav-toc"><code>toc nav</code> element</a> [[epub-33]].</p>
 				</li>
 				<li>
-					<p id="confreq-nav-alt-text">MUST, when generating non-HTML based navigation widgets, replace
+					<p id="confreq-nav-alt-text" data-tests="#nav-non-text_img,#nav-non-text_img_title">MUST, when generating non-HTML based navigation widgets, replace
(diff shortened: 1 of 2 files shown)


iherman (approved):
My only worry is that, by introducing a new _MUST_ statement in the RS spec, we must create a test asap (which is of course easy), but we will have to nudge implementers to run it. As we are in the danger not to find enough implementations implementing this MUST, it puts us into a strange spot. But we may have to do that anyway.

I have (again) some serious home networking issues, so I am not sure if I can create those tests easily now (although the tests are probably easy to do). Hopefully I will be able to do that by the middle of next week, unless someone beats me into it...

@wareid @dauwhe

mattgarrish:
I think I'd be fine with a "should" given the time constraints, but it's such a terrible user experience to have entries with no labels, or where you get the value of a src attribute. Unless you know of a reading system that will support images and multimedia and target only it, you're just asking for trouble to use the elements.

I'd probably favour banning embedded and interactive content, given there's no support for these anyway. That the fallback text also isn't supported only worsens the situation. There's just a risk that we could invalidate some content if we change the model, but in this case I'd have to think it'd be pretty small.

mattgarrish:
> given there's no support for these anyway

Found one major reading system that displays images, at least, so leaving the content model alone is probably wisest.

iherman:
> > given there's no support for these anyway
> 
> Found one major reading system that displays images, at least, so leaving the content model alone is probably wisest.

Great. Then my proposal is to merge the text as proposed, and I will come up with 2-3 tests for it.

epub33/rs/index.html
```diff
@@ -1419,6 +1419,14 @@ <h3>Navigation document processing</h3>
 							data-cite="epub-33#confreq-nav-a">links and headings</a> in the <a
 							data-cite="epub-33#sec-nav-toc"><code>toc nav</code> element</a> [[epub-33]].</p>
 				</li>
+				<li>
+					<p id="confreq-nav-alt-text">MUST, when generating non-HTML based navigation widgets, replace
```
```suggestion
					<p id="confreq-nav-alt-text" data-tests="#nav-non-text_img,#nav-non-text_img_title">MUST, when generating non-HTML based navigation widgets, replace
```

iherman (approved):

epub33/rs/index.html
iherman: The tests are in https://github.com/w3c/epub-tests/pull/251.

wareid (approved):

epub33/rs/index.html
iherman: The tests have been approved, I will merge this comment that is bound to the tests.

PR #2493, opened 2022-11-29
Clarify where epub:type is not allowed

### LINKED ISSUE #2486: `epub:type` restricted to palpable content may cause backward incompatibilities
In EPUB 3.3, [XHTML requirements](https://www.w3.org/TR/2022/CRD-epub-33-20221124/#sec-xhtml-structural-semantics) says the `epub:type` attribute MUST only be used on [palpable content](https://html.spec.whatwg.org/multipage/dom.html#palpable-content-2).

This list is restrictive and will likely cause a lot of existing material being rejected. Several HTML elements like `body` or `li` are not palpable content but are often associated an `epub:type` attribute (they're defined as relevant usage context in the [structural semantics vocabulary](https://www.w3.org/TR/epub-ssv-11/)).

Also, it means `epub:type` and `hidden` cannot co-exist, I'm not sure how that's relevant.

Was the idea was to not allow `epub:type` on some elements like `head` and `style`?

If yes, we could either:
- restrict `epub:type` to any element except [metadata content](https://html.spec.whatwg.org/multipage/dom.html#metadata-content) and the `html` element.
- restrict `epub:type` to [flow content](https://html.spec.whatwg.org/multipage/dom.html#flow-content) and the `body` element.

mattgarrish:
Restricts is from head element and metadata content to better match the RS restriction.

Fixes #2486

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2493.html" title="Last updated on Nov 29, 2022, 5:49 PM UTC (03519f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2493/27fbf42...03519f5.html" title="Last updated on Nov 29, 2022, 5:49 PM UTC (03519f5)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -5775,7 +5775,8 @@ No Entry</pre>
 						<p>[=EPUB creators=] MAY use the [^/epub:type^] attribute in [=XHTML content documents=] to
 							express <a href="#sec-structural-semantics-intro">structural semantics</a>.</p>
 
-						<p>The attribute MUST only be used on [=palpable content=] [[html]].</p>
+						<p>The attribute MUST NOT be used on the <a data-cite="html#the-head-element"
+								><code>head</code></a> element or [=metadata content=] [[html]].</p>
 					</section>
 
 					<section id="sec-xhtml-rdfa">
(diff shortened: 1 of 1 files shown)


rdeltour (approved):
Looks OK to me.

Note that this allows `epub:type` on the `html` element (not sure if that is intentional or if it should be disallowed there too).

mattgarrish:
> Note that this allows `epub:type` on the `html` element (not sure if that is intentional or if it should be disallowed there too).

Intentional. I'm hesitant to change where it's allowed. The restriction pre-3.3 did not forbid it on the root element.

iherman (approved):

PR #2494, opened 2022-11-29
Clarify data URLs are not publication resources

### LINKED ISSUE #2485: some confusion around data URLs
I find the spec slightly confusing around data URLs:

- the ["Data URLs" section](https://www.w3.org/TR/epub-33/#sec-data-urls) says resources in data URLs **are not** publication resources.
- the [_publication resource_ definition](https://www.w3.org/TR/epub-33/#dfn-publication-resource) says data URLs are an exception to the requirement to list publication resources in the manifest, which kinda implies resources in data URLs **are** publication resources.

Which is true?  
Can a data URL be listed in the manifest? (the manifest MUST only list publication resources)  
If yes, a navigation document that is not in the spine can be defined as a data URL… do we need more restrictions?

mattgarrish:
This one got me pretty twisted up trying to solve.

There's a second case I thought of where you could technically use a data URL in the manifest: fallback resources. You could encode a CMT image as a data URL for an img element fallback, for example. But do we really want to support this or having a nav doc not in the spine as a data URL? I'm guessing not.

Assuming that's the case, what I've done is try to make clearer that data URLs are not unique from their container resource. They must not be listed in the manifest, but because they have media types separate from their container they are still subject to fallback restrictions.

I've also assumed that we don't care what people do with linked resources, so data URLs are allowed in `link` elements in the metadata.

Anyway, take a good look at this PR as it's a bit of a convoluted problem.

Fixes #2485

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2494.html" title="Last updated on Dec 3, 2022, 2:34 PM UTC (ab0261a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2494/58879d0...ab0261a.html" title="Last updated on Dec 3, 2022, 2:34 PM UTC (ab0261a)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -540,9 +540,10 @@
 							render the EPUB publication as the [=EPUB creator=] intends. Examples of publication
 							resources include the [=package document=], [=EPUB content documents=], CSS Style Sheets,
 							audio, video, images, embedded fonts, and scripts.</p>
-						<p>EPUB creators list publication resources in the package document [=EPUB manifest | manifest=]
-							and typically bundle them all in the [=EPUB container=] (the exception being they may locate
-							resources listed in <a href="#sec-resource-locations"></a> outside the EPUB container).</p>
+						<p>EPUB creators must list publication resources in the package document [=EPUB manifest |
+							manifest=] and typically bundle them all in the [=EPUB container=] (the exception being they
+							may locate resources listed in <a href="#sec-resource-locations"></a> outside the EPUB
+							container).</p>
 						<div class="note">
 							<p>Resources on the web identified in outbound hyperlinks (e.g., referenced from the
 									<code>href</code> attribute of an [[html]] [^a^] element) are not publication
@@ -1463,9 +1464,9 @@
 					directly into a URL string. The advantage of this scheme is that it allows [=EPUB creators=] to
 					embed a resource within another, avoiding the need for an external file.</p>
 
-				<p>A consequence of embedding is that data URLs are not considered their own unique publication
-					resources (i.e., they are part of their containing publication resource). As a result, EPUB creators
-					MUST NOT declare data URLs in the manifest [^item^] element's <a href="#attrdef-item-href"
+				<p>A consequence of embedding is that data URLs are not considered their own unique [=publication
+					resources=] (i.e., they are part of their containing publication resource). As a result, EPUB
+					creators MUST NOT declare data URLs in the manifest [^item^] element's <a href="#attrdef-item-href"
 							><code>href</code></a> attribute.</p>
 
 				<p>EPUB creators MAY use data URLs anywhere else in [=EPUB publications=] provided their use does not
@@ -1488,14 +1489,14 @@
 
 				<p>Although data URLs are considered part of their containing resource, because they have their own
 					media types they are still subject to the <a href="#sec-foreign-resources">foreign resource
-						restrictions</a>. EPUB creators MUST therefore encode data URLs as core media type resources or
-					use them where they can provide a fallback.</p>
-
-				<p>These restrictions on the use of data URLs are to prevent security issues and also to ensure that
-					[=reading systems=] can determine where to take a user next (i.e., because data URLs cannot be
-					referenced from the [=EPUB spine | spine=]).</p>
+						restrictions</a>. EPUB creators MUST therefore encode data URLs as [=core media type resources=]
+					or provide a fallback using the intrinsic fallback mechanisms of the host format.</p>
 
 				<div class="note">
+					<p>These restrictions on the use of data URLs are to prevent security issues and also to ensure that
+						[=reading systems=] can determine where to take a user next (i.e., because data URLs cannot be
+						referenced from the [=EPUB spine | spine=]).</p>
+
 					<p>The list of prohibited uses for data URLs is subject to change as the respective standards that
 						allow their use evolve.</p>
 				</div>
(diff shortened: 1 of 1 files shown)


epub33/core/index.html
```diff
@@ -540,22 +540,15 @@ <h3>Terminology</h3>
...
-							<li>
-								<p>they do not have to list resources encoded as <a href="#sec-data-urls">data URLs</a>
-									in the manifest; and</p>
-							</li>
-							<li>
-								<p>they may locate resources listed in <a href="#sec-resource-locations"></a> outside
-									the EPUB container.</p>
-							</li>
-						</ul>
+						<p>EPUB creators list publication resources in the package document [=EPUB manifest | manifest=]
```
rdeltour: nitpick: s/EPUB creators list/EPUB creators must list/

(I assume a lower case must would be informative here? it doesn't hurt being extra explicit about that requirement of the `manifest` element)

epub33/core/index.html
```diff
@@ -1470,15 +1463,16 @@ <h3>Data URLs</h3>
 					directly into a URL string. The advantage of this scheme is that it allows [=EPUB creators=] to
 					embed a resource within another, avoiding the need for an external file.</p>
 
-				<p>EPUB creators MAY use data URLs in [=EPUB publications=] provided their use does not result in a
-					[=top-level content document=] or [=top-level browsing context=] [[html]]. This restriction applies
-					to data URLs used in the following scenarios:</p>
+				<p>A consequence of embedding is that data URLs are not considered their own unique publication
+					resources (i.e., they are part of their containing publication resource). As a result, EPUB creators
```
rdeltour: add link to the _publication resources_ definition here?

epub33/core/index.html
```diff
@@ -1492,21 +1486,24 @@ <h3>Data URLs</h3>
 					</li>
 				</ul>
 
+				<p>Although data URLs are considered part of their containing resource, because they have their own
+					media types they are still subject to the <a href="#sec-foreign-resources">foreign resource
+						restrictions</a>. EPUB creators MUST therefore encode data URLs as core media type resources or
+					use them where they can provide a fallback.</p>
```
rdeltour: s/or use them where they can provide a fallback/or provide a fallback using intrinsic fallback mechanisms of the host format/

(reason: it is not because they can that they MUST 😉, and the suggested rewording is a bit more explicit).

We might add a note on the implicit consequence that data URLs cannot be used for non-CMT resources on elements that do not have intrinsic fallback capabilities, but I think it is clear enough already.

epub33/core/index.html
```diff
@@ -1492,21 +1486,24 @@ <h3>Data URLs</h3>
...
+
+				<p>These restrictions on the use of data URLs are to prevent security issues and also to ensure that
+					[=reading systems=] can determine where to take a user next (i.e., because data URLs cannot be
+					referenced from the [=spine=]).</p>
```
rdeltour: Shouldn't that be a note?

rdeltour (changes requested):
This looks OK to me, pending some (minor) edit suggestions.

Also, if we allow data URLs on linked resources, we may want to change the following (from the `link` element definition):

> The `media-type` attribute is OPTIONAL when a linked resource is located outside the EPUB container, as more than one media type could be served from the same URL [url]. EPUB creators MUST specify the attribute for all linked resources within the EPUB container

to 

> The media-type attribute is OPTIONAL when a linked resource is located outside the EPUB container (as more than one media type could be served from the same URL [url]), or when the resource is encoded as a data URL (in the `href` attribute). EPUB creators MUST otherwise specify the attribute for all linked resources located within the EPUB container.

epub33/core/index.html
mattgarrish: Ya, I think that's fine now that we reference rfc 8174. It's also terminology, so I wouldn't expect people will be looking for normative statements there.

mattgarrish:
> we may want to change the following (from the `link` element definition):

Ya, that's fine with me. I just reworked it into a bullet list as there's a lot going on in the sentence.

Since linked resources are kind of a niche anyway, allowing data URLs for them doesn't bother me. But if we don't want them at all in the package document I can go that way, too.

epub33/core/index.html
```diff
@@ -1470,15 +1464,16 @@ <h3>Data URLs</h3>
 					directly into a URL string. The advantage of this scheme is that it allows [=EPUB creators=] to
 					embed a resource within another, avoiding the need for an external file.</p>
 
-				<p>EPUB creators MAY use data URLs in [=EPUB publications=] provided their use does not result in a
-					[=top-level content document=] or [=top-level browsing context=] [[html]]. This restriction applies
-					to data URLs used in the following scenarios:</p>
+				<p>A consequence of embedding is that data URLs are not considered their own unique [=publication
+					resources=] (i.e., they are part of their containing publication resource). As a result, EPUB
```
iherman: Purely editorial issue: a data URL is, conceptually, a URL which also _defines_ some data. So the question whether it is a publication resource or not does not apply to the data URL itself, but to the data it defines.

But this is really a nuance and it may be simpler to ignore it here.

iherman (approved):

rdeltour (approved):

epub33/core/index.html
mattgarrish: We could say: "A consequence of embedding is that the data in a data URL is not considered a unique publication resource (i.e., it is part of the containing publication resource)."

I don't think that's particularly hard to parse, if that's what you're getting after?

mattgarrish:
Should we ban data URLs in `link` elements and be done with them in the package document? Given the problematic security of data URLs, and the limited uses of links, it seems like a prudent thing to do.

rdeltour:
> Should we ban data URLs in `link` elements and be done with them in the package document?

Yeah, no strong opinion here, but I'm pretty confident no one will complain if we ban them.

epub33/core/index.html
iherman: Yep, that sounds good to me

iherman:
> Should we ban data URLs in `link` elements and be done with them in the package document? Given the problematic security of data URLs, and the limited uses of links, it seems like a prudent thing to do.

Same as @rdeltour.

mattgarrish:
I've updated the PR to disallow data urls in the package document, but it's meant some rewriting of the data URLs section.

It didn't make sense to lead into the restriction in the package document by talking about publication resources, so I added the restriction to the list of scenarios that can result in insecure rendering. Instead of saying these will result in top-level content documents/browsing context, I shifted it to "can result" to justify having the link element in there. It's a bit of an outlier, as we don't know how a reading system will present a link.

I then shifted the text about data URLs not being unique publication resources down to the paragraph about their still having their own media types so being subject to fallback requirements.

Otherwise, I just undid the bits that were dealing with the link element - the one note about allowing data urls and the paragraph about linked resources not needing a media-type attribute when expressed as data urls.

mattgarrish:
As this has been open for a week now, I'm going to merge.

PR #2496, opened 2022-11-30
Intrinsic fallbacks for audio/video

### LINKED ISSUE #2490: intrinsic fallback should be clarified for the HTML `video` element too
An audio resource can be used in an HTML `video` element. In this case, the element's `source` children can define the intrinsic audio fallback.

By only mentioning the `audio` element in the intrinsic fallback clarification, the EPUB spec confuses the element with the underlying resource: the **resource** requires a fallback, regardless of where it appears.

This can easily be fixed by replacing any reference to "[audio](https://html.spec.whatwg.org/multipage/#the-audio-element) element" by "[media element](https://html.spec.whatwg.org/multipage/media.html#media-element) (namely, [`audio`](https://html.spec.whatwg.org/multipage/media.html#the-audio-element) and [`video`](https://html.spec.whatwg.org/multipage/media.html#the-video-element))".

iherman:
This is to Fix #2490.

I do have a problem, however: I am not sure I understand the note. What does "this requirement" means? Does it mean that the authors may add intrinsic fallback for a video element, at least in case a video resource is played? This seems to contradict with the HTML spec, which says (in the [video element](https://html.spec.whatwg.org/multipage/media.html#the-video-element):

> Content may be provided inside the video element. User agents should not show this content to the user; it is intended for older web browsers which do not support video, so that legacy video plugins can be tried, or to show text to the users of these older browsers informing them of how to access the video contents.

(This is the same text as for the audio element.) Ie, the restrictions seem to be identical to audio and video; I wonder if, with this change, the note should not simply be removed altogether, as not providing any further information...

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2496.html" title="Last updated on Dec 1, 2022, 5:15 AM UTC (bd90dde)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2496/27fbf42...bd90dde.html" title="Last updated on Dec 1, 2022, 5:15 AM UTC (bd90dde)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1323,21 +1323,25 @@
 						requirements of specific elements.</p>
 
 					<section id="sec-fallbacks-audio">
-						<h5>HTML <code>audio</code> fallbacks</h5>
+						<h5>HTML <code>audio</code> and <code>video</code> fallbacks</h5>
 
 						<p id="confreq-resources-cd-fallback-media">[=EPUB creators=] MUST NOT use embedded
-							[[html]] [=flow content=] within the <a data-cite="html#the-audio-element"
-									><code>audio</code></a> element as an intrinsic fallback for [=foreign resources=].
+							[[html]] [=flow content=] within the <a data-cite="html#the-media-element"
+									>media element</a> (i.e, <a data-cite="html#the-audio-element"
+									><code>audio</code></a> or <a data-cite="html#the-vido-element"
+									><code>video</code></a>) as an intrinsic
+									fallback for [=foreign resources=].
 							Only child [^source^] elements [[html]] provide intrinsic fallback capabilities.</p>
 
-						<p>Only older [=reading systems=] that do not recognize the <code>audio</code> element (e.g.,
-							EPUB 2 reading systems) will render the embedded content. When reading systems support the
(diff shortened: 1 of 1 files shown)


mattgarrish:
> What does "this requirement" means?

I assume it meant that because video do not require fallbacks, what does and doesn't count as a fallback doesn't apply. But now that we're talking about the `video` element, I agree it doesn't make sense. What we should be clarifying is that we're not talking about video resources here, but audio played in a `video` element. Otherwise, I can see people getting confused by this section.

So perhaps the note should now say something like:

> NOTE
> The requirement for fallbacks only applies to audio foreign resources referenced from `audio` and `video` elements. Fallbacks are not required for video resources; they are exempt resources.

It might help to emphasize this earlier, too, by saying: "as an intrinsic fallback for **audio** [=foreign resources=]" at line 1332-33. The note might then be shortened a bit to avoid the redundancy.

iherman:
Made the changes per @mattgarrish's comment.

rdeltour (approved):

iherman:
@mattgarrish are you ok merging this?

PR #2500, opened 2022-12-08
Remove the acquire and xmp properties

### LINKED ISSUE #2489: Unused metadata properties
In the end, we only have two properties that don't meet the minimum two implementation threshold in the [usage report](https://github.com/w3c/epub-specs/blob/main/epub33/reports/epub-properties-use.md):

- acquire
- xmp

The question now is what do we do with these? Despite our best efforts, we haven't turned up anyone using either of these properties, so they are better categorized as having no implementations rather than being under-implemented.

One possibility is that we remove both from the specification but leave them as valid in epubcheck:

- the acquire property is defined in the Previews specification which links to its definition in a 3.0.1 vocabulary. That makes it valid even if it is no longer in 3.3.
- the xmp doesn't have an equivalent spec that would make it valid, but we could leave it valid for historical reasons without continuing to promote it for use.

The alternative, of course, is to put "under-implemented" labels on them and leave them in the specification, but given that there are no known users of these properties, it's a bit of clutter that we arguably don't need.

mattgarrish:
Assuming everyone is fine with dropping these rather than keeping them around as under-implemented.

Fixes #2489

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2500.html" title="Last updated on Dec 8, 2022, 1:00 PM UTC (7767f0d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2500/0d6432d...7767f0d.html" title="Last updated on Dec 8, 2022, 1:00 PM UTC (7767f0d)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3302,9 +3302,9 @@
       …
       &lt;link
           rel="record"
-          href="http://example.org/meta/12389347?format=xmp"
+          href="http://example.org/meta/12389347?format=onix"
           media-type="application/xml"
-          properties="xmp"/>
+          properties="onix"/>
       …
    &lt;/metadata>
    …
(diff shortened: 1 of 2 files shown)


murata2makoto:
"to a ONIX record" should be "to an ONIX record"

dauwhe (approved):

iherman (approved):

PR #2504, opened 2022-12-15
Remove editorializing from the AAA note

mattgarrish:
This PR writes out the text about the ability to meet AAA as described in https://github.com/w3c/publ-a11y/issues/141#issuecomment-1353345225.

Encouraging people to follow AAA SC is more important than wading into whether conforming at AAA is possible.


clapierre (approved):

iherman (approved):

avneeshsingh (approved):
Looks good. We should encourage people to meet level AAA SC wherever possible.

PR #2505, opened 2022-12-16
Update workflows to use checkout@v3

mattgarrish:
The accessibility spec failed to run this morning (for no particular reason as it resolved itself on the next run). When I opened the action, though, it had this new warning after the failure message:

> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout@v2

If you follow the provided URL to [the page on what to update](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-versioned-actions), we need to switch all the workflows to "actions/checkout@v3", so that's what this PR does.

### DIFF
--- a/.github/workflows/epub-a11y-eaa-mapping.yml
+++ b/.github/workflows/epub-a11y-eaa-mapping.yml
@@ -15,7 +15,7 @@ jobs:
     name: Publish EPUB Accessibility - EU Accessibility Act Mapping 
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: w3c/spec-prod@v2
         with:
           SOURCE: epub33/epub-a11y-eaa-mapping/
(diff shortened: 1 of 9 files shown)


iherman:
I am a bit worried about these things, and always have an "ain't broken, don't fix it" reaction. I would prefer to hear from @deniak before doing any change like that.

(@deniak there was a time when echidna failed because we had to use specific virtual hosts to run the actions and we (well, you) had to manually add new allowed hosts to a magic list for echidna. The experience was that such runs were often successful the second time. I do not know whether this is what happened here...)

deniak:
That PR shouldn't break anything. It's a major release just because of the node version support (see [changelog](https://github.com/actions/checkout/blob/main/CHANGELOG.md)) but in your case, this should work just fine.

Echidna only checks that the requests with a token comes from a GitHub server but since you are running spec-prod via a GH action, there's no problem.

deniak (approved):

iherman (approved):

PR #2506, opened 2022-12-16
Clarify that ocf processing files are not listed in the manifest

mattgarrish:
As raised in https://github.com/w3c/epubcheck/issues/1452, I don't know that we can ever perfectly clarify what publication resources are, but PR makes clear that the mimetype file and meta-inf files are not, so they are not listed in the manifest.

That should make the statement about the manifest only listing publication resources redundant, as the other cases are now explained (linked resources and the above files).

There will still be impossibilities for epubcheck to determine, like whether a publication resource without any reference to it is intentionally travelling in the container or was forgotten to be removed, but as noted in the epubcheck issue we can add usage messages to flag those.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2506.html" title="Last updated on Dec 16, 2022, 5:47 PM UTC (37cb0cd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2506/564fd05...37cb0cd.html" title="Last updated on Dec 16, 2022, 5:47 PM UTC (37cb0cd)">Diff</a>

PR #2507, opened 2022-12-16
Add property for exempt resources

mattgarrish:
With 3.3, we now allow resources in the container, and listed in the manifest, that may or may not be used in rendering the publication. These are defined at the end of the [exempt resources ](https://www.w3.org/TR/epub-33/#sec-exempt-resources) section - to allow data sets, etc. to travel with an EPUB.

A problem raised in https://github.com/w3c/epubcheck/issues/1452, however, is that epubcheck has for a long time reported unused resources in the container because it's been an issue with vendors rejecting publications with these.

There's no way of determining whether a resource is unused or was intentionally meant to travel in the container when a reading system or validator cannot find a reference to it in content documents, so if we don't do anything then exempt resources will result in warnings that they appear unused.

This PR represents one option to solve this problem, which is to add a new manifest item `properties` value for tagging exempt resources. This would clarify that the author intended them to be included. Since no one has yet been able to use this feature, adding this requirement is not problematic from an existing content perspective.

I'm opening this now, but I don't know we'll be able to resolve this until people are back in the new year. It's open for discussion by anyone who desperately needs an epub fix over the next couple of weeks! 😉

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2507.html" title="Last updated on Dec 17, 2022, 1:25 AM UTC (fa2b580)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2507/27599ee...fa2b580.html" title="Last updated on Dec 17, 2022, 1:25 AM UTC (fa2b580)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1247,11 +1247,11 @@
 					resources (i.e., the requirement for a fallback for the foreign content document covers any
 					rendering issues within it). As the resource is not referenced from an EPUB content document, it
 					automatically becomes exempt from fallbacks.</p>
-				
+
 				<div class="note">
 					<p>Resources that fall under this latter exemption should be identified in the manifest using the
-						<code>exempt</code> property on their [^item^] entry to avoid their being flagged by EPUB conformance
-						checkers. Refer to <a href="#sec-exempt"></a> for more information.</p>
+							<code>exempt</code> property on their [^item^] entry to avoid their being flagged by EPUB
+						conformance checkers. Refer to <a href="#sec-exempt"></a> for more information.</p>
 				</div>
 			</section>
 
@@ -4937,6 +4937,9 @@ XHTML:
 							<li><a href="#sec-svg">svg</a></li>
 							<li><a href="#sec-switch">switch</a></li>
 						</ul>
+
+						<p>EPUB creators SHOULD set the <a>exempt</a> property whenever a resource referenced by an
+								<code>item</code> element matches its definition.</p>
 
 						<aside class="example" id="example-item-properties-scripted-mathml"
 							title="Identifying a scripted content document with embedded MathML">
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -4938,8 +4938,8 @@ XHTML:
 							<li><a href="#sec-switch">switch</a></li>
 						</ul>
 
-						<p>EPUB creators SHOULD set the <a>exempt</a> property whenever a resource referenced by an
-								<code>item</code> element matches its definition.</p>
+						<p>EPUB creators SHOULD set the <a href="#sec-exempt">exempt</a> property whenever a resource
+							referenced by an <code>item</code> element matches its definition.</p>
 
 						<aside class="example" id="example-item-properties-scripted-mathml"
 							title="Identifying a scripted content document with embedded MathML">
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1249,9 +1249,9 @@
 					automatically becomes exempt from fallbacks.</p>
 				
 				<div class="note">
-					<p>Resources that fall under this latter exemption have to be identified in the manifest using the
-						<code>exempt</code> property on their [^item^] entry. Refer to <a href="#sec-exempt"></a>
-						for more information.</p>
+					<p>Resources that fall under this latter exemption should be identified in the manifest using the
+						<code>exempt</code> property on their [^item^] entry to avoid their being flagged by EPUB conformance
+						checkers. Refer to <a href="#sec-exempt"></a> for more information.</p>
 				</div>
 			</section>
 
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1247,6 +1247,12 @@
 					resources (i.e., the requirement for a fallback for the foreign content document covers any
 					rendering issues within it). As the resource is not referenced from an EPUB content document, it
 					automatically becomes exempt from fallbacks.</p>
+				
+				<div class="note">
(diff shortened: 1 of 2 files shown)


mattgarrish:
One additional comment is that I've made this only a recommendation to set so that it matches the warning that epubcheck already emits for unused resources in the container. That way the warning could be to either set the exempt property or remove the resource.

epub33/core/vocab/item-properties.html
```diff
@@ -37,6 +37,43 @@ <h4>cover-image</h4>
...
+				</tr>
+				<tr>
+					<th>Description:</th>
+					<td>
+						<p>The <code>exempt</code> property indicates that the [=publication resource=]
+							is an [=exempt resource=] (e.g., is used as script input or is included in the
+							[=EPUB container=] for non-rendering purposes).</p>
+						<p>If [=EPUB creators=] do not set this property, acknowledging the resources are
+							included by design, reading systems and [=EPUB conformance checkers=] may flag the
+							resources as unused if they cannot find references to them.</p>
```
iherman, line 59: I am not sure I understand the "acknowledging the resources are included by design" remark here.

epub33/core/index.html
```diff
@@ -1247,6 +1247,12 @@ <h3>Exempt resources</h3>
 					resources (i.e., the requirement for a fallback for the foreign content document covers any
 					rendering issues within it). As the resource is not referenced from an EPUB content document, it
 					automatically becomes exempt from fallbacks.</p>
+
+				<div class="note">
+					<p>Resources that fall under this latter exemption should be identified in the manifest using the
+							<code>exempt</code> property on their [^item^] entry to avoid their being flagged by EPUB
```
iherman, line 1253: This "fall under this latter exemption" is a bit vague. Isn't it better to make it more explicit and refer to exceptions that are not content-specific?

iherman (approved):
I was worried about the formal consequences, but I see that the property is set as a SHOULD (in line 4941), ie, it does _not_ invalidate our testing results. That being said, I would hope that, if accepted, epubcheck will add tests on whether it warns or not with this property being set, so this check will appear in our final report. @rdeltour ?

rdeltour:
A couple things:

## 1. how this relates to EPUBCheck

> A problem raised in https://github.com/w3c/epubcheck/issues/1452, however, is that epubcheck has for a long time reported unused resources in the container because it's been an issue with vendors rejecting publications with these.

To clarify: what EPUBCheck warns about is not unused resources, but resources that are present in the container and not listed in the manifest.

This is reported as `OPF-003`, with the following message:

> Item "_some-resource-path_" exists in the EPUB, but is not declared in the OPF manifest.

EPUBCheck does not (currently) check if the resource is used or not, for that specific report.

So EPUBCheck basically requires authors to list everything in the manifest if they do not want `OPF-003` to be reported. That includes unused resources and exempt resources travelling in the container.

If the spec requires that only publication resources must be listed in the manifest, this conflicts with what EPBUCheck has historically recommended for more than a decade (i.e. list everything in the manifest, even unused resources).  
#2506 fixes this conflict by removing the requirement that only publication resources are listed.

By removing the requirement that only publication resources are listed in the manifest (as proposed in #2506), unused and exempt resources are in a gray area: EPUB does not recommend that they are listed or not, so they can be. It makes the overzealous-but-historical EPUBCheck `OPF-003` report acceptable: authors will keep on listing everything in the manifest so that `OPF-003` is not reported.

## 2. does this PR help?

If I understand correctly, the intent of this PR is to provide an explicit way out of the gray area I mentioned above. Basically, it allows author to say "do not warn about this resource, it has a stamp of approval".

EPUBCheck would:
- keep on warning about resources in the container but not in the manifest (these resources can be listed in the manifest with an `exempt` property, to remove the warning)
- start warning about resources in the manifest but not used, if they do not have the `exempt` property

That may be a little improvement, but in practice I am concerned about two things:
- EPUBCheck can only check if a resource is used or not for unscripted content. As soon as scripting is involved, a resource may be used by a script, an image added to the DOM, etc, etc, so EPUBCheck cannot issue any warning.
- If EPUBCheck starts warning about unused resources that are not explicitly flagged as `exempt`, then it will potentially start issuing warnings for many existing EPUBs which passed validation before. In other words, this change would introduce a backward-incompatible warning check.

mattgarrish:
> To clarify: what EPUBCheck warns about is not unused resources, but resources that are present in the container and not listed in the manifest.

Yes, but don't you also want to check resources in the manifest to see if they're used in the publication? That's why I opened this issue, as there's no way to differentiate a publication resource you can't find a use of vs. one of these exempt resources.

I know once scripting gets added to the equation things get even murkier, but a data file used by a script is one of the cases that makes these resources "exempt". It's an explicit way of identifying the resources that can't otherwise be determined (it doesn't only have to apply to foreign resources).

> then it will potentially start issuing warnings for many existing EPUBs which passed validation before

I'm not sure about this, at least where it matters. Scripting doesn't get significant use, and the legitimate resources that would be flagged would be data files for input, which is probably a niche case given the list of CMTs for which you wouldn't need a fallback.

Where it'll definitely raise warnings is for publications that have resources that are no longer used but are still floating around in the manifest/container. All that requires is removing the unused resource, so we're not forcing publishers to figure out a workaround to keep content working as intended.

But that's one of the questions that needs answering from the group: is this worth getting into? If we choose now not to address what belongs in the manifest, then it will be too late in the future to do anything about it (except list all resource that epubcheck can make no determination about as usage messages).

I don't have a strong opinion about this, so it would be helpful to get some publisher input.

PR #2515, opened 2023-01-06
Mark the rtl dir value under-implemented

mattgarrish:
As discussed on the 2023-01-05 telecon, there are support gaps for the `dir` attribute in the package document. Looking at the tests, these failures appear to all be centred around support for the `rtl` value.

Consequently, rather than mark the whole attribute under-implemented, this PR only adds the label to the `rtl` value. This is in keeping with how the `portrait` value for `rendition:spread` is deprecated.

If I've misread the test results and we need to consider a whole-attribute labelling, please feel free to say so, though.

And as this looks like it could be the only under-implemented feature we'll have, and we want reading systems to support it moving forward, it doesn't make sense to me to label it in the Reading Systems specification. It seems to undermine the goal of getting support improved. Can we limit ourselves only to warning authors? And, similarly, can we remove the text in the RS appendix about not having to support under-implemented features?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2515.html" title="Last updated on Jan 11, 2023, 1:25 PM UTC (0ff1bd6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2515/6293601...0ff1bd6.html" title="Last updated on Jan 11, 2023, 1:25 PM UTC (0ff1bd6)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3217,7 +3217,8 @@
 
 					<ul>
 						<li><code>ltr</code> &#8212; left-to-right base direction;</li>
-						<li><code>rtl</code> &#8212; right-to-left base direction; and</li>
+						<li><code>rtl</code> &#8212; right-to-left base direction (<a href="#under-implemented"
+								>under-implemented</a>); and</li>
 						<li><code>auto</code> &#8212; base direction is determined using the Unicode Bidi Algorithm
 							[[bidi]].</li>
 					</ul>
@@ -11789,6 +11790,9 @@ EPUB/images/cover.png</pre>
 						Recommendation</a></h3>
 
 				<ul>
+					<li>06-Jan-2023: Marked the <code>rtl</code> value of the <code>dir</code> attribute
+						under-implemented due to lack of support in reading systems. See <a
+							href="https://github.com/w3c/epub-specs/pull/2515">pull request 2515</a>.</li>
 					<li>16-Dec-2022: Clarified that the special files for processing the OCF container are not listed in
 						the manifest, so the restriction that the manifest only list publication resources is not
 						needed. See <a href="https://github.com/w3c/epub-specs/pull/2506">pull request 2506</a>.</li>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3206,7 +3206,15 @@
 					elements).</p>
 
 				<section id="attrdef-dir">
-					<h4>The <code>dir</code> attribute</h4>
+					<h4>The <code>dir</code> attribute (under-implemented)</h4>
+
+					<div class="note">
+						<p>The <code>dir</code> attribute is marked <a href="#under-implemented">under-implemented</a>
+							as [=reading systems=] often only support a single default directionality for text display.
+							[=EPUB creators=] are still strongly encouraged to set the proper directionality of text
+							values in the [=package document=] to ensure proper rendering once this situation
+							improves.</p>
+					</div>
 
 					<p
 						data-tests="#pkg-dir_creator-rtl,#pkg-dir_rtl-root-ltr,#pkg-dir_rtl-root-unset,#pkg-dir_unset-root-rtl,#pkg-dir_unset-root-unset,#pkg-dir-auto_root-rtl"
@@ -3217,14 +3225,13 @@
 
 					<ul>
 						<li><code>ltr</code> &#8212; left-to-right base direction;</li>
-						<li><code>rtl</code> &#8212; right-to-left base direction (<a href="#under-implemented"
-								>under-implemented</a>); and</li>
+						<li><code>rtl</code> &#8212; right-to-left base direction; and</li>
 						<li><code>auto</code> &#8212; base direction is determined using the Unicode Bidi Algorithm
 							[[bidi]].</li>
 					</ul>
 
-					<p data-tests="#pkg-dir-auto_root-rtl,#pkg-dir-auto_root-unset">[=Reading systems=] will assume the
-						value <code>auto</code> when [=EPUB creators=] omit the attribute or use an invalid value.</p>
+					<p data-tests="#pkg-dir-auto_root-rtl,#pkg-dir-auto_root-unset">Reading systems will assume the
+						value <code>auto</code> when EPUB creators omit the attribute or use an invalid value.</p>
 
 					<div class="note">
 						<p>The base direction specified in the <code>dir</code> attribute does not affect the ordering
@@ -11790,9 +11797,9 @@ EPUB/images/cover.png</pre>
 						Recommendation</a></h3>
 
 				<ul>
-					<li>06-Jan-2023: Marked the <code>rtl</code> value of the <code>dir</code> attribute
-						under-implemented due to lack of support in reading systems. See <a
-							href="https://github.com/w3c/epub-specs/pull/2515">pull request 2515</a>.</li>
+					<li>11-Jan-2023: Marked the <code>dir</code> attribute under-implemented due to lack of support in
+						reading systems. See <a href="https://github.com/w3c/epub-specs/pull/2515">pull request
+						2515</a>.</li>
 					<li>16-Dec-2022: Clarified that the special files for processing the OCF container are not listed in
 						the manifest, so the restriction that the manifest only list publication resources is not
 						needed. See <a href="https://github.com/w3c/epub-specs/pull/2506">pull request 2506</a>.</li>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -11788,9 +11788,6 @@ EPUB/images/cover.png</pre>
 					<li>11-Jan-2023: Marked the <code>dir</code> attribute under-implemented due to lack of support in
 						reading systems. See <a href="https://github.com/w3c/epub-specs/pull/2515">pull request
 						2515</a>.</li>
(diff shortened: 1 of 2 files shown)


OriKuttner:
Where in the package document there is a dir="rtl"?

mattgarrish:
The root element and all elements that accept text allow dir. See the diff.

iherman (approved):

epub33/core/index.html
```diff
@@ -3217,7 +3217,8 @@ <h4>The <code>dir</code> attribute</h4>
 
 					<ul>
 						<li><code>ltr</code> &#8212; left-to-right base direction;</li>
```
```suggestion
						<li><code>ltr</code> &#8212; left-to-right base direction (<a href="#under-implemented"
 								>under-implemented</a>); and</li>
```

Alas! Per the [report](https://w3c.github.io/epub-tests/results.html#sec-consolidated-internationalization-results) the ltr test also fails, so that should be set as under-implemented, too. Only the `auto` value passes

iherman (approved):

epub33/core/index.html
mattgarrish: Is there an ltr test that fails? The only failing one I see that meantions ltr is overridding ltr on the root with rtl, but that sounds like an rtl failure.

To be honest, though, I wonder if all of these tests are accurate. If RTL isn't supported, for example, then how can you determine if a later override is working? auto requires using the BIDI algorithm, but if it looks like LTR because that's all the reading system supports how do you know auto is really working?

OriKuttner: All publications in Hebrew (RTL) that I know use dir=rtl only in XHTML files and page-progression-direction="rtl" in spine.

mattgarrish: I'm wondering if we should just label the whole attribute underimplemented until there's more consistent support, since it seems reading systems have a default way of rendering and don't support much else.

mattgarrish: But there's still the question of what to do about the RS spec. Do we make an i18n feature not a requirement because it's not yet supported, or do we leave its requirements intact because we anticipate that this feature should be supported?

iherman: > Is there an ltr test that fails? The only failing one I see that meantions ltr is overridding ltr on the root with rtl, but that sounds like an rtl failure.

You are right, the labeling is misleading.

> I'm wondering if we should just label the whole attribute underimplemented until there's more consistent support, since it seems reading systems have a default way of rendering and don't support much else.

+1 to label the attribute under implemented as a whole.

> But there's still the question of what to do about the RS spec. Do we make an i18n feature not a requirement because it's not yet supported, or do we leave its requirements intact because we anticipate that this feature should be supported?

Doesn't the 'under-implemented' apply to the RS as well? If we mark it as under-implemented in the core spec, we should also mark it under-implemented in the RS. That actually should be the general rule whenever applicable.

mattgarrish: > Doesn't the 'under-implemented' apply to the RS as well?

I haven't been able to wrap my head around that.

If we mark the feature as under-implemented in the RS spec, aren't we just undermining future support? It takes a requirement that we still want implemented, in this case, and turns it into a recommendation. So moving forward reading system developers are no longer obligated to follow the requirement (if they haven't done it, why should I?).

I can't see how that keeps pushing developers to update their failing interfaces? Do we add a caution box that the under-implemented label will be removed?

It made more sense to me when we were thinking the label might be applied to features that had some support but that were unlikely to ever be fully implemented, so we had to keep them around somehow. The `dir` attribute seems a little more valuable than that.

If you warn authors about under-implemented features, however, then at least they know not to rely on them until there is more support.

iherman: Hm. I see your points.

I was looking at this from a W3C process point of view, although the EPUB spec is very special in this respect. In general, the CR->PR transition also means to show that a _feature can be implemented_ (by several, independent implementations). I.e., the question is not only a matter of good will from the RS developers' point of view, but it is also whether the spec contains major technical flaws. That is what triggered my remark: we do not have the proof that what is defined as a _MUST_ statement in [§3.7](https://www.w3.org/TR/epub-rs-33/#sec-epub-rs-i18n) is realizable.

But maybe that is all right: we have a special derogation from the process after all...

In this very case, I wonder whether it is nevertheless worth adding an extra (non-normative) text in the aforementioned section to instigate RS developers to do the work at some point drawing that attention to the fact that only partial implementations exist for the `dir` feature and it would be very important for different markets to get this thing done right...

mattgarrish: Playing devil's advocate for a minute, perhaps there's a case that we've set the requirement too high (remembering previous versions didn't specify any support requirements). Knowing that the dir attribute rarely works, that means it's been the case for over a decade now and no one's complained. Is that an argument to move support to a "should" and add a note that RS developers concerned with i18n support should strongly consider supporting the feature?

It's less ideal, but would be more in line with how we're handling other cases. We could then do away with labelling anything under-implemented, which has its positives.

iherman: > Playing devil's advocate for a minute, perhaps there's a case that we've set the requirement too high (remembering previous versions didn't specify any support requirements). Knowing that the dir attribute rarely works, that means it's been the case for over a decade now and no one's complained. Is that an argument to move support to a "should" and add a note that RS developers concerned with i18n support should strongly consider supporting the feature?

With my W3C hat 'on': this is not really an option. The "usual" approach taken in W3C specs is that if a feature does not pass the bar of the CR exit criteria, it is to be removed altogether from the spec. We have the special 'under-implemented' arrangement for EPUB 3.3, this is what we should use.

(Yes, we moved the EPUB-CSS features into the SHOULD category as well, but I would not expect that to raise eyebrows. The `dir` attribute is different.)

iherman: Also: keeping the 'under-implemented' flag for 'dir' in general keeps up a pressure on RS-s. I think that is an advantage.

> Knowing that the dir attribute rarely works, that means it's been the case for over a decade now and no one's complained.

The `rtl` feature is special for some languages which, unfortunately, do not necessarily have a very strong "voice" in the community. But I believe keeping the pressure is really the right thing to do.

mattgarrish: > The "usual" approach taken in W3C specs is that if a feature does not pass the bar of the CR exit criteria, it is to be removed altogether from the spec.

Right, but that applies if the feature is new and required, right? We haven't tested all the existing recommended features of epub to see if they're supported, so if we had set support to "should" instead of "must" when we added the requirements, this would never have been flagged. It was our decision to make them musts for an existing feature.

> But I believe keeping the pressure is really the right thing to do.

Right, I don't know that I believe strongly that reducing the requirement helps, but it's another option to consider. What we were planning for under-implemented features (indirect recommendations) effectively would have done the same thing.

shiestyle (approved):

dauwhe (approved):

epub33/core/index.html
mattgarrish: In any case, I've added new notes and expanded the scope of the under-implemented label as we discussed earlier. If everyone is fine with that, let's not drag this out.

epub33/core/index.html
```diff
@@ -11777,6 +11785,12 @@ <h3>Substantive changes since <a href="https://www.w3.org/TR/2022/CR-epub-33-202
 						Recommendation</a></h3>
 
 				<ul>
+					<li>11-Jan-2023: Marked the <code>dir</code> attribute under-implemented due to lack of support in
+						reading systems. See <a href="https://github.com/w3c/epub-specs/pull/2515">pull request
+						2515</a>.</li>
+					<li>09-Jan-2023: Marked the <code>rtl</code> value of the <code>dir</code> attribute
+						under-implemented due to lack of support in reading systems. See <a
+							href="https://github.com/w3c/epub-specs/pull/2515">pull request 2515</a>.</li>
```
iherman: I presume this entry should be removed from the list of changes; the entry of 11-Jan-2023 supersedes it.

epub33/rs/index.html
```diff
@@ -2697,6 +2697,11 @@ <h3>Substantive changes since <a href="https://www.w3.org/TR/2022/CR-epub-rs-33-
 						Recommendation</a></h3>
 
 				<ul>
+					<li>11-Jan-2023: Added note advising reading system developers to support package metadata
+						directionality even though marked as under-implemented for authoring. See <a
+							href="https://github.com/w3c/epub-specs/pull/2515">pull request 2515</a>.</li>
+					<li>11-Jan-2023: Removed support description for under-implemented features. See <a
+							href="https://github.com/w3c/epub-specs/pull/2515">pull request 2515</a>.</li>
```
iherman, line 2704: I guess this is based on getting to an equilibrium point on MO, ie, that there is no under-implemented notion for MO at all (which is probably all right, because MO is not a required feature as a whole).

iherman (changes requested):

epub33/core/index.html
mattgarrish: Ya, not sure what happened there. I modified the existing entry so I must have missed the duplication when resolving the merge conflict. Gone now.

epub33/rs/index.html
mattgarrish: Either that or we do something similar where we under-implement the authoring feature. I made that change after the last call when it sounded like we wouldn't have to label MO because it's only a recommended feature. It's easy to get back the prose if we need it, though.

iherman (approved):

PR #2529, opened 2023-01-24
Fix incorrect use of roles on body

### LINKED ISSUE #2528: Sample codes use roles on html body
Sample codes in EPUB Accessibility Techniques 1.1 [Section 4.3.2](https://www.w3.org/TR/epub-a11y-tech-11/#sem-002) use roles on html body. 
```
<html … >
   …
   <body
       role="doc-part">
      <h1>Part 1</h1>
   </body>
</html>
```
Should there be a section wrapper to apply roles? And I wonder if it is necessary to add another heading ahead of the section wrapper to get a conforming outline.

mattgarrish:
This updates the techniques document and fixes #2528.

I also made some general tweaks to the section, like removing "semantics" from the headings. I also added a note to explain why aria-labelledby is used for the example with all the content in one file and not used for the chunked examples.

- Accessibility Techniques [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2528/epub33/a11y-tech/index.html)
- Accessibility Techniques [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2528/epub33/a11y-tech/index.html)

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -1158,30 +1158,10 @@
    &lt;/body>
 &lt;/html></pre>
 
-					<p>It is not necessary to add a part wrapper to each chapter when breaking them out into separate
-						files, as in the following example:</p>
-
-					<pre>&lt;html … >
-   …
-   &lt;body>
-      &lt;section
-          role="doc-part"
-          aria-label="Part 1">
-          
-         &lt;section
-             role="doc-chapter">
-            
-            &lt;h2>Chapter 1&lt;/h2>
-            …
-         &lt;/section>
-      &lt;/section>
-   &lt;/body>
-&lt;/html></pre>
-
-					<p>Doing so introduces a new <code>part</code> landmark into each document, which will cause an
-						assistive technology to announce the landmark each time (e.g., the user will hear "Part 1 Part 1
-						Chapter 1" in the preceding example because the "Part 1" is also the heading content of the
-						preceding document).</p>
+					<p>If another <code>section</code> tag were added around the chapter in the preceding example to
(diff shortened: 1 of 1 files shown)


clapierre (commented):
I am wondering if we should make this even clearer after the example where you have
Doing so introduces a new part landmark into each document, which will cause an assistive technology to announce the landmark each time (e.g., the user will hear "Part 1 Part 1 Chapter 1" in the preceding example because the "Part 1" is also the heading content of the preceding document).

something like
Instead of
```
      <section
          role="doc-part"
          aria-label="Part 1">
```

that should be removed completely. and only have the `<section role="doc-chapter">.`

mattgarrish:
The last example is what not to do. I hate making those examples, as people can mistake them as proper techniques, but I don't think there's any way to differentiate a "bad" example.

In the issue I suggested maybe cutting out the example entirely. Maybe it would be better to explain in words not to put a wrapper section around, rather than show it. Would that make it more palatable for you @clapierre?

clapierre:
That was my concern exactly @mattgarrish I think what you suggest is great.

mattgarrish:
See if it's better now. The paragraphs above the example already explain why it's bad to repeat roles, so all I've left after the example is an explanation of the particular repetition problem it would create if an extra wrapper were present.

clapierre (approved):

PR #2530, opened 2023-02-01
Minor clarification to summary technique

mattgarrish:
Changes the following sentence from:

> As this metadata is required by the EPUB Accessibility specification,

to:

> When other accessibility metadata is present in the package document,

as it's been noted that the meaning of "this metadata" may not be clear.


gregoriopellegrino (approved):
Thank you!

clapierre (approved):

wareid (approved):

PR #2546, opened 2023-03-22
Changed the paragraph referring to sideloading

### LINKED ISSUE #2542: "sideloading" and "untrustworthy" references are unclear given lack of authentication and integrity
The reading system spec refers to "untrustworthy" EPUB files as files that are not signed and were sideloaded. But assurances of authenticity and integrity are known to be lacking for EPUB 3.3 generally; package-wide signatures are not possible. 

I don't think we should suggest that "sideloaded" files (not defined in the spec, but I think the implication is files that weren't accessed through an integrated store) are likely to be less trustworthy. Indeed, books purchased through stores seem more likely to come with DRM, which has substantial known violations of user privacy. And signatures on individual files within an EPUB package are not a good signal to use in determining whether content is trustworthy or not.

It might be that this paragraph should just be removed until #2265 is addressed. But if reading systems should be recommended to prompt for access about untrusted content, that recommendation should apply to content from any source (perhaps on the first use of that source).

iherman:
This is to fix #2542

### DIFF
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2478,7 +2478,7 @@
 
 				<p id="security-privacy-recommendations-sideloaded-consent" 
 					data-tests="#sec-untrusted-consent_network,#sec-untrusted-consent_scripting">Reading systems 
-					SHOULD treat content from any new source, including  from an integrated bookstore, as insecure 
+					SHOULD treat content from any new source, including from an integrated bookstore, as insecure 
 					(e.g., prompt users to allow scripting and network access the first time the source is accessed).
 					If a reading system allows users to load their own content (e.g., through the process 
 					of "sideloading"), each instance SHOULD be treated as insecure.
--- a/epub33/rs/index.html
+++ b/epub33/rs/index.html
@@ -2476,9 +2476,13 @@
 					the validation steps that usually occur. This could include running virus scans, validating external
 					links and remote resources, and other precautions.</p>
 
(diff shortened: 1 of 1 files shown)


mattgarrish (approved):

epub33/rs/index.html
```diff
@@ -2476,9 +2476,13 @@ <h3>Recommendations</h3>
 					the validation steps that usually occur. This could include running virus scans, validating external
 					links and remote resources, and other precautions.</p>
 
-				<p id="security-privacy-recommendations-sideloaded-consent" data-tests="#sec-untrusted-consent_network,#sec-untrusted-consent_scripting">Reading systems that allow users to load untrustworthy EPUB publications (e.g., unsigned EPUB
-					publications through the process of "sideloading") SHOULD treat such content as insecure (e.g.,
-					prompt users to allow scripting and network access).</p>
+				<p id="security-privacy-recommendations-sideloaded-consent" 
+					data-tests="#sec-untrusted-consent_network,#sec-untrusted-consent_scripting">Reading systems 
+					SHOULD treat content from any new source, including  from an integrated bookstore, as insecure
```
npdoty: two spaces after including should be one

iherman: Yep, done

iherman:
@npdoty is it good to go?

wareid (approved):

npdoty:
While it wouldn't be my preference to include a sideloading reference at all, this does address the concern in #2542 about levels of trust. Thanks.

PR #2550, opened 2023-03-29
Use respec includes for copyright statements

mattgarrish:
Uses data-oninclude to run the processing step on the included copyrights.

I've also stripped fixIncludes from all the other imports. It looks like I copied the import markup for acknowledgements from dpub-aria, as that's where I finally found the function defined, and then it got copied around wherever we've needed another import. It's correcting references to the ARIA core spec there, but obviously does nothing at all here.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2550.html" title="Last updated on Mar 30, 2023, 8:54 AM UTC (344158b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2550/15a9f05...344158b.html" title="Last updated on Mar 30, 2023, 8:54 AM UTC (344158b)">Diff</a>

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -50,6 +50,7 @@
 					branch: "main"
 				},
 				profile: "web-platform",
+				preProcess:[modifyCopyright],
 				localBiblio: {
 					"a11y-discov-vocab": {
 						"title": "Schema.org Accessibility Properties for Discoverability Vocabulary",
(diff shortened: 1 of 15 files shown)


mattgarrish:
I think we might have to abandon this idea, unless you want to report it to get it fixed @iherman?

For whatever reason, the conversion process run by the github action makes a mess of the files when the copyright is included using the respec functions. It doesn't appear to be related to the modifyCopyright function call, either, as I get the same errors using a static import.

I'm guessing from the errors that the paragraph is being moved into the header, but why is beyond me as it must be happening in a step I can't reproduce. If I serve the files locally and export, there's no issue.

mattgarrish:
Sigh... well, at least I realized what a doofus I was before completely drifting off to sleep.

I forgot to commit the shared copyright file. I have no idea why that resulted in validation errors about duplicate meta charset tags and a title tag in the body, but whatever. All good now. Day thoroughly wasted chasing phantom bugs...

If you go ahead and merge this, it might be good to pick "squash and merge" from the options so all my test commits don't get combined into the spec history.

PR #2551, opened 2023-03-30
Reactivate the TR doc builds

mattgarrish:
I noticed the merges earlier today didn't result in new versions of the three standards getting published and then remembered that the yml files got renamed while we were waiting to publish the updated CR snapshots.

Can you merge this if nothing else needs changing but the extension @iherman ?

Not sure if adding the actions back will trigger them to run. Maybe we should consider adding a workflow_dispatch event to the yml files so we can manually trigger builds from within github? See https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/


iherman:
Do we really want/need this, @mattgarrish? The idea is to produce a PR version of the rec-track documents with a date of a few weeks for now. After that date, we are not supposed to use echidna any more; if we change the yml files now this means we will publish a very short-lived CR draft, that will stand between the latest CR snapshot and the PR. Is it worth the trouble?

mattgarrish:
> this means we will publish a very short-lived CR draft, that will stand between the latest CR snapshot and the PR. Is it worth the trouble?

You tell me! 😄 

I thought we had to publish any changes after the CR snapshot before we could publish the Proposed Rec. If that's not required, then, sure, close this without merging.

iherman:
> 
> You tell me! 😄
> 
> I thought we had to publish any changes after the CR snapshot before we could publish the Proposed Rec. If that's not required, then, sure, close this without merging.

Not if we have made minor, editorial changes. None of the changes we have done are major; the comments of Nick were the most substantial but, after all, even those changes are really just editorial.

PR #2554, opened 2023-04-19
Preparation for REC

iherman:
This is a preparation for the REC publication

- added an errata.html file and linked to it from all the documents
- added a link to an epub version of the spec (epub file to be provided at the last minute)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2554.html" title="Last updated on Apr 19, 2023, 2:38 PM UTC (e4e2606)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2554/f2d2e46...e4e2606.html" title="Last updated on Apr 19, 2023, 2:38 PM UTC (e4e2606)">Diff</a>


iherman:
Obviously, this PR should not be merged until we have a final publication date...

epub33/errata.html
```diff
@@ -0,0 +1,259 @@
...
+        </dl>
+        <section>
+          <h2>Substantial Issues</h2>
+        </section>
+        <section>
+          <h2>Editorial Issues</h2>
+        </section>
+      </section>
+
+      <section data-erratalabel="Spec-A11Techniques">
```
mattgarrish:
```suggestion
      <section data-erratalabel="Spec-A11YTechniques">
```

epub33/errata.html
```diff
@@ -0,0 +1,259 @@
...
+        <p>If you have problems following this process, but you want nevertheless to report an error, you can also contact the staff contact of the Working Group, <a href="mailto:ivan@w3.org">Ivan Herman</a>.</p>
+      </section>
+    </header>
+
+    <div class="toc" id="toc"></div>
+
+    <main>
+      <!-- The data-erratalabel should include one label that filters the errata -->
+      <section data-erratalabel="Spec-Overview">
+        <h1>Open Errata on the “EPUB 3 Overview” Working Group Note</h1>
```
mattgarrish, line 57: Do we need errata for notes? Isn't it faster to just fix and republish them since they don't have the same involved update process as REC-track documents?

(Just thinking it might make for a less busy page.)

mattgarrish:
Should we also do something about https://www.w3.org/publishing/epub3/ ?

epub33/errata.html
iherman: I am a bit neutral about it... and would like to hear the reaction of the chairs.

My only reason to include it (apart from the fact that it does not cost anything to do it:-) is that, I presume, the new WG will not meet that frequently, so a number of errata may pile up between two calls, and this mechanism may help us keeping track of those.

iherman:
> Should we also do something about https://www.w3.org/publishing/epub3/ ?

Yes, we will have to. But that should be a separate issue somwhere; it cannot be handled in a PR on this repository. It is a matter of changing the .htaccess file in the /publishing page (I will have to look at that).

epub33/errata.html
mattgarrish: > the new WG will not meet that frequently, so a number of errata may pile up between two calls

Doesn't the WG have to agree to publish errata, though? I thought the sticking point for REC-track documents was getting them republished, so publishing errata is quicker, whereas notes we can freely revise and republish once we know what needs changing?

iherman: It is all true. But will we look at the possible errata frequently? That is really the issue.

Again, I am not against removing the note references, but I would like to see how Wendy & Co imagine running the show...

wareid: The only notes I see potentially receiving attention are the accessibility related ones, particularly as we approach EAA. I'm ok with errata and keeping track, but we can always revise the process if it becomes arduous?

wareid (approved):

epub33/errata.html
iherman: Yes, we can touch that whenever we want.

So let us leave it as is now.

PR #2561, opened 2023-05-18
Changed the reference to MO

iherman:
It pointed to an erronous (read: non-existing) fragment ID in the A11 spec; changed it to the relevant section in the core spec.


mattgarrish:
That's a better reference for media overlays, but I'm not sure if @gregoriopellegrino also wanted to reference the [accessibility objectives for them](https://www.w3.org/TR/epub-a11y-11/#sec-sync-text-audio) in the accessibility spec.

gregoriopellegrino (approved):
For me it's fine, thank you

PR #2564, opened 2023-05-24
Fix aria-level values in caution boxes

### LINKED ISSUE #2567: algorithm error generating aria-level for caution headings
Seems, there is a coding error in the realization of heading levels for caution-notes on some W3C sites:
https://github.com/w3c/epub-specs/blob/181828b283ee5e20546746057076d970e78ce1ea/epub33/common/js/add-caution-hd.js#L16

Example:
https://www.w3.org/TR/epub-33/#sec-intro-relations
aria-heading with an extraordinary value 31 for aria-level:
&lt;div class="caution" id="caution-0">
&lt;div
  role="heading"
  aria-level="31"
  class="caution-title marker">
&lt;span>Caution&lt;/span>
&lt;/div>

Two digits concatenated together, but should be added mathematically.

mattgarrish:
The number extracted from the preceding heading level is being concatenated with "1" rather than added to it (e.g., `aria-level="31"` instead of the correct `aria-level="4"`).

This fix forces the string to a number and also prevents the level from going over 6.

Fixes #2567

### DIFF
--- a/epub33/common/js/add-caution-hd.js
+++ b/epub33/common/js/add-caution-hd.js
@@ -11,9 +11,13 @@ function addCautionHeaders() {
 		var parent_hd = parent_section.querySelector('h2,h3,h4,h5,h6');
 		var parent_num = parent_hd.localName.replace('h','');
 		
+		var aria_level = Number(parent_num) + 1;
+		
+		if (aria_level > 6) { aria_level = 6; }
+		
 		var caution_hd = document.createElement('div');
 			caution_hd.setAttribute('role', 'heading');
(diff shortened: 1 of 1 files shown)


jnurthen (approved):

iherman (approved):
Procedurally (with the Rec being published right now), I believe
- we should raise a separate issue as an Erratum to signal the problem (this is a typical Class 1 change) and bind it with this PR;
- for the time being, this PR should be flagged as not-to-be-merged, because it has to go through the admin hoopla of the Maintenance WG.

PR #2565, opened 2023-05-25
Added a redirect function to the index page

iherman:
At makes it possible to redirect (through the .htaccess trick), e.g.

`https://www.w3.org/publishing/epub33/epub-33` to `https://www.w3.org/TR/epub-33`

Just as

`https://www.w3.org/publishing/epub32/epub-contentdocs` went to the relevant 32 spec.

### DIFF
--- a/epub33/index-page/index.html
+++ b/epub33/index-page/index.html
@@ -5,6 +5,30 @@
         <meta charset="utf-8" />
         <title>EPUB 3.3</title>
         <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
+        <script>
+            const specs = [
+                "epub-33",
+                "epub-a11y-11",
+                "epub-a11y-eaa-mapping",
+                "epub-a11y-tech-11",
+                "epub-aria-authoring-11",
+                "epub-multi-rend-11",
+                "epub-overview-33",
+                "epub-rs-33",
+                "epub-ssv-11",
+                "epub-tts-10"
+            ];
+            function redirect() {
+                const ref = window.location.hash;
+                if (ref.length > 0 && ref.startsWith('#')) {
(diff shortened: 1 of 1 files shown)


mattgarrish:
I don't think I follow what's going on. Where are the hashes in the URL coming from that this script picks up on?

If I put in https://www.w3.org/publishing/epub33/epub-33, are you saying there's already a redirect on that URL that will put "epub-33" into the hash, so it redirects first here as https://w3c.github.io/epub33/#epub-33 and then this page will grab the hash and do another redirect?

iherman:
> I don't think I follow what's going on. Where are the hashes in the URL coming from that this script picks up on?
> 
> If I put in https://www.w3.org/publishing/epub33/epub-33, are you saying there's already a redirect on that URL that will put "epub-33" into the hash, so it redirects first here as https://w3c.github.io/epub33/#epub-33 and then this page will grab the hash and do another redirect?

Yes (well, almost; it redirects to https://w3c.github.io/epub33/index-page/#epub-33). And it is very convoluted, I know, and it would be more direct to just add a set of boring `.htaccess` lines. The only thing that motivated me to do it this way is that this script can be modified at any time by, e.g., you (if we publish a new document, for example) and you would not depend on me (or a team member) to go back to the `.htaccess` file (which is strictly team access only). 

If you think that is over the top, I am happy to use `.htaccess` directly...

iherman:
(And, b.t.w., we had the same problems with some a11 reports; we may use the same approach to avoid massaging `.htaccess` for those as well.)

mattgarrish:
Ah, okay, that makes sense now. I'm fine with keeping it; I just wasn't sure what the first part of getting URLs here was.

mattgarrish (approved):

PR #2572, opened 2023-07-20
Add "none" as conformance option

### LINKED ISSUE #2569: Publications that don't meet accessibility requirements
The need to differentiate a publication that hasn't been checked for accessibility versus one that doesn't meet the EPUB accessibility standard came up again while trying to work out the display metadata in the CG's accessibility task force editors call.

This sounds like it will be important to know in EU for regulatory purposes, so parsing a summary for claims isn't optimal. @gregoriopellegrino has also indicated that ONIX is updating the "inaccessible" code point value to reflect that such books aren't necessarily inaccessible to everyone (the definition will also apply to books with "known accessibility limitations").

The idea from the call is to recommend/note that authors use the conformance value "none" to indicate this:

```
<meta property="dcterms:conformsTo">none</meta>
```

This could be a requirement for publishing in the EU.

mattgarrish:
This pull request adds text on using "none" as a value of a dcterms:conformsTo property when a publication does not meet accessibility standards.

This fits into the existing note about claiming conformance to WCAG or other standards, as I don't think we should try to write a normative requirement for something we can't enforce. Adding to that note would make it very large, however. Instead, I've changed the note into an informative subsection called "Other accessibility claims".

(Hopefully, changing an informative note to an informative section also keeps us in the scope of a class 2 change rather than a class 3.)

I was also able to work in a link to the exemption property we're defining in the other pull request, but we'll have to publish that first to make the reference valid.

I didn't add a change log entry, as I don't know if calling it a "substantive change" would trigger people to question the classification. Or at this stage do we need to document all changes in the logs @iherman?

Fixes #2569

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -1209,22 +1209,36 @@
 							Recommendation</a>.</p>
 					</div>
 
-					<div class="note">
+					<section class="informative">
+						<h5>Other conformance claims</h5>
+
 						<p>The requirement to include a <code>dcterms:conformsTo</code> identifer does not prevent EPUB
 							publications from conforming to other standards, including other accessibility standards and
 							guidelines (e.g., a specification that covers specific natural language issues). An EPUB
-							publication may have multiple <code>dcterms:conformsTo</code> statements.</p>
+							publication can have multiple <code>dcterms:conformsTo</code> statements.</p>
 
-						<p>EPUB creators may also use a conformance URL, as defined in the <a
+						<p>For example, if an EPUB publication only meets WCAG conformance requirements (i.e., does not
+							fully conform to this specification), EPUB creators can add a <code>conformsTo</code>
+							property with a conformance URL defined in the <a
 								href="https://www.w3.org/TR/WCAG/#conformance-required">Required Components of a
-								Conformance Claim</a> [[wcag2]], with the <code>conformsTo</code> property for EPUB
-							publications that only meet WCAG conformance requirements (i.e., that do not fully conform
-							to this specification).</p>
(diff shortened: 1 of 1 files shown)


mattgarrish:
Forgot the preview and diff links:

- [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/conformsto-none/epub33/a11y/index.html)
- [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/conformsto-none/epub33/a11y/index.html)

avneeshsingh:
Looks good!

iherman:
@mattgarrish let me talk to PlH next week (I will have a 1-1 with him) regarding your editorial questions, @mattgarrish. I _expect_ that if we do a republish with classes 1-2, then we can directly publish as we wish, and I expect it is all right to add a change log as we did. We know that if we publish class 3 we must be careful, because we have to publish a document with the diffs clearly marked (and then reviewed by the AC). I do not know what happens if we publish a new version with classes 1, 2, _**and**_ 3…

iherman:
> (Hopefully, changing an informative note to an informative section also keeps us in the scope of a class 2 change rather than a class 3.)

I would think it is class 2.

I have added a label on this, we will have to do that systematically...

clapierre (approved):

gregoriopellegrino (approved):

PR #2573, opened 2023-07-20
Clarify decimal values are allowed in viewport meta declarations

### LINKED ISSUE #2557: Clarify that viewport meta allows decimal values
As raised in https://github.com/w3c/epubcheck/issues/1481, there are rare occurrences where the value of the viewport meta is expressed with a fractional part.  The current definition, however, says the value must be a positive number, suggesting only positive integers are allowed.

We should probably make clear that decimal values are allowed but reading systems may ignore the fractional part.

mattgarrish:
Adds links to CSS2 definition of "number" for the viewport height and width.

Fixes #2557

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2573.html" title="Last updated on Jul 20, 2023, 7:24 PM UTC (72c568a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2573/0dd3160...72c568a.html" title="Last updated on Jul 20, 2023, 7:24 PM UTC (72c568a)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1340,8 +1340,8 @@
 						<h5>HTML <code>audio</code> and <code>video</code> fallbacks</h5>
 
 						<p id="confreq-resources-cd-fallback-media">[=EPUB creators=] MUST NOT use embedded
-							[[html]]&#160;[=flow content=] within a <a data-cite="html#media-elements">media
-								element</a> (i.e, <a data-cite="html#the-audio-element"><code>audio</code></a> or <a
+							[[html]]&#160;[=flow content=] within a <a data-cite="html#media-elements">media element</a>
+							(i.e, <a data-cite="html#the-audio-element"><code>audio</code></a> or <a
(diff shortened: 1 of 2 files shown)


mattgarrish:
I've updated the pull request and revised the first comment to reflect the changes. All this does now is add a link to the CSS definition of number as additional clarification. These values are defining the CSS ICB, so it's not really changing anything substantive.

iherman (approved):

shiestyle (approved):

rdeltour (changes requested):
I'm confused about which `number` definition should be referenced here.

The PR currently refers to the CSS 2.1 definition, but as EPUB refers to CSS snapshot, isn't this definition overridden by the one in [CSS Values and Units Level 3](https://www.w3.org/TR/css-values-3/#numbers)? (which is being slightly corrected in [Level 4](https://drafts.csswg.org/css-values/#numbers)).

Note by the way that the new definition allows scientific notation (for instance `10e3` is a valid CSS number). Do we know if these are supported in viewport dimensions?

PR #2574, opened 2023-07-21
Clarify epub:type use and restrictions on svgs

### LINKED ISSUE #2555: Clarify epub:type use with embedded SVG
As reported in https://github.com/w3c/epubcheck/issues/1497#issuecomment-1518634823, the specification is not clear that the restriction on where `epub:type` can be used with svg content documents also applies to embedded svg. The restriction should either be in the referenced restrictions section or there should be a separate section for semantics that embedded svg definition can reference.

### LINKED ISSUE #2556: Restriction on epub:type in svg is too strict
As reported in https://github.com/w3c/epubcheck/issues/1497, the current [restriction on epub:type in svg content documents](https://www.w3.org/TR/epub-33/#confreq-svg-structural-semantics) does not allow the attribute on `a` elements.

The definition of [renderable elements](https://www.w3.org/TR/SVG2/render.html#TermRenderableElement) is probably a better alternative than the three categories listed.

(If the current three categories are kept, however, note that they link to the svg 1.1 specification so should be updated.)

mattgarrish:
This largely implements the text in https://github.com/w3c/epub-specs/issues/2555#issuecomment-1630989881 but I tried to further clean up the section to make it more readable.

The big difference is I split the embedding definitions out into a list and included the requirements for each in the bullets. The definitions were hard to read in a single paragraph and having the requirements follow them made the section sound jumpy.

It also makes the fix in issue #2555 to allow the `epub:type` attribute on renderable elements.

Fixes #2555
Fixes #2556

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2574.html" title="Last updated on Aug 8, 2023, 9:01 PM UTC (0d07401)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2574/0dd3160...0d07401.html" title="Last updated on Aug 8, 2023, 9:01 PM UTC (0d07401)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -91,7 +91,9 @@
 			<p>This specification defines the authoring requirements for [=EPUB publications=] and represents the third
 				major revision of the standard.</p>
 		</section>
-		<section id="sotd" class="updateable-rec"></section>
+		<section id="sotd" class="updateable-rec">
+			<p class="proposed correction">Proposed corrections are marked in the document.</p>
+		</section>
 		<section id="sec-introduction">
 			<h2>Introduction</h2>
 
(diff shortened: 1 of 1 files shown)


mattgarrish:
I was looking at the rest of the spec and found a couple of places where we referred to "embedded SVGs" generally when we meant included SVGs.

I also noticed that the svg content documents section requires standalone svg file conformance, so that conflicted with the embedding section referring to svg document fragments for both embedding types. I moved the fragment reference into the inclusion definition.

iherman:
+1 on the content change.

Process question to @plehegar. This is a class-2 change. My understanding is that such change can be made on the document without further ado (publishing the result through the webmaster, that is). However, if this version is _combined_ with another, class-3 change (#2575 in this case), does my remark in https://github.com/w3c/epub-specs/pull/2575#issuecomment-1646519990 concern the class-2 changes as well, i.e., they should also be marked via ins/del elements, or is it o.k. to do direct change on the document for these?

epub33/core/index.html
```diff
@@ -5971,18 +5971,33 @@ <h5>Embedded MathML</h5>
 						data-tests="https://w3c.github.io/epub-structural-tests/#06-content-document_content-document-xhtml.feature_L910,https://w3c.github.io/epub-structural-tests/#06-content-document_content-document-xhtml.feature_L923">
 						<h5>Embedded SVG</h5>
 
-						<p>[=XHTML content documents=] support the embedding of <a
-								href="https://www.w3.org/TR/SVG/conform.html#ConformingSVGXMLFragments">SVG document
-								fragments</a> [[svg]] <em>by reference</em> (embedding via reference, for example, from
-							an <code>img</code> or <code>object</code> element) and <em>by inclusion</em> (embedding via
-							direct inclusion of the <code>svg</code> element in the XHTML content document).</p>
+						<p>[=XHTML content documents=] support the embedding SVG:</p>
```
rdeltour: probably a typo there?
s/support the embedding SVG/support the embedding of SVG/
or
s/support the embedding SVG/support embedded SVG/

epub33/core/index.html
```diff
@@ -5971,18 +5971,33 @@ <h5>Embedded MathML</h5>
...
+							<li id="sec-xhtml-svg-inclusion"><em>by inclusion</em> &#8212; via direct inclusion of an <a
+									href="https://www.w3.org/TR/SVG/conform.html#ConformingSVGXMLFragments">SVG document
+									fragment</a> [[svg]] in an XHTML content document. SVGs embedded by inclusion have
+								the same content conformance constraints as those defined for [=SVG content documents=]
+								in <a href="#sec-svg-restrictions"></a>.</li>
+						</ul>
+
+						<p>Although the XHTML content document definition allows the use of the [^/epub:type^] attribute
+							on [^svg^] elements, the attribute's <a href="#confreq-svg-structural-semantics">use is
+								restricted</a> within an included SVG in the same way as for SVG content documents.</p>
```
rdeltour: nitpicking: elsewhere we now use "SVG embedded by inclusion" instead of "included SVG".

rdeltour (commented):
The change looks OK to me.

About the specific issue of how the `epub:type` is defined for included SVG, I still find the reading a little mazy:

- the "by inclusion" bullet says content conformance constraints are defined by 6.2.3… but not only, wait for the next paragraph.
- we need to make a special case for `epub:type`, by pointing to a bullet in the restrictions for SVG content documents.

What about moving instead the `epub:type` SVG restrictions from the bullet in "6.2.2 SVG Requirements" to a new paragraph in "6.2.3 Restrictions on SVG"?

To summarize, what I'm suggesting is:

- in 6.2.2:

replace 
> MAY include the `epub:type` attribute for expressing structural semantics. When specified, the attribute MUST only be included on structural, shape, or text elements [svg].

by
> MAY include the `epub:type` attribute for expressing structural semantics. When specified, the attribute MUST conform to constraints expressed in 6.2.3 Restrictions on SVG

- in 6.2.3, add a new bullet in the restrictions list:
> the `epub:type` attribute MUST only be included on structural, shape, or text elements [svg].

- in 6.1.4.2, the paragraph "Although the XHTML content (…) in the same way as for SVG content documents." can now be removed.

mattgarrish:
Maybe we can do without mentioning epub:type in 6.2.2. I hate splitting the allowed and restricted parts of the attribute. It's sort of like if we had separate sections in the HTML extensions and constraint subsections to define each part of how you can use epub:type.

SVG already allows foreign namespaced attributes and elements, so we're not actually allowing anything in 6.2.2 that the SVG spec doesn't already allow. The two "MAY" bullets could be expressed as a note without changing anything normative.

I'm going to play around with this a bit more.

rdeltour (approved):
Looks cleaner and more readable to me! 👍

mattgarrish:
@iherman I'm wondering if we should merge this with #2575?

I'm not sure it qualifies as a class 3 change, as we're only taking a couple of unnecessary MAY statements and turning them into a note, but it's going to make a merge conflict because the bullet that 2575 is changing is now in a different section. It might be simpler to make all these epub:type in svg changes together.

mattgarrish:
I've updated this pull request to incorporate the changes that were in #2575.

PR #2576, opened 2023-07-22
Fix wg metadata

mattgarrish:
Only for the three rec-track specs.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2576.html" title="Last updated on Jul 25, 2023, 6:29 PM UTC (d12b913)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2576/0dd3160...d12b913.html" title="Last updated on Jul 25, 2023, 6:29 PM UTC (d12b913)">Diff</a>

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -1874,19 +1874,17 @@
 		<section id="change-log" class="appendix informative">
 			<h2>Change log</h2>
 
-			<p>Note that this change log only identifies substantive changes since <a href="https://idpf.org/epub/a11y/"
-					>EPUB Accessibility 1.0</a> &#8212; those that affect the conformance of <a
-					href="https://www.w3.org/TR/epub/#dfn-epub-publication">EPUB publications</a> or are similarly
-				noteworthy.</p>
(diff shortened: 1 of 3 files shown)


iherman (approved):

mattgarrish:
So does this qualify as a class 2 change, too, since we're modifying the text of the specifications? (i.e., I assume I can't just merge this, either.)

mattgarrish:
In other words, should this PR only update the core and a11y specs because those are the only two we're updating with the other issues, and the RS spec would change it's metadata whenever we have some other change for it?

iherman:
> So does this qualify as a class 2 change, too, since we're modifying the text of the specifications? 

I think so

> (i.e., I assume I can't just merge this, either.)

I do not think so. I believe any merge for rec track should have a more "visible" approval from the WG, or at least from the WG Chairs

cc @w3c/w3c-group-145018-members

mattgarrish:
I'm going to close and re-open this for the two specs we're updating. Seems some unrelated changes got in an earlier commit, and undoing is proving a pain.

PR #2577, opened 2023-07-25
Change respec metadata to PM working group

mattgarrish:
As the title says.

Do we also want to republish the RS spec at the same time just to update the group info?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2577.html" title="Last updated on Aug 9, 2023, 10:06 AM UTC (53748b4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2577/0dd3160...53748b4.html" title="Last updated on Aug 9, 2023, 10:06 AM UTC (53748b4)">Diff</a>

### DIFF
--- a/epub33/a11y/index.html
+++ b/epub33/a11y/index.html
@@ -11,8 +11,8 @@
 		<script class="remove">
 			var respecConfig = {
 				subtitle: "Conformance and Discoverability Requirements for EPUB publications",
-				group: "epub",
-				wgPublicList: "public-epub3",
+				group: "pm",
+                wgPublicList: "public-pm-wg",
 				specStatus: "ED",
 				shortName: "epub-a11y-11",
 				edDraftURI: "https://w3c.github.io/epub-specs/epub33/a11y/",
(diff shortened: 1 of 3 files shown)


shiestyle (approved):

iherman (commented):
If we do that, it should definitely be done for the RS and A11y as well.

Any republication, even class 1 and 2, require an overall WG resolution, though...

iherman (approved):

mattgarrish:
> If we do that, it should definitely be done for the RS

Ya, there are just no other changes to that spec in the bunch we're proposing. But it will eventually have to be done.

mattgarrish:
@iherman I've made a few other tweaks to this PR after looking at the metadata again:

- I set the previousMaturity to REC (from PR)
- I changed the previousPublishDate to the May 25, 2023
- I removed the CR/PR end dates, as I assume those have no meaning now

Since we were saying on the last call that we may merge PRs without publishing for a while, I've also updated the RS spec. Even if we don't publish it with the others, we might as well get this functional stuff out of the way. I'm going to find the PR that modified the change logs and adjust it to use details elements, too (but not add a new section for changes since REC, since we don't have any yet).

iherman:
I think you should also comment out the EPUB format reference. There is no use of regenerating it every time, only when we publish a consolidated version...

Otherwise 👍

PR #2584, opened 2023-09-19
[FXL-A11Y] Switching to HTML format and Adding New Section

wareid:
In preparation for the work to finalize the note on fixed-layout accessibility, I've converted the doc from markdown to HTML and added the sections on reading system accessibility.

### DIFF
--- a/epub33/fxl-a11y/index.html
+++ b/epub33/fxl-a11y/index.html
@@ -9,7 +9,6 @@
 		<script class="remove">
 			//<![CDATA[
             var respecConfig = {
-                format: "markdown",
                 group: "epub",
                 wgPublicList: "public-epub-wg",
                 specStatus: "ED",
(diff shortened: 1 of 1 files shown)


epub33/fxl-a11y/index.html
```diff
@@ -51,522 +50,701 @@
...
+            </ul>
 
-> [EPUB Region-Based Navigation](http://idpf.org/epub/renditions/region-nav/epub-region-nav-20150826.html), introduces conventions for region-based navigation through a visual rendition of a Publication based on Regions of Interest.
+            <section id="legibility-fonts-sizing">
+                <h4>Font Sizing</h4>
 
-<aside class="note">
-EPUB Region-Based Navigation is currently only supported by a few reading systems. Content creators are advised to test any EPUBs that use this feature.
-</aside>
+                <p>There is no font size guideline in [[wcag2]], however the standard default font size in most desktop and mobile browsers is 16pt for body text (I.e. in a <code>p</code> element). This size is sufficient for most content, and headings should be based off of it by using <code>em</code> or <code>rem</code> sizing in [[css]]. If content contains a great deal of text, it is also recommended to consider a larger body font size like 18pt to ensure readability.</p>
```
mattgarrish: The reference to "css" is broken. Probably should be css2.

mattgarrish:
Since you're reformatting it, it might be good to go through and change all the titles to sentence case. Or did that stick as a requirement, @iherman?

Might be good to give it a run through the link checker, too. One of the quotes caught my eye because it talked about removing content from the "accessibility API" when it should be the "accessibility tree". When I clicked through to the referenced MDN page I got a 404.

wareid:
> Since you're reformatting it, it might be good to go through and change all the titles to sentence case. Or did that stick as a requirement, @iherman?
> 
> Might be good to give it a run through the link checker, too. One of the quotes caught my eye because it talked about removing content from the "accessibility API" when it should be the "accessibility tree". When I clicked through to the referenced MDN page I got a 404.

It needs a lot of link checking, thanks for the reminder! It's not been given a hard look, and we still have a lot to write.

iherman (commented):
Some editorial issues (I did not go into the technical content)

- > (@mattgarrish) Since you're reformatting it, it might be good to go through and change all the titles to sentence case. Or did that stick as a requirement, @iherman?

    Yes, indeed, sentence case seems to be the new style requirement.
- Before we forget, we should re-assign this document to the PM WG. This means that, in the respec header, it should now say:
  - `group: "pm"`
  - `wgPublicList: "public-pm-wg`
- In §2.1 there is a link at the end of the text without explanation. In other sections this is preceded by a remark "useful resources" or something similar. Maybe worth having it here as well
- It would be good, in the HTML/SVG examples, to indent the structures of the extracts, like it is done in the spec. Also, maybe have them as "complete" codes for larger examples, ie, with `<html>....</html>` (see, e.g., [Example 1 in the core spec](https://www.w3.org/TR/epub-33/#example-referencing-a-container-resource))
- The reference formats in quote blocks at the start of sections should be consistent (compare §2. vs. §2.2)
- I suspect the `<svg>...</svg>` pair is missing in the last example of §3.3
- For examples in general: it seems that respec prefers the `<aside>` tag instead of the `<div>` tag for the examples. If you also add a `title` attribute the examples all get a title, and the a number with a hyperlink anchor. Makes it easier to refer to later.

wareid:
@mattgarrish @iherman Made the suggested changes and ran the document through link checker to catch any bad links. Is this ok to merge for now?

iherman (approved):

mattgarrish:
It's just a draft, so I'm fine with you merging whenever you want. Other than the CSS ref, the rest were just suggestions.

PR #2591, opened 2023-11-08
Add links to wcag level definitions

### LINKED ISSUE #2589: Add links to level A/AA/AAA definitions
Now that WCAG 2.2 is a recommendation, we can finally link our references to the three levels in the accessibility specs to their definitions.

mattgarrish:
Adds `a` tags to the references in the Accessibility and Techniques documents and fixes a link that only referenced the parent section.

Fixes #2589 

- [Accessibility preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/link-wcag-levels/epub33/a11y/index.html)
- [Accessibility diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/link-wcag-levels/epub33/a11y/index.html)
- [Techniques preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/link-wcag-levels/epub33/a11y-tech/index.html)
- [Techniques diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/link-wcag-levels/epub33/a11y-tech/index.html)

Edit: Fixed the broken diff links from the first post.

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -268,8 +268,8 @@
 					<li><code>textual</code> &#8212; setting this value indicates that the publication is screen-reader
 						friendly (i.e., all the content is available for text-to-speech rendering by an assistive
 						technology). EPUB creators can set this value for EPUB publications that conform to [[WCAG2]] <a
-							href="https://www.w3.org/TR/WCAG2/#cc1">Level A</a> or higher, for example, as there must be
-						textual alternatives for non-text content to meet these thresholds.</li>
+							href="https://www.w3.org/TR/WCAG2/#cc1_A">Level A</a> or higher, for example, as there must
+						be textual alternatives for non-text content to meet these thresholds.</li>
 					<li><code>auditory</code> &#8212; setting this value indicates that pre-recorded audio is available
 						for the publication. EPUB creators can set this value for EPUB publications that provide media
 						overlays for all the content.</li>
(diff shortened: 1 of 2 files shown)


iherman (approved):
Shouldn't we add this to the changes' section? At least for the Rec document...

gregoriopellegrino (approved):

mattgarrish:
> Shouldn't we add this to the changes' section? At least for the Rec document...

We haven't documented class 1 or 2 changes to date. If we start with this one, should we go back and add the other non-substantive changes?

iherman:
> > Shouldn't we add this to the changes' section? At least for the Rec document...
> 
> We haven't documented class 1 or 2 changes to date. If we start with this one, should we go back and add the other non-substantive changes?

Class 1 can be safely ignored. For class 2, I think it is a bit of a case-by-case basis. At first, I had the feeling that it is better to add this one, although I am not that sure any more... :-)

mattgarrish:
> Class 1 can be safely ignored.

Isn't this a class 1 change?

We're only adding a hyperlink to existing text -- and it doesn't alter the meaning of that text -- so I assume it falls into the "No changes to text content" bucket.

iherman:
> > Class 1 can be safely ignored.
> 
> Isn't this a class 1 change?
> 
> We're only adding a hyperlink to existing text -- and it doesn't alter the meaning of that text -- so I assume it falls into the "No changes to text content" bucket.

Heh, just for the sake of arguments... I could argue that it adds info to the text because it now opens up the hyperlinks to the reader, which was not there before... Who knows where the boundaries are? But I think we all have better things to do than to discuss these minute things...😀

mattgarrish:
> But I think we all have better things to do than to discuss these minute things...

Ya, my perspective on the change log is that it should only contain things people need to know. That you can now link directly to the level definitions instead of going through the wcag reference doesn't affect anyone's understanding of the specification. Saves you one whole click.. 😄

clapierre (approved):
Looks good

mattgarrish:
Merging this one, as well, as these are just links.

PR #2598, opened 2023-12-07
Add note and replace uri reference

### LINKED ISSUE #2594: URI mention without reference
[Section 5.5.3.1](https://www.w3.org/TR/epub-33/#sec-opf-dcidentifier) says:

> EPUB creators MAY specify additional identifiers. The identifiers should be fully qualified URIs.

But we dropped URI references during 3.3 in favour of URL.

I think this could be rewritten as:

> EPUB creators MAY specify additional identifiers. The identifiers should be [absolute-url strings](https://url.spec.whatwg.org/#absolute-url-string) [url].

Since it's an informative recommendation, this change shouldn't rise above class 2, at least.

mattgarrish:
This PR splits the sentence about using URIs for identifiers off into a separate note, fixes it to reference absolute-url strings, and adds some explanation for why these are preferable.

Looking at the Process document, clarifications of non-normative text - which is all this is - are called out in the class 2 description, so I think we're fine with the current issue categorization.

Fixes #2594

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2598.html" title="Last updated on Dec 7, 2023, 2:41 PM UTC (1315407)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2598/932fe2c...1315407.html" title="Last updated on Dec 7, 2023, 2:41 PM UTC (1315407)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3836,8 +3836,16 @@
 							identifiers when making minor revisions such as updating metadata, fixing errata, or making
 							similar minor changes.</p>
 
-						<p>EPUB creators MAY specify additional identifiers. The identifiers should be fully qualified
-							URIs.</p>
+						<p>EPUB creators MAY specify additional identifiers.</p>
+
+						<div class="note">
+							<p>EPUB creators are advised to use <a
+									href="https://url.spec.whatwg.org/#absolute-url-string">absolute-URL strings</a>
+								[url] for identifiers whenever possible. The inclusion of a domain owned by the EPUB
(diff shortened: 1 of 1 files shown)


iherman (approved):

shiestyle (approved):

mattgarrish:
Let's give this until the middle of next week for feedback and then merge it before the holidays set in.

mattgarrish:
Looks like this got forgotten. Going to merge it now.

PR #2599, opened 2023-12-08
Update copyright links

mattgarrish:
The accessibility guidelines failed to build/deploy because the URLs used in the copyright have changed: https://github.com/w3c/epub-specs/actions/runs/7145045877/job/19459898614#step:3:720

This PR will fix the links, but I'm concerned that because the file is in /common it will trigger all the notes to be republished. Do we just live with that?

### DIFF
--- a/epub33/common/copyright.html
+++ b/epub33/common/copyright.html
@@ -1,5 +1,5 @@
 <p class="copyright">
-	<a href="https://www.w3.org/policies/#copyright">Copyright</a> © 1999-2023 
+	<a href="https://www.w3.org/policies/#copyright">Copyright</a> © 1999-%thisyear%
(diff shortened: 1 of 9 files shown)


iherman:
> This PR will fix the links, but I'm concerned that because the file is in /common it will trigger all the notes to be republished. Do we just live with that?

I think this is long overdue, actually. Not only because of the copyright link, but also to make these documents explicitly "owned" by the PM WG and not the EPUB3 WG any more.

Note that I have committed a change on this branch making the header changes for all documents. (Echidna reported an error on the first run for almost all of the documents, except the ones that were recently changed. They are all o.k. now.)

iherman (approved):

mattgarrish:
I'm going to go ahead and merge this and it only updates the prelim for the documents.

PR #2600, opened 2023-12-14
[FXLA11Y] Re-adding font section and adding WCAG success criteria

wareid:
As discussed in the last meeting, this PR re-adds the font selection recommendations. This PR also begins the work of restructuring the document to include the success criteria as they are in the legibility section.

### DIFF
--- a/epub33/fxl-a11y/index.html
+++ b/epub33/fxl-a11y/index.html
@@ -65,7 +65,7 @@
 
             <p>EPUB fixed layout (FXL) publications, or publications where the print layout is preserved in the digital edition, have been around since before EPUB 3.0.1. These publications span a number of genres and types, from comic books, cook books, children's books, and more.</p>
 
-            <p>An accessible fixed layout EPUB file is one that meets the accessibility requirements of the Web Content Accessibility Guidelines 2 level AA[[wcag2]] and EPUB Accessibility 1.1 [[epub-a11y-11]], including color contrast, reading order, font layout, structural navigation, metadata, and text alternatives.</p>
+            <p>An accessible fixed layout EPUB file is one that meets the accessibility requirements of the Web Content Accessibility Guidelines 2 level AA [[wcag2]] and EPUB Accessibility 1.1 [[epub-a11y-11]], including color contrast, reading order, font layout, structural navigation, metadata, and text alternatives.</p>
 
             <p>The main motivation behind creating fixed layout publications is the need to preserve the print layout of the book, either because of it's importance to the text (i.e. complex diagrams) or it's artistic purpose (i.e. illustrated text). However, these publications are often partially or completely inaccessible to people with print disabilities.</p>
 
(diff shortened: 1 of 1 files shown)


epub33/fxl-a11y/index.html
```diff
@@ -347,6 +378,40 @@ <h4>Non-text content or images of text</h4>
...
+                <ul>
+                    <li>Font sizing</li>
+                    <li>Font weight</li>
+                    <li>Font face or style</li>
+                </ul>
+
+                <section id="legibility-fonts-sizing">
+                    <h5>Font sizing</h5>
+
+                    <p>There is no font size guideline in WCAG [[wcag2]], however the standard default font size in most desktop and mobile browsers is 16pt for body text (i.e. in a <code>p</code> element). This size is sufficient for most content, and headings should be based off of it by using <code>em</code> or <code>rem</code> sizing in CSS [[css2]]. If content contains a great deal of text, it is also recommended to consider a larger body font size like 18pt to ensure readability.</p>
```
mattgarrish: I'm not sure it's accurate to say there isn't a font size requirement. 1.4.4 says that text needs to be able to resize to 200% of its default size without loss of readability or functionality.

Reading systems that run on tablets or similar devices should let you zoom fxl pages, but desktop-based apps don't, at least not when I last checked (which was some time back). So long as the text scales with the page on the RSes that do zoom, this SC won't be an issue. You can expect that users who need the functionality will seek out those RSes. But if reading systems don't scale the text, and you can't modify the text properties as a user, then there'd be a problem.

(This also leads into the ever problematic reflow SC where zooming fxls doesn't reflow the text as it enlarges, so you end up having to scroll to read lines.)

mattgarrish: And to be clear, I'm not expecting this to be changed before you merge. Just wanted to mention it.

gregoriopellegrino:
Thanks for the great work done and sorry for the late analysis: I have had a bad flu the past few days.

The section on readability -- net of Matt's comments -- looks fine to me. Perhaps we could add a note to section _2.5.4 Font selection_ to say that this section is not normative, even though the whole document is not, but I would like to say that these are recommendations, not tied to international guidelines.

With respect to the linkage to the WCAG Success Criteria, it seems to me a very good job. I had also envisioned a reorganization of the content, following the structure of the SCs:

1. Perceivable
1.1 Text Alternatives
1.2 Time-based Media
1.3 Adaptable
1.4 Distinguishable
2. Operable
2.1 Keyboard Accessible
2.2 Enough Time
2.3 Seizures and Physical Reactions
2.4 Navigable
2.5 Input Modalities
3. Understandable
3.1 Readable
3.2 Predictable
3.3 Input Assistance
4. Robust
4.1 Compatible

But maybe this is a bit too much, I would leave it to the group discussion.

In the meantime, thank you for the work!

gregoriopellegrino (approved):

epub33/fxl-a11y/index.html
jonaslil: The default font size in browsers is 16px, not 16pt, as far as I know.

Another thing I thought about:

If the RS allows scaling the page, the text size in a fixed layout book will be influenced by two things, when read on a smaller screen:

1. font size setting
1. page size and/or line length

The larger the page, the smaller the text will be when the page is scaled down to fit the viewport. If the lines flow out of the visible area, so that you have to scroll horizontally, the text will be very difficult to read.

wareid:
See:

    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/fxl-editorial-3/epub33/fxl-a11y/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/fxl-a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/fxl-editorial-3/epub33/fxl-a11y/index.html)

gregoriopellegrino (approved):
Great improvement for me, thanks!

PR #2601, opened 2023-12-19
Removed Luc as editor for the eu mapping

iherman:
The republication of the Note was rejected, because each editor must be member of the WG. To get the publication through this Thursday, I have created a new field for "authors", with Cristina, Luc, and Gregorio, and the "editors" field only refers to Cristina and Gregorio.

We can modify this later if we want, keeping the publication constraint in mind. But this setup reflects the re-publication of the Note under the PM WG's auspices.

### DIFF
--- a/epub33/epub-a11y-eaa-mapping/index.html
+++ b/epub33/epub-a11y-eaa-mapping/index.html
@@ -16,7 +16,21 @@
 				edDraftURI: "https://w3c.github.io/epub-specs/epub33/epub-a11y-eaa-mapping/",
 				errata: "https://w3c.github.io/epub-specs/epub33/errata.html",
                 copyrightStart: "2021",
-				editors:[
+				editors: [
+					{
+						name: "Cristina Mussinelli",
+						company: "Fondazione LIA",
+						companyURL: "https://www.fondazionelia.org",
+						w3cid: 66100
+					},
+					{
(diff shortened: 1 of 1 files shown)


mattgarrish (approved):

gregoriopellegrino:
Thank you for this proposal. We think that for historical matters we should keep track and mention that Luc was co-editor of the first version. Is there any way to do this? Do we have similar situations in other notes?

iherman:
@gregoriopellegrino 

I have just realized that respec has an undocumented feature of "formerEditors"; it behaves like authors, ie, it can list as a number of persons as part of the header

I am happy to modify this. The current PR says:

Editors: Cristina and you
Authors: Cristina, Luc and you

We can do a

Editors: Cristina and you
Former editors: Cristina, Luc and you

or

Editors: Cristina and you
Former editors: Luc

Which one would you prefer?

As for the other question: pubrules checker would have shouted at me if there was any problem with any of the other documents.

gregoriopellegrino:
We prefer:

Editors: Cristina and you
Former editors: Luc

iherman:
Done. 

See the preview: https://cdn.statically.io/gh/w3c/epub-specs/moving-Luc/epub33/epub-a11y-eaa-mapping/index.html

gregoriopellegrino (approved):

PR #2604, opened 2024-01-11
[FXLA11Y] Adding Reading System Guidance 

wareid:
Adding Gautier's section on reading system guidance. I made some editorial changes, and have opted not to include some sections that were in the original document. 

I think for now we need to avoid mentioning features that don't yet exist (converting FXL to reflow), as I don't want to confuse publishers or any other readers (like the EU commission). I do think the ideas and concepts mentioned are worth keeping and exploring elsewhere. 

Open to discussing it though!

### DIFF
--- a/epub33/fxl-a11y/index.html
+++ b/epub33/fxl-a11y/index.html
@@ -675,42 +675,130 @@ a textual description to aid in accessibility.
         <section id="rs-a11y-intro">
             <h3>Introduction</h3>
 
-            <p>An EPUB reading system can take many forms. It might have a visual display area for rendering the content to users, for example, or it might only provide audio playback of the content. Therefore, there is no single set of rules that applies to all reading systems.</p>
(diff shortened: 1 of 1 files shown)


gautierchomel:
This is a [preview link](https://rawcdn.githack.com/w3c/epub-specs/69c6641fef434c31aa4982ec9b8fbdaf4c0e3bbd/epub33/fxl-a11y/index.html#rs-a11y).

gautierchomel:
Adding here to keep track, more feedback should be collected on the originally proposed following section: 

### Display transform options

The best way to provide the user with the ability to modify the display of textual content (e.g., to change the font family, font size, line height and word spacing) is to extract and display the content as text. It comes down to exposing the content sent to the speech synthesis through TTS.

All styling options provided in the authored content should be ignored as positionings and absolute length units (e.g., pixels and points) may lead to superimposed content.

PR #2605, opened 2024-01-12
[FXLA11Y] Media Overlays Section

wareid:
Adding the text provided by Susan for the Media Overlays section.

### DIFF
--- a/epub33/fxl-a11y/index.html
+++ b/epub33/fxl-a11y/index.html
@@ -479,33 +479,33 @@
             </section>
         </section>
 
-        <section id="media-overlays">
-            <h3>Media overlays</h3>  
+        <section id="media">
+            <h3>Media in fixed layout</h3>  
             
-            <p>Synchronized Media. This can be audio or video that is synchronized within the EPUB. This format can be used to present information or with interactive components that are based on time.</p>
-            
-            <p>Time-based media is prerecorded media that has time as a component, such as an audio recording, a video, or audio or video combined with interaction. Source: <a href="https://accessibleweb.com/question-answer/what-is-synchronized-media/">Accessible Web</a></p> 
-            
-            <section id="media-text-alternatives">
-                <h4>Media Alternatives for Text</h4>
-                
-                <p>This section refers to media that presents no more information than is already presented in text (directly or via text alternatives).</p>
-                
-                <p>EPUB creators must ensure that all text content is available in audio.</p>
-                
-                <aside class="note">EPUB creators do not have to include synchronized text-audio playback in their EPUB publications, only ensure it conforms to these requirements when present. </aside>
-            
-                <note><em>Add how to specify media alternative for text?</em></note>
-            </section>
+            <p>Media content such as audio, video, and interactive elements present both an opportunity and a challenge for fixed layout content. When media content is implemented with accessibility in mind, it can greatly improve the reading experience of a book. Done poorly, it can be a major barrier to access.</p>
 
-            <section id="media-playback">
-                <h4>Synchronized text-audio playback</h4>
+            <p>Media content can be incorporated into fixed layout EPUBs in a variety of ways. When working with media content in EPUB, it is important to consider WCAG [[wcag2]] criteria relating to media accessibility:</p>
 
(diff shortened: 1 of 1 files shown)


sueneu:
Transposed letters:
<li>Ebook creators must ensure auditory playback for epub navigation. This gives reading systems the ability to **sue** auditory labels for links.</li>
<li>Ebook creators must ensure auditory playback for epub navigation. This gives reading systems the ability to **use** auditory labels for links.</li>

gautierchomel:
[Preview](https://rawcdn.githack.com/w3c/epub-specs/f3c33af54a8577b8ed3789dadc1fb9ec72907129/epub33/fxl-a11y/index.html#media-overlays)

wareid:
    * [Preview](https://raw.githack.com/w3c/epub-specs/fxl-media-overlays/epub33/fxl-a11y/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/fxl-a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fxl-media-overlays/epub33/fxl-a11y/index.html)

sueneu:
Here are the comments from my colleague who specializes in media overlays for children's picture books. He's pretty exasperated with the current state of accessibility in fxl. 

`EPUB creators must ensure that all text content is available in audio.`
_Feedback: Better to describe what the EPUB should have in it than to assign a task to a specific role. Not all EPUB creators are the content creators._ 
**Propose change to:**
`All text within the EPUB must be available in audio`

_Feedback: "All text" available as audio overlays is unrealistic/unreasonable. Asking publishers to create custom audio for things like text in images like a talk bubble, graphical text treatments, and text in logos on the copyright page would be too high a burden. He suggests levels of audio availability. 1 = main text available as audio, 2 = main text + captions, front, and back matter  3 = main text + captions, front matter, back matter + word bubbles and graphical text treatments. TTS is not within an EPUB creator's control._
**Question for the group:** Is a combination of recorded audio/overlays for the main story with TTS for the copyright page and back matter acceptable in this case? Who decides that?

`In order to offer users greater control over content presentation, EPUB creators need to add structure and semantics so that the reading system has the necessary context to enable this type of user experience. With greater context, a reading system can provide the ability to skip past secondary content that interferes with the primary narrative and escape users from deeply nested structures like tables.`
_Feedback: Children's picture books are rarely this complex. Tables are rarely used and semantic markup is never used in them. In picture books, there is little text and children skip by simply turning a page_
**Suggestion:** since children's picture books with small amounts of text are only one type of fxl title, let's keep the reference to tables and deeply nested structures. We could continue with EPUB attributes without assigning roles.
**Propose Change to:**
`An accessible fixed layout EPUB should offer users greater control over presentation. Users should be able to skip past secondary content that interferes with the primary narrative and escape deeply nested structures like tables. `

`Ebook creators must ensure auditory playback for epub navigation. This gives reading systems the ability to use auditory labels for links.`
_Feedback: No, reading systems must ensure that they use already-standardized requirements, eg “the <title> tag must contain useful information,” to accomplish this on the e-reader side._
**Suggestion:** This is another good place to describe attributes of the EPUB without assigning tasks or responsibilities in this document.
**Propose change to:**
`An accessible EPUB must have auditory navigation`

_**General Feedback:** The encouragement of cells of reflowable content within the fixed-layout structure would solve a lot. It would allow tables to be run as tables, not broken into individual words over a JPEG of the table lines, and also make adding “aside” or “table” or whatever more possible._

_Current accessibility in fixed layout picture books is, and there’s no other word, pathetic. Fix the actual issues instead of dreaming up new features, half of which don’t apply to fixed layout EPUBS. 
The best workaround for the inability of screen readers to handle a big block of words with individual spans without breaking the audio playback is to include aria-hidden=”true” on each word (so the screen reader does NOT read it) and then use role=”image” and aria-label=”text of the paragraph goes here” in a containing div, so the reading system reads that instead. The reading system then says “Image!” and reads the text with much less issue._

_Any level of synchronized audio should be considered a level of accessibility feature._

PR #2608, opened 2024-03-18
[FXLA11Y] Clean-up based on feedback

wareid:
Based on the feedback I received from everyone, here is some clean up of the document!

I have also added an acknowledgements section, if I am missing any names, please let me know! 

See:

    * [Preview](https://raw.githack.com/w3c/epub-specs/fxl-a11y-editorial-march24/epub33/fxl-a11y/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/fxl-a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fxl-a11y-editorial-march24/epub33/fxl-a11y/index.html)


gregoriopellegrino:
Some of my pointed comments:
- 1.3.1 Perceivable:
    - the Success Criteria for multimedia (audio/video) are not mentioned, but in the multimedia section they are present
    - 1.4.10 Reflow is mentioned: but further down it is indicated that it is virtually impossible to meet this requirement
- 2.2.1 Altering the reading order: it seems to me that the content is not related to altering the reading order
- 2.3 Images in fixed layout: perhaps there is a missing point about decorative images?
- in some places in the document the Success Criteria are mentioned with the level in parentheses, in other places without level, maybe it is better to unify?
- In 2.5.1 Visual adjustments:
    - it says "this content is exempt from some of these requirements" I don't know if "exempt" is the best term, maybe say that in their current state they cannot be met?
    - as for Orientation I would go more decisive: it is not recommended to indicate a fixed orientation
- to section 2.5.2.1 Relating content through placement maybe we could add a note clearly stating that these are non-mandatory suggestions? I'm afraid we'll get feedback saying we're "limiting creativity"
- in the third paragraph of 2.7 Tables we talk about using alt-text and ARIA to make a table image accessible. What is meant by this?
- 4.1 Accessibility metadata for a children's picture book:
    - I would suggest changing the title to "Common accessibility metadata for a children's picture book"
    - the accessibility summary in my opinion should be revised: for EPUB Accessibility 1.1 should not contain information already in machine-readable metadata
- 4.2 Accessibility metadata for a manga EPUB: I would suggest changing the title to "Common accessibility metadata for a manga EPUB"
- 5.3 Reading system accessibility requirements: it seems to me there is a mix of "mandatory" requirements from WCAG and EPUB Accessibility and EPUB 3.3, while others are features that might help the user, but not "mandatory"... Perhaps it would be better to distinguish them?
- 5.3.6 Enable access to alternative text and extended descriptions: what is meant?
- 5.3.9 Support fallbacks: what is meant by fallbacks for MathML?

PR #2625, opened 2024-05-30
Enabling echidna for fxl a11y

iherman:
Secret for echidna has been set, the PR adds the action file

### DIFF
--- a/.github/workflows/epub-fxl-a11y
+++ b/.github/workflows/epub-fxl-a11y
@@ -1,0 +1,27 @@
+# .github/workflows/epub-multi-rend-11.yml
+name: CI (EPUB FXL A11Y)
+on:
+  pull_request:
+    paths: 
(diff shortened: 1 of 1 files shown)


.github/workflows/epub-fxl-a11y
```diff
@@ -0,0 +1,27 @@
+# .github/workflows/epub-multi-rend-11.yml
```
mattgarrish: I don't think this does anything, but still...

```suggestion
# .github/workflows/epub-fxl-a11y.yml
```

mattgarrish (approved):

iherman:
Oops... thanks @mattgarrish

iherman:
@wareid ready to merge this? Means that any merge on fxl a11y will yield a new version of the note...

wareid (approved):

PR #2628, opened 2024-06-20
Updated techniques

### LINKED ISSUE #2621: Recommendation to use epub:type for page breaks
The techniques document for page breaks says:

> It is recommended that both semantics be applied to EPUB 3 content to ensure maximum compatibility with [reading systems](https://www.w3.org/TR/epub/#dfn-epub-reading-system) and [assistive technologies](https://www.w3.org/TR/epub-a11y-11/#dfn-assistive-technology).

https://www.w3.org/TR/epub-a11y-tech-11/#page-001

But as raised in https://github.com/daisy/ace/issues/414, is there a specific reason for this advice? I can't remember anymore if we put this in to deal with a known reading system or if we just weren't sure if some reading system somewhere might be making use of the semantic. As mentioned in the other issue, I don't know of any reading systems that do anything with the epub:type semantic.

Should we change this advice to require the aria role and make the epub:type optional for reading system compatibility? That would be more in keeping with not using epub:type for accessibility.

### LINKED ISSUE #2626: Update accessibility metadata guidance
There have been a few changes to the accessibility metadata since we last published the techniques. We need to update the document to note:

- that there are now individual unknown hazard values
- that the `pageNavigation` feature should be used when a page list is included
- that the `pageBreakMarkers` feature replaces `printPageNumbers`
- that the `a11y:pageBreakSource` property should now be used instead of the `source-of` refinement property.

### LINKED ISSUE #2627: Remove restriction on page numbers as text
The techniques document currently includes this text under the page break markers section:

> Do not include the page number as text content, as this practice forces the user to hear it announced wherever it occurs (e.g., without any context in the middle of a sentence).

While the concern is probably still valid, this seems too strongly worded. I think we can warn about the issue of page numbers being confusing when within the text without a strong ban, which I'm not sure is even followed.

mattgarrish:
This pull request updates the techniques to reflect the variety of changes over the last year, as detailed in issues #2621, #2626, and #2627.

I also made a minor editorial change to make the section ids more easily referenceable. I've kept the old ids for now on spans, but ideally these should disappear in time. (They used to match text identifiers we had for each technique but then we dropped those identifiers in 1.1.)

I'm going to give the document another pass to check that the other guidance we have still makes sense.

Fixes #2621 
Fixes #2626 
Fixes #2627 

- [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/technique-updates/epub33/a11y-tech/index.html)
- [diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/technique-updates/epub33/a11y-tech/index.html)

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -14,7 +14,7 @@
 				shortName: "epub-a11y-tech-11",
 				noRecTrack: true,
 				edDraftURI: "https://w3c.github.io/epub-specs/epub33/a11y-tech/",
-                previousPublishDate: "2021-07-12",
+                previousPublishDate: "2024-04-12",
 				errata: "https://w3c.github.io/epub-specs/epub33/errata.html",
                 previousMaturity: "NOTE",
                 copyrightStart: "2017",
@@ -77,6 +77,13 @@
 						"href": "http://idpf.org/epub/20/spec/OPF_2.0.1_draft.htm",
 						"date": "04 September 2010",
 						"publisher": "IDPF"
+					},
+					"page-source-id": {
+						"title": "Page Source Identification",
+						"href": "https://www.w3.org/publishing/a11y/page-source-id/",
+						"editors": [
+							"Matt Garrish"
+						]
 					},
 					"svg": {
 						"title": "SVG",
@@ -176,7 +183,8 @@
 		<section id="sec-discovery">
 			<h2>Discovery metadata techniques</h2>
 
-			<section id="meta-001">
+			<section id="accessMode">
+				<span id="meta-001"></span>
 				<h3>Identify primary access modes</h3>
 
 				<p>An access mode is defined as a "human sense perceptual system or cognitive faculty through which a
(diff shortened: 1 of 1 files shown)


clapierre (approved):
Great clarifications and improvements Matt.

mattgarrish:
The last commit changes this sentence in the hazards section:

>  Individual unknown hazards must  be definitively determined, report  specified alongside "no hazard" values. 

This isn't what I meant to say, as you could have hazard or no hazard values along with the individual unknowns. What I wanted to convey is that when setting individual hazards, all three need to be covered. I've replaced that sentence with this at the end of the section:

> EPUB creators must ensure that information about all three types of hazards is included when not using the "<code>unknown</code>" or "<code>none</code>" values.

I also made a minor tweak to list each no hazard and unknown hazard value so you don't have to go searching in the vocabulary.

gregoriopellegrino (approved):

gregoriopellegrino (approved):

mattgarrish:
As we're heading into summer break time for a lot of people, it's probably worth merging this now as there hasn't been any additional feedback. We can always announce the changes to the lists if we want to broadcast the updates a bit further.

PR #2640, opened 2024-07-15
Switch from infra to xml definition of whitespace for viewport meta

### LINKED ISSUE #2637: Invalid use of infra definition of whitespace
As noted in #2636, we mix the xml and infra definitions of whitespace and whitespace handling for the viewport meta tag value in [this paragraph](https://www.w3.org/TR/epub-33/#app-viewport-meta-syntax):

> The authoring requirements in this section apply after [whitespace normalization](https://www.w3.org/TR/xml/#AVNormalize) [[xml](https://www.w3.org/TR/epub-33/#bib-xml)] (i.e., after a reading system strips leading and trailing whitespace and compacts all instances of multiple whitespace within the attribute to single spaces). EPUB creators MAY include any valid [ascii whitespace](https://infra.spec.whatwg.org/#ascii-whitespace) [[infra](https://www.w3.org/TR/epub-33/#bib-infra)] in the authored tag so long as the result is valid to this definition.

But the infra definition allows form feed while xml's does not. If you put a form feed character in an xhtml document, it will cause the parser to halt. You can try and twist the definition that form feed is not a "valid" whitespace character, so it's excluded, but why even refer to a definition with invalid whitespace characters? The infra reference only works for html. Since we refer to xml's definition of normalization everywhere else in that section, we should be consistent and refer to [xml's whitespace definition](https://www.w3.org/TR/xml/#NT-S), too.

(The other concern I mentioned with [package document values](https://www.w3.org/TR/epub-33/#sec-metadata-values) is probably fine as is. The two references to infra functions in that section - for leading and trailing whitespace removal and collapsing whitespace in a string - don't change the rules of whitespace authoring in xml. They just add an extra character for processing that isn't ever going to appear. Since xml passes through whitespace in elements, we need some definition for this handling.)

mattgarrish:
If we agree to merge this, I guess I'll have to go back and configure it with the necessary class 3 markup/descriptions. That's assuming we can't justify it as a class 2 change of reference only -- based on form feed not being a "valid" xml whitespace character in the old wording. (It's not like you could use form feed without halting xml processing even if you found some odd reason to author it, regardless of what the spec seemed to be suggesting, so there's a zero likelihood it's ever been used in content.)

Fixes #2637

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2640.html" title="Last updated on Jul 16, 2024, 3:50 PM UTC (5d23b86)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2640/cf1af14...5d23b86.html" title="Last updated on Jul 16, 2024, 3:50 PM UTC (5d23b86)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -11011,8 +11011,9 @@ html.my-document-playing * {
 				<p>The authoring requirements in this section apply <em>after</em>
 					<a data-cite="xml#AVNormalize">whitespace normalization</a>&#160;[[xml]] (i.e., after a reading
 					system strips leading and trailing whitespace and compacts all instances of multiple whitespace
-					within the attribute to single spaces). EPUB creators MAY include any valid [=ascii whitespace=]
-					[[infra]] in the authored tag so long as the result is valid to this definition.</p>
+					within the attribute to single spaces). EPUB creators MAY include any <a data-cite="xml#NT-S"
+						>whitespace characters</a> [[xml]] in the authored tag so long as the result is valid to this
+					definition.</p>
 
 				<p>There are no restrictions on any other attributes allowed on the <a data-cite="html#the-meta-element"
 							><code>meta</code></a> element by the [[html]] grammar.</p>
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -11008,12 +11008,33 @@ html.my-document-playing * {
 						href="#viewport.ebnf.sep-char">separator characters</a> or the <a href="#viewport.ebnf.assign"
 						>assignment character</a>.</p>
 
+				<div class="proposed correction" id="change_4">
+					<span class="marker">Proposed Correction 4</span>
+					<p data-cite="svg2">The following paragraph mixed in a reference to the [[infra]] specification's
+						definition of whitespace characters, which is only valid for HTML documents. This would have
+						allowed the Form Feed character in <code>viewport meta</code> tags even though Form Feed is not
+						valid XML whitespace character. This change corrects the reference to point to the [[xml]]
+						definition and adds an explanatory note about the difference between the two definitions. For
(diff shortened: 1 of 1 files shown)


iherman (approved):
Oops, sorry, I did not realize you already submitted this PR...

See also my additional note in https://github.com/w3c/epub-specs/issues/2637#issuecomment-2230108179.

iherman:
> I guess I'll have to go back and configure it with the necessary class 3 markup/descriptions. That's assuming we can't justify it as a class 2 change of reference only

Unfortunately, yes. Although I doubt it happens in practice, in theory currently valid EPUB instances may become invalid, which does mean class 3...

mattgarrish:
That last commit adds the note, a change log entry, and formats the proposed correction.

Going off on a bit of a tangent, but should we try and publish an update to the specs before TPAC? Not sure how long that process requires, but we've already kind of overshot our original plan of publishing changes somewhat frequently.

iherman:
> That last commit adds the note, a change log entry, and formats the proposed correction.

👍

> 
> Going off on a bit of a tangent, but should we try and publish an update to the specs before TPAC? Not sure how long that process requires, but we've already kind of overshot our original plan of publishing changes somewhat frequently.

You can see the transition steps here: https://www.w3.org/Guide/transitions/?profile=REC&rec=proposed-corrections (I presume you have access to it).

TL;DR it is a not-trivial documentation, with an AC round, etc, due to the class 3 changes. The problem is that the internal approval cycle, as well as the availability of the AC slows all this down.

My personal take is to start the process at or right after TPAC; alternatively, do it at the beginning of September. I do not know what the chances are to get into new class 3 features during TPAC itself...

@wareid @shiestyle @TzviyaSiegman wdyt?

shiestyle (approved):
I agree with Ivan, let's start the process from September and explanation at TPAC if we can.

PR #2642, opened 2024-07-30
Add IDs to EAA exemption value definitions

mattgarrish:
While looking to [update the schema.org crosswalk](https://github.com/w3c/a11y-discov-vocab/pull/115), we noticed that the individual values don't have linkable IDs. Currently, you can only link to the section that contains the table. This pull request fixes that.

- [preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/exemption-ids/epub33/a11y-exemption/index.html)

### DIFF
--- a/epub33/a11y-exemption/index.html
+++ b/epub33/a11y-exemption/index.html
@@ -204,7 +204,7 @@
 						<th>Definition</th>
 					</tr>
 					<tr>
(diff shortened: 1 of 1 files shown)


clapierre (approved):
Great now more work for me to do :)

Issue #2645, opened 2024-08-22
Wrong link to technique in EPUB Accessibility 1.1 note


gregoriopellegrino:
I realized that in the note in [3.4.2.3.1 Completeness](https://www.w3.org/TR/epub-a11y-11/#sec-sync-complete) the name of the link is **Ensuring complete text coverage**, but the link leads to **5.2.4 Identifying escapable structures**, the correct link should be https://www.w3.org/TR/epub-a11y-tech-11/#sync-coverage

iherman:
Will you provide a PR, @gregoriopellegrino, or should I do it?

gregoriopellegrino:
I'll do it

mattgarrish:
Closing this with https://github.com/w3c/epub-specs/pull/2646

PR #2646, opened 2024-08-26
Fixing link in EPUB A11y 1.1

gregoriopellegrino:
As discussed in issue #2645


iherman (approved):

mattgarrish:
This change is fine, but if we're fixing this one link we should probably go ahead and update the others. We changed the techniques document to use more understandable IDs, so while the old numbered IDs are still in there for compatibility with old links, it would be good to stop using them. I can make an update to this PR with the fixes.

mattgarrish (approved):

PR #2648, opened 2024-09-10
Update the generic issue template

iherman:
Let us see if this works...

Using this template it may be possible to add a separate template for [BUG] or [FEATURE REQUEST]; with the volume of issues I am not sure if it is worth doing...

(Wendy, Shinya: we regularly get issues that are irrelevant for this repo, and we have to close. Matt and I wondered if it helps to set up an issue template to deter people from doing so; this is this PR. Note: I have never created issue templates, I hope it will work...)

### DIFF
--- a/.github/ISSUE_TEMPLATE/general-issue-template.md
+++ b/.github/ISSUE_TEMPLATE/general-issue-template.md
@@ -28,4 +28,13 @@ Some possible alternative locations get help to your problems:
 * the [epub forum](https://www.mobileread.com/forums/forumdisplay.php?f=179) at mobileread
 * the [W3C publishing community group](https://github.com/w3c/publishingcg/issues)
 * the [ebooks section](https://ebooks.stackexchange.com/questions/tagged/epub) at stackexchange
+## Which standard or note is this issue related to?
+
+Please include the title and version number. For bug reports, include the relevant section.
+
+## Describe the problem? 
+
+What is not working? What would you like to see improved? For bug reports, cite the relevant text from the specification.
+
+## Describe the fix or new feature you propose?
 
--- a/.github/ISSUE_TEMPLATE/general-issue-template.md
+++ b/.github/ISSUE_TEMPLATE/general-issue-template.md
@@ -1,0 +1,25 @@
+---
+name: General issue template
+about: Issue on the EPUB 3.3 specifications or related Working Group Notes
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+This repository is dedicated to the maintenance of the following W3C documents:
+
+* [EPUB 3 Overview](https://www.w3.org/TR/epub-overview-33/), WG Note
+* [EPUB 3.3](https://www.w3.org/TR/epub-33/), Recommendation
+* [EPUB Reading Systems 3.3](https://www.w3.org/TR/epub-rs-33/), Recommendation
+* [EPUB 3 Multiple-Rendition Publications 1.1](https://www.w3.org/TR/epub-multi-rend-11/), WG Note
+* [EPUB 3 Text-to-Speech Enhancements 1.0](https://www.w3.org/TR/epub-tts-10/), WG Note
+* [EPUB 3 Structural Semantics Vocabulary 1.1](https://www.w3.org/TR/epub-ssv-11/), WG Note
+* [EPUB Accessibility 1.1](https://www.w3.org/TR/epub-a11y-11/), Recommendation
+* [EPUB Accessibility Techniques 1.1](https://www.w3.org/TR/epub-a11y-tech-11/), WG Note
+* [EPUB Accessibility - EU Accessibility Act Mapping](https://www.w3.org/TR/epub-a11y-eaa-mapping/), WG Note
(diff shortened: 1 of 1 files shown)


mattgarrish:
I wonder if listing all the documents might be overload in an issue. Maybe we could shorten it to something like: "This repository is for reporting issues in the standards and notes produced by the Publishing Maintenance Working Group."

Should we provide any questions to guide users after the text? I don't think it has to be a lot, but maybe:

- Which standard or note is this issue related to? (For bug reports, include the relevant section.)
- Describe the problem? (For bug reports, cite the problematic text.)
- Describe the fix or new feature you propose?

It might help reinforce that we don't want how-to questions here.

iherman:
> I wonder if listing all the documents might be overload in an issue. Maybe we could shorten it to something like: "This repository is for reporting issues in the standards and notes produced by the Publishing Maintenance Working Group."

... in which case people may (a) try to find the references to the documents in the PM WG and (b) submit comments on audiobooks or the publishing manifest...

I think it is better to keep the list. It saves the reader to go to step (a) and make the mistake of (b).

> 
> Should we provide any questions to guide users after the text? I don't think it has to be a lot, but maybe:
> 
> * Which standard or note is this issue related to? (For bug reports, include the relevant section.)
> * Describe the problem? (For bug reports, cite the problematic text.)
> * Describe the fix or new feature you propose?
> 
> It might help reinforce that we don't want how-to questions here.

It might be a simple version of the question I had at the beginning (having separate issue templates for bugs/features).

Do you think you could make an inline comment with the question? It might make it simpler to process...

iherman:
B.t.w., the magic of setting up issue templates seems to be, simply, to have markdown files in the `.github/ISSUE_TEMPLATE/` directory, with some general metadata at the top. I presume anyone with the right to merge/commit can do that. I have learned something, again, about github :-)

mattgarrish:
> I think it is better to keep the list.

I'm just worried that it will push the important part we want people to read off screen. The input box isn't all that big, so someone might just read that it's for issues with the following specs and still presume that means any how-to question.

Could we can merge the paragraph after the list into the top one? So maybe have:

> This repository is only for issues with the following W3C documents. Questions or issues with specific reading systems, achieving specific styling, understanding validation errors, and similar will be considered as out of scope and will be closed without further action.

iherman:
> > I think it is better to keep the list.
> 
> I'm just worried that it will push the important part we want people to read off screen. The input box isn't all that big, so someone might just read that it's for issues with the following specs and still presume that means any how-to question.
> 
> Could we can merge the paragraph after the list into the top one? So maybe have:
> 
> > This repository is only for issues with the following W3C documents. Questions or issues with specific reading systems, achieving specific styling, understanding validation errors, and similar will be considered as out of scope and will be closed without further action.

Yep, I like that; I've made the changes.

.github/ISSUE_TEMPLATE/general-issue-template.md
```diff
@@ -0,0 +1,30 @@
+---
+name: General issue template
+about: Issue on the EPUB 3.3 specifications or related Working Group Notes
+title: ''
+labels: ''
+assignees: ''
+
+---
+This repository is only for issues with the following W3C documents. Questions or issues with specific reading systems, achieving specific styling, understanding validation errors, and similar will be considered as out of scope and will be closed without further action.
```
mattgarrish, line 10: If we're going to add questions, which I'll pose in another comment, maybe we should put a heading here.
```suggestion
## Before opening an issue
This repository is only for issues with the following W3C documents. Questions or issues with specific reading systems, achieving specific styling, understanding validation errors, and similar will be considered as out of scope and will be closed without further action.
```

.github/ISSUE_TEMPLATE/general-issue-template.md
```diff
@@ -0,0 +1,30 @@
...
+* [EPUB Type to ARIA Role Authoring Guide 1.1](https://www.w3.org/TR/epub-aria-authoring-11/), WG Note
+* [EPUB Fixed Layout Accessibility](https://www.w3.org/TR/epub-fxl-a11y/), WG Note
+
+
+Some possible alternative locations get help to your problems:
+
+* the [epub forum](https://www.mobileread.com/forums/forumdisplay.php?f=179) at mobileread
+* the [W3C publishing community group](https://github.com/w3c/publishingcg/issues)
+* the [ebooks section](https://ebooks.stackexchange.com/questions/tagged/epub) at stackexchange
+
```
mattgarrish, line 40: It might be useful to add starter questions.
```suggestion
## Which standard or note is this issue related to?

Please include the title and version number. For bug reports, include the relevant section.

## Describe the problem? 

What is not working? What would you like to see improved? For bug reports, cite the relevant text from the specification.

## Describe the fix or new feature you propose?

```

iherman:
@wareid, @shiestyle, could you give your blessing (or opposition to) this? Just to have at least another pair of eyes...

shiestyle (approved):
I agree with this update, let's try.

iherman:
I will go ahead and merge this (and see if it works). We can always change it later, the real question is whether it works as intended...

mattgarrish:
Looks good, although seeing it in action I now wonder if we should have done separate feature and bug templates. The generic template doesn't load by default, so you still have to pick it to begin.

PR #2652, opened 2024-09-20
Reorder issue templates

mattgarrish:
As we learn more about issue templates, it seems github uses the file names to order the templates when you click to create a new issue. That puts the EU mapping document first and EPUB 3.3 second last, among other oddities.

This PR just puts numbers at the start of the file names so that we can sort them easily. The three RECs are first followed by an alphabetized list of notes (with the accessibility techniques intentionally out of order so that they appear after the spec). This should give us the following order when opening an issue:

- EPUB 3.3
- EPUB Reading Systems 3.3
- EPUB Accessibility 1.1
- EPUB Accessibility Techniques 1.1
- EPUB Accessibility exemption property
- EPUB Accessibility - EU Accessibility Act Mapping
- EPUB Type to ARIA Role Authoring
- EPUB 3 Fixed Layout Accessibility
- EPUB 3 Multiple Rendition
- EPUB 3 Overview
- General issue template


iherman:
As discussed separately... I will merge this.

iherman (approved):

PR #2654, opened 2024-10-03
Allow xml:lang and dir on all optional DC elements

### LINKED ISSUE #2650: lang and dir of dc:date
This is similar to #2635 in that dc:date is recommended to use an ISO8601 date but it's not a requirement. @clapierre has had a case of a publisher putting in language-specific date. It will inherit the language and direction of the package document, which is probably good enough if that metadata is set, but for consistency it seems like we should allow the lang and dir attributes on the element. 

At this point, I'm wondering why we even have these restrictions on the attributes when we don't enforce the values of any of the dc optional elements. dc:format and dc:type could have a human-readable values and those are the only ones left excluded.

mattgarrish:
This pull request removes the lists of allowed elements from the optional DC definition. It also simplifies the global attribute definitions for xml:lang, dir, and id by referring to the required and optional element sections rather than list out every individual element.

Fixes #2650

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2654.html" title="Last updated on Oct 3, 2024, 6:44 PM UTC (2dab955)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2654/4969dd3...2dab955.html" title="Last updated on Oct 3, 2024, 6:44 PM UTC (2dab955)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -3282,8 +3282,8 @@
 &lt;/package&gt;</pre>
 					</aside>
 
-					<p>Allowed on: [^collection^], Dublin Core <a href="#sec-opf-dcmes-required">required</a> and <a
-							href="#sec-opf-dcmes-optional">optional</a> elements, [^meta^], and [^package^].</p>
+					<p>Allowed on: [^collection^], <a href="#sec-opf-dcmes">Dublin Core elements</a>, [^meta^], and
+						[^package^].</p>
 				</section>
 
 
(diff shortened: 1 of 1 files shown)


iherman (approved):
I am just picky, and this comment is not really on this PR, so I approve this either way...

Why is the title of §5.5.3 "Dublin Core required elements" and not "Required Dublin Core elements"? The latter sounds better to my ears. (Similar comment for optional in §5.5.4.)

If you choose to change then, alas! all the references in this PR must change, too...

mattgarrish:
I actually don't like having two "Dublin Core" sections at the same level. What if we switched it to:

- Dublin Core
  - Required elements
  - Optional elements

?

<strike>The current text would still make sense this way.</strike>

Actually, we could simplify the current text by just linking the parent DC section since it would contain all the definitions.

iherman:
Yes, that would make sense. The only problem I see is the heading levels for each entry would be even 'deeper'; we will break records in our spec;-) but I guess that is all right...

mattgarrish:
> we will break records in our spec;-)

You got my hopes up, but, alas, no. The OCF reserved files still hold the depth record by a level.

iherman:
> > we will break records in our spec;-)
> 
> You got my hopes up, but, alas, no. The OCF reserved files still hold the depth record by a level.

Oh well, we can still work on that 😀

Anyway, for the records, I am fine with the latest changes in https://github.com/w3c/epub-specs/pull/2654/commits/2dab9555b2ecf8198f2639a5468b1ace6e74e2c8. Thanks.

shiestyle (approved):

PR #2527, opened 2023-01-23
Add section about compatibility with Accessibility 1.0

mattgarrish:
This pull request becomes valid assuming we agree to publish the erratum noted in issue #2526.

If we can vote on moving ahead with that erratum on this week's call, this pull request doesn't add any new normative requirements.

If it's too late to change the specification and vote on moving to PR in the same call, we can defer integrating this and leave it for the maintenance group. Getting the erratum agreed on is more important and it doesn't affect the transition of the specifications.

PR #2655, opened 2024-10-03
Clarify that script and script resources are exempt

### LINKED ISSUE #2649: wasm support
sorry if this is the wrong place but i literally could not find a single place online mentioning epub and wasm together

will future epub specifications include wasm support for scripting?

mattgarrish:
The current prose about exempt resources including resources not referenced from the spine or embedded in content documents already excludes these resources, but this creates a formal exemption group to make this clear.

The pull request does not get into whether WASM should be a core media type or be supported by reading systems, so it only rises to a class 3 change. (For that reason, I'm not setting this to auto-close #2649.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2655.html" title="Last updated on Oct 7, 2024, 5:50 PM UTC (1784159)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2655/4969dd3...1784159.html" title="Last updated on Oct 7, 2024, 5:50 PM UTC (1784159)">Diff</a>

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1197,8 +1197,8 @@
 				<dl>
 					<dt id="exempt-fonts">Fonts</dt>
 					<dd id="confreq-resources-cd-fonts">
-						<p>All font resources not already covered as <a href="#cmt-grp-font">font core media types</a>
-							are exempt resources.</p>
+						<p>Font resources not covered as <a href="#cmt-grp-font">font core media types</a> are exempt
+							resources.</p>
 						<p>This exemption allows EPUB creators to use any font format without a fallback, regardless of
 							reading system support expectations, as CSS rules will ensure a fallback font in case of no
 							support.</p>
@@ -1209,42 +1209,46 @@
 					<dt id="exempt-links">Linked resources</dt>
 					<dd id="confreq-resources-cd-fallback-link">
 						<p>Any resource referenced from the [[html]] <a data-cite="html#the-link-element"
-									><code>link</code></a> element that is not already a core media type resource (e.g.,
-							CSS style sheets) is an exempt resource.</p>
+									><code>link</code></a> element that is not a core media type resource (e.g., CSS
+							style sheets) is an exempt resource.</p>
 					</dd>
 
 					<div class="proposed correction" id="change_1">
 						<span class="marker">Proposed Correction 1</span>
-						<p>The exemption for resources not referenced from the spine or embedded in content documents
-							(refer to the paragraphs following note below), includes resources referenced from
-								<code>script</code> elements, as well as resources retrieved by scripts for processing.
-							The description focused on data sets travelling in the EPUB container, but that was only one
-							use case. This change adds scripting to the major exemption groups and removes it from the
-							case for resources travelling in the container. For more information, refer to <a
+						<p>The <a href="#confreq-foreign-no-fallback">general exemption</a> defined below is ambiguous
+							about whether it applies to scripts and any modules they may use (the <code>script</code>
+							element is not in the list of embedding elements cited as not being exempt). This change
+							makes it clear that these resources are exempt. It also adds the use case of resources being
+							retrieved for processing. For more information, refer to <a
 								href="https://github.com/w3c/epub-specs/issues/2649">issue 2649</a>.</p>
 					</div>
 
 					<ins cite="#change_1">
-						<dt id="exempt-script">Scripts and script resources</dt>
+						<dt id="exempt-script">Scripts and modules</dt>
 						<dd id="confreq-resources-cd-fallback-script">
 							<p>Any resource referenced from the [[html]] <a data-cite="html#the-script-element"
-										><code>script</code></a> element that is not already a core media type resource
-								(e.g., JavaScript) is an exempt resource.</p>
+										><code>script</code></a> element that is not a core media type resource (e.g.,
+								JavaScript) is an exempt resource.</p>
 							<p>In addition, any resources retrieved by a script (e.g., using an <code>import</code>
-								declaration [[ecmascript]] or the XmlHttpRequest [[xhr]] or Fetch [[fetch]] APIs) are
+								declaration [[ecmascript]], or the XmlHttpRequest [[xhr]] or Fetch [[fetch]] APIs) are
 								exempt.</p>
+							<div class="note">
+								<p>These resources may still be subject to fallback requirements depending on how the
+									script uses them. For more information, refer to <a href="#confreq-cd-scripted-flbk"
+									></a>.</p>
+							</div>
 						</dd>
 					</ins>
 
 					<dt id="exempt-track">Tracks</dt>
 					<dd id="confreq-resources-cd-fallback-track">
-						<p>All audio and video tracks (e.g., [[?webvtt]] captions, subtitles and descriptions)
-							referenced from the [[html]]&#160;[^track^] element are exempt resources.</p>
+						<p>Audio and video tracks (e.g., [[?webvtt]] captions, subtitles and descriptions) referenced
+							from the [[html]]&#160;[^track^] element are exempt resources.</p>
 					</dd>
 
 					<dt id="exempt-video">Video</dt>
 					<dd id="confreq-resources-cd-fallback-video">
-						<p>All video codecs referenced from the [[html]] [^video^] —&#160;including any child [^source^]
+						<p>Video codecs referenced from the [[html]] [^video^] —&#160;including any child [^source^]
 							elements&#160;— are exempt resources.</p>
 						<div class="note">
 							<p>Although reading systems are encouraged to support at least one of the H.264 [[?h264]]
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -1213,15 +1213,28 @@
 							CSS style sheets) is an exempt resource.</p>
 					</dd>
 
-					<dt id="exempt-script">Scripts and script resources</dt>
-					<dd id="confreq-resources-cd-fallback-script">
-						<p>Any resource referenced from the [[html]] <a data-cite="html#the-script-element"
-									><code>script</code></a> element that is not already a core media type resource
-							(e.g., JavaScript) is an exempt resource.</p>
-						<p>In addition, any resources retrieved by a script (e.g., using an <code>import</code>
-							declaration [[ecmascript]] or the XmlHttpRequest [[xhr]] or Fetch [[fetch]] APIs) are
-							exempt.</p>
-					</dd>
+					<div class="proposed correction" id="change_1">
+						<span class="marker">Proposed Correction 1</span>
+						<p>The exemption for resources not referenced from the spine or embedded in content documents
+							(refer to the paragraphs following note below), includes resources referenced from
+								<code>script</code> elements, as well as resources retrieved by scripts for processing.
+							The description focused on data sets travelling in the EPUB container, but that was only one
+							use case. This change adds scripting to the major exemption groups and removes it from the
+							case for resources travelling in the container. For more information, refer to <a
+								href="https://github.com/w3c/epub-specs/issues/2649">issue 2649</a>.</p>
+					</div>
+
+					<ins cite="#change_1">
+						<dt id="exempt-script">Scripts and script resources</dt>
+						<dd id="confreq-resources-cd-fallback-script">
+							<p>Any resource referenced from the [[html]] <a data-cite="html#the-script-element"
+										><code>script</code></a> element that is not already a core media type resource
+								(e.g., JavaScript) is an exempt resource.</p>
+							<p>In addition, any resources retrieved by a script (e.g., using an <code>import</code>
+								declaration [[ecmascript]] or the XmlHttpRequest [[xhr]] or Fetch [[fetch]] APIs) are
+								exempt.</p>
+						</dd>
+					</ins>
 
 					<dt id="exempt-track">Tracks</dt>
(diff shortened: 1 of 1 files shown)


shiestyle (approved):

epub33/core/index.html
```diff
@@ -1213,6 +1213,29 @@ <h3>Exempt resources</h3>
 							CSS style sheets) is an exempt resource.</p>
 					</dd>
 
+					<div class="proposed correction" id="change_1">
+						<span class="marker">Proposed Correction 1</span>
+						<p>The exemption for resources not referenced from the spine or embedded in content documents
+							(refer to the paragraphs following note below), includes resources referenced from
```
bduga: This isn't really true - resources referenced from the script element itself did require a fallback before this change, so we are in fact loosening the restriction here. I am not sure if that still falls in class 3, maybe it does? But it isn't really a required change at this time, wasm is not technically importable via the script element (yet). Happy leaving this if we do decide this is still class 3.

epub33/core/index.html
```diff
@@ -1213,6 +1213,29 @@ <h3>Exempt resources</h3>
...
+							use case. This change adds scripting to the major exemption groups and removes it from the
+							case for resources travelling in the container. For more information, refer to <a
+								href="https://github.com/w3c/epub-specs/issues/2649">issue 2649</a>.</p>
+					</div>
+
+					<ins cite="#change_1">
+						<dt id="exempt-script">Scripts and script resources</dt>
+						<dd id="confreq-resources-cd-fallback-script">
+							<p>Any resource referenced from the [[html]] <a data-cite="html#the-script-element"
+										><code>script</code></a> element that is not already a core media type resource
```
bduga: nit: s/already// for brevity, but you are the wordsmith so final authority is yours :)

epub33/core/index.html
```diff
@@ -1213,6 +1213,29 @@ <h3>Exempt resources</h3>
...
+								(e.g., JavaScript) is an exempt resource.</p>
+							<p>In addition, any resources retrieved by a script (e.g., using an <code>import</code>
+								declaration [[ecmascript]] or the XmlHttpRequest [[xhr]] or Fetch [[fetch]] APIs) are
```
bduga: I am not super happy with the examples here, since it seems to confuse exempt and foreign resources. At the very least it is a bit murky. Maybe just the import example? Or drop the examples since people seem to get very confused by them and decide it is an exhaustive list.

epub33/core/index.html
mattgarrish: > resources referenced from the script element itself did require a fallback before this change

I'd argue not this change but the change that added the exemption below. It was probably too broad in hindsight, as the script element doesn't render the text of its script (it's not embedding the script in the content document; it's there to execute). So if a script is not embedded content and not referenced from the spine, then it's exempt.

If we don't include this exemption explicitly, the only way to make script require core media types again is to do a different class 3 change that adds it along with the embeddable content elements. I'd rather just move ahead this way.

epub33/core/index.html
mattgarrish: Ya, I copied the text from exemption above but agree it doesn't add anything. Will take it out of both.

epub33/core/index.html
mattgarrish: I suppose a script could read a foreign resource that is embedded somewhere else in the publication, which would make it bypass the exemption.

It might be better to leave the second paragraph out and let the exemption below continue to handle exempt files for processing. I can't think of how we make this clearer without duplicating the two exemption scenarios.

epub33/core/index.html
bduga: I guess I am getting lost in the reviews here. Which is the change that added the exemption? At least in the github UI I see "Any resource referenced from the html script element that is not already a core media type resource [snip] is an exempt resource." That seems to be a newly added change. Are you saying that it was already implied because referencing from the script element isn't really embedding? It's an interesting argument, and hard to dispute since we don't really define what it means to embed something.

epub33/core/index.html
bduga: Rereading it this is probably fine as is. Do people really have to list remote resources loaded via XmlHttpRequest in the manifest? I guess they do. Scripting is horrible.

epub33/core/index.html
mattgarrish: > It's an interesting argument, and hard to dispute since we don't really define what it means to embed something.

Right, in 3.3 we added the new exemption for data files and such. The first condition is the resource not be in the spine and the second that:

> it is not embedded directly in EPUB content documents (e.g., via [[html](https://www.w3.org/TR/epub-33/#bib-html)] [embedded content](https://html.spec.whatwg.org/multipage/dom.html#embedded-content-category) and [[svg](https://www.w3.org/TR/epub-33/#bib-svg)] [image](https://www.w3.org/TR/SVG/embedded.html#ImageElement) and [foreignObject](https://www.w3.org/TR/SVG/embedded.html#ForeignObjectElement) elements).

Embedding has always been the term we use for rendering a resource within a content document (via `object`, `img`, `iframe`, etc.). We usually use some form of "referenced from" if we mean any kind of reference (`script`, `link`, `a`, etc.).

`script` and `link` both allow references to resources without embedding them. That's partly why we made explicit that resources referenced from `link` exempt in 3.3 (plus there's a stronger case that nothing uses them than with script).

It's also why, at least to me, we've already (accidentally) opened the door to the `script` element not needing backups in the current spec. If we opt not to keep it exempt, we'd need to make that second bullet clearer.

Another reason it probably makes more sense to exempt `script` is that the only way you can provide an alternative for a resource referenced from it is through a manifest fallback. The less of that we have the better. :smile:

epub33/core/index.html
mattgarrish: > Do people really have to list remote resources loaded via XmlHttpRequest in the manifest? I guess they do.

Yup, that's my take, too. Used in the rendering = publication resource = manifest listing required.

> Scripting is horrible.

Ya, hoping we could get by in 3.0 underspecifying how scripting works wasn't the best of ideas.

What makes it more confusing is that you can't host most resources outside the container, unless, seemingly, you fetch them by script. So, if you want remote images, just fetch them and write them into a data url in an `img` tag. Spec rules circumvented.

epub33/core/index.html
bduga: I just did a review of embed use in the spec, and I can't say the use is as clear as you make it sound, It is certainly never defined that way. We do talk about embedded scripts (i.e scripts that are written directly in a script element), so those scripts are embedded, but if you move it to another file they are referenced? But we don't have that distinction for images (a data url embedded image does not become referenced when we move it to a new file). And what about CSS? If it is in another file is it referenced and therefore exempt? 

FWIW, my understanding of embedded vs referenced is embedded content is used by the current document in the (continuing) process of rendering it, while referenced content does not participate in the rendering of the current document, it is simply mentioned (links, etc). Of course, that isn't defined anywhere either.

mattgarrish: > It is certainly never defined that way.

No, that's a big part of the problem, but that might be a useful cleanup step for a future update.

But in this case we've selectively picked definitions and elements that are for rendering a resource directly. I can't square that with how script works.

mattgarrish: Are you arguing against having the exemption for script, though, or is this just about whether the resources were exempt before or not? (I'll get on editing out some of that prose now, since it's arguable both ways.)

I don't see how making script resources exempt would go up to a class 4 change, for example, even if we disagree over the current exemptions. Exempt resources are not new and exempting resources only gets them out of fallbacks. It doesn't add a new feature to the specification.

iherman:
(I comment on the version after https://github.com/w3c/epub-specs/pull/2655/commits/1784159270393637ef1d1ede47eec1ada45d4cd0; @bduga, @mattgarrish it would be better to officially "resolve" the conversations, to make any further discussion unambiguous.)

I am fine with the text as is now. I agree it also settles the original issue in #2655: after some cursory look at some examples out there, it seems that `.wasm` files are to be fetched in a JavaScript module, which is covered by the new text and firmly places these files as exempts. (Without even mentioning them, although it may be a good idea to add it as an example to the text to make things absolutely clear)

However, I continue to feel uneasy on declaring this new text as class 3, i.e., include it in the upcoming publishing round. I may buy the argument that this is what we always intended to happen. However, it looks like we simply forgot to make an explicit case for the `script` element alongside the `link` element, and this PR takes care of this mistake. It very much "smells" a new feature, which may lead to disagreeable discussions on process issues if we want to push it through the current PM WG route. I do not think we need a formal objection on our hands, even if it is, at the end, unjustified.

If we were at the beginning of the WG's life span, I would say "ok, let's try it". However, with the discussions on the [new charter in the making](https://w3c.github.io/publ-maintenance-wg-charter/), with the hopeful new incarnation of the PM WG aiming at EPUB 3.X, I would really prefer to consider this as a class 4. We should find a proper formulation to replace bullet point 4 in [the charter scope](https://w3c.github.io/publ-maintenance-wg-charter/#scope), and keep this PR open until we begin to work on the new version. We could live without `.wasm` files up until now, I do not think that is a highly urgent issue.

cc @wareid @shiestyle

mattgarrish:
> We could live without `.wasm` files up until now, I do not think that is a highly urgent issue.

Well, this change has no effect on the ability to include and use wasm files right now, which you can do.

Epubcheck doesn't inspect js code, so fetching/instantiating a wasm file isn't going to be flagged. And as it's not referenced from one of the embedding elements in the current exemption, the wasm file will fly under the radar in the manifest as just travelling in the container.

The only practical effect this would have is to allow any resource from `script` without a fallback, something that epubcheck can and will flag. But that's hardly a critical case since js is the only scripting language with broad support (and seems unlikely reading systems would support any of the more obscure possibilities given js support isn't even universal).

In other words, I don't think it really matters when we formally change anything. And as this is dragging on our timeframe to publish, I can get onboard with leaving it until we get the updated versions out.

bduga:
> The only practical effect this would have is to allow any resource from `script` without a fallback, something that epubcheck can and will flag. But that's hardly a critical case since js is the only scripting language with broad support (and seems unlikely reading systems would support any of the more obscure possibilities given js support isn't even universal).

+100

> 
> In other words, I don't think it really matters when we formally change anything. And as this is dragging on our timeframe to publish, I can get onboard with leaving it until we get the updated versions out.

Agreed. I don't disagree with this change is principle, but it seems like extra credit for some future potential scripting spec and not worth blocking on.

bduga:
> 
> Epubcheck doesn't inspect js code, so fetching/instantiating a wasm file isn't going to be flagged. And as it's not referenced from one of the embedding elements in the current exemption, the wasm file will fly under the radar in the manifest as just travelling in the container.

I would make the stronger argument that it is explicitly allowed today, and called out in the spec with "The primary case for this exemption is to **allow data files** to travel with an EPUB publication, whether **for scripts** to use in their constituent EPUB content documents ..."

mattgarrish:
> The primary case for this exemption is to **allow data files** to travel with an EPUB publication

And it ends with the case of a data file in a journal. That's the only case I remember us taking up, not that data file means any code file. It's also the only purpose noted in the use cases that were the basis of this exemption: https://docs.google.com/document/d/1msVuFxWk7ALQZ6PAaOy6MvO__s6L5r6vy-icqLWZRy4/edit

bduga:
> > The primary case for this exemption is to **allow data files** to travel with an EPUB publication
> 
> And it ends with the case of a data file in a journal. That's the only case I remember us taking up, not that data file means any code file.

But that is just an example, not an exhaustive list. And I don't think the journal file was meant to be related to scripting, it was just a bonus file users could find. So there are no examples of data files for scripts, but I don't see how that changes anything. A wasm file pretty clearly falls into the category of data file used by scripts.

mattgarrish:
> And I don't think the journal file was meant to be related to scripting,

Except that the file I pointed to says:

> Use case 3: As a journals publisher, I want to embed data files in the package. I don’t expect the RS to be able to render them natively, but I want them there either a)to be rendered by a JS-based renderer that I supply, or b) to be there for completeness/longevity, and/or for the user to retrieve and render in another application on the platform. 

That's exactly the use case that the exemption talks about.

I could also throw out the definition of a data file from wikipedia:

> A data file is a [computer file](https://en.wikipedia.org/wiki/Computer_file) which stores data to be used by a computer [application](https://en.wikipedia.org/wiki/Application_software) or [system](https://en.wikipedia.org/wiki/System_software), including input and output data. A data file usually does not contain instructions or code to be executed (that is, a [computer program](https://en.wikipedia.org/wiki/Computer_program)).

But it seems we're forever going to disagree until we revisit it and decide what we want. 

(And for the record, I'm all on board with code being exempt, and accidentally being exempt now; I just don't agree it was intended to be captured in the current exemption. We're kind of fighting over nothing! 😄 )

mattgarrish:
Closing this off, as if we're not going ahead with it now then who knows what state the spec might be in when we return to it, or if we want to add other clarifications. It's easy enough to reopen it or lift the changes if we can use it.

iherman:
Actually, also with regard to https://github.com/w3c/publ-maintenance-wg-charter/pull/30, it may be better to leave this PR open and label it as status-deferred. @mattgarrish wdyt?

mattgarrish:
> PR open and label it as status-deferred

Maybe it's my OCD kicking in, but I hate having open pull requests like #2527 that sit around. Plus it required the other proposed corrections to be renumbered, so I just think it's going to be a mess compared to starting over once we work out anything and everything that needs fixing.

It's probably better to refer to #2649 and #2656 than this pr.

iherman:
ok

PR #2666, opened 2024-11-21
Fix name conflict in templates

mattgarrish:
The structural semantics vocabulary template has the same name as the TTS template so neither is appearing in the list of options when you try to open a new issue. This should fix the problem.


iherman:
The beauties of cut and paste...

iherman (approved):

PR #2668, opened 2024-12-02
Remove the a11y prefix from the pageBreakSource property

### LINKED ISSUE #2667: Incorrect prefixing of pageBreakSource property
## Issue on the [“EPUB Accessibility Techniques 1.1”](https://www.w3.org/TR/epub-a11y-tech-11/) WG Note

[Section 5.1.4](https://www.w3.org/TR/epub-a11y-tech-11/#pageSource)

## Describe the problem 

The pageBreakSource descriptions show it with an `a11y` prefix but the property is defined as part of the meta properties vocabulary in epub 33 to replace `source-of`

## Describe the fix or new feature you propose

Remove the prefix.

mattgarrish:
Fixes #2667 

***

[Preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/pbs-property-prefix/epub33/a11y-tech/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/pbs-property-prefix/epub33/a11y-tech/index.html)

### DIFF
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -1993,7 +1993,7 @@
 						<p>The <code>pageBreakSource</code> property is the proposed replacement for this property. It
 							is still being incubated for inclusion in the <a
 								href="https://www.w3.org/TR/epub/#app-meta-property-vocab">Meta properties
-								vocabulary</a> [[epub-3]] but it is anticipated it will be added in the upcoming
+								vocabulary</a> [[epub-3]] but it is anticipated that it will be added in the upcoming
 							revision. That is why these techniques no longer refer to the <code>source-of</code>
 							property, even though it remains the official property in the specification.</p>
 
--- a/epub33/a11y-tech/index.html
+++ b/epub33/a11y-tech/index.html
@@ -1980,6 +1980,29 @@
 &lt;/metadata></pre>
 					</aside>
 
+					<div class="note">
+						<p>The <a href="https://www.w3.org/TR/epub/#sec-source-of"><code>source-of</code> property</a>
+							was previously recommended in these techniques to identify the <a
+								href="https://www.w3.org/TR/epub/#sec-opf-dcmes-optional-def"><code>dc:source</code>
+								element</a> containing the source of the pagination [[epub-3]]. This method came with a
+							number of limitations, however. It was not future proof, for example, as it relied on the
+							EPUB 3 <a href="https://www.w3.org/TR/epub/#attrdef-refines"><code>refines</code>
+								attribute</a> [[epub-3]] and it could not handle pagination without a source due to the
+							reliance on a <code>dc:source</code> element.</p>
+
+						<p>The <code>pageBreakSource</code> property is the proposed replacement for this property. It
+							is still being incubated for inclusion in the <a
+								href="https://www.w3.org/TR/epub/#app-meta-property-vocab">Meta properties
+								vocabulary</a> [[epub-3]] but it is anticipated it will be added in the upcoming
+							revision. That is why these techniques no longer refer to the <code>source-of</code>
+							property, even though it remains the official property in the specification.</p>
+
+						<p>It is strongly recommended that EPUB creators switch to using the
+								<code>pageBreakSource</code> property. EPUBCheck already accepts it so it should not
+							cause validation issues. Although the <code>source-of</code> method will remain valid for
+							backwards compatibility purposes, it is no longer recommended for new publications.</p>
+					</div>
+
 					<p>If an ISBN is not available, include as much information as possible about the source publication
 						(e.g., the publisher, date, edition, and binding).</p>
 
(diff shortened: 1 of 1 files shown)


gregoriopellegrino (approved):

iherman:
Copying my comment from the mailing list:

I do not think I agree with the argument.

The EPUB spec says: 

> A default vocabulary is one that EPUB creators do not have to declare a prefix for in order to use its terms and properties where a property value is expected. 

The elements of the default vocabulary are defined in the appendix D of the spec[1].

Following up the references for `pageBreakSource`, we arrive to [2] (a CG Report) which describes the origin of this property and defines it.  But it does not even refer to the EPUB 3.3 Default Vocabulary, ie, it does not say that this property should be considered as part of the EPUB default vocabulary. Nor should it: as a CG Note it cannot normatively "change" the Recommendation. I.e., the "a11y" prefix should be used to refer to it.

I am (of course) not against adding this term to the default vocabulary of the spec under the new charter, thereby indeed making the "a11y" prefix obsolete. (We can do that, under the charter's item referring to the Publishing CG's A11y task force.)

[1] https://www.w3.org/TR/epub/#sec-default-vocab  
[2] https://www.w3.org/publishing/a11y/page-source-id/  

P.S. I am not sure I understand the "and properties where a property value is expected" remark in the aforementioned definition. Is this the source of our misunderstanding?

mattgarrish:
> The elements of the default vocabulary are defined in the appendix D of the spec[1].

Yes, and this is a proposed addition to it. It doesn't make sense to replace a property in the meta vocabulary with a prefixed property in another specification. Identifying pagination source isn't exclusively an accessibility need.

It also doesn't make sense to incubate a property in another vocabulary space only to change it when it gets adopted. That just messes everyone up, from authors to developers.

The same logic can be applied that the note doesn't mention the accessibility namespace or vocabulary anywhere. That's a failing of the note that it doesn't refer to the meta properties vocabulary, but the note never proposes or uses a prefix. It was always how it was intended.

Moreover, we had to get epubcheck to allow it for incubation purposes: https://github.com/w3c/epubcheck/issues/1491 We wouldn't have done that if it was supposed to carry a prefix.

This is also the property behind the proposal to add features developed by the publishing cg into the epub spec. If it helps, I can add a note, or append to one, that the property is still being incubated but is expected to be adopted in the next revision. But we updated the techniques because we've already pushed people to move on from the old method and having conflicting guidance was just causing confusion.

No one has been using a prefix for it as the techniques were only recently updated to mention this property. To have to get everyone to switch to using a prefixed property for a few months until we can revise the specs and then switch back really isn't going to fly.

danielweck (approved):

mattgarrish:
> I am not sure I understand the "and properties where a property value is expected"

Default vocabularies don't apply to attributes that don't accept property values (dir, xml:lang, etc.).

The "property value" is essentially a definition of a curie. So the meta `property` attribute requires a `property` value. Confusing terminology, but it just means the attribute value has to be a prefixed property name unless it's one of the ones from the default vocabulary.

iherman:
@mattgarrish fine, I believe we are on the same opinion.

>  If it helps, I can add a note, or append to one, that the property is still being incubated but is expected to be adopted in the next revision.

Yes, I think that would make things clearer. And we will take care of this in the new version as soon as we can.

mattgarrish:
> Yes, I think that would make things clearer.

Okay, I've updated the note at the end of the section to try and explain the status of the property more clearly and I also moved it up to where the property is first mentioned so that it's not lost at the end of the section. Let me know if you think that's better or if it needs any other tweaking.

epub33/a11y-tech/index.html
```diff
@@ -1965,60 +1965,69 @@ <h4>Identify the pagination source</h4>
...
+								element</a> containing the source of the pagination [[epub-3]]. This method came with a
+							number of limitations, however. It was not future proof, for example, as it relied on the
+							EPUB 3 <a href="https://www.w3.org/TR/epub/#attrdef-refines"><code>refines</code>
+								attribute</a> [[epub-3]] and it could not handle pagination without a source due to the
+							reliance on a <code>dc:source</code> element.</p>
+
+						<p>The <code>pageBreakSource</code> property is the proposed replacement for this property. It
+							is still being incubated for inclusion in the <a
+								href="https://www.w3.org/TR/epub/#app-meta-property-vocab">Meta properties
+								vocabulary</a> [[epub-3]] but it is anticipated it will be added in the upcoming
```
iherman:
```suggestion
								vocabulary</a> [[epub-3]] but it is anticipated that it will be added in the upcoming
```

iherman (approved):

PR #2671, opened 2024-12-12
Remove normative keywords and conformance section from the exemption note

mattgarrish:
Concerns have been raised in the charter process about the use of normative language in the exemption note. There are only three instances where we've used normative keywords, primarily in defining the value of the property. None of these is essential and this pull request rewords them out of the document.

I've also added a note to the overview that the we don't plan to take this property any further down the standardization track; that's it's defined as a convenience for those who need it.

Let me know if there are any concerns with this before we take it any further.

---

See [Preview](https://raw.githack.com/w3c/epub-specs/fix/exempt-note-non-norm/epub33/a11y-exemption/index.html), [Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2Fepub-a11y-exemption%2F&doc2=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fraw.githack.com%2Fw3c%2Fepub-specs%2Ffix%2Fexempt-note-non-norm%2Fepub33%2Fa11y-exemption%2Findex.html)

### DIFF
--- a/epub33/a11y-exemption/index.html
+++ b/epub33/a11y-exemption/index.html
@@ -81,13 +81,19 @@
 					would require a fundamental alteration of the product.</p>
 
 				<p>Although it is possible to explain these exemptions in the accessibility summary for an [=EPUB
-					publication=], plain language statements are not easily processed by machines. A vendor may want to
-					know that a publication is exempt before allowing it in their bookstore without an accessibility
+					publication=], plain language statements are not easily processed by machines. A vendor might want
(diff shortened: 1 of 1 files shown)


iherman (approved):

shiestyle (approved):

iherman:
Overtaken by events per https://github.com/w3c/strategy/issues/481#issuecomment-2551661056 and subsequent change in the proposed charter. Closing without merge.

PR #2672, opened 2025-01-02
Added dark mode to the core spec

iherman:
I played with this for the core spec. The necessary changes are:

- add a specific meta statement to the header
- added some extra css statements to our common css file, related to the classes we define. It is mostly making sure that the characters are black for a light background.

We can play with a more elaborate change in colours for those entries (caution, explanation, etc), but I think @mattgarrish has more experience in making properly contrasted colours. I did the minimal that worked for me.

I do not think we should _merge_ this PR now, just make it ready to merge. First, we should have all documents prepared to dark mode. Also, we may want to wait until we begin working on the EPUB 3.4 line...

See:

* For EPUB 3.3:
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/dark-mode/epub33/core/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub33/core/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/dark-mode/epub33/core/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2672.html" title="Last updated on Jan 17, 2025, 2:52 PM UTC (4f5e6a9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2672/d792e06...4f5e6a9.html" title="Last updated on Jan 17, 2025, 2:52 PM UTC (4f5e6a9)">Diff</a>


iherman:
Actually, I did some more elaborate colour settings in the common.css file. Also checked them for the WAI colour contrast requirements.

mattgarrish:
Would it make sense to start this off on the notes and then migrate to the rec docs once we're good with it? (Or maybe those don't have the same scale of custom styling?)

iherman:
> Would it make sense to start this off on the notes and then migrate to the rec docs once we're good with it? 

I would prefer to change all our publications in a big bang but, pragmatically, doing it first on notes is doable, while doing it on the rec track documents becomes more complicated because it requires a more complicated publication process. So, yours may be the right pragmatic approach... 

I actually think that this should be a group decision and not ours as editors. (@wareid @shiestyle something to discuss at a call?) 

> (Or maybe those don't have the same scale of custom styling?)

I actually did not check that; I do not know whether they all include a reference to `custom.css`. However, it may not matter: the changes I made on CSS are only of concern for the specific styles introduced in that file. Other than that, the W3C standard environment covers everything; the only thing to be done is to add the `<meta>` element in the header.

mattgarrish:
> I actually did not check that; I do not know whether they all include a reference to `custom.css`.

I'm not sure, either. I think, unfortunately, some may inline the custom css even though they don't need it. That's why I'm not sure we could even test all the styling even if the notes purport to use it.

At some point we should try and clean out the unnecessary use of custom styling, and maybe look at ways of optimizing it for the docs that need it. I suspect it's often called on only to get the zebra table styling.

iherman:
> > I actually did not check that; I do not know whether they all include a reference to `custom.css`.
> 
> I'm not sure, either. I think, unfortunately, some may inline the custom css even though they don't need it. That's why I'm not sure we could even test all the styling even if the notes purport to use it.
> 
> At some point we should try and clean out the unnecessary use of custom styling, and maybe look at ways of optimizing it for the docs that need it. I suspect it's often called on only to get the zebra table styling.

Yep. And we should get all the notes to use the same `custom.css` rather than have a local copy...

Let us try to get an agreement with the WG about the dark theme (or not) and we can take care of that cleanup at the same time.

wareid (approved):
Working now on my end!

wareid:
Just did an additional check as I realized I did not look at the styling for examples and images. It's not converting code samples properly and figures look a bit strange (acceptable but could be better). 
Screenshot 2025-01-15 at 3 39 03 PM
Screenshot 2025-01-15 at 3 38 54 PM

iherman:
The code sample colours are set by the "official" W3C style sheet. I would not change that (just to be consistent with other documents), although, if we want, we can do it.

The official style sheet put the image background to white. I was a bit bothered by this glaring white spot, so I set it to slightly yellowish, but we can change that.

The reason of the white background for the images is that the figures are usually prepared for a white background, and they do not necessarily look good on black background. In the VC WG I spent quite some time on creating images that work on both backgrounds, and it is extremely difficult, so I can see why the choice was to keep it white.

mattgarrish:
Should we update this to include the RS and A11y specs so we have a reason to republish all three when the proposed corrections are added? (To harmonize their final publish dates.)

iherman:
> Should we update this to include the RS and A11y specs so we have a reason to republish all three when the proposed corrections are added? (To harmonize their final publish dates.)

Yes, absolutely!

Thx

iherman:
Also: I see that core is now on dark mode, but it seems that some css additions I did for dark mode have not been 'transferred'. Meaning that the boxes with a yellow background white characters...

mattgarrish:
I've added the meta tag to the other two specs. If you need to commit some additional styles here, it'd be good to add them to the css for the 3.4 docs at the same time.

iherman:
> I've added the meta tag to the other two specs. If you need to commit some additional styles here, it'd be good to add them to the css for the 3.4 docs at the same time.

No we are good after all. It is just funny: if I read, e.g., https://w3c.github.io/epub-specs/epub34/authoring/ on my dark mode, and look at the caution box in https://w3c.github.io/epub-specs/epub34/authoring/#sec-intro-relations, it first appears without the css changes I made and then, after 1-2 seconds, it does. It seems that the css changes of common.css need some time. We may have to look at this issue at some point...

mattgarrish:
> it first appears without the css changes I made and then, after 1-2 seconds, it does

Ya, the custom.css is inlined in a `style` element by a respec preprocessing function call. When I load in dark mode I also get the washed out orange background while the respec button is spinning and then it's replaced by the styles when it stops. It must be lag fetching the css.

I don't know how we work around that, but it'll only affect the editor drafts so it's not a huge issue. The exported/published drafts will have the static/inlined styles.

mattgarrish (approved):

iherman:
Matt, I have to run now... could you merge it, and keep an eye that everything goes well?

PR #2678, opened 2025-01-28
Fixed layout techniques accessibility metadata

clapierre:
Work on Metadata techniques with examples


clapierre:
Here is a [preview ](https://raw.githack.com/w3c/epub-specs/Fixed-Layout-Techniques-Accessibility-Metadata/epub33/fxl-a11y-tech/index.html)of what I have done thus far.

clapierre:
@wareid Let me know what you think.  Is this what you were expecting?  Should I include all possible values, like I have outlined?

gautierchomel (commented):
It's a wonderful comprehensive list! I want it as a technique document for [Schema.org Accessibility Properties for Discoverability Vocabulary](https://www.w3.org/2021/a11y-discov-vocab/latest/).

I'm not sure how useful all the vocabulary is for fxl, but I also think it's good to expose content creators to all the possible values, which may spark ideas. 

Also, I would suggest putting access modes at the first subsection level.

gautierchomel (approved):

epub33/fxl-a11y-tech/index.html
```diff
@@ -315,6 +342,921 @@ <h3>Legibility</h3>
...
+                    </aside>
+                </section>
+            </section>
+
+            <section id="metadata-access-mode-sufficient">
+                <h3>Access Mode Sufficient</h3>
+                <section>
+                    <h4>Single Values</h4>
+                    
+                    <p>Having a single accessMode implys that the entire content within the EPUB can be consumed with only that mode of accessing the content within the book.</p>
```
wareid: accessMode or accessModeSufficient? This sentence is a little confusing and these properties are already really confusing!

epub33/fxl-a11y-tech/index.html
```diff
@@ -315,6 +342,922 @@ <h3>Legibility</h3>
...
+                            &lt;p&gt;
+                                &lt;img src="images/hawk.jpg"&gt;
+                            &lt;/p&gt;
+                            
+                            or
+                            
+                            &lt;div style="background-image: url('images/hawk.jpg')"&gt;&lt;/div&gt;
+                            </code>
+                        </pre>
+                        <p>The metadata to included in the OPF file would be:</p>
```
mattgarrish: OPF file is epub 2 terminology.

wareid: Ack, good catch.

mattgarrish:
Similar to @gautierchomel comment above, is this useful to put in a document on fixed layouts? There don't seem to be a lot of specifics to fixed layouts, and things like braille properties I'm not even sure are all that likely to ever be found in these kinds of documents.

You might want to consider splitting this off into a separate guide on implementing accessibility metadata in epub and focus only on properties that are problematic for fixed layouts, like reading order and display transformability.

wareid (approved):
I like how this is progressing and I think we should merge it and continue building on it. I have only two editorial things I want to get feedback on: 
1. Editors/authors, I want to make sure everyone who contributed text to this extent is credited, so I wonder if using "authors" here would work? I don't think everyone contributing is editing, but I also want to make sure proper credit is attributed. 
2. Use of notes, my thinking with notes is that they should be used sparingly to provide context or additional information that is not necessary but helpful. I think there's places here where essential information is included in notes and I want to remove them, but I wanted to get feedback first since it's a significant editorial change.

mattgarrish:
FYI, we're looking to move all the notes to a new version-less directory so we don't get caught having to migrate them each time we do a new revision. To do that, we need to get the FXL notes to a stable place so they can be migrated.

I think there's value in having an epub accessibility metadata guide, but this looks like a big project. What I'd like to do is take the bulk of the work done here and move it into a new document (we can then look at how best to continue that work). I can try and separate out the FXL-specific stuff and keep it in this document at the same time.

Does that work for you @clapierre? That way we can get to moving the notes without having to rush this document in before it's finished.

PR #2682, opened 2025-02-12
Create epub 3.4 folder with baseline documents

mattgarrish:
This pull request preps a new /epub34 folder for the coming revision. I've only copied over the REC documents plus the overview and accessibility techniques notes for now. We can pull other notes over as we need to amend them.

I've only updated metadata, stripped down the change logs, and put in a placeholder acknowledgement file. There's also a new index page and an empty errata file, but I imagine we won't properly update those until later (sort of like the overview).

It might be good to look at all the scripts we have and whether we need shared styles anymore, but I don't want to take away all the fun times ahead with the first commit. 😉 

A few things of note:

- /core has been renamed /authoring. We often call epub 3.x, rs 3.x and a11y 1.x the core specs, so it's probably better to call this the authoring spec behind the scenes and match it up with who it's for.  It won't affect the formal name. (I finally concede to @iherman's nit with the directory!)
- I've intentionally not updated the respec metadata for the authoring spec to 3.4 yet, but integrated all of the proposed corrections. The goal is that I can come back later and grab this version from the revision history and not have to make the corrections over again when the proposed rec is accepted next month. I'll open another pull request momentarily that can be merged after this one.
- I fixed the revision history links in the change logs for the EPUB 3.3 and Reading Systems 3.3 specs. The labels they were matching (or negating) weren't getting the right lists of closed issues. I figure since these are class 1 changes it's okay to push them whenever we do the proposed rec update.
- I deleted the release folders from the 3.3 directory. Not sure how these survived my earlier moving of them all to the /releases subdirectory, but there's no need for them to be polluting the root folder anymore.

There will be lots of respec errors in these files until the -34 versions are picked up by respec and until all the shared terminology gets harvested again. I can maybe look at putting in placeholder biblio entries to get rid of those errors, but I expect we'll be stuck with the terminology ones until we get a first working draft published.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2682.html" title="Last updated on Feb 13, 2025, 3:20 PM UTC (e75fe9b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2682/ba7198d...e75fe9b.html" title="Last updated on Feb 13, 2025, 3:20 PM UTC (e75fe9b)">Diff</a>


iherman (approved):
It is probably futile to go into all the details of such huge PR. I believe the PR should be merged and deal with minutiae separately.

Just a few general things that we should do while renewing/restructuring things. (I put them here not to forget them. They can be part of a separate PR or done in this one, as you prefer.)

- We should systematically use the common.css in all our documents, including the notes. This means using the `<script src="../common/js/css-inline.js" class="remove"></script>` entry in all our headers. That would avoid us to repeat ourselves.
- We should look at PR #2672, and try to incorporate the style and html header changes in all our document. (There was a respec/css error at the time of doing that PR which should be settled by now.) Note that the changes in #2672 are mostly in the common.css file, so if the previous item is taken care of, it should be easy to do those without bothering about PR merges
- There is a pending PR for the workflow scripts: #2681. This should be merged into main as soon as possible (which will trigger a general republication of all our non-normative documents) and this branch should be (probably) rebased on that new version of 'main'.
- If we want to republish a note with a new short name, we should not forget to ask for short name approval and do the publication magic which ensures that the history simply flows back to the 3.3 version. I presume that is what we want for notes (as opposed to Rec-track documents which start a new journey from WD to REC).
- I actually wonder whether we need to keep those releases in the first place. The content of those releases are copied and pushed onto the W3C Web site verbatim, and then kept them for eternity. I know history is important, but duplicating things is not really necessary. In other words, once a document is published, I do not mind simply removing it from the repository; this admin issue less on our plate.

iherman:
One more thing we should not forget (just noting it to myself...): the secret tokens used for echidna are bound to a short name. As soon as a short name changes, I will have to regenerate the corresponding token...

mattgarrish:
> It is probably futile to go into all the details of such huge PR. I believe the PR should be merged and deal with minutiae separately.

Yup, I was hoping to share the pain... 😉 

And it's easier and faster to start working on the issues in parallel, and review them properly, rather than have one ginormous pull request that keeps sitting open as we add more and more to it.

mattgarrish:
> * I actually wonder whether we need to keep those releases in the first place.

I'm more than happy to remove them. They're really only in the repository because it's an easy way for you to grab and publish them.

shiestyle (approved):

iherman:
> > * I actually wonder whether we need to keep those releases in the first place.
> 
> I'm more than happy to remove them. They're really only in the repository because it's an easy way for you to grab and publish them.

Ok. So the workflow would be:
- you produce a snapshot _somewhere_; you use that to check with pubrules
- I grab it, push it to cvs, and get it published via the webmaster
- I remove it from the repository

mattgarrish:
> * This should be merged into main as soon as possible (which will trigger a general republication of all our non-normative documents) and this branch should be (probably) rebased on that new version of 'main'.

I don't think we need to rebase as that one only affected the yml files and I didn't modify them yet.

I'm not sure how that works, as do we need to wait until the documents are in an initial FPWD state for making actual changes? (i.e., do we want working drafts published while we're still doing basic cleanup steps, and can we even do FPWD through echnidna?)

Also, do we need to go through that process of getting the group to agree to use echidna to publish since we've rechartered?

mattgarrish:
But let's merge this and then we can get working on the other issues.

iherman:
> I'm not sure how that works, as do we need to wait until the documents are in an initial FPWD state for making actual changes? (i.e., do we want working drafts published while we're still doing basic cleanup steps, and can we even do FPWD through echnidna?)

We should avoid doing any change on the EPUB 3.3 Recs, except for what has been accepted through the AC review and possible low level changes. No Echidna there. 

Once that is done (cca. March 15), we should go and officially establish an EPUB 3.4 line via FPWD's. Then, and only then, can we begin to use Echidna again.

(Of course we can publish Notes through Echidna except if we change the short name.)

PR #2684, opened 2025-02-14
Start accessibility spec at 1.1.1

mattgarrish:
It was noted in the working group meeting yesterday that we may not have substantive enough changes to warrant a new 1.2 version number, which would require a new conformance string and all that change would entail.

The more prudent course of action is to start the accessibility specification and techniques as a 1.1.1 revision and then upgrade them later if we determine we need to bump the number to accommodate substantive changes to conformance.

To that end, this pull request changes all the numbering in the /epub34 directory to 1.1.1.

(There are also some unrelated fixes to add temporary spec info to local bibliographies to get rid of some respec errors.)

### DIFF
--- a/epub34/a11y/index.html
+++ b/epub34/a11y/index.html
@@ -2,7 +2,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml" lang="en-US" xml:lang="en-US">
 	<head>
 		<meta charset="utf-8" />
-		<title>EPUB Accessibility 1.2</title>
+		<title>EPUB Accessibility 1.1.1</title>
 		<script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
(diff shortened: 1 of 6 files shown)


iherman (approved):

avneeshsingh (approved):

epub34/a11y/index.html
```diff
@@ -106,17 +106,17 @@
...
-				//		]
-				//	}
-				//]
+				/* otherLinks: [
+					{
+						key: "This document is also available in this non-normative format:",
+						data: [
+							{
+								value: "EPUB 3",
+								href: "./epub-a11y-12.epub"
```
clapierre: shouldn't this be epub-a11y-111.epub?

mattgarrish: Yup, that code's commented out because we don't generate the epub version of the spec until the final stages of publishing the new version, so I wasn't paying attention to it.

PR #2686, opened 2025-02-19
Clarify the types of resources exempt by not being used by reading systems

### LINKED ISSUE #2649: wasm support

### LINKED ISSUE #2656: Ambiguous exemption for "embedded" resources
In section 3.4 "Exempt resources" we have the following bullet point for a blanket exemption:

"it is not embedded directly in EPUB content documents (e.g., via [html] embedded content and [svg] image and foreignObject elements)."

It is unclear what is meant by embedded as we never define the term. Valid interpretations of the term would make scripts, style sheets, and/or fonts exempt (or not). We should clarify what the term means, or just better explain the types of exempt resources. 

See discussion in #2655 for background.

mattgarrish:
As discussed on the last WG call, this pull request expands on the types of resources that fall under the exemption of not being used by reading systems. I broke the existing paragraph out into a list and added script modules and made script inputs its own bullet rather folded into data sets for external apps.

I've also tried to better address the question about embedded resources so there isn't overlap with this exemption. I've promoted the references to the HTML and SVG concepts of embedded content so that they aren't just examples of embedding but the defining definitions.

Finally, I've added the exemption for resources referenced from the script element.

We started to consider these changes last year in #2655 and I've kept what I could from that pull request here. This one is different in that I didn't try to merge the script inputs and module exemptions with the exemption for the script element itself. It feels like overloading and it starts to duplicate the existing exemption when all we really needed to do was clarify it.

As always, feedback welcome.

Fixes #2649
Fixes #2656

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2686.html" title="Last updated on Feb 21, 2025, 8:54 PM UTC (f076e27)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2686/7b2e40a...f076e27.html" title="Last updated on Feb 21, 2025, 8:54 PM UTC (f076e27)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -1184,6 +1184,13 @@
 					resources=]. It is most similar to a [=foreign resource=] in that it is not guaranteed [=reading
 					system=] support, but, like a core media type resource, does not require a fallback.</p>
 
+				<p>In some cases, the exemptions defined in this section overlap with how existing core media type
+					resources are included in [=EPUB content documents=] (e.g., using the [[html]] <a
+						data-cite="html#the-link-element"><code>link</code></a> element to include CSS style sheets,
+					importing fonts into CSS, or using the [[html]] [^script^] element for JavaScripts). Although this
+					superficially makes the resources similar in terms of not requiring fallbacks, core media types are
+					differentiated by having stronger support expectations in reading systems.</p>
+
 				<p>There are only a small set of special cases for exempt resources. Video, for example, are exempt from
 					fallbacks because there is no consensus on a core media type video format at this time (i.e., there
 					is no format to fallback to). Similarly, audio and video tracks are exempt to allow [=EPUB
@@ -1196,8 +1203,7 @@
 				<dl>
 					<dt id="exempt-fonts">Fonts</dt>
 					<dd id="confreq-resources-cd-fonts">
-						<p>Font resources not covered as <a href="#cmt-grp-font">font core media types</a> are exempt
-							resources.</p>
+						<p>Font resources are exempt resources.</p>
 						<p>This exemption allows EPUB creators to use any font format without a fallback, regardless of
 							reading system support expectations, as CSS rules will ensure a fallback font in case of no
 							support.</p>
@@ -1208,14 +1214,12 @@
 					<dt id="exempt-links">Linked resources</dt>
 					<dd id="confreq-resources-cd-fallback-link">
 						<p>Any resource referenced from the [[html]] <a data-cite="html#the-link-element"
-									><code>link</code></a> element that is not a core media type resource (e.g., CSS
-							style sheets) is an exempt resource.</p>
+									><code>link</code></a> element is an exempt resource.</p>
 					</dd>
 
 					<dt id="exempt-script">Scripts</dt>
 					<dd id="confreq-resources-cd-fallback-script">
-						<p>Any resource referenced from the [[html]] [^script^] element that is not a core media type
-							resource (e.g., JavaScript) is an exempt resource.</p>
+						<p>Any resource [[html]] allows the [^script^] element to reference is an exempt resource.</p>
 					</dd>
 
 					<dt id="exempt-track">Tracks</dt>
@@ -1239,8 +1243,8 @@
 
 				<div class="note">
 					<p>The exemptions made above do not apply to the [=EPUB spine | spine=]. If an exempt resource is
-						used in the spine, and it is not also an [=EPUB content document=], it will require a fallback
-						in that context.</p>
+						used in the spine, and it is not also an EPUB content document, it will require a fallback in
+						that context.</p>
 				</div>
 
 				<p id="confreq-foreign-no-fallback">In addition to the content-specific exemptions, a resource is
@@ -1252,9 +1256,7 @@
 							document=]); <strong><em>and</em></strong></p>
 					</li>
 					<li>
-						<p>it is not embedded content in an EPUB content document &#8212; refer to the [[html]]
-							[=embedded content=] and [[svg]] <a href="https://www.w3.org/TR/SVG/embedded.html">embedded
-								content</a> definitions for more information.</p>
+						<p>it is not directly rendered in its native format in an EPUB content document.</p>
 					</li>
 				</ul>
 
@@ -1269,6 +1271,12 @@
 					<li>data files for use by external applications (e.g., a scientific journal might include a data set
 						with instructions on how to extract it from the EPUB container).</li>
 				</ul>
+
+				<div class="note">
+					<p>An example of resource not exempted by this rule would be a image referenced by a <a
+							data-cite="css-values#urls"><code>url</code> function</a> [[css-values]] in a CSS style
+						sheet, as this would not result in core media type resource being displayed.</p>
+				</div>
 
 				<p>The exemption also allows EPUB creators to use foreign resources in foreign content documents without
 					reading systems or [=EPUB conformance checkers=] having to understand the fallback capabilities of
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -1196,8 +1196,8 @@
 				<dl>
 					<dt id="exempt-fonts">Fonts</dt>
 					<dd id="confreq-resources-cd-fonts">
-						<p>All font resources not already covered as <a href="#cmt-grp-font">font core media types</a>
-							are exempt resources.</p>
+						<p>Font resources not covered as <a href="#cmt-grp-font">font core media types</a> are exempt
+							resources.</p>
 						<p>This exemption allows EPUB creators to use any font format without a fallback, regardless of
 							reading system support expectations, as CSS rules will ensure a fallback font in case of no
 							support.</p>
@@ -1208,19 +1208,25 @@
 					<dt id="exempt-links">Linked resources</dt>
 					<dd id="confreq-resources-cd-fallback-link">
 						<p>Any resource referenced from the [[html]] <a data-cite="html#the-link-element"
-									><code>link</code></a> element that is not already a core media type resource (e.g.,
-							CSS style sheets) is an exempt resource.</p>
+									><code>link</code></a> element that is not a core media type resource (e.g., CSS
+							style sheets) is an exempt resource.</p>
+					</dd>
+
+					<dt id="exempt-script">Scripts</dt>
+					<dd id="confreq-resources-cd-fallback-script">
+						<p>Any resource referenced from the [[html]] [^script^] element that is not a core media type
+							resource (e.g., JavaScript) is an exempt resource.</p>
 					</dd>
 
 					<dt id="exempt-track">Tracks</dt>
 					<dd id="confreq-resources-cd-fallback-track">
-						<p>All audio and video tracks (e.g., [[?webvtt]] captions, subtitles and descriptions)
-							referenced from the [[html]]&#160;[^track^] element are exempt resources.</p>
+						<p>Audio and video tracks (e.g., [[?webvtt]] captions, subtitles and descriptions) referenced
+							from the [[html]]&#160;[^track^] element are exempt resources.</p>
 					</dd>
 
 					<dt id="exempt-video">Video</dt>
 					<dd id="confreq-resources-cd-fallback-video">
-						<p>All video codecs referenced from the [[html]] [^video^] —&#160;including any child [^source^]
+						<p>Video codecs referenced from the [[html]] [^video^] —&#160;including any child [^source^]
 							elements&#160;— are exempt resources.</p>
 						<div class="note">
 							<p>Although reading systems are encouraged to support at least one of the H.264 [[?h264]]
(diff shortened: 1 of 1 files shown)


epub34/authoring/index.html
```diff
@@ -1208,19 +1208,25 @@ <h3>Exempt resources</h3>
...
 						<p>Any resource referenced from the [[html]] <a data-cite="html#the-link-element"
-									><code>link</code></a> element that is not already a core media type resource (e.g.,
-							CSS style sheets) is an exempt resource.</p>
+									><code>link</code></a> element that is not a core media type resource (e.g., CSS
+							style sheets) is an exempt resource.</p>
+					</dd>
+
+					<dt id="exempt-script">Scripts</dt>
+					<dd id="confreq-resources-cd-fallback-script">
+						<p>Any resource referenced from the [[html]] [^script^] element that is not a core media type
```
bduga: I remain ambivalent here. This feels like we are encroaching on HTML by allowing non-Javascript scripts, which really should be up to that spec. It is also confusing, in that we are allowing things that are currently forbidden, so it would give me pause when reading - I would end up questioning my understanding of the text. The only reason I can think of for allowing this is to let things like WASM be referenced using a polyfill, which I guess is handy? But also seems somewhat niche. @iherman I think this is something you wanted - what do you think?

epub34/authoring/index.html
```diff
@@ -1246,23 +1252,27 @@ <h3>Exempt resources</h3>
 							document=]); <strong><em>and</em></strong></p>
 					</li>
 					<li>
-						<p>it is not embedded directly in EPUB content documents (e.g., via [[?html]]&#160;[=embedded
-							content=] and [[?svg]]&#160;<a href="https://www.w3.org/TR/SVG/embedded.html#ImageElement"
-									><code>image</code></a> and <a
-								href="https://www.w3.org/TR/SVG/embedded.html#ForeignObjectElement"
-									><code>foreignObject</code></a> elements).</p>
+						<p>it is not embedded content in an EPUB content document &#8212; refer to the [[html]]
```
bduga: This is perhaps more clear, but I think it remains too broad. For instance, this means that all style content is an exempt resource, as is all content reference by style sheets (e.g. a background image referenced from CSS does not need to be a core media type, nor does it need a fallback). Is this really what the WG intended when we made this change? And if it is, I think we should be clear about that.

epub34/authoring/index.html
mattgarrish: The html spec already mentions you can reference external css and json module scripts from the element, but the latter wouldn't currently be allowed. If we want to make sure we stay within the bounds of the html spec, could we just make that a qualifier:

> Any resource [html] allows the `script` element to reference is an exempt resource.

(I think it might be helpful to rework the core media type references in all these definitions, like maybe putting a single note before the list that these exemptions are in addition to them.)

epub34/authoring/index.html
mattgarrish: The original goal was to not flag resources harmlessly travelling in the container as having to meet core media type fallback requirements. It then morphed into also using these for scripts. The part of about exempting foreign resources was a convenient side effect when we were mapping out the resource types.

What if we go the rendering route instead of focusing on the methods? So something like:

> it is not directly rendered in its native format in an EPUB content document

A data file used by a script to write to canvas, or an image added to the dom with a fallback, would be fine. As would wasm files.

But it wouldn't work for static content like images referenced from styles rules as they'd attempt to present the image in whatever format it's in.

It would also keep the foreign resource exemption stuff intact.

epub34/authoring/index.html
bduga: I like this a lot better - it defers everything to html. The old wording felt very much like "you could if you could, but you can't so you can't".

epub34/authoring/index.html
bduga: Yeah, this matches my intuitive feeling for embedded in this context, and feels like it captures the intent better. I would support this wording.

epub34/authoring/index.html
iherman: I like the proposed text, but shouldn't it say

> Any resource [html] allows the script element to reference that is nota core media type is an exempt resource.

But, in general, deferring to the HTML (or other spec) is a good idea...

epub34/authoring/index.html
iherman: Although... someone may trick the system. Say I have an image in jpg and, for whatever reason, I want it to be treated as an exemption. I then include a jpg->png transformation somewhere, which is used to render the image. Ie, the jpg image is not rendered in its native format, therefore it is an exemption...

Maybe

> It cannot be directly rendered in its native format in an EPUB content document

epub34/authoring/index.html
mattgarrish: > that is nota core media type

That was the parenthesis I had below the suggested text. I find stating this makes reading the exemptions more confusing than they need to be, since core media types are just explicit exemptions from fallbacks by media type. If you don't know the details, it can leave you wondering why core media types aren't exempt from fallbacks.

I think a general note before the list that these exemption are in addition to any core media type designations would serve the same purpose.

epub34/authoring/index.html
mattgarrish: > Say I have an image in jpg and, for whatever reason, I want it to be treated as an exemption.

But a jpeg is already a core media type so it shouldn't need an exemption to travel in the container. It's not going to get flagged as a foreign resource without a fallback, which is what happens when you list these non-cmt data files in the manifest.

Epubcheck might warn you that you appear to have an unused publication resource, but that's only a usage message so it's not a problem for publishing.

> It cannot be directly rendered in its native format in an EPUB content document

There's a lot of stuff that will render inside an iframe. Wouldn't this exclude json, xml, etc.?

iherman: > There's a lot of stuff that will render inside an iframe. Wouldn't this exclude json, xml, etc.?

But those are not rendered "natively" but through a transformation process. It is not in the default rendering engine's capability to render them. So they would be covered as well.

mattgarrish: > But those are not rendered "natively" but through a transformation process.

I'm not sure I follow. What do you mean by a transformation process? If I reference a json or xml file from the src attribute of an iframe, that's what is loaded into the iframe (with some differences between browsers).

Your exemption says it must not be possible for the format to display at all in xhtml. Or is "it cannot be" an instruction to the author and not a quality of the rendering engine?

iherman: > > But those are not rendered "natively" but through a transformation process.
> 
> I'm not sure I follow. What do you mean by a transformation process? If I reference a json or xml file from the src attribute of an iframe, that's what is loaded into the iframe (with some differences between browsers).
> 

Is that formally defined somewhere, or is it just "done" by browsers? Ie, is it an added functionality by browsers (or webviews)? If the former, then of course it is clearly 'inherited' by an RS. But if it is the latter, then we may be on a slippery ground: do we have to take it for granted that a RS would do that?

(The only exception is of course XHTML. That is "natively" rendered.)

All that being said, we are talking about something very minor here, so I won't object to do what you originally proposed to do. The fact that we are having this discussion shows that the line I was proposing raises more questions than what is worth...

mattgarrish: > Is that formally defined somewhere, or is it just "done" by browsers?

This is what confuses me about making it dependent on what can render natively in an xhtml (or svg) document. Even reading systems don't have to render everything epub supports. (Would image data be illegal to use in a reading system without a viewport?) I think we're opening a can of worms that's open to all kinds of interpretation.

We're probably also getting a little picky over an edge case that we have little to no control over anyway. The bigger problem was, as Brady suggested, not opening a loophole for unintended cases, like images called from css. We can't check if scripts are behaving honestly, so any restriction we come up with will depend on the author being aware of the rule and adhering to it.

mattgarrish:
I just pushed some changes per the discussions. I added the text we discussed and also added a note after the examples of exempt resources to describe the css case of a non-exempt resource. Maybe not critical to have, but I think it's helpful to explain the limits of the exemption.

I also took out the repeating "that is not already a core media type" references out and added a new second paragraph to the section to explain the overlap but how core media types are better supported.

iherman:
I am not sure I understand the remark

> An example of resource not exempted by this rule would be a image referenced by a [url function](https://www.w3.org/TR/css-values-3/#urls) [[css-values](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2686.html#bib-css-values)] in a CSS style sheet, as this would not result in core media type resource being displayed.

If I use css url to get a jpg file to set a bullet item marker, isn't that resource displayed? Or the same for a background?

mattgarrish:
> If I use css url to get a jpg file to set a bullet item marker, isn't that resource displayed?

Exemptions are for what would otherwise be foreign resources. A jpeg is a core media type.

mattgarrish:
But if it helps, I can change it to: "... a foreign image resource reference by ... "

epub34/authoring/index.html
```diff
@@ -1184,6 +1184,13 @@ <h3>Exempt resources</h3>
 					resources=]. It is most similar to a [=foreign resource=] in that it is not guaranteed [=reading
 					system=] support, but, like a core media type resource, does not require a fallback.</p>
 
+				<p>In some cases, the exemptions defined in this section overlap with how existing core media type
+					resources are included in [=EPUB content documents=] (e.g., using the [[html]] <a
+						data-cite="html#the-link-element"><code>link</code></a> element to include CSS style sheets,
+					importing fonts into CSS, or using the [[html]] [^script^] element for JavaScripts). Although this
+					superficially makes the resources similar in terms of not requiring fallbacks, core media types are
+					differentiated by having stronger support expectations in reading systems.</p>
```
bduga: Does this need a clarification about how such resources are defined? Something like "Resources of core media types that match an exemption remain core media types and are not reclassified as exempt resources." That is, a jpg referenced from a script is not an exempt resource, it remains a core media type. Or maybe that isn't true - I guess it depends if core media types can also be exempt, although we seem to consider these two distinct classes.

epub34/authoring/index.html
```diff
@@ -1246,23 +1256,31 @@ <h3>Exempt resources</h3>
...
+					<li>script code modules, such as WebAssembly [[?wasm-core-2]];</li>
+					<li>script inputs, such as fetched [[?fetch]] data or images (refer to <a
+							href="#confreq-cd-scripted-flbk"></a> for restrictions on dynamically adding such resources
+						to EPUB content documents);</li>
+					<li>data files for use by external applications (e.g., a scientific journal might include a data set
+						with instructions on how to extract it from the EPUB container).</li>
+				</ul>
+
+				<div class="note">
+					<p>An example of resource not exempted by this rule would be a image referenced by a <a
```
bduga: Nit: "a resource" or perhaps "resources"

epub34/authoring/index.html
```diff
@@ -1246,23 +1256,31 @@ <h3>Exempt resources</h3>
...
+							data-cite="css-values#urls"><code>url</code> function</a> [[css-values]] in a CSS style
+						sheet, as this would not result in core media type resource being displayed.</p>
```
bduga: I am having a little trouble with the double negatives ("not exempt"/"would not result"), but I have no specific suggestion to make it clearer.

bduga (commented):
Nice, I like this version, just a few comments.

epub34/authoring/index.html
mattgarrish: Oops, thanks!

epub34/authoring/index.html
mattgarrish: Agree, the more we can clarify here the better. I've added a couple of additional sentences to make clear that CMTs do not become exempt resources because they share the same inclusion mechanisms and exempt resources without the exemption would be foreign resources. Hopefully, that improves it a bit.

epub34/authoring/index.html
mattgarrish: I've tweaked this a bit in the last commit and added the additional context of it being a foreign image resource.

epub34/authoring/index.html
```diff
@@ -1189,7 +1189,9 @@ <h3>Exempt resources</h3>
 						data-cite="html#the-link-element"><code>link</code></a> element to include CSS style sheets,
 					importing fonts into CSS, or using the [[html]] [^script^] element for JavaScripts). Although this
 					superficially makes the resources similar in terms of not requiring fallbacks, core media types are
-					differentiated by having stronger support expectations in reading systems.</p>
+					differentiated by having stronger support expectations in reading systems. A core media type
+					resource does not become an exempt resource just because they share the same inclusion mechanism.
+					And without an exemption, an exempt resource would be [=foreign resource=].</p>
```
bduga: Nit: "a [=foreign resource=]"?

bduga (approved):
Nice! As a previous manager used to say - "ship it!"

iherman (approved):

iherman:
For the records: WG Approval in https://w3c.github.io/pm-wg/minutes/2025-02-27.html#3fff

iherman:
This was discussed during the [pmwg meeting on 01 March 2025](https://w3c.github.io/pm-wg/minutes/2025-02-27.html#d6c2).

<h3><a href="https://github.com/w3c/epub-specs/pull/2686" rel="noopener noreferrer">WASM resolution (w3c/epub-specs#2686)</a></h3>
<p><cite>Wendy:</cite> We have a packed agenda today</p>
<p><cite>&lt;wendyreid&gt;</cite> <a href="https://github.com/w3c/epub-specs/pull/2686" rel="noopener noreferrer">w3c/<wbr>epub-specs#2686</a></p>
<p><cite>Wendy:</cite> We would like to have a resolution affirming 2686, which includes webassembly…</p>
<p><cite>&lt;wendyreid&gt;</cite> Proposed: Merge PR 2686.</p>
<p><cite>&lt;Sueneu&gt;</cite> +1</p>
<p><cite>&lt;gpellegrino&gt;</cite> 0</p>
<p><cite>duga:</cite> There were some changes from out discussion…</p>
<p>something of a backdoor for fallbacks for embedded content. We’ve closed an open door and removed embedded content</p>
<p><cite>mgarrish:</cite> mostly it maps to what we talked about, added some clarifications and examples…</p>
<p>there is a summary in the PR</p>
<p><cite>&lt;George&gt;</cite> +1</p>
<p><strong>RESOLUTION:</strong> Merge PR 2686.</p>

<hr /></details>

PR #2693, opened 2025-03-14
Add action to publish epub:type to aria roles authoring guide

mattgarrish:
While looking at moving the notes, I spotted that the authoring guide has a postprocess function call with no source script include for it. Guessing it accidentally got copied in while adding metadata.

Made me wonder why this was never flagged when the notes have been republished, and sure enough we're missing an action to publish it.

@iherman the only thing I'm unsure of here is the echidna token. I put the short name on the end for now, but please use your secret w3c powers to make sure it's correct.


mattgarrish:
I've updated this PR to rename the yml files for the REC track documents back to .yml files.

Github allows you to disable workflows from the actions page, so we shouldn't need to resort to this hack anymore: https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/disabling-and-enabling-a-workflow

But after integrating this PR, we'll need to disable the REC publishing actions before merging anything else.

The paths in the yml files are all going to need updating, too, but leaving that part until we rearrange the notes.

iherman:
> 
> @iherman the only thing I'm unsure of here is the echidna token. I put the short name on the end for now, but please use your secret w3c powers to make sure it's correct.

I am at a loss here, my apologies. Has that short name been approved and already in use before? If so, things should be o.k.; if not, we have to officially ask for it.

As for echidna, we need a secret per short name. If the short name changes, the secret token has to be regenerated (this will be the case for any note that we republish, as well as the EPUB 3.4 publications.

mattgarrish:
> Has that short name been approved and already in use before?

Yes, the note was published: https://www.w3.org/TR/epub-aria-authoring-11/

And we have the approval: https://www.w3.org/2023/01/05-epub-minutes.html#resolution04

But it doesn't appear that a github action was ever created to have echnida auto-publish it along with the other notes. That's why I have no idea what the token should be. I copied a yml file for another note, but I don't know if there's a public way to look up what the token should be.

mattgarrish:
> As for echidna, we need a secret per short name.

So it sounds like if we never set up auto-publishing for this note then we need to get one created? I was assuming tokens were created automatically whenever new documents are published.

.github/workflows/epub-aria-authoring.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: w3c/spec-prod@v2
+        with:
+          SOURCE: epub33/epub-aria-authoring/
+          TOOLCHAIN: respec
+          W3C_ECHIDNA_TOKEN: ${{ secrets.W3C_TR_TOKEN_EPUB_ARIA_AUTHORING }}
+          W3C_WG_DECISION_URL: https://www.w3.org/2023/01/05-epub-minutes.html#resolution04
+          W3C_BUILD_OVERRIDE: |
+             shortName: epub-aria-authoring
```
iherman:
```suggestion
             shortName: epub-aria-authoring-11
```

iherman (approved):

iherman:
> > As for echidna, we need a secret per short name.
> 
> So it sounds like if we never set up auto-publishing for this note then we need to get one created? I was assuming tokens were created automatically whenever new documents are published.

No, it was set up; I mean the secret key has been generated. But the short name was wrong, see my comment above.

PR #2695, opened 2025-03-17
Update workflow paths

mattgarrish:
Updates the yml files to use the new /epub34, /common, and /wg-notes paths.

I also updated the REC track yml files to link to the resolutions from the last meeting, but we'll need to add the same later for the overview and a11y techniques.


mattgarrish:
I also renamed core.yml to authoring.yml, so leaving a mental note here to disable the new workflow after merging this.

iherman (approved):
I hope it won't lead to confusion that all the workflow entries listed to the epub34 folder will fail for now... But the changes themselves look good to me.

mattgarrish:
You and I are probably the only ones who pay attention to this stuff... 😄

PR #2697, opened 2025-03-17
Split meta guide from fxl techniques

mattgarrish:
To get the repository reconfigured for publishing next week, this pull request just pulls out the metadata guide for now.

After the documents are moved, it would be good to review what metadata may be confusing to implement for fixed layouts or that might have a different interpretation and only cover that. Let the metadata guide cover everything.

Just as one possible example, textOnVisual could be misunderstood in an FXL context as absolute positioned text over an image if you assume the meaning only from the name. It might be worth noting in a discussion on it to also set accessMode=visual, since this is the mode most likely to be recognized. When to claim a textual sufficient access mode is likely another complex issue for fxls that could be of value to cover. I'd focus on those kinds of issues.


mattgarrish:
I'm going to merge this so the changes and new locations can be reviewed in the other pull request

PR #2703, opened 2025-04-01
Refresh the Overview to 34

iherman:
It just mostly updated the references. There is also a small addition to the history layers, and an a minor change in the figure to dark mode.

See:

* For EPUB 3 Overview:
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/refresh-overview-to-34/epub34/overview/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/overview/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/refresh-overview-to-34/epub34/overview/index.html)


epub34/overview/index.html
```diff
@@ -728,6 +746,20 @@ <h3>EPUB 3.3: 2022</h3>
 
 				<p>The separate <a data-cite="epub-33#change-log">section</a> in [[epub-33]] provides a more detailed
 					overview of the changes.</p>
+
+				<p>The W3C Recommendation was published in 2023, with a minor update in 2025.</p>
+			</section>
+			<section id="epub34">
```
mattgarrish: I'm not sure about adding the recap in advance of work really beginning. This could easily get forgotten. What about putting the paragraph inside an editor's note so that it's more likely to get flagged as needing finalizing later?

epub34/overview/index.html
```diff
@@ -728,6 +746,20 @@ <h3>EPUB 3.3: 2022</h3>
...
+				<h3>EPUB 3.4: 2027</h3>
+
+				<p>The work on EPUB 3.4 [[epub-34]] started in 2025. The goal is to add some new features 
+					to EPUB 3.3, and to clarify some existing features. The new features include the ability to
+					add standard annotating features to EPUB publications, and the ability to include 
+					"webtoons" in EPUB publications. The new features are expected to be published as
+					 a W3C Recommendation in 2027.</p>
+
+				</p>
```
mattgarrish:
```suggestion
```

epub34/overview/index.html
```diff
@@ -103,15 +103,21 @@ <h3>Reading this document</h3>
...
+						style="background-color: transparent;" />
+				</figure>
 
 				<details id="fig-epub-structure-diagram" class="desc">
 					<summary style="font-weight: normal; font-style: italic">Image description</summary>
 					<p>'EPUB (OCF) container' as the outer most component which encapsulates the 'EPUB publication'
 						containing two documents 'Package Document' and 'EPUB navigation document' along with a
 						inner component labelled 'publication resources' which contains multiple 'EPUB content
-						documents' (XHTML, SVG), and multiple 'Other resources' (CSS, png, mp3, mov, etc.). </p>
+						documents (XHTML, SVG)', and multiple 'Other resources (CSS, png, mp3, mov, etc.)'. </p>
```
mattgarrish: The use of quotes here suggests direct citation, but the text in the summary and in the image are slightly different. Shouldn't the image use "EPUB content documents", for example? It also has the singular "Other resource" and an ellipsis instead of "etc.". I hate to make image manipulation work for you, but having these in sync would be best.

iherman: I decided not to touch the graphics, but went through the alt text to sync them. This is an old thing that nobody discovered so far... I believe whatever is in quote here is now the same as in the figure (except for the new line in the diagram).

epub34/overview/index.html
iherman: You are right. I have put the whole section as an "ednote"

iherman:
@mattgarrish I think it is o.k. to merge this, too...

PR #2704, opened 2025-04-03
Fix typo in a11y-tech note

martinpub:
Found a typo in the accessibility techniques document (missing closing quote), this PR is a fix.

PR #2705, opened 2025-04-04
Add missing quote

mattgarrish:
Fixes the issue from #2704 in the new version of the note.


iherman (approved):

iherman:
@mattgarrish I believe this PR can be merged, the echidna script is inactive.

b.t.w. there is now a resolution to transition this and the overview to 3.4; I guess we can make that transition in about 10-15 days. Things will then be back to order as far as echidna goes...

mattgarrish:
Oh, right, I was forgetting these are technically new notes so will have to be republished anyway. I was thinking they'd only publish after approval once we started merging PRs again. Early mornings are hell... 😉

PR #2709, opened 2025-04-08
Add basic support for html syntax

mattgarrish:
As resolved as the face-to-face, this pull request adds very preliminary support for html to the specification.

Until we know for sure that this will be part of the revision, all it does for now is

- add text/html as a core media type and create a new HTML grouping in the CMT table
- renames the xhtml content document section to html content documents and makes xml syntax conformance a requirement only when the media type is set in the manifest

I've also added a large editor's note to the start of the html content document section to note that more substantive changes will be needed, and issues with xml-only features will need to be resolved, once we're sure this is going to happen. Aside from general feedback on the wording of the note, one question I have is should we open a specific issue for people to comment into? I've pointed people generally to the tracker for now as we don't appear to have one.

(The idea with this PR is that it gets the ball rolling without putting us in a place where it'll be very difficult to undo should this change not happen.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2709.html" title="Last updated on Apr 25, 2025, 1:48 AM UTC (f3ea437)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2709/f444801...f3ea437.html" title="Last updated on Apr 25, 2025, 1:48 AM UTC (f3ea437)">Diff</a>


iherman:
I think it is worth documenting, even as a note for now, that in the HTML case the `epub:type` is replaced by `epub-type`. In practice, looking at [§C.2](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2709/01d4330...82a6beb.html#sec-epub-type-attribute) we should

- make a note that `epub:type` is for the XML serialization only
- add a similar section §C.3 introducing `epub-type` for HTML making it clear that the semantics of that attribute is identical to `epub:type` (and it requires no namespace declaration).
- I presume it is also true that (1) XHTML _SHOULD NOT_ use `epub-type` and (2) HTML _SHOULD NOT_ use `epub-type`. (Or is it a _MUST NOT_?)

iherman:
Admin comment: @wareid @shiestyle I believe this PR is so fundamental, that it should require a WG call presentation/discussion/resolution before merging...

epub34/authoring/index.html
```diff
@@ -5795,27 +5807,38 @@ <h4>NCX</h4>
 			<h2>EPUB content documents</h2>
 
 			<section id="sec-xhtml">
-				<h3>XHTML content documents</h3>
+				<h3>HTML content documents</h3>
+
+				<div class="ednote">
+					<p>Support for the HTML syntax is still an open issue in the EPUB 3.4 revision. This specification
+						introduces it as a core media type and also minimally adds requirements for its authoring in
+						this section to begin making readers aware of the proposed change.</p>
```
```suggestion
					<p>Support for the HTML syntax is still an open issue in the EPUB 3.4 revision. It is <b><i>at risk</i></b>, depending on authors' and implementers' feedbacks . This specification
						introduces it as a core media type and also minimally adds requirements for its authoring in
						this section to begin making readers aware of the proposed change.</p>
```

Declaring a feature officially "at risk" means that we can enter CR with this feature with the possibility to pull it at the last minute if implementations fail. I realize it is very early to talk about CR, but it may be worth emphasizing this as soon as possible.

epub34/authoring/index.html
```diff
@@ -5795,27 +5807,38 @@ <h4>NCX</h4>
...
+
+					<p>Much of this specification, and the Reading Systems specification [[epub-rs-34]], continues to
+						refer only to XHTML content documents and their authoring requirements, a number of which need
+						resolving (e.g., providing an alternative to the <code>epub:type</code> attribute). These
```
iherman: As I commented elsewhere, I would prefer to document the `epub:type` alternative right away, as it is probably the most fundamental question that needs resolution, and put it out there early is the best. I am not sure what other example we would put here (maybe whether we allow the nav file to be HTML?)

mattgarrish:
> * add a similar section §C.3 introducing `epub-type`

Ya, I wasn't sure if we wanted to go this far yet. It's probably not an issue to do this as the likelihood we make changes around epub:type are small.

> * (1) XHTML _SHOULD NOT_ use `epub-type` and (2) HTML _SHOULD NOT_ use `epub-type`. (Or is it a _MUST NOT_?)

I think these should be hard restrictions, although I don't think we need to phrase it this way. XML namespaced attributes aren't allowed in the HTML syntax, so they'll already cause errors. We'd have to extend HTML to allow them in that syntax, and I can't imagine it's worth the bother. If your document isn't valid xml, having the xml attributes as some kind of fallback isn't going to make a program trying to parse it as xml work any better.

And I'd wait and see if there's a real need to phase out epub:type in xhtml content documents before introducing another way of doing the same thing. If we allow epub-type in xhtml, it should also come with a warning to stop using epub:type (even if that has to be an informative instruction because we could never disallow it considering all the existing content that would break).

mattgarrish:
For now, I've added a new serialization `epub-type` to the existing section on `epub:type`. It's technically one attribute (the type attribute), so duplicating the section seemed like bloat. But we can always come back to this later when we have to iron out the differences.

iherman (approved):
For some reason the "Caution" block in the renewed section on (epub:)type has lost its special styling as a result of preceding it by an editor's note. But that is only an editorial knit-picking.

mattgarrish:
I think it's a bug in the preview. It must not have retrieved the inline css styles as the attribute definitions are also supposed to be in those blue definition boxes. Everything looks fine on my local version.

shiestyle (approved):

bduga (commented):
I started to do a deep review, but I think it may be too early for that, as the text notes there are still mixed references to HTML and XHTML, so I won't bother with that yet. At a higher level, do we need to introduce HTML alongside XHTML content documents? I guess the question is, is HTML a new supported content document type, or is HTML the new content document type and XHTML is a subset of HTML? It looks like this pr takes the latter approach. Is that something we need to decide?

iherman:
> I started to do a deep review, but I think it may be too early for that, as the text notes there are still mixed references to HTML and XHTML, so I won't bother with that yet. At a higher level, do we need to introduce HTML alongside XHTML content documents? I guess the question is, is HTML a new supported content document type, or is HTML the new content document type and XHTML is a subset of HTML? 

My feeling it should be the former. The reason is more a question of communication rather than technical: at the moment we are proposing that, if you publish an EPUB, HTML _may_ be used instead of XHTML but, if you want, you can ignore HTML altogether. 

> It looks like this pr takes the latter approach. Is that something we need to decide?

I do not know whether that was @mattgarrish's intention. Why do you think this is what this PR is based on the latter approach?

mattgarrish:
> > It looks like this pr takes the latter approach. Is that something we need to decide?
> 
> I do not know whether that was @mattgarrish's intention.

I was thinking of adapting the HTML spec approach, so we define "HTML content documents" but that encompasses both syntaxes. Not that the HTML syntax is primary or replacing XHTML, or that XHTML is a subset, but when we need to call out one or the other format specifically we can say the "XML syntax of " or the "HTML syntax of " HTML content documents.

I think it will avoid a lot of rewriting madness if we stick to one term. If we add HTML content documents then presumably we'll also have to add (X)HTML content documents to then talk about the two cases together, or else we start listing both content document types everywhere we have to refer to both. Both those approaches sound ugly. And then we have the headache of rewriting all the other specs and notes that have to be revised to introduce this change. A global swap of xhtml to html that encompasses both syntaxes will be a lot easier to pull off.

But at this stage, which way we go with naming is completely open. Minimizing the changes while introducing HTML was my only goal, as I don't want to back out entire specs worth of changes at the last minute. The final changeover will require rewriting a lot of sections, so hopefully we can get wider review in the publishing community, plus TAG review of our epub-type proposal, asap so we can start the rewriting and updating process in earnest without the risk of a massive undo looming over our heads.

wareid (approved):

iherman:
This was discussed during the [pmwg meeting on 17 April 2025](https://w3c.github.io/pm-wg/minutes/2025-04-17.html#2de4).

<h4><a href="https://github.com/w3c/epub-specs/pull/2709" rel="noopener noreferrer">w3c/<wbr>epub-specs#2709</a></h4>
<p><cite>mgarrish:</cite> start to supporting HTML.  Adds media type starts to configure the content document to identify it with a Note, define epub-type to use it in the HTML representation.  There is a defintion and NOTE to under discussion potentially *At Risk* to get on everyone's radar.</p>
<p><cite>duga:</cite> 1. this will be fun for testing, this will take a lot of tests, not sure yet how. could be 100s of tests then to delete them at the end if we pull it.</p>
<p><cite>ivan:</cite> this is an exception to create tests first. since we are requiring feedback from the larger community. just put a toe in not a complete dive.<br>
<span>… merge as soon as we are happy and see if we can put it in EPUBCheck and see what publishers will do, as long as we can rollback at the present. we could add a couple tests turning some into html from XHTML.</span></p>
<p><cite>duga:</cite> I do think we should have a hand full of tests for RS to tests.</p>
<p><cite>gpellegrino:</cite> clarification: all HTML is all in addition to XHTML.  not removing XHTML.</p>
<p><cite>wendyreid:</cite> we would be allowing EPUB creators to choose XHTML or HTML.</p>
<p><cite>duga:</cite> important to answer it will impact what this PR look likes. we can have 2 types of content documents (HTML &amp; XHTML) maybe replace xhtml with html and allow for the serialization of HTML in XML, but maybe this is not the case but where do we land with this PR.</p>
<p><cite>ivan:</cite> we have a very strong requirement of backward compatibility so my view is HTML is alongside XHTML.<br>
<span>… is it ok to mix the two inside the same publication or restrict it to one for a publication.</span><br>
<span>… this is what the document does but leave it to Matt.</span></p>
<p><cite>mgarrish:</cite> I was trying to achieve to start to align us with HTML and there is an different syntax for each. requirements are the same a lot of duplicates. we have a lot of notes/specs so we want to try to minimize the change.<br>
<span>… do we wont to get into defining XHTML or HTML documents or could have an XHTML nav doc since its better supported for example.  I would let folks figure it out on their own. I would be surprised if they switch syntaxs.</span></p>
<p><cite>duga:</cite> I agree mixed is fine, so lets not try to restrict it.  XHTML is not really a thing we reference a spec that says it is not really a thing. we are saying you can use any HTML its up to you to use HTML, we require support for the XML serialization but we are fully backward compatible.</p>
<p><cite>mgarrish:</cite> I agree we have HTML content and don't use say XHTML. I don't think we want two termsmoving forward.</p>
<p><cite>wendyreid:</cite> Sounds like this is a cleaner approach, but will need some explanation to our community if we remove XHTML and replace it with HTML but its just the XML serialization.  Change in terminology.</p>
<p><cite>Hadrien:</cite> when we reference a media type when declare resources where they will be treated differently.</p>
<p><cite>ivan:</cite> current draft maintains both so that's fine. define epub-type not epub:type similar to aria-*<br>
<span>… when we merge this we should have a discussion on outreach on this right away. Could lead to misunderstandings.  Blog Mailing List announcements etc.</span></p>
<p><cite>duga:</cite> if we dont' decide not use HTML in the content we could remove the term XHTML everywhere.</p>
<p><cite>ivan:</cite> I think its a good idea.</p>
<p><cite>SueNeu:</cite> as a book publisher I am not sure, I could get myself into trouble.</p>
<p><cite>duga:</cite> xhtml is not a term that is used anymore in the HTML specs. we could get rid of the term xhtml but xhtml doesn't define it. we require XHTML everywhere but just the term XHTML doesn't exist.</p>
<p><cite>rdeltour:</cite> split the PR one side implements what Brady suggests, and the other PR for extends the spec to use the HTML syntax.</p>
<p><cite>George:</cite> tools I use when I switch between the two with the validation. that will be an important distinction.</p>
<p><cite>George:</cite> not just EPUBCheck but all the other tools out there.</p>
<p><cite>duga:</cite> I agree it is a concern. Maybe that is a good idea. Not sure if we should remove the XHTML term in our spec, not sure that could lead to real confusion.  Might be the right thing to do, may need more time or coffee :)</p>
<p><cite>DaleR:</cite> sounds like 2 audiences, authors how do I create a product that will be rendered and doesn't produce errors, the the authors of the creators of the reading systems.</p>
<p><cite>mgarrish:</cite> sitting on the fence getting rid of X. safer if we have both syntax's if the spec just says HTML but the content still requires XHTML. just could be confusing until we really do support both.</p>
<p><cite>ivan:</cite> there are different audiences. change in the reading system, is easier since only RS developers.</p>
<p><cite>mgarrish:</cite> except the RS does refer to the authoring document so they need to be in sync.</p>
<p><cite>wendyreid:</cite> help to see what if we remove the X to see what it looks like. I have an EPUB that uses HTML content documents. and see what breaks.</p>
<p><cite>mgarrish:</cite> what is there now is easy to back out.<br>
<span>… we won't get any feedback until we merge this in our draft.</span></p>
<p><cite>shiestyle:</cite> I agree we should merge this then we can wide review. This will become a possiblity once its merged.</p>
<p><cite>SueNeu:</cite> is this a Task Force to explore this? test document failures??</p>
<p><cite>ivan:</cite> I don't think a TF is needed we have been discussing this for 2 1/2 years or more. its time to merge and get wide review / discussions.<br>
<span>… if there is an acceptance, then we can restructure the document in the long term is a good thing to do. use epub-type instead of epub:type but lets see.</span></p>

<hr /></details>

iherman:
This was discussed during the [pmwg meeting on 24 April 2025](https://w3c.github.io/pm-wg/minutes/2025-04-24.html#b741).

<h3>Changing XHTML to HTML - <a href="https://github.com/w3c/epub-specs/pull/2709" rel="noopener noreferrer">w3c/<wbr>epub-specs#2709</a></h3>
<p><cite>wendyreid:</cite> Started this discussion last week, there is a open PR<br>
<span>… There was an idea to just replace XHTML with HTML and change mention to the XML serialization</span></p>
<p><cite>duga:</cite> I haven't commented more on the PR, I think Ivan was looking for something to add for HTML to see if people reacted<br>
<span>… if there was a problem, we could just pull it out before it gets released. Easy to revert.</span><br>
<span>… I'm worried no matter what we do, it'll be hard to pull this out, particularly the way its written now</span><br>
<span>… the current PR is only changing some of the stuff, its resulting in a broken build, a broken spc</span><br>
<span>… we want to get the changes in, but the more we breakup the change, the harder it is to pull it out later</span><br>
<span>… the spec is broken in the midst of the changes, that's less important, it leaves us in a weird place</span><br>
<span>… the downside of doing anything is that it will be hard to pull it out</span><br>
<span>… we'll add things to the spec over time, it'll be difficult to fix in the end</span><br>
<span>… and room for significant errors if we pull it out</span><br>
<span>… I don't think it's possible for us to do this in a way that's easy to revert later</span><br>
<span>… I just worry about rushing to this change and doing it in multiple parts, but we might be fooling ourselves if we think it's fixable later</span><br>
<span>… it's a radical solution too to change XHTML to HTML</span></p>
<p><cite>ivan:</cite> Let me pick it up from there<br>
<span>… My understanding is that our current spec is wrong</span><br>
<span>… because at w3c there is no such thing as XHTML</span><br>
<span>… there is just html, just xml serialization of xhtml</span><br>
<span>… So the right thing to do is to refer to html</span><br>
<span>… and then say that only the xml serialization is accepted</span><br>
<span>… This is more work, but even if we don't do raw html, as it makes the spec more in line with the other w3c specs</span></p>
<p><cite>mgarrish:</cite> I disagree, we aren't wrong, as we define xhtml ourselves<br>
<span>… But if we suddenly take the X out after so long, people won't read the definition and will think html is allowed</span><br>
<span>… But it requires people to pay attention to definitions to understand what really happened</span><br>
<span>… I do agree that the farther we go down that harder it is to back out</span><br>
<span>… But I am not that worried to back out what we have, it is just a few changes at the moment</span><br>
<span>… but yes once it is everything this will be really hard to back out</span></p>
<p><cite>SueNue:</cite> I think there are two competing ideas, one is supporting our community, the other is matching the  w3c<br>
<span>… I think we owe it to the community to leave the X in</span></p>
<p><cite>George:</cite> We should move forward until we get responses from the survey<br>
<span>… And then we can have tools to support our decision (e.g. epubcheck)</span></p>
<p><cite>wendyreid:</cite> This spec is used by people who aren't web people. What is a survey here asking?<br>
<span>… I think a lot of people who get such a survey won't know the difference between XHTML and html</span><br>
<span>… The big problem is toolchains, tools, and rendering systems</span><br>
<span>… Is it just pulling a few blocks of html? Or bigger?</span></p>
<p><cite>AvneeshSingh:</cite> I think we need to get the survey out ASAP<br>
<span>… we need to get this to aggregators</span><br>
<span>… Focus on them and reading systems</span></p>
<p><cite>Dale_R:</cite> Back to the notion of who is the audience<br>
<span>… When I first looked at making my book, I looked at the spec and saw xhtml</span><br>
<span>… I was surprised that I had to go back to xhtml</span><br>
<span>… So from my point of view it threw me off</span><br>
<span>… I would love to go forward with html</span><br>
<span>… I think there is a question if people are using an authoring system they may never see the difference</span></p>
<p><cite>SueNue:</cite> That was really interesting to hear that perspective<br>
<span>… I wonder if we change the doc and remove the X, will they make a document that works</span><br>
<span>… do we have people that make large quantities of epubs?</span></p>
<p><cite>duga:</cite> Interesting, and I want to home in on something SueNue said, and question for Dale, if you had read the spec it said HTML and buried somewhere it said "XML serialization", what would you have done with your first book ?</p>
<p><cite>Dale_R:</cite> I would have used modern HTML tools and code, and just followed the rest of the spec for other files</p>
<p><cite>duga:</cite> And I think that's the problem, it would not work<br>
<span>… the XHTML is such a speed bump because it says it everywhere, you know you have to do it. If we say HTML we'd get a lot of upset authors and broken content</span><br>
<span>… as much as I like the purity of saying HTML, I just worry we'll create more confusion than we want</span><br>
<span>… I also disagree that we're wrong, we could call it anything we want, as long as we define it we're fine</span></p>
<p><cite>CharlesL1:</cite> To SueNue point, are the major players who are tool providers<br>
<span>… Adobe is one</span><br>
<span>… as far as the other vendors, we have agreements with a lot of them</span><br>
<span>… if we do a survey, I can make sure they see it, as well as gca publishers</span></p>
<p><cite>mgarrish:</cite> I just threw out the idea of the survey this morning<br>
<span>… We did this before.</span><br>
<span>… Are authors using xhtml just because they have to?</span><br>
<span>… It would be great to have data that shows that</span></p>
<p><cite>wendyreid:</cite> I like the idea of a survey<br>
<span>… We want to target to specific stake holders</span><br>
<span>… for instance targets are content authors, another for conversion houses, distributors, renderers</span></p>
<p><cite>ivan:</cite> I understand all the arguments<br>
<span>… to be precise we do not define xhtml (I just checked)</span><br>
<span>… We just refer to the html spec, which at the end of a link chain takes us to a place in the html specification that says don't use it</span><br>
<span>… The spec itself still looks strange</span></p>
<p><cite>wendyreid:</cite> One thing we can clarify now for the survey is what a no-go point is<br>
<span>… we will hear that people don't want to do this</span><br>
<span>… tools and tech evolve, so more companies may be forced into it eventually</span><br>
<span>… is it too hard to really do, vs we can, but we don't want to</span></p>
<p><cite>AvneeshSingh:</cite> Let's get it out, maybe there will be less pushback<br>
<span>… This isn't really a show stopper as I understand it, maybe it will just inform how we approach the change</span><br>
<span>… back to what SueNue said, people have to pay to be on this group, hence the need for outreach</span></p>
<p><cite>duga:</cite> Thinking about the survey, thinking about the change, and the reality, the reality is that XHTML is dead, it's been dying for a while, no new features are making it from HTML to XHTML<br>
<span>… so if anything changes in HTML, it won't make it to the XML serialization</span><br>
<span>… because XHTML is dead, we need to make the change, and it will only make using the spec harder if we don't make the change</span><br>
<span>… the survey is also a PR or marketing push to tell people "we're doing this eventually, now might be the time" and we need them on board</span><br>
<span>… unless there are literal fires, telling them this is coming</span><br>
<span>… we need to use the wording of the survey to convince people to do it</span></p>
<p><cite>Dale_R:</cite> Listening to this, it reminds me of current news events (e.g. cobol used in government)<br>
<span>… And it reminds me of the browser wars, and sticking to the spec regardless of outcome</span><br>
<span>… Is our job to say how it is done, or how it should be done?</span></p>
<p><cite>George:</cite> two things - there are a11y innovations around complex images, etc that would benefit from html<br>
<span>… And people working on that are in html world</span><br>
<span>… Second, maybe our survey should be targeted at "what is the difficulty of switching" instead of "s hould we"</span></p>
<p><cite>wendyreid:</cite> we should use this as a pr tool. E.g. as we are making this change how will this impact you? What will the problems be? How can we help?<br>
<span>… I will take up the task of making a doc to brainstorm on this</span></p>
<p><cite>ivan:</cite> I propose we merge the current pr with the minimal change<br>
<span>… Then we can reference that</span><br>
<span>… Then someone has to take point on the survey, and we need to decide how long we have for submittals</span><br>
<span>… No matter what we have to make some changes</span><br>
<span>… as the current spec is a bit dirty in this regard</span></p>
<p><cite>&lt;AvneeshSingh&gt;</cite> Just to remind July and August are holidays season in Europe and US. So, ideally we should have most responses within June</p>
<p><cite>CharlesL1:</cite> Some publishers are making EPUB 2 files.<br>
<span>… So this will be a big shock</span><br>
<span>… If we did this is it an epub 3.4 vs 4 change</span></p>
<p><cite>SueNue:</cite> I like the idea of this is happening and how can we help as the focus</p>
<p><cite>SueNue:</cite> As a group it would help to have our talking points ready</p>
<p><cite>tzviya:</cite> I understand this is a big processing change, but if we market this as a "don't panic" change<br>
<span>… then we are covered from the publisher side. For reading system side we need to document how to do this</span><br>
<span>… We can't stay in 1988 forever! Documentation is our friend, and we have to do this</span></p>
<p><cite>ivan:</cite> One thing, if this is epub 3.4 or 4, I push back on it being 4</p>
<p><cite>&lt;tzviya&gt;</cite> +1 - this is not EPUB 4</p>
<p><cite>ivan:</cite> It is still a gradual improvement<br>
<span>… We need to make sure it is not major change, XHTML is still allowed, 3.3 doc is still valid</span></p>
<p><cite>wendyreid:</cite> I will start the doc and share it for comments and addition<br>
<span>… We will help wrap peoples heads around what the change means</span></p>
<p><cite>duga:</cite> If we merge the PR can I review first?</p>
<p><cite>&lt;George&gt;</cite> +1</p>
<p><cite>wendyreid:</cite> AOB?</p>

<hr /></details>

epub34/authoring/index.html
```diff
@@ -5798,27 +5810,39 @@ <h4>NCX</h4>
 			<h2>EPUB content documents</h2>
 
 			<section id="sec-xhtml">
-				<h3>XHTML content documents</h3>
+				<h3>HTML content documents</h3>
+
+				<div class="ednote">
+					<p>Support for the HTML syntax is still an open issue in the EPUB 3.4 revision. It is <b><i>at
```
bduga: Nit: Should there be a reference to the issue? The may just be no.

epub34/authoring/index.html
```diff
@@ -5798,27 +5810,39 @@ <h4>NCX</h4>
...
+					<p>To provide feedback on this change, please open issues in the <a
+							href="https://github.com/w3c/epub-specs/issues">GitHub tracker for the EPUB
+							specifications</a>.</p>
+				</div>
 
 				<section id="sec-xhtml-intro" class="informative">
 					<h4>Introduction</h4>
 
-					<p>This section defines a profile of [[html]] for creating [=XHTML content documents=]. An instance
-						of an XML document that conforms to this profile is a [=core media type resource=] and is
```
bduga, line 5807: This loses the part about it being a core media type resource. Was that intentional? Even if it was, perhaps leave it in since it is unrelated to the change to html. So drop "...and is referred to in this specification as an XHTML content document", and change xml to html.

epub34/authoring/index.html
```diff
@@ -5829,14 +5853,19 @@ <h4>XHTML requirements</h4>
 								defined in <a href="#sec-xhtml-extensions"></a>, and MUST conform to all content
 								conformance constraints defined therein.</p>
 						</li>
+						<li>
+							<p id="confreq-cd-html-docprops-syntax">MUST conform to the <a
```
bduga: Does this need a complementary conformance requirement for html syntax documents conforming to the to that syntax? I don't think we need to add it for this review, it is probably a question for the group, but I am not really sure where or how to add issues regarding the switch to html. Do you know or have preferences? 

More notes: html breaks this up into writing and parsing sections. The writing section doesn't seem to talk about conformance, if we add a conformance requirement we should be explicit about what it means. That is, does a document have to pass some specific conformance criteria?

bduga (approved):
Should the entry around line 640 be updated? Or do you want to wait for a later change to do that? And is there really no way to comment on an unchanged line in a github review?!

epub34/authoring/index.html
mattgarrish: We don't have one. It's an issue in the charter only. It might be good to open one so that people have a place for general comments about whether to include it or not.

epub34/authoring/index.html
mattgarrish: Intentional in that it was an xml-only statement that seems to only be there to reinforce that only the xml syntax is valid. I don't know that we need it when we support both (the normativeness of the statement is defined elsewhere), but I'm fine with generalizing it and keeping it, as we're left with a pretty lonely sentence of an introduction without it. But I assumed that might change later if we write a better intro to describe support for both.

mattgarrish:
> Should the entry around line 640 be updated? Or do you want to wait for a later change to do that?

I'd rather wait to change it. If we change the name of the entry to "html content document" then all the references to it have to change, too, because respec uses the text for linking. If we change the definition to be broader but keep it called "xhtml content document" then that's probably going to be more confusing.

> And is there really no way to comment on an unchanged line in a github review?!

Not if it's more than a few lines past a change, nope.

epub34/authoring/index.html
mattgarrish: > I am not really sure where or how to add issues regarding the switch to html. Do you know or have preferences?

I'd open them as regular issue on the epub spec but add a new label so we can group and track them (something like "HTML Transition"). We probably can't implement any resolutions until we know we're moving ahead, but that shouldn't stop us from trying to figure out solutions.

To your question, though, when I looked at the requirements I was thinking that the bullet "MUST conform to the conformance criteria for all document constructs defined by [html] unless explicitly overridden" already encompassed the syntax, but that's probably more generally about content models than syntax itself. Adding a requirement to follow [13.1](https://html.spec.whatwg.org/multipage/syntax.html#writing) sounds like a better approach.

mattgarrish: And I think we can implement that change now, since we've already modified this section a bit. It's not a section that usually gets modified, so backing these changes out should be easy.

mattgarrish: > I'd open them as regular issue on the epub spec but add a new label so we can group and track them

Or maybe this is an opportunity to try the sub issues feature of github? If we open a general "move to html" issue, maybe we try opening sub issues on that?  (I haven't seen sub issues done before, so no idea if it will be practical or not.)

bduga: Let me mess around and see what tools we have. I kind of like having nested bugs, but tags would work as well.

epub34/authoring/index.html
bduga: Created issue #2715 which you could reference.

mattgarrish: > Created issue #2715 which you could reference.

Thanks, I've add links to it in the last commit. I also added the html syntax bullet (and updated the xml syntax to also point to its authoring section), and restored the line to the intro.

mattgarrish:
> If we change the name of the entry to "html content document" then all the references to it have to change, too

Come to think of it, we can use the data-lt attribute to make "xhtml content document" and "xhtml content documents" aliases of "html content document". That would let us revise the definition and not have to change references.

mattgarrish:
Any other comments on this? I'll merge it around 21UTC if I don't hear anything.

PR #2710, opened 2025-04-08
Remove version numbers from issue templates

mattgarrish:
Putting specific version numbers in titles and urls was probably a bad idea in hindsight. I've stripped them down so that we don't have to revise these every time we bump a version number as the issues are unlikely to ever be version bound.


iherman:
But what should one do to raise an issue against 3.4? Isn't it better to keep the 3.3 specific entries untouched and clone them for 3.4? (The problem does not arise for the notes that are not version bound...)

mattgarrish:
> But what should one do to raise an issue against 3.4?

In 15 years the only issue I've seen opened that was specific to a single version of a spec is the still-open one about making the Accessibility 1.0 requirement for a summary match the 1.1.

Nothing else was version specific. Someone may be looking at the 3.3 REC and open an issue, but it's going to be resolved in 3.4 so why add to the confusion about which template to pick?

In any case, if there's any value to state the version it's simpler to ask for it along with the section number in the first question.

mattgarrish:
What if we switch to using github forms for these? It would be possible to add a dropdown list of version numbers to pick from that way. Here's a test I did on the EPUB 3 template: https://github.com/mattgarrish/epub-specs/issues/new?template=01-epub.yml

iherman:
> > But what should one do to raise an issue against 3.4?
> 
> In 15 years the only issue I've seen opened that was specific to a single version of a spec is the still-open one about making the Accessibility 1.0 requirement for a summary match the 1.1.
> 
> Nothing else was version specific. Someone may be looking at the 3.3 REC and open an issue, but it's going to be resolved in 3.4 so why add to the confusion about which template to pick?

That was not my point. You replaced the references to the spec to `https://www.w3.org/TR/epub/`. But, until we get to CR, that URL points to the static 3.3 version and not to 3.4. That can lead to confusion...

mattgarrish:
> But, until we get to CR, that URL points to the static 3.3 version and not to 3.4.

What's the point of that link, though? I'd strip those out entirely. Who opens an issue and needs the issue itself to tell them how to get to the specification?

iherman:
> > But, until we get to CR, that URL points to the static 3.3 version and not to 3.4.
> 
> What's the point of that link, though? I'd strip those out entirely. Who opens an issue and needs the issue itself to tell them how to get to the specification?

It might entice them look for a more specific statement in the spec beyond something like "EPUB does not work!" 😀

Actually... it would force people to raise issues in 3.4 and _not_ 3.3 which we will never touch again... even if they are not familiar with the exact status (or the very existence) of 3.4.

mattgarrish:
> It might entice them look for a more specific statement in the spec beyond something like "EPUB does not work!" 😀

That's why I'm wondering if we should move to the forms. You can make fields required before a person can submit, so we could force the person to put something in the section field. The validation is currently pretty basic as you can only require fields, but maybe in the future they'll add things like the pattern attribute so we could make it a required link.

mattgarrish:
> Actually... it would force people to raise issues in 3.4 and _not_ 3.3

Maybe that should be a warning in the prose instead of a link in the heading? Please check the current editor's draft to ensure your issue is still relevant before opening an issue.

It leaves us stuck back in the spot where we have to modify the template every time we release a REC or start a new revision, but at least that text/link wouldn't wind up in the issue itself. It would only exist in the form template.

wareid (approved):

iherman:
Two points:

- For the documents with "revisions" (the Recs, plus the overview and the a11y techniques) I am still worried that we will get issues that would refer to sections, headings, etc., extracted from the 3.3 versions, because they will ignore the separate warning. But I guess the best is to see if this indeed happens, or whether I am just paranoiac. Ie, I am o.k. merging this, and we will cross the bridge if and when we get there.

    That being said, I wonder whether it is not clearer to refer to the Working Draft, using the URL on /TR. After all, the notion of a separate editors' draft has become pretty meaningless with the usage of echidna.
- My last remark on echidna also applies for the notes without "revisions" (TTL, FXL-A11y, etc). For those, the reference to the editors' draft is superfluous: whatever issue one finds in the official document is also valid for the editors' draft, because the two are (apart from the header) identical...

mattgarrish:
> * I am still worried that we will get issues that would refer to sections, headings, etc., extracted from the 3.3 versions, because they will ignore the separate warning

Ya, but I don't know that we can stop that. And it's not really wrong to open an issue that cites a problem in 3.3. That's the current recommendation, so it ultimately will fall on us to figure out how to remap the problem. I'd be more worried if we were planning a huge reorg like we did in 3.3 and nothing from 3.2 mapped directly to the new documents. Things are pretty stable now.

> That being said, I wonder whether it is not clearer to refer to the Working Draft, using the URL on /TR. After all, the notion of a separate editors' draft has become pretty meaningless with the usage of echidna.

True. Good point. I'll tweak them.

Beyond the linking stuff, which we can revisit as we go, what I prefer about using the forms is that it takes all the extraneous stuff that's guidance out of the final issue instead of requiring the commenter to delete it. Only the labels and what text the commenter adds appear in the issue. That's huge for the general template where we have all those spec links and references to other help resources.

mattgarrish:
For the last commit, I removed the links from the headings since they're mostly duplicating the links to the latest version that are now in the warning and changed the links to either the latest numbered version (for 3.4-specific notes and recs) or their unversioned short names (for all other notes).

iherman (approved):
> For the last commit, I removed the links from the headings since they're mostly duplicating the links to the latest version that are now in the warning and changed the links to either the latest numbered version (for 3.4-specific notes and recs) or their unversioned short names (for all other notes).

The only one that made me pause is the last one (60-general.yml): shouldn't we use the 3.4 version references for those as well?

mattgarrish:
Fixed. I also rearranged the documents so the RECs are first followed by the core notes followed by the notes in general order of their being worked on and/or likely to still get issues.

PR #2722, opened 2025-05-14
Updated the ack script configuration files

iherman:
I have also generated new acknowledgement files. Some notes

- I have put everything into `/common` (it used to be `epub33/common`, but we moved the other files, like css and js, into `/common`. That means the generated acknowledgement files are also there.
- I have kept the three versions: simple ack file, one dedicated to a11 with a text on Ben, and one dedicated to the core EPUB specs and dedicated to Garth. I just could not get myself to remove these; what do others think. (Matt, you may want to look at the template file whether it needs a different text.)


mattgarrish:
> * I just could not get myself to remove these; what do others think.

Maybe we should move these to a special dedications subsection and try to merge/shorten them so we can use a single file for all the documents? Not to sound morbid, but what if we need to make other dedications?

mattgarrish:
If we have two different acknowledgement files, we'll also need to fix the accessibility specification to point to the new one and update its build script to watch for changes. That's why I'd like to get us back to one file if we can.

mattgarrish:
Here's a possibility for a paragraph we might include after the list of current contributors, leveraging as much of the existing language as seemed appropriate:

> The Publishing Maintenance Working Group would also like to make a special acknowledgment to notable
contributors, and friends, we have lost along the way. In particular, EPUB, not just EPUB 3, would not be what is
today without the vision, knowledge, and preternatural good nature of Garth Conboy. Similarly, Ben Schroeter
will always be remembered for his tireless work to help EPUB reach its potential as a universally accessible
format. We are forever in their debt.

iherman:
I like the text. My understanding of your proposal is that we would have a _single_ acknowledgment file (and template), ie, all docs would include this text; for epub3.3, we had a separate text referring to Garth for the content and RS, and a separate one for Ben in a11y. Is that correct? I am happy to make that change (from hindsight, it would have been better for EPUB 3.3 as well...).

mattgarrish:
> we would have a _single_ acknowledgment file (and template), ie, all docs would include this text;

Right, the revised text keeps the memorials without making the dedications specific to a particular document/revision number. This way we can carry them forward and use a single file for all the REC documents.

iherman:
Ok, I am fine with that. I will take care of the update for the PR later today.

iherman:
> Here's a possibility for a paragraph we might include after the list of current contributors, leveraging as much of the existing language as seemed appropriate:
> 
> > The Publishing Maintenance Working Group would also like to make a special acknowledgment to notable
> > contributors, and friends, we have lost along the way. In particular, EPUB, not just EPUB 3, would not be what is
> > today without the vision, knowledge, and preternatural good nature of Garth Conboy. Similarly, Ben Schroeter
> > will always be remembered for his tireless work to help EPUB reach its potential as a universally accessible
> > format. We are forever in their debt.

Done in https://github.com/w3c/epub-specs/pull/2722/commits/a6304b90c45c0ea5a8674a339f600c1678aaf6e7

mattgarrish:
I wrote it thinking it would go after the list of current contributors; that's why I put "would also like to".

If you want it before the list, then that part should be reworded as there's no "also" if there are no acknowledgements before it.

iherman:
Oops.  I changed it by removing "also". (I followed the same order as was in the epub33 spec.)

mattgarrish (approved):

iherman:
@sueneu @shiestyle @wareid can you look at this PR? It is mostly editors' stuff, except for the issue of how to handle memorials to Ben and Garth in 3.4 (and possibly onwards). That should not be a purely editors' decision...

shiestyle (approved):
I have no objection to this.

sueneu:
The memorials are lovely, I’m fine with the text and the order of the acknowledgments

iherman:
No problems raised, comments incorporated, merging.

PR #2723, opened 2025-05-15
Add erratum to 1.0 and backwards compatibility statement to 1.1.1

### LINKED ISSUE #2526: Make accessibility summaries recommended in 1.0
It's been noted that the one issue that prevents a publication that conforms to 1.1 at WCAG 2.0 Level A, AA or AAA from being fully backwards compatible with 1.0 (both the IDPF and ISO versions) is that we reduced the requirement to have a summary to a recommendation. Consequently, only if your 1.1 conforming publication has a summary are you compatible.

This lack of full backwards compatibility with the ISO version may be an issue in some regions.

To fix this situation, all we should need to do is issue an erratum for the 1.0 specification to lower the requirement to a recommendation. It might be worth noting compatibility in the 1.1 specification, although this isn't strictly required.

mattgarrish:
This pull request updates the errata page for the EPUB Accessibility 1.0 specification to add an entry making the inclusion of an accessibilitySummary a recommendation instead of a requirement. It also notes that repeating information in the other metadata is not necessary, but this was never a requirement so publications that did repeat the info aren't non-conforming.

(The errata page will be published through a different process than the W3C specs.)

I've also re-integrated the changes from pull request #2527 that add a section on backwards compatibility.

Fixes #2526 

* * *

Accessibility 1.0 Errata: [Preview](https://raw.githack.com/w3c/epub-specs/errata/a11y10/archive/a11y/errata/index.html)
Accessibility 1.1.1: [Preview](https://raw.githack.com/w3c/epub-specs/errata/a11y10/epub34/a11y/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/errata/a11y10/epub34/a11y/index.html)


mattgarrish:
For reference, I copied the formatting of the errata from the content documents page: https://idpf.org/epub/301/spec/epub-contentdocs-errata/

iherman (approved):

iherman (approved):

mattgarrish:
Hm, I just remembered that we switched the requirements for access mode and access mode sufficient in 1.2.

That means that 1.2 content isn't backwards compatible, although it's not critical because each version has its own identifiers. But it means I need to amend this PR to take out the proposed section on compatibility.

PR #2731, opened 2025-06-02
Specify the format of the evaluation date

mattgarrish:
This pull request clarifies that the dc:date value for the evaluation has to <strike>match one of the patterns: YYYY-MM-DD, YYYY-MM, or YYYY</strike> be an iso 8601 conformant datetime.

***

Accessibility 1.1.1: [Preview](https://raw.githack.com/w3c/epub-specs/fix/issue-2725/epub34/a11y/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fix/issue-2725/epub34/a11y/index.html)

### DIFF
--- a/epub34/a11y/index.html
+++ b/epub34/a11y/index.html
@@ -1909,8 +1909,8 @@
 				<summary>Substantive changes since <a href="https://w3.org/TR/epub-a11y-11/">EPUB Accessibility
 					1.1</a></summary>
 				<ul>
-					<li>02-June-2025: Make explicit that the evaluation date has to match a year-month-day pattern.
-						Refer to <a href="https://github.com/w3c/epub-specs/issues/2725">issue 2725</a>.</li>
+					<li>02-June-2025: Make explicit that the evaluation date has to conform to iso 8601. Refer to <a
+							href="https://github.com/w3c/epub-specs/issues/2725">issue 2725</a>.</li>
 				</ul>
 			</details>
 		</section>
--- a/epub34/a11y/index.html
+++ b/epub34/a11y/index.html
@@ -1369,7 +1369,7 @@
 							expressed in W3C Date and Time Formats [[datetime]], as such strings are both human and
 							machine readable.</p>
 
-						<div class="issue" data-number="1725">
+						<div class="issue" data-number="1725" title="Evaluation date format">
 							<p>The change to require the form of the evaluation date depends on the standard version
 								number changing, and as a consequence there being new conformance identifiers.
 								Otherwise, this change could invalidate existing content. The requirement could be
--- a/epub34/a11y/index.html
+++ b/epub34/a11y/index.html
@@ -1370,11 +1370,11 @@
 
 						<dl>
 							<dt><code>YYYY-MM-DD</code></dt>
-							<dd> The year, month, and date of the evalation. This format is preferred. </dd>
+							<dd>The year, month, and day of the evaluation. This format is preferred.</dd>
 							<dt><code>YYYY-MM</code></dt>
-							<dd> If only the year and month of the evaluation are known. </dd>
+							<dd>If only the year and month of the evaluation are known.</dd>
 							<dt><code>YYYY</code></dt>
-							<dd> If only the year of the evaluation is known. </dd>
+							<dd>If only the year of the evaluation is known.</dd>
 						</dl>
 
 						<aside class="example" title="Expressing the evaluation date">
--- a/epub34/a11y/index.html
+++ b/epub34/a11y/index.html
@@ -1365,17 +1365,16 @@
 								href="https://www.w3.org/TR/epub/#subexpression">associated with</a> [[epub-3]] the <a
 								href="#sec-evaluator-name">evaluator's name</a>.</p>
 
-						<p>For human readability and localization purposes, the value of the property MUST be a string
-							that matches one of the following [[iso8601-1]] patterns:</p>
+						<p>It is REQUIRED that the date string conform to [[iso8601-1]], particularly the subset
+							expressed in W3C Date and Time Formats [[datetime]], as such strings are both human and
+							machine readable.</p>
 
-						<dl>
-							<dt><code>YYYY-MM-DD</code></dt>
-							<dd>The year, month, and day of the evaluation. This format is preferred.</dd>
-							<dt><code>YYYY-MM</code></dt>
-							<dd>If only the year and month of the evaluation are known.</dd>
-							<dt><code>YYYY</code></dt>
-							<dd>If only the year of the evaluation is known.</dd>
-						</dl>
+						<div class="issue" data-number="1725">
+							<p>The change to require the form of the evaluation date depends on the standard version
+								number changing, and as a consequence there being new conformance identifiers.
+								Otherwise, this change could invalidate existing content. The requirement could be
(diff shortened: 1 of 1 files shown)


clapierre (approved):

epub34/a11y/index.html
```diff
@@ -1365,6 +1365,18 @@ <h5>Evaluation date</h5>
 								href="https://www.w3.org/TR/epub/#subexpression">associated with</a> [[epub-3]] the <a
 								href="#sec-evaluator-name">evaluator's name</a>.</p>
 
+						<p>For human readability and localization purposes, the value of the property MUST be a string
+							that matches one of the following [[iso8601-1]] patterns:</p>
+
+						<dl>
+							<dt><code>YYYY-MM-DD</code></dt>
+							<dd> The year, month, and date of the evalation. This format is preferred. </dd>
```
bduga: s/evalation/evaluation/

bduga: I think ISO 8601 refers to the DD portion as "day" or "calendar day", or even more precisely "ordinal number of a calendar day". Maybe change this to "year, month, and day" instead of "year, month, and date".

bduga (approved):
lgtm

epub34/a11y/index.html
mattgarrish: Done, thanks!

iherman:
This was discussed during the [pmwg meeting on 26 June 2025](https://w3c.github.io/pm-wg/minutes/2025-06-26.html#7586).

<h3>PR 2731 - <a href="https://github.com/w3c/epub-specs/pull/2731" rel="noopener noreferrer">w3c/<wbr>epub-specs#2731</a></h3>
<p><cite>wendy:</cite> seeing none, we will move on</p>
<p><cite>wendyreid:</cite> formulating the evaluation dates</p>
<p><cite>mgarrish:</cite> we didn't formalize a format for the date<br>
<span>… the problem is how to localize the date when we don't know the format of the date</span><br>
<span>… we could choose an existing format, and likely exclude the time</span><br>
<span>… and will this affect existing content</span><br>
<span>… and have we already committed to bumping the conformance number</span><br>
<span>… if not, this affect all existing content that conforms to the current version</span><br>
<span>… are we proceeding on requiring access mode sufficient and is it OK to add this when we bump up the number?</span></p>
<p><cite>AvneeshSingh:</cite> we will discuss this in an upcoming meeting in the WG<br>
<span>… we will elevate it to a must follow and see how the community responds</span><br>
<span>… we are moving ahead with precautions ACE will flag it</span><br>
<span>… I am worried if we make the date a must at this point, we would get downstream errors</span></p>
<p><cite>mgarrish:</cite> maybe we keep a separate list of items that we would return to if we change conformance numbers</p>
<p><cite>Hadrien:</cite> I don't see the point of this change as a must statement<br>
<span>… what is the value of this change to the user? When you localize the date</span><br>
<span>… you can choose what to display. At best this could be a "should"</span><br>
<span>… especially since there are potential validation problems</span></p>
<p><cite>mgarrish:</cite> I see what you're saying, but at this time we have no suggestion about how to express the date</p>
<p><cite>Hadrien:</cite> as long as it is based on an existing format, we can work with it</p>
<p><cite>George:</cite> identifying the date format seems like standards work<br>
<span>… if we change the conformance number, it gives us more freedom</span></p>
<p><cite>CharlesL:</cite> the reading system or bookstore can always display it how they choose<br>
<span>… having a standard format will help it be changed to the local language</span><br>
<span>… so we need to go with the ISO</span><br>
<span>… as far as requiring it, we can start requiring, it has been optional, but we tell everyone to put it in</span><br>
<span>… we can discuss this in our next WG call</span><br>
<span>… I agree with George that we should plan to update the conformance number</span></p>
<p><cite>mgarrish:</cite> if we don't make it a requirement we will never really be sure what we get when processing this data<br>
<span>… if we made it a must we would know we could trust the required date format after a certain point</span><br>
<span>… we don't know what the date will contain if we don't do some testing</span></p>
<p><cite>AvneeshSingh:</cite> does the backward compatibility clause get violated if we make these kinds of changes</p>
<p><cite>ivan:</cite> the conformance in the charter never really explicitly refers to the accessibility metadata, we are in a gray area<br>
<span>… it refers to in practice what epub check checks</span></p>
<p><cite>CharlesL:</cite> similar to "conforms to" if you have that then you must put in "certified by"<br>
<span>… maybe it could be if you include a certified date it should conform to an ISO standard</span></p>
<p><cite>mgarrish:</cite> that is what this change does</p>
<p><cite>wendyreid:</cite> This change is OK, but we shouldn't make it a must?</p>
<p><cite>mgarrish:</cite> right now we will accept any date format from the date section, make it a "should" and see what happens</p>
<p><cite>CharlesL:</cite> I don't know that we need to include the date/time, but its OK as long as the time is ISO standard</p>
<p><cite>Hadrien:</cite> One of the reasons we should be looser about the format is we don't know what people have in their systems<br>
<span>… if people use a timezone for instance, we can deal with that</span><br>
<span>… we shouldn't force things that aren't necessary</span></p>
<p><cite>&lt;CharlesL&gt;</cite> +1 to Hadrien</p>
<p><cite>Wendy:</cite> any opposition to making this a "should" and adding date/time</p>
<p><cite>AvneeshSingh:</cite> let's wait to merge this until after the Accessibility WG</p>
<p><cite>&lt;CharlesL&gt;</cite> I also agree this should be a MUST</p>
<p><cite>duga:</cite> I prefer making it a must and updating the conformance number</p>
<p><cite>mgarrish:</cite> if it isn't a must, ACE can bump it up as a serious error</p>
<p><cite>shiestyle:</cite> This date is not mandatory, right?<br>
<span>… from a publisher's point of view we generate epub, and we have to change it because the date appeared, many publishers will use this, but this looks like tricky metadata</span></p>
<p><cite>mgarrish:</cite> it isn't required, but it is a good record, for knowing when their accessibility was last done</p>
<p><cite>mgarrish:</cite> the more important part is getting the accessibility review is done</p>
<p><cite>wendyreid:</cite> if you're running ACE, you'll get an error, but not from EPUB check</p>
<p><cite>CharlesL:</cite> does ONIX have a similar field?</p>
<p><cite>&lt;wendyreid&gt;</cite> <a href="https://ns.editeur.org/onix/en/196/91" rel="noopener noreferrer">https://<wbr>ns.editeur.org/<wbr>onix/<wbr>en/<wbr>196/<wbr>91</a></p>
<p><cite>wendyreid:</cite> it is number 91, latest accessibility assessment date<br>
<span>… so we'll let the Accessibility TF talk about this first, if they are happy, then we can merge it</span></p>

<hr /></details>

mattgarrish:
After discussing with both the WG and CG, the current proposal is to keep the requirement for now but modify it to allow both date and time. We're also adding an issue to the specification to alert people that the change depends on the version number changing so if there is pushback against that change the requirement would have to drop.

mattgarrish:
I'm going to merge this and #2746 as we've discussed pretty thoroughly now. Until we merge them, we can't get any further feedback from the issue markers that will stay in them.

PR #2734, opened 2025-06-04
Add the prefix URL for the structural semantics vocabulary

### LINKED ISSUE #2733: Prefix URL for the structure vocab is not listed
### Section

https://www.w3.org/TR/epub-34/#sec-epub-type-attribute

### Describe the problem

When we specify default vocabularies, we typically provide their prefix URL, but epub:type in content documents just says that the structural semantics vocabulary is the default vocabulary. The section doesn't include a prefix URL for expanding the terms.

If I go back to 3.0.1, although there still isn't an explicit statement of the URL, it's clear from the references section that we intended the prefix URL to be:

> http://www.idpf.org/epub/vocab/structure/#

That matches the pattern used for the other default vocabs we established.

We should make this clearer, though, for anyone looking to create full URIs from the values.

(The omission was brought to my attention by the Readium folks, so it's not just a hypothetical need.)

mattgarrish:
Makes the idpf.org URL for expanding terms explicit in the document so there is no confusion about what to use now that the vocabulary is hosted under a W3C URL.

We should probably see about getting http://www.idpf.org/epub/vocab/structure/# to redirect to the new SSV location. It currently goes to the old idpf.github.io location. Even though that document has an obsolete message attached to it, it doesn't need to be accessible anymore.

A simple fix would be to temporarily unarchive that repository and replace that old document with a meta refresh directive so that readers are automatically bumped forward. The more complicated fix would be to get the old idpf.org htaccess file updated.

Fixes #2733

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2734.html" title="Last updated on Jun 9, 2025, 3:35 PM UTC (2925651)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2734/0b5be5d...2925651.html" title="Last updated on Jun 9, 2025, 3:35 PM UTC (2925651)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -10034,7 +10034,7 @@ html.my-document-playing * {
 
 				<p>The <a href="#sec-default-vocab">default vocabulary</a> for the <code>epub:type</code> attribute is
 					the EPUB&#160;3 Structural Semantics Vocabulary&#160;[[?epub-ssv-11]]. The prefix URL for
-					referencing its properties is <code>http://www.idpf.org/epub/vocab/structure/#</code>.</p>
+					referencing its properties is <code>http://idpf.org/epub/vocab/structure/#</code>.</p>
 
 				<p>EPUB creators MAY include unprefixed terms that are not part of this vocabulary, but the preferred
 					method for adding custom semantics is to use <a href="#sec-prefix-attr">prefixes</a> for them. Refer
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -10033,10 +10033,12 @@ html.my-document-playing * {
 					document instance.</p>
 
 				<p>The <a href="#sec-default-vocab">default vocabulary</a> for the <code>epub:type</code> attribute is
-					the EPUB&#160;3 Structural Semantics Vocabulary&#160;[[?epub-ssv-11]]. EPUB creators MAY include
-					unprefixed terms that are not part of this vocabulary, but the preferred method for adding custom
-					semantics is to use <a href="#sec-prefix-attr">prefixes</a> for them. Refer to <a
-						href="#sec-vocab-assoc"></a> for more information.</p>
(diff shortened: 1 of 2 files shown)


iherman (approved):

mickael-menu:
Thank you @mattgarrish, this is helpful.

I have one question regarding the presence of `www` in the URL. Are they expected? I saw different versions online, but this document references an URL without `www`:

https://idpf.github.io/epub-vocabs/package/meta/#sec-ref

> Properties without a prefix are referenceable using the base IRI http://idpf.org/epub/vocab/package/#.

mattgarrish:
> I have one question regarding the presence of `www` in the URL. Are they expected?

Ya, it looks like we've flipped between having www and not having it with the vocabularies. Even searching in 3.4 specification I can find some vocabulary URLs that have it and others that don't.

I'm inclined to standardize them all without www. in them.

It has no effect on the authoring side, so the only impact is on tools that want to expand the values to full URLs. Doing that, and for what purpose, is at your discretion, so it's not like changing xml namespaces where the wrong URL could lead to failed rendering. I've also only heard of Readium expanding the values, so if this works for you I don't think this change will have any wider impacts.

mickael-menu:
Yes, we went ahead and used the version without `www`.
It's not really used yet, but it's a way for people to access the original EPUB values in the Readium Web Publication Manifest when we don't have a native mapping. It doesn't really matter as long as we (Readium) stay consistent. However, we prefer to follow the specification accurately.

mattgarrish:
And I've been forgetting that we wrote an algorithm in the reading systems specification last revision on how to obtain an expanded URL from a value: https://www.w3.org/TR/epub-rs-34/#dfn-obtain-an-expanded-url

That algorithm omits www from all the vocabularies, so the inconsistency is only in the authoring specification. Changing the references in the authoring document to match is more justifiable given that parsing expectation. (I really like it when changes are not seen as arbitrary!)

Thanks again for pointing this out.

mattgarrish:
Merging this one, as well, as it's a minor and obvious clarification.

PR #2735, opened 2025-06-04
Allow ITS attributes in content documents

### LINKED ISSUE #2732: Add the ITS Attribute set to the as an allowed HTML extension item
### Section

https://www.w3.org/TR/epub-34/#sec-xhtml-extensions

### Describe the problem

The [§6.1.3 section](https://www.w3.org/TR/epub-34/#sec-xhtml-extensions) on HTML extensions lists some features that can be used in an HTML content document without further ado (ie, without additional definitions like namespaces). Epubcheck checks these and rejects attributes that are not listed in this section. Currently, the (non-deprecated) extensions are the usage of `epub:type` (`epub-type` for non-XML) attribute and the RDFa attributes. Additionally to these, though not listed in the sections, the `aria-*` attributes are also usable the same way (as defined in the HTML specification proper).

However, the [Internationalization Tag Set](https://www.w3.org/TR/its20/) W3C standard defines an attribute set which is not listed in EPUB. These attributes are used, among other things, for the creation and the processing of multilingual content, giving hints for translations, etc. For example, the [Requirement for Chinese Text Layout Note](https://www.w3.org/TR/clreq/) uses one of the ITS attributes to switch the text between traditional and simplified Chinese characters, as well as English. The switch is done via a tiny local JavaScript that relies on that attribute. (The example is an HTML page, but it can be converted into EPUB.)

The problem therefore is that an EPUB instance using the ITS attribute is invalid. I believe that is a problem

The proposal is to add a separate sub section to [§6.1.3 section](https://www.w3.org/TR/epub-34/#sec-xhtml-extensions) allowing the usage of all the `its-*` attributes in an HTML content document.

(Note that other, similar specifications at W3C may come to the fore at some point. We may think about a somewhat more general way of making these extensions possible, without being forced to take each individual extension one-by-one. But we may decide to control this in the future, too, and handling these one-by-one.)

@mattgarrish @rdeltour @xfq

mattgarrish:
With optional reading system processing support.

We should wait on merging this until the group has a say, though.

Fixes #2732 

***

Reading Systems: [Preview](https://raw.githack.com/w3c/epub-specs/feature/issue-2732/epub34/rs/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/feature/issue-2732/epub34/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2735.html" title="Last updated on Jun 25, 2025, 7:27 PM UTC (0e1ff55)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2735/39712aa...0e1ff55.html" title="Last updated on Jun 25, 2025, 7:27 PM UTC (0e1ff55)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -5888,7 +5888,21 @@ No Entry</pre>
 								both are natively supported.</p>
 						</div>
 					</section>
-
+					
+					<section id="sec-xhtml-its">
+						<h5>International tag set (ITS)</h5>
+						
+						<p>The [[its20]] specification defines a set of attributes that [=EPUB creators=] MAY use in
+							[=XHTML content documents=] to add support for internationalization, translation, and
+							localization.</p>
+						
+						<p>EPUB creators MUST only use ITS attributes as they are defined in <a
+								data-cite="its20#html5-markup">Using ITS markup in HTML</a> [[its20]] (i.e., EPUB 3 does
+							not support the namespaced attributes).</p>
+						
+						<p>The use of these attributes MUST conform to the requirements defined in [[its20]].</p>
+					</section>
+					
 					<section id="sec-xhtml-custom-attributes" data-epubcheck="true"
 						data-tests="https://w3c.github.io/epub-structural-tests/#06-content-document_content-document-xhtml.feature_L790,https://w3c.github.io/epub-structural-tests/#06-content-document_content-document-xhtml.feature_L795">
 						<h5>Custom attributes</h5>
@@ -5911,20 +5925,6 @@ No Entry</pre>
 								intended for use by other reading systems. This specification should be extended to
 								provide extensions that multiple independent reading systems can use.</p>
 						</div>
-					</section>
-
-					<section id="sec-xhtml-its">
-						<h5>International tag set (ITS)</h5>
-
-						<p>The [[its20]] specification defines a set of attributes that [=EPUB creators=] MAY use in
-							[=XHTML content documents=] to add support for internationalization, translation, and
-							localization.</p>
-
-						<p>EPUB creators MUST only use ITS attributes as they are defined in <a
-								data-cite="its20#html5-markup">Using ITS markup in HTML</a> [[its20]] (i.e., EPUB 3 does
-							not support the namespaced attributes).</p>
-
-						<p>The use of these attributes MUST conform to the requirements defined in [[its20]].</p>
 					</section>
 				</section>
 
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -12038,7 +12038,7 @@ EPUB/images/cover.png</pre>
 				<summary>Substantive changes since <a href="https://www.w3.org/TR/epub-33/">EPUB 3.3</a></summary>
 				<ul>
 					<li>04-June-2025: Added support for Internationalization Tag Set. See <a
-							href="https://github.com/w3c/epub-specs/issues/2732">issue 2732</a></li>
+							href="https://github.com/w3c/epub-specs/issues/2732">issue 2732</a>.</li>
 					<li>26-May-2025: Added <code>application/x-font-ttf</code> to the list of core media types for
 						identifying TTF fonts. See <a href="https://github.com/w3c/epub-specs/issues/667">issue 667</a>. </li>
 					<li>08-Apr-2025: Added early support for HTML syntax. See <a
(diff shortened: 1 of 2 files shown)


iherman (approved):
Tiny comment: I would reorder the subsections by pushing the deprecated extensions to the end of the sections. Ie, having something like: RDFa, ITS, Custom Attributes, content switching and trigger (and the order in the RS should be the same, of course). It de-emphasizes the deprecated attributes in favor of the "valid" ones.

mattgarrish:
Or should we move these sections (and all the other deprecated features) to [A.2 Deprecated features](https://www.w3.org/TR/epub-34/#deprecated)?

It seems like most specifications these days don't leave their deprecated features in the main body where they'll continue to draw attention, even if they are marked as deprecated.

But if we go this route, I'd prefer to do it in a different pull request.

iherman:
> Or should we move these sections (and all the other deprecated features) to [A.2 Deprecated features](https://www.w3.org/TR/epub-34/#deprecated)?
> 
> It seems like most specifications these days don't leave their deprecated features in the main body where they'll continue to draw attention, even if they are marked as deprecated.

Actually, this sounds like a great idea!

> 
> But if we go this route, I'd prefer to do it in a different pull request.

Agreed.

epub34/authoring/index.html
```diff
@@ -12023,12 +12037,12 @@ <h2>Change log</h2>
...
 				<summary>Substantive changes since <a href="https://www.w3.org/TR/epub-33/">EPUB 3.3</a></summary>
 				<ul>
-					<li>26-May-2025: Added <code>application/x-font-ttf</code> to the list of core media types for identifying TTF fonts. 
-					    See <a href="https://github.com/w3c/epub-specs/issues/667">issue 667</a>.
-					</li>
-					<li>08-Apr-2025: Added early support for HTML syntax.
-						See <a href="https://github.com/w3c/epub-specs/issues/2715">issue 2715</a>.
-					</li>
+					<li>04-June-2025: Added support for Internationalization Tag Set. See <a
+							href="https://github.com/w3c/epub-specs/issues/2732">issue 2732</a></li>
```
```suggestion
					<li>04-June-2025: Added support for Internationalization Tag Set. See <a
							href="https://github.com/w3c/epub-specs/issues/2732">issue 2732</a>.</li>
```

mattgarrish:
Even though the deprecated sections are now gone, I'm going to swap ITS and custom attributes so that the extensions from actual standards come first.

iherman:
This was discussed during the [pmwg meeting on 26 June 2025](https://w3c.github.io/pm-wg/minutes/2025-06-26.html#ea9a).

<h3>PR 2735 - <a href="https://github.com/w3c/epub-specs/pull/2735" rel="noopener noreferrer">w3c/<wbr>epub-specs#2735</a></h3>
<p><cite>wendyreid:</cite> the next is about ITS attributes</p>
<p><cite>ivan:</cite> there is a set of attributes that look like the ARIA attributes defined by the W3C for internationalization<br>
<span>… like setting the translation of terms, setting aside language portions</span><br>
<span>… these attributes are not recognized in epub or an extension to HTML</span><br>
<span>… i ran into this problem changing a document into Chinese</span><br>
<span>… it was written with 3 different director sets</span><br>
<span>… epubcheck rejects this because it is not accepted</span></p>
<p><cite>mgarrish:</cite> we were wondering if it would be easier to add extensions going forward</p>
<p><cite>mgarrish:</cite> we would just add another extension section with optional reading system support</p>
<p><cite>wendy:</cite> so we want to change this to an extension section?<br>
<span>… any opposition to merging this?</span></p>

<hr /></details>

epub34/authoring/index.html
```diff
@@ -5888,7 +5888,21 @@ <h5>RDFa</h5>
 								both are natively supported.</p>
 						</div>
 					</section>
-
+					
+					<section id="sec-xhtml-its">
+						<h5>International tag set (ITS)</h5>
```
xfq, line 5893: ITS should be "Internationalization Tag Set"?

mattgarrish: Yes, thanks, I've pushed a fix for that in https://github.com/w3c/epub-specs/commit/30e4e2e3212a737cc86893d7407409d0a8949edd

PR #2736, opened 2025-06-05
Move all deprecated features to the appendix

mattgarrish:
As noted in https://github.com/w3c/epub-specs/pull/2735#issuecomment-2944066035, it would make the specifications cleaner if we removed the deprecated features from the body of the specification.

This pull request moves all the authoring features into the deprecated features section. All I've retained are the names of the features and the links/references to where they were last defined. If there is a replacement to use instead, I've kept that, too.

A couple of changes that people might disagree with are:

- I took bindings out of the content model definition for the package element in addition to removing the section about them. I've added that features must be used as they were last defined, so this still allows bindings to appear between the spine and any collections.
- I removed -epub-text-combine and left -epub-text-combine-horizontal since only the former is deprecated. They seem to have both been spun out of text-combine-upright, but I assume we can live without the deprecated one being mentioned anymore.

For the reading system specification, I've only kept the features that were unique to that document listed. It's easier to refer across to the authoring specification for all the other features, especially since any reading system requirements will also be linked from their references since they generally all predate the split we did in 3.3.

***

Reading Systems: [Preview](https://raw.githack.com/w3c/epub-specs/reorg/deprecated-features/epub34/rs/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/reorg/deprecated-features/epub34/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2736.html" title="Last updated on Jun 9, 2025, 1:05 AM UTC (bbc3079)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2736/266bd71...bbc3079.html" title="Last updated on Jun 9, 2025, 1:05 AM UTC (bbc3079)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -9692,7 +9692,7 @@ html.my-document-playing * {
 					Although it lacks the necessary reading system support, it is integral to the content model on which
 					EPUB is built (i.e., for internationalization support in the package document).</p>
 
-				<p>EPUB creators MAY use this features as described.</p>
+				<p>EPUB creators MAY use this feature as described.</p>
 
 				<div class="note">
 					<p>[=EPUB conformance checkers=] should alert EPUB creators to the presence of under-implemented
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -3592,15 +3592,6 @@
 								</p>
 							</li>
 							<li>
-								<p>
-									<a href="#sec-opf-bindings">
-										<code>bindings</code>
-									</a>
-									<code>[0 or 1]</code>
-									<a href="#deprecated" class="deprecated">(deprecated)</a>
-								</p>
-							</li>
-							<li>
 								<p> [^collection^] <code>[0 or more]</code>
 								</p>
 							</li>
@@ -5337,20 +5328,6 @@ No Entry</pre>
 						</aside>
 					</section>
 				</section>
-
-				<section id="sec-opf-bindings">
-					<h4>The <code>bindings</code> element (deprecated)</h4>
-
-					<p>The <code>bindings</code> element defines a set of custom handlers for media types not supported
-						by this specification.</p>
-
-					<p>Use of the element is <a href="#deprecated">deprecated</a>.</p>
-
-					<p>Refer to the <a
-							href="https://idpf.org/epub/301/spec/epub-publications-20140626.html#sec-bindings-elem"
-								><code>bindings</code> element definition</a> in [[epubpublications-301]] for more
-						information.</p>
-				</section>
 			</section>
 
 			<section id="sec-pkg-spine">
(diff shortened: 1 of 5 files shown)


iherman (approved):
Thanks!

iherman:
A related comment, which might require an easy fix that could be added to this PR: in view of the changes, the section §A.1 on under-implemented features looks a bit strange. It has a description of what under-implemented means, but there is no listing of what they are. At this moment, I believe we have only one of those (the `dir` attribute). Could we add a _reference_ to that attribute here (not _moving_ the attribute here, it would make it less visible, which we do not want)?

mattgarrish:
I just did some rewriting of that section to focus it solely on the dir attribute. It didn't make sense to have it read like there could be many under-implemented features since we've already gone through 3.3 and we firmly stated that we'd never use the label again. That's the only feature that ended up with the label.

It had also incorrectly had the epub version reference bumped to 3.4 when this was a one-time event for pre-3.3 features.

But let me know if you spot anything with the new wording that needs adjusting.

iherman:
The new text says, for the dir attribute:

> Although [setting the directionality of package document metadata](https://www.w3.org/TR/epub-34/#attrdef-dir) is marked as [under-implemented](https://www.w3.org/TR/epub-34/#under-implemented) in [[epub-34](https://raw.githack.com/w3c/epub-specs/reorg/deprecated-features/epub34/rs/index.html#bib-epub-34)], reading system with an international audience, or that claim support for international content, are strongly advised to implement this feature when exposing that metadata to users. Ignoring the directionality of text can cause readability issues.

Did you mean to make a reference to EPUB 3.3 here?

mattgarrish:
> Did you mean to make a reference to EPUB 3.3 here?

I think that reference is fine because it's one spec referring across to the definition in the other. It continues to be under-implemented in 3.4 (so far as we know yet, anyway).

But I think we can strip "from [epub-34]" from this following paragraph:

> The under-implemented label will be removed from [[epub-34]] when the necessary baseline of support in reading systems has been achieved. 

The label will be removed from whatever version we finally get enough support in, so no reason to keep calling out what version we're currently at.

mattgarrish:
I'm kind of wondering now why we have a note about it being under-implemented in the reading systems spec. Under-implemented is an authoring consideration, not one for developers as we want them to implement it regardless of the label. I'm going to push a change to take out the talk of it being under-implemented and leave only the rationale for why it's important.

iherman:
> I'm kind of wondering now why we have a note about it being under-implemented in the reading systems spec. Under-implemented is an authoring consideration, not one for developers as we want them to implement it regardless of the label. 

I am not sure you are right. It is a feature that is not, or almost not, implemented by reading systems, i.e., authors should, alas!, not count on its availability. Yes, we want RS-s to implement it, but that has not happened yet.

> I'm going to push a change to take out the talk of it being under-implemented and leave only the rationale for why it's important.

See above. I am not sure this should be changed...

mattgarrish:
Right, the authoring spec retains the label and note about lack of support in the definition, but what purpose does it serve in the reading systems note to simply say that it is under-implemented and that the label will be removed when it's not? That's a given and irrelevant to someone implementing the standard. It almost comes off like we're warning developers that it's not implemented by other developers so maybe they can wait until it is better developed. That's not what we want, though.

The bit of text in the rs note around why dir matters is all I keep in the last commit.

iherman:
> Right, the authoring spec retains the label and note about lack of support in the definition, but what purpose does it serve in the reading systems note to simply say that it is under-implemented and that the label will be removed when it's not? 

Oops, I am sorry, I misread your text. I missed that your remark was for the RS spec, not the authoring one. All my remarks are withdrawn :-) 

> The bit of text in the rs note around why dir matters is all I keep in the last commit.

+1

mattgarrish:
Okay, let's go ahead and merge this one since it's primarily just a shuffling of existing content.

PR #2737, opened 2025-06-06
Media type contact should be the PM WG...

iherman:
... instead of the EPUB WG. Also, the reference in the registration should be epub 3.4 and not 3.3.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2737.html" title="Last updated on Jun 25, 2025, 6:55 AM UTC (b6e7d9c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2737/266bd71...b6e7d9c.html" title="Last updated on Jun 25, 2025, 6:55 AM UTC (b6e7d9c)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -11237,7 +11237,7 @@ html.my-document-playing * {
 
 					<dt><code>page.xhtml</code></dt>
 					<dd>
-						<p>The resource is an XHTML document. It is listed in the [[EPUB spine | spine=]. It is a
+						<p>The resource is an XHTML document. It is listed in the [=EPUB spine | spine=]. It is a
 							[=publication resource=] on the manifest plane, a [=container resource=], an [=EPUB content
 							document=] on the [=spine plane=], and is not present on the [=content plane=]. No fallback
 							is necessary.</p>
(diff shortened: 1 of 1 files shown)


mattgarrish (approved):
Does this mean we have to submit it to IANA again?

iherman:
> Does this mean we have to submit it to IANA again?

I am not sure. I believe they refer to the latest EPUB version to get to this section, but maybe I am wrong. In any case, it is not a new submission, just a request to change.

iherman:
I have checked: the [IETF admin page](https://www.iana.org/assignments/media-types/media-types.xhtml#application) refers to the EPUB WG's email address, which bounces back with a message referring to the new WG. Works, but not ideal.

I think we have two possibilities:

1. Leave the text as is, with the adage of "who really cares?" :-) 
2. Do the proper job and (a) update the reference in the specification and (b) get the IETF to change the entry 

I am torn, but more inclined to do (2). @plehegar your advice would be welcome here: what is the administration to get IETF to change the email/WG reference for a registered media type?

mattgarrish:
> I am torn, but more inclined to do (2).

It's the proper thing to do, but it means we have to remember to make the change after we publish as I don't think they'll accept changes before the specification is final.

iherman:
Ok, I have discussed this with PlH, and here is what we agreed to do

- let us create a separate email alias (e.g., epub-maintenance@w3.org) which, as of now, points at the PM WG list. That means this thing becomes future-proof
- we contact IANA to change the email address for the EPUB Media types to this address. This does not require a completely new procedure, PlH found a channel whereby we can ask for a non-substantial change in the media type registration. In other words, there is no need for a completely new registration, nor does it require a full new specification
- this PR should be changed to use the email alias.

Unless there is a major issue with this, I will go ahead this week to get that behind us.

WDYT?

mattgarrish:
That could work, but don't we have to update the registrations at the end anyway because the current ones refer to epub-33?

Maybe we should change those urls to use TR/epub so they don't keep going out of date? (At least until an EPUB 4 comes around and probably invalidates that URL, but I digress... 😉 )

iherman:
> That could work, but don't we have to update the registrations at the end anyway because the current ones refer to epub-33?
> 
> Maybe we should change those urls to use TR/epub so they don't keep going out of date? 

Yep, I will do that. It is cleaner and makes things better for the future.

> (At least until an EPUB 4 comes around and probably invalidates that URL, but I digress... 😉 )

You indeed digress 😀

(EPUB 4 would upset so much things that this will be the least of our problems...)

reidmore-online (approved):

iherman:
The email address at IANA has changed to [epub-maintenance@w3.org](mailto:epub-maintenance@w3.org). Merging this PR.

PR #2741, opened 2025-06-17
Added borders for each table

clapierre:
Also centered the heading for the complex Table heading spanning 3 columns

### DIFF
--- a/wg-notes/fxl-a11y-tech/index.html
+++ b/wg-notes/fxl-a11y-tech/index.html
@@ -402,7 +402,7 @@
             &lt;/table&gt;
                     </pre>
 					<p>That will render as:</p>
-					<table>
+					<table border="1">
 						<caption>Sample Data Table</caption>
 						<thead>
 							<tr>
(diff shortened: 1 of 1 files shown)


wg-notes/fxl-a11y-tech/index.html
```diff
@@ -402,7 +402,7 @@ <h3>Tables</h3>
             &lt;/table&gt;
                     </pre>
 					<p>That will render as:</p>
-					<table>
+					<table border="1">
```
mattgarrish: Non-conforming border attribute alert! 😄 

If you want nicer borders, CSS would be the way to go. You could use a rule like:

```
table {
    border-collapse: collapse;
}
th, td {
    border: 1px solid black;
}
```

to avoid the double borders that using the attribute generates.

clapierre: OK ok ;)

reidmore-online (approved):

PR #2756, opened 2025-07-03
Update the overview document on xhtml vs html change

iherman:
Also following the discussion on the 2025-07-03 call, some minimal clarification on the XHTML 1.1 vs. HTML change...

Issue #2715, opened 2025-04-24
Adopt the HTML syntax in addition to the XML syntax for content documents


bduga:
### Describe the problem

EPUB 3.3 and earlier require the use of the XML serialization of HTML (AKA the XML syntax, formerly known as XHTML). HTML no longer provides new feature drops to the XML syntax and [explicitly warns against its use](https://html.spec.whatwg.org/multipage/xhtml.html#the-xhtml-syntax). There are already some issues with scripting and XML, and some potential limitations with upcoming extended accessibility techniques. In addition, most off the shelf tools produce the HTML syntax, and most developers are familiar with it. To roughly quote a working group member "Finding out EPUB is based on XHTML is like finding out Social Security runs on COBOL".

### Describe the fix or new feature you propose

Add the HTML syntax as an option for content creators, and define requirements around its support for Reading Systems, distributors, aggregators, etc. while explicitly retaining the XML syntax as a valid format and mandating its continued support.

murata2makoto:
XML was never intended as a passing trend; it was designed with universal, long-term principles in mind.

Like a Rolls-Royce, it was built for enduring reliability — structured, neutrally extensible, and deeply internationalized.

Unlike HTML, where any input yields a parseable document but future parsing results are not guaranteed to remain the same, XML enforces strict well-formedness and guarantees that well-formed documents will always produce the same structure — ensuring the stability and trust essential for the long-term preservation, authenticity, and interoperability of digital publications.

Some may argue that, as a surviving member of the XML Working Group, my views are outdated.
However, the strict requirement of well-formedness in XML — and the guarantee that parsing results will remain consistent for all time — are not relics of the past.  They are vital strengths that safeguard the long-term viability of digital publications, which are invaluable assets to publishers.

Stability, predictability, and fidelity across decades are not luxuries for publications — they are necessities.

iherman:
> Unlike HTML, where any input yields a parseable document but future parsing results are not guaranteed to remain the same,

[...]

PR #2757, opened 2025-07-03
Update the relationship to html section

mattgarrish:
As a complement to #2756, this pull request updates the Relationship to HTML section to describe the integration of both syntaxes in the HTML standard we refer to.

I've skipped noting the change is still a proposal since this section links to the html docs section where we already have an extensive note about the status.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2757.html" title="Last updated on Jul 5, 2025, 1:06 PM UTC (2abf806)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2757/5fe557b...2abf806.html" title="Last updated on Jul 5, 2025, 1:06 PM UTC (2abf806)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -213,8 +213,8 @@
 					<h4>Relationship to HTML</h4>
 
 					<p>The [[html]] standard is continuously evolving — there are no longer versioned releases of it.
-						That standard, in turn, references various technologies that continue to evolve, such as MathML,
-						SVG, CSS, and JavaScript.</p>
+						That standard, in turn, references various technologies that also continue to evolve, such as
+						MathML, SVG, CSS, and JavaScript.</p>
 
 					<p>The benefit of this approach for EPUB is that [=EPUB publications=] always keep pace with changes
 						to the web without the need for new revisions. [=EPUB creators=], however, must keep track of
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -527,8 +527,7 @@
 							<p>EPUB 3 previously supported only the XML syntax of [[html]] via what were called XHTML
 								content documents. The change of name to HTML content documents is not to give priority
 								to the HTML syntax but to better reflect that [[html]] is the common standard for both
-								syntaxes. XHTML has not been a separate standalone technology since [[xhtml11]] and the
-								XML syntax has replaced it in [[html]].</p>
+								syntaxes. The term "XHTML" used to refer to [[xhtml11]], but that was replaced by the XML syntax of [[html]] in 2011, which also officially superceded [[xhtml11]] in 2018.</p>
 						</div>
 						<div class="issue" data-number="2715" title="Adoption of the HTML syntax">
 							<p>The change to allow both syntaxes of HTML is an open issue in EPUB 3.4. It is important
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -521,7 +521,7 @@
 					<dd>
 						<p>An [=EPUB content document=] that conforms to the profile of [[html]] defined in <a
 								href="#sec-xhtml"></a>.</p>
-						<p>HTML content documents can be authored in either the <a data-cite="html#syntax">HTML
+						<p>HTML content documents can be expressed in either the <a data-cite="html#syntax">HTML
 								syntax</a> or the <a data-cite="html#the-xhtml-syntax">XML syntax</a> [[html]].</p>
 						<div class="note">
 							<p>EPUB 3 previously supported only the XML syntax of [[html]] via what were called XHTML
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -223,19 +223,18 @@
 
 					<p>The [[html]] standard defines a single content model with rules for expressing a tag set using
 						either the <a data-cite="html#syntax">HTML syntax</a> (what has always been considered HTML) or
-						the <a data-cite="html#the-xhtml-syntax">XML syntax</a> (what was known as XHTML). Up until
-						[[epub-33]], EPUB 3 only supported the XML syntax via what were called "XHTML content
-						documents." It now allows the authoring of both syntaxes under the common name of [=HTML content
-						documents=].</p>
+						the <a data-cite="html#the-xhtml-syntax">XML syntax</a> (what was formerly known as XHTML). EPUB
+						3 allows authoring of content documents using either syntax even though the [[html]] standard <a
+							data-cite="html/xhtml.html#the-xhtml-syntax">no longer recommends the use of the XML
+							syntax</a>. The Working Group recognizes that XML remains an integral technology in the
+						publishing ecosystem and will not remove support for the XML syntax from EPUB 3. Regardless,
+						publishers that prefer to keep using the XML syntax will need to monitor future support for it,
+						and may have to adapt to the HTML syntax to gain access to some features of [[html]].</p>
 
 					<div class="issue" data-number="2715" title="Adoption of the HTML syntax">
 						<p>The change to allow both syntaxes of HTML is an open issue in EPUB 3.4. The above paragraph
 							was added to help explain the change, but would be reverted if the HTML syntax is not
 							adopted.</p>
-
-						<p>It is important to note that the change is only to allow another syntax of [[html]]. Because
-							EPUB 2's XHTML content documents were implemented using [[xhtml11]], some authors mistakenly
-							assume EPUB 3's XHTML content documents use the same.</p>
 					</div>
 
 					<p>The <a href="#sec-xhtml">HTML profile defined by this specification</a> inherits all definitions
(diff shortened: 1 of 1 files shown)


epub34/authoring/index.html
```diff
@@ -221,12 +221,23 @@ <h4>Relationship to HTML</h4>
 						the various changes to HTML and the technologies it references to ensure they keep their
 						processes up to date.</p>
 
+					<p>The [[html]] standard also combines what used to be two separate technologies: [[?html401]] and
+						[[?xhtml11]]. These two standards differed not only in their syntaxes, but also in what elements
+						and attributes authors could use, as well as how the syntaxes were identified (their
+							<code>DOCTYPE</code> declarations).</p>
```
iherman: To be honest, I am not an expert regarding all the minutia of the differences between html 4.01 and xhtml 1.1, i.e., I am not sure whether this statement is 100% correct. Even if this is correct, I am not sure if it is worth adding this to the discussion. My proposal would be to remove this paragraph.

(Adding the deletion to the comment so that it could be committed by github easily, in case we agree...)

```suggestion
					
```

epub34/authoring/index.html
```diff
@@ -221,12 +221,23 @@ <h4>Relationship to HTML</h4>
...
+
+					<p>The [[html]] standard now defines a single content model with rules for expressing this tag set
+						using either the <a data-cite="html#syntax">HTML syntax</a> (what has always been considered
```
```suggestion
					<p>The [[html]] standard defines a single content model with rules for expressing a tag set
						using either the <a data-cite="html#syntax">HTML syntax</a> (what has always been considered
```

(This is relevant only if the previous paragraph is removed.)

epub34/authoring/index.html
```diff
@@ -221,12 +221,23 @@ <h4>Relationship to HTML</h4>
...
+						HTML) or the <a data-cite="html#the-xhtml-syntax">XML syntax</a> (what was known as XHTML). Up
+						until [[epub-33]], EPUB 3 only supported the XML syntax via what were called "XHTML content
+						documents." It now allows the authoring of both syntaxes under the common name of [=HTML content
+						documents=].</p>
+
```
iherman, line 239: I think we should be careful: the paragraph reads as if this was a done deal, and a cursory look at the draft might raise eyebrows among those who intend to complete our survey.

I think we should add an editorial issue reference to #2715, and possibly to the questionnaire.

Also, if we agree on the removal of the paragraph above, adding an editorial note (here or somewhere else, I do not mind) making it explicit that what we call "XHTML Content Document" is _not_ referring to XHTML 1.1.

epub34/authoring/index.html
mattgarrish: Ya, that's fine. We don't need a history lesson.

epub34/authoring/index.html
mattgarrish: Ya, I was trying to avoid overloading notes, but that's fine. Worst case scenario, it leaves a marker where we have to undo things. I'll add an editor's note with another explanation that this is still a pending change.

iherman: Issue markers can be removed easily, at no cost... but having it ensures that the readers are not led to a wrong conclusion.

mattgarrish: > making it explicit that what we call "XHTML Content Document" is _not_ referring to XHTML 1.1

I've added this to the issue box in the last commit, but explaining this difference becomes even less relevant long term if we drop "xhtml content document". I'm not sure where it would meaningfully fit.

The only thought I've had would be to add a "differences from epub 2" section somewhere to briefly overview the major changes. But that might make the most sense as an appendix in the overview. Since we already have the revision history stuff there, maybe I'll see if I can combine the two.

iherman: I think it is o.k. as it is for now.

Whether we will keep the "XHTML Content Document" terminology, or whether we would drop the "X": I am not sure yet. Editorially, it is probably the good move. Keeping the "X", with appropriate warnings, may be socially better, avoiding scaring people who might think that the requirement is to use HTML _and not_ XML...

Let us cross the bridge when we get there.

iherman (approved):

mattgarrish:
I made some pretty significant changes in the last commit, so have another look. I didn't like that it was mixing explaining the syntaxes in the relationship section when that's more of a content document issue. The text also didn't really make clear why this is an issue to be aware of in our adaptation of html.

I've rewritten the relationship section to try and explain that even though html doesn't recommend the xml syntax anymore, we don't have any plans to remove it. But if you're going to stick with it you need to monitor support and may need to adapt to the html syntax to get some features. I think that keeps it balanced while trying to point out why we're making this change.

For the HTML content document definition, I was worried about not mentioning xhtml content documents anymore after a changeover, so I added a second dt/dfn with "(formerly XHTML content documents)" and added a note to explain that the name reflects the standard and isn't a change in preference to the HTML syntax. Again, just trying to avoid any potential confusion/criticism that our naming is intended to give preference to a syntax.

epub34/authoring/index.html
```diff
@@ -221,12 +221,27 @@ <h4>Relationship to HTML</h4>
 						the various changes to HTML and the technologies it references to ensure they keep their
```
iherman, line 221: (GitHub does not allow me to directly comment on earlier lines, hence anchoring this comment here.)

The text says:

>  That standard, in turn, references various technologies that continue to evolve, such as MathML, SVG, CSS, and JavaScript

Shouldn't it be "also continue to evolve"?

epub34/authoring/index.html
```diff
@@ -221,12 +221,27 @@ <h4>Relationship to HTML</h4>
...
 						processes up to date.</p>
 
+					<p>The [[html]] standard defines a single content model with rules for expressing a tag set using
+						either the <a data-cite="html#syntax">HTML syntax</a> (what has always been considered HTML) or
```
```suggestion
						either the <a data-cite="html#syntax">HTML syntax</a> (what has always been considered as the default for HTML) or
```

I am not sure about the exact details of the proposed change, but somehow it did not sound good without something there...

epub34/authoring/index.html
```diff
@@ -221,12 +221,27 @@ <h4>Relationship to HTML</h4>
...
+						3 allows authoring of content documents using either syntax even though the [[html]] standard <a
+							data-cite="html/xhtml.html#the-xhtml-syntax">no longer recommends the use of the XML
+							syntax</a>. The Working Group recognizes that XML remains an integral technology in the
+						publishing ecosystem and will not remove support for the XML syntax from EPUB 3. Regardless,
+						publishers that prefer to keep using the XML syntax will need to monitor future support for it,
+						and may have to adapt to the HTML syntax to gain access to some features of [[html]].</p>
+
+					<div class="issue" data-number="2715" title="Adoption of the HTML syntax">
+						<p>The change to allow both syntaxes of HTML is an open issue in EPUB 3.4. The above paragraph
+							was added to help explain the change, but would be reverted if the HTML syntax is not
```
```suggestion
							was added to help explain the change, but would be amended if the HTML syntax is not
```

(Not sure that we would just revert to the original text. I believe the emphasis on the fact that XHTML is 'just' the XML syntax to the moving target called "html" is important.)

epub34/authoring/index.html
```diff
@@ -498,14 +513,31 @@ <h3>Terminology</h3>
 					</dd>
 
 					<dt>
-						<dfn class="export" data-lt="xhtml content document|xhtml content documents">HTML content
-							document</dfn>
+						<dfn class="export">HTML content document</dfn>
```
iherman, line 516:
```suggestion
						<dfn class="export"  data-lt="xhtml content document|xhtml content documents">HTML content document</dfn>
```

Shouldn't we export the `data-lt` for the new text as well? We may want to add, if it becomes final, 'html content documents' as well

epub34/authoring/index.html
```diff
@@ -498,14 +513,31 @@ <h3>Terminology</h3>
...
+					</dt>
+					<dt>
+						<small>(formerly <dfn class="export">XHTML content document</dfn>)</small>
 					</dt>
 					<dd>
 						<p>An [=EPUB content document=] that conforms to the profile of [[html]] defined in <a
 								href="#sec-xhtml"></a>.</p>
 						<p>HTML content documents can be authored in either the <a data-cite="html#syntax">HTML
```
```suggestion
						<p>HTML content documents can be expressed in either the <a data-cite="html#syntax">HTML
```

The word "authored" sounded funny to my ears here. Feel free to ignore me...

epub34/authoring/index.html
```diff
@@ -498,14 +513,31 @@ <h3>Terminology</h3>
...
-								syntax</a> or <a data-cite="html#the-xhtml-syntax">XML syntax</a> [[html]].</p>
+								syntax</a> or the <a data-cite="html#the-xhtml-syntax">XML syntax</a> [[html]].</p>
+						<div class="note">
+							<p>EPUB 3 previously supported only the XML syntax of [[html]] via what were called XHTML
+								content documents. The change of name to HTML content documents is not to give priority
+								to the HTML syntax but to better reflect that [[html]] is the common standard for both
+								syntaxes. XHTML has not been a separate standalone technology since [[xhtml11]] and the
+								XML syntax has replaced it in [[html]].</p>
```
```suggestion
								syntaxes. The term "XHTML" used to refer to [[xhtml11]], but that was replaced by the XML syntax of [[html]] in 2011, which also officially superceded [[xhtml11]] in 2018.</p>
```

epub34/authoring/index.html
```diff
@@ -498,14 +513,31 @@ <h3>Terminology</h3>
...
+						</div>
+						<div class="issue" data-number="2715" title="Adoption of the HTML syntax">
+							<p>The change to allow both syntaxes of HTML is an open issue in EPUB 3.4. It is important
+								to note that EPUB 3 is already based on [[html]] and this change only makes the HTML
+								syntax valid to use. Because EPUB 2's XHTML content documents were implemented using
+								[[xhtml11]], some authors mistakenly assume EPUB 3's XHTML content documents use the
+								same technology and that this change represents a move to a new standard.</p>
+							<p>This new definition will be reverted if the HTML syntax is not adopted.</p>
```
```suggestion
							<p>This new definition will be amended if the HTML syntax is not adopted.</p>
```

(See my comments above.)

epub34/authoring/index.html
mattgarrish: I added a new dfn to export the term xhtml content document. We can't have that dfn exported and have a data-lt on this term to export the same terms.

But the choice is whether to land all links at the primary term or to have documents still linking to xhtml content document land at the "formerly" line. I don't have a preference on that.

epub34/authoring/index.html
mattgarrish: The act is authoring but what you write is expressing, I suppose. Doesn't really make a difference to me.

epub34/authoring/index.html
mattgarrish: I've removed these parentheticals as they don't seem relevant. The note for html content documents already explains that xhtml content documents was never xhtml 1.1. This is just doing a similar equating of the syntaxes to the old standards.

epub34/authoring/index.html
iherman: Ah. Are we sure that the 'dfn' would cover the plural form as well? Isn't it why we used the data-lt attribute?

(I am fine leaving the xhtml ones on the 'formerly' line, b.t.w.)

mattgarrish: > Ah. Are we sure that the 'dfn' would cover the plural form as well?

Yes, respec [handles plurals automatically](https://respec.org/docs/#automatic-pluralization) so long as the term is in dfn tags. That's why we had to supply both forms of xhtml content doc to layer it onto the html content doc definition.

mattgarrish: Here's the tag respec creates for xhtml content document for example:

```
<dfn class="export" data-export="" data-plurals="xhtml content documents" id="dfn-xhtml-content-document" tabindex="0" aria-haspopup="dialog" data-dfn-type="dfn">XHTML content document</dfn>
```

I don't know how it does the matching on data-plurals, but none of our terms have plurals in data-lt attributes and they've always linked up fine in plural form.

reidmore-online (approved):

shiestyle (approved):

PR #2767, opened 2025-07-30
Add new use cases for annotations and bookmarks,

llemeurfr:
following EPUB WG and PBG discussions.

---

See:

* For EPUB Annotations Use Cases and Requirements:
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/july-2025/wg-notes/annotations-ucr/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/wg-notes/annotations-ucr/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/july-2025/wg-notes/annotations-ucr/index.html)

### DIFF
--- a/wg-notes/annotations-ucr/index.html
+++ b/wg-notes/annotations-ucr/index.html
@@ -62,9 +62,15 @@
 			</p>
 		</section>
 		<section>
+			<h2>Formatting annotations</h2>
+			<p>
+				A user adds an annotation to an EPUB. He adds some formatting to the text of the annotation: emphasize, italic, underline, subtitles.
+			</p>
+		</section>
+		<section>
 			<h2>Bookmarking</h2>
 			<p>
-				A user decides to bookmark a location in a reflowable EPUB. The cursor is located where the user clicks on the screen, or at a default location. 
+				A user decides to bookmark a location in a reflowable EPUB. The cursor is located where the user clicks on the screen, or at a default location (often the top-left corner of the screen, for left-to-right content). 
 				The user triggers the bookmark affordance, optionally selects a highlight mode & color, and saves the bookmark. 
 				A bookmark icon appears on the page, near the line where the cursor was positioned when the bookmark was created.
 			</p>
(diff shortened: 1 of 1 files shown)


sueneu:
@iherman These are largely clarifications, making use cases clearer and more detailed. Is it OK to commit? Then we can solicit more comments and refine the documentation further.

Is that the right next step here?

iherman:
> @iherman These are largely clarifications, making use cases clearer and more detailed. Is it OK to commit? Then we can solicit more comments and refine the documentation further.
> 
> Is that the right next step here?

The netiquette is usually that people give their approval (or not) to the PR using the 'add your review' button up there and then, if there are enough approvals and there is a consensus, the person having submitted merges the PR (if he/she has the right credentials, which is the case for @llemeurfr).

As part of the review one can also comment on specific texts, propose changes, etc. which usually leads to an interaction between the reviewer and the submitter...

iherman (commented):
Two editorial comments:

1. I think we have to use subsections so that a section would only contain a single use case (e.g., section 4 should consist of subsections, one for each paragraph). This will help later to refer to the right use case
2. The "he" pronoun should alternate among use cases between "he" and "she". We can do that later when the list is final, or do it now, and it may become a bit random later (which is fine).

sueneu:
@iherman many thanks for clarifying the process

+1 to using both “he/him" and “she/her" in the text.

> Today, he loads a new version of the EPUB in his reading system. The spine item in which the annotations were created has been modified, and some words have been added to the leaf HTML element in which an annotation was created, before the start of the annotation and in the annotated text itself. 

> Still, the user gets the annotation at the proper location.

What if the HTML element containing an annotation is deleted in a subsequent Ebook version? There should still be a way to access the annotation. Perhaps a non-linear HTML file concatenating unanchored annotations that can be accessed from the TOC?

llemeurfr:
> What if the HTML element containing an annotation is deleted in a subsequent Ebook version? There should still be a way to access the annotation. Perhaps a non-linear HTML file concatenating unanchored annotations that can be accessed from the TOC?

Modifying the structure of the EPUB is not possible. 
If the annotation has lost its "anchor", it will still appear in the list of annotations, with its context (parent chapter/section).  Highlighting the annotated text (or image portion of the audio section) is no longer possible. An RS MAY still present an affordance (an active icon ...) around the original position if a good-enough selector has been provided. 

I could add this as a new use case if you feel it useful.

llemeurfr:
ok, I'll merge this one, and in a later stage, we'll alternate he/she and decide if each use case has its own section (it is right that identifying a section is semantically more correct than a paragraph.

PR #2771, opened 2025-08-06
Made the changes on CSS management discussed in the issue 2762 

### LINKED ISSUE #2762: Relationship between the rendering control of the RS and the publisher's settings?
### Describe the problem

This is an "abstraction" of issue #2700, see also https://github.com/w3c/epub-specs/issues/2700#issuecomment-3038977670 and the comments that follow. Also, in some respect, #2754, #2753, etc. are related to the same underlying principles.

The question is: what is the relationship between rendering aspects that are controlled by the publisher's on the one hand, and what are controlled by the user via some UI on the RS on the other? Examples are font setting, justification, hyphenation, but also spread control, or theme/background color. Almost all Reading Systems have some ways to override some of the publisher's settings, but nothing is documented in our specifications on what the industry practice is, or should be. What (if any) is the value of those publisher's settings (let that be settings in the package document or via CSS) if an RS may, essentially, ignore it and render the content based on UI interaction with the user? 

We have to realize that the possibility of overwriting the publishers' settings is a major difference between a browser and a RS. Providing such user control is a good thing in terms of, e.g., accessibility, so we should keep that. But providing a publisher fallback on the RS might be useful/necessary to make this difference explicit.

Many (but not all!) reading systems provide a publisher's fallback: the UI panel may have a "default" choice, it may have a checkbox to set the, say, choice of fonts (and, presumably, uncheck means "keep to the publisher's choice"). But none of that is clearly documented, made it explicit, etc. EPUB authors do not have a real clue of what to expect.

At the minimum, there should be some statement in the RS (and possibly in the Authoring spec, too), saying something like:

> A RS _MAY_ provide user control over some rendering aspects, like font choice, background color, text justification, or flow control, but it also _SHOULD_ (_MUST_?) provide an easily identifiable way for the reader to switch all local setting off and rely on the publisher's choices.
>

We may also come up with additional mechanisms that would, essentially, disallow overwriting the publisher's settings (Defining a specific metadata? Use CSS' media query for some special media?) but I am not sure. Maybe worth discussing.

iherman:
I have made the PR as discussed in https://github.com/w3c/epub-specs/issues/2762#issuecomment-3151116277 and onwards. 

See also my comments below on the discussion in #2762.

Fix #2762

***
See:

* For EPUB 3.4:
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/css-sections-review-(issue-2762)/epub34/authoring/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/authoring/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/css-sections-review-(issue-2762)/epub34/authoring/index.html)
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/css-sections-review-(issue-2762)/epub34/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/css-sections-review-(issue-2762)/epub34/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2771.html" title="Last updated on Sep 5, 2025, 1:56 PM UTC (d8c5510)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2771/da1c954...d8c5510.html" title="Last updated on Sep 5, 2025, 1:56 PM UTC (d8c5510)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -6215,6 +6215,22 @@ No Entry</pre>
 									<p>Some types of screens will render animations and transitions poorly (e.g., those
 										with high latency).</p>
 								</li>
+								<li>
+									<p>
+										Creators should be aware that a reading systems may override some aspects of 
+										the creator's style settings; creators may want to consult the reading systems' 
+										user agent style sheets, if publicly available. Furthermore, reading systems may
+										also modify some rendering aspects as a result of user interaction (e.g., choice 
+										of fonts, text justification, or foreground and background colors).
+									</p>
+								</li>
+								<li>
+									<p>
+										Creators should avoid using [[html]] [^html-global/style^] attributes in content documents. 
+										Styling should always be done via CSS files, otherwise a reading system will 
+										not be able to easily override them when interacting with the user.
+									</p>
+								</li>
 							</ul>
 						</div>
 					</section>
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -6201,7 +6201,7 @@ No Entry</pre>
 							section, EPUB defers to the W3C to define CSS and expects reading systems to support it at
 							level of major browsers.</p>
 
-						<p>Although [=Reading systems=] are expected to support <a href="confreq-css-rs-support">CSS as
+						<p>Although [=reading systems=] are expected to support <a href="confreq-css-rs-support">CSS as
 								defined in the CSS snapshot</a>, the reality is that most reading systems currently do
 							not support all desired features of CSS, and often will modify, and allow users to change,
 							the default presentation defined by the [=EPUB creator=]. As a result, EPUB creators will
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -6217,7 +6217,7 @@ No Entry</pre>
 								</li>
 								<li>
 									<p>
-										Creators should be aware that a reading systems may override some aspects of 
+										Creators should be aware that a reading systems usually override some aspects of 
 										the creator's style settings; creators may want to consult the reading systems' 
 										user agent style sheets, if publicly available. Furthermore, reading systems may
 										also modify some rendering aspects as a result of user interaction (e.g., choice 
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -6198,15 +6198,16 @@ No Entry</pre>
 							provided prefixed versions of numerous other properties. Although the CSS Working Group no
 							longer recommends the use of prefixed properties, this specification maintains some prefixed
 							properties to avoid breaking existing content. But with the minor exceptions defined in this
-							section, EPUB defers to the W3C to define CSS and expects reading systems to support it at
-							level of major browsers.</p>
+							section, EPUB defers to the W3C to define CSS and expects [=reading systems=] to support it
+							at level of major browsers.</p>
 
-						<p>Although [=reading systems=] are expected to support <a href="confreq-css-rs-support">CSS as
-								defined in the CSS snapshot</a>, the reality is that most reading systems currently do
-							not support all desired features of CSS, and often will modify, and allow users to change,
-							the default presentation defined by the [=EPUB creator=]. As a result, EPUB creators will
-							need to adapt their styling to these realities. For more information, refer to <a
-								href="#sec-css-rs"></a>.</p>
+						<p>Although reading systems are expected to support <a
+								data-cite="epub-rs-34#confreq-css-rs-support">CSS as defined in the CSS snapshot</a>
+							[[epub-rs-34]], the reality is that most reading systems currently do not support all
+							desired features of CSS, and often will modify, and allow users to change, the default
+							presentation defined by the [=EPUB creator=]. As a result, EPUB creators will need to adapt
+							their styling to these realities. For more information, refer to <a href="#sec-css-rs"
+							></a>.</p>
 					</section>
 
 					<section id="sec-css-req" data-epubcheck="true"
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -6199,7 +6199,7 @@ No Entry</pre>
 							longer recommends the use of prefixed properties, this specification maintains some prefixed
 							properties to avoid breaking existing content. But with the minor exceptions defined in this
 							section, EPUB defers to the W3C to define CSS and expects [=reading systems=] to support it
-							at level of major browsers.</p>
+							at the level of major browsers.</p>
 
 						<p>Although reading systems are expected to support <a
 								data-cite="epub-rs-34#confreq-css-rs-support">CSS as defined in the CSS snapshot</a>
@@ -6284,7 +6284,7 @@ No Entry</pre>
 						</ul>
 
 						<p>Reading systems will typically set some aspects of an [=EPUB publication | EPUB
-							publication's=] style (e.g., setting borders appropriate to the application), potentially
+							publication's=] style (e.g., setting margins appropriate to the application), potentially
 							overriding the [=EPUB creator | EPUB creator's=] instructions. To mitigate conflicts, and
 							any potential rendering problems, EPUB creators are advised to consult reading systems' user
 							agent style sheets, when made publicly available, and adapt their styling choices for
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -6198,7 +6198,15 @@ No Entry</pre>
 							provided prefixed versions of numerous other properties. Although the CSS Working Group no
 							longer recommends the use of prefixed properties, this specification maintains some prefixed
 							properties to avoid breaking existing content. But with the minor exceptions defined in this
-							section, EPUB defers to the W3C to define CSS.</p>
+							section, EPUB defers to the W3C to define CSS and expects reading systems to support it at
+							level of major browsers.</p>
+
+						<p>Although [=Reading systems=] are expected to support <a href="confreq-css-rs-support">CSS as
+								defined in the CSS snapshot</a>, the reality is that most reading systems currently do
+							not support all desired features of CSS, and often will modify, and allow users to change,
+							the default presentation defined by the [=EPUB creator=]. As a result, EPUB creators will
+							need to adapt their styling to these realities. For more information, refer to <a
+								href="#sec-css-rs"></a>.</p>
 					</section>
 
 					<section id="sec-css-req" data-epubcheck="true"
@@ -6257,10 +6265,10 @@ No Entry</pre>
 					</section>
 
 					<section id="sec-css-rs" class="informative">
-						<h4>Reading system considerations</h4>
+						<h4>Reading system support considerations</h4>
 
-						<p>[=Reading systems=] do not support all desired features of CSS. The following are known to be
-							particularly problematic:</p>
+						<p>Support for the following CSS features are known to be particularly problematic in [=EPUB
+							reading systems=]:</p>
 
 						<ul>
 							<li>
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -6199,40 +6199,6 @@ No Entry</pre>
 							longer recommends the use of prefixed properties, this specification maintains some prefixed
 							properties to avoid breaking existing content. But with the minor exceptions defined in this
 							section, EPUB defers to the W3C to define CSS.</p>
-
-						<div class="note">
-							<p>Keep in mind that some [=reading systems=] will not support all desired features of CSS.
-								The following are known to be particularly problematic:</p>
-
-							<ul>
-								<li>
-									<p>Reading system-induced pagination can interact poorly with style sheets as
-										reading systems sometimes paginate using columns. This may result in incorrect
-										values for viewport sizes. Fixed and absolute positioning are particularly
-										problematic.</p>
-								</li>
-								<li>
-									<p>Some types of screens will render animations and transitions poorly (e.g., those
-										with high latency).</p>
-								</li>
-								<li>
-									<p>
-										Creators should be aware that a reading systems usually override some aspects of 
-										the creator's style settings; creators may want to consult the reading systems' 
-										user agent style sheets, if publicly available. Furthermore, reading systems may
-										also modify some rendering aspects as a result of user interaction (e.g., choice 
-										of fonts, text justification, or foreground and background colors).
-									</p>
-								</li>
-								<li>
-									<p>
-										Creators should avoid using [[html]] [^html-global/style^] attributes in content documents. 
-										Styling should always be done via CSS files, otherwise a reading system will 
-										not be able to easily override them when interacting with the user.
-									</p>
-								</li>
-							</ul>
-						</div>
 					</section>
 
 					<section id="sec-css-req" data-epubcheck="true"
@@ -6288,6 +6254,40 @@ No Entry</pre>
 								</li>
 							</ul>
 						</div>
+					</section>
+
+					<section id="sec-css-rs" class="informative">
+						<h4>Reading system considerations</h4>
+
+						<p>[=Reading systems=] do not support all desired features of CSS. The following are known to be
+							particularly problematic:</p>
+
+						<ul>
+							<li>
+								<p>Reading system-induced pagination can interact poorly with style sheets as reading
+									systems sometimes paginate using columns. This may result in incorrect values for
+									viewport sizes. Fixed and absolute positioning are particularly problematic.</p>
+							</li>
+							<li>
+								<p>Some types of screens will render animations and transitions poorly (e.g., those with
+									high latency).</p>
+							</li>
+						</ul>
+
+						<p>Reading systems will typically set some aspects of an [=EPUB publication | EPUB
+							publication's=] style (e.g., setting borders appropriate to the application), potentially
+							overriding the [=EPUB creator | EPUB creator's=] instructions. To mitigate conflicts, and
+							any potential rendering problems, EPUB creators are advised to consult reading systems' user
+							agent style sheets, when made publicly available, and adapt their styling choices for
+							optimal display.</p>
+
+						<p>Furthermore, reading systems that allow users to change the appearance (e.g., choice of
+							fonts, text justification, or foreground and background colors) will also modify some CSS
+							rendering directions in an EPUB publication. Due to CSS precedence rules, it is advisable to
+							limit the use of [[html]] [^html-global/style^] attributes in EPUB content documents so that
+							the user experience is not negatively impacted (i.e., the users' style choices not being
+							universally applied). Although some reading systems will override inline styles, such
+							behavior is not guaranteed.</p>
 					</section>
 
 					<section id="sec-css-prefixed">
--- a/epub34/rs/index.html
+++ b/epub34/rs/index.html
@@ -1219,13 +1219,16 @@
 					<h4>Author styling overrides</h4>
 
 					<p id="confreq-css-overrides">A reading system MAY override parts of an [=EPUB publication|EPUB
-						publication's] styling (e.g., for ergonomic or user interface reasons), but SHOULD only do so in
-						a way that allow the author's original design to be restored.</p>
+						publication's] styling (e.g., for ergonomic or user interface reasons). Morevoer, they MAY allow
+						users to set default themes and style preferences (e.g., for more accessible reading) that
+						further alter the original design of a publication.</p>
 
-					<p id="confreq-css-user-styles">Reading systems MAY provide users with an affordance that removes
-						any reading system-applied styling, thereby allowing users to view a publication as designed by
-						the author (e.g., so users can check the original design if a publication refers to location,
-						color, etc. that gets overriden by the reading system styling).</p>
+					<p id="confreq-css-user-styles">These changes often make it impractical to render an EPUB
+						publication exactly as designed by an author, even if that is what the user wants. To assist
+						users in viewing an EPUB publication as closely as possible to the author intent, however,
+						reading systems MAY provide an affordance that reverts as much reading system-applied styling as
+						feasible. For example, it could reset any user-applied styling and themes back to the reading
+						system defaults.</p>
 				</section>
 			</section>
 
--- a/epub34/rs/index.html
+++ b/epub34/rs/index.html
@@ -1223,12 +1223,11 @@
 						users to set default themes and style preferences (e.g., for more accessible reading) that
 						further alter the original design of a publication.</p>
 
-					<p id="confreq-css-user-styles">These changes often make it impractical to render an EPUB
+					<!-- <p id="confreq-css-user-styles">These changes often make it impractical to render an EPUB
 						publication exactly as designed by an author, even if that is what the user wants. To assist
 						users in viewing an EPUB publication as closely as possible to the author intent, however,
 						reading systems MAY provide an affordance that reverts as much reading system-applied styling as
-						feasible. For example, it could reset any user-applied styling and themes back to the reading
-						system defaults.</p>
+						feasible.</p> -->
 				</section>
 			</section>
 
--- a/epub34/rs/index.html
+++ b/epub34/rs/index.html
@@ -1205,13 +1205,16 @@
 							[=EPUB content documents=].</p>
 					</li>
 					<li>
-						<p id="confreq-css-overrides">SHOULD NOT override the EPUB creator's style sheets, but SHOULD do
-							so in a way that preserves the Cascade when necessary: through a user agent style sheet, or
-							[[html]] [^html-global/style^] attributes.</p>
+						<p id="confreq-css-overrides">MAY override parts of the creator's style sheets (for, e.g., ergonomic or 
+							user interface reasons). If it does, it SHOULD do so in a way that preserves the cascade model
+							of CSS: through a user agent style sheet, or [[html]] [^html-global/style^] attributes.
+							Reading system developers should also publicly document the user agent style sheets and how they
+							interact with EPUB creator's style settings.</p>
 					</li>
 					<li>
-						<p id="confreq-css-user-styles">It MAY override parts of the EPUB creator's style sheet because
-							of user interaction.</p>
+						<p id="confreq-css-user-styles">MAY override parts of the EPUB creator's style sheet as a result
+							of user interaction.  It MAY also provide additional user interaction to revert to the default
+							style sheets, with or without the effects of the user agent style sheets.</p>
 					</li>
 				</ul>
 
@@ -1219,8 +1222,7 @@
 					system's user agent style sheet SHOULD support the [[html]] <a data-cite="html#rendering">suggested
 						default rendering</a>.</p>
 
-				<p>Reading system developers should implement CSS support at the level of major browsers and publicly
-					document their user agent style sheets and how they interact with EPUB creator's style sheets.</p>
+				<!-- <p>Reading system developers should implement CSS support at the level of major browsers.</p> -->
 			</section>
 
 			<section id="sec-scripted-content">
@@ -2710,6 +2712,10 @@ alert("Feature " + feature + " supported?: " + conformTest);</pre>
 				<summary>Substantive changes since <a href="https://www.w3.org/TR/epub-rs-33/">EPUB Reading Systems
 						3.3</a></summary>
 				<ul>
+					<li>06-August-2025: Updated the <a href="#sec-css">section on handling CSS</a>, in particular 
+						the relationships between the user agent's and the creators' style sheets. 
+						See discussion in <a href="https://github.com/w3c/epub-specs/issues/2762">issue 2762</a>.
+					</li>
 					<li>05-June-2025: Consolidated all deprecated features under the unsupported features section. See
 						the comments in <a href="https://github.com/w3c/epub-specs/pull/2735#issuecomment-2944066035"
 							>pull request 2735</a>.</li>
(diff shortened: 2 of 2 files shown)


iherman:
### Comments on the proposed RS changes

> This isn't a requirement on a reading system, so I don't know how it fits in the standard. It's a request to the developers to document how their reading system works. I'm sure that's why we went with a non-normative should. (@mattgarrish)

Agreed; I use lower case should and refer to developers.

> [On the penultimate bullet point] These are reasonable to revisit. […] Saying SHOULD NOT for things that will almost certainly happen seems like a spec error - are there any Reading Systems that do not override the creator's style sheets in some way? (@bduga)

Yes, I agree, I tried to change the whole item.

> In practice, it is really hard to do this, and the cascade isn't a very good way to accomplish it. (@bduga)

Isn't this an implementation issue? In "abstract", the cascade model should still be valid, shouldn't it? If not, it becomes a complete break that authors will never understand… I have therefore kept the text for now.

> [On the last bullet item] LGTM, though perhaps WILL ABSOLUTELY would be more accurate. (@bduga)

I must admit I do not understand that remark. I kept the text as proposed; I also still maintain that the _MAY_ in the second sentence should be a _SHOULD_…

I have also removed the sentence "Reading system developers should implement CSS support at the level of major browsers." It sounds like an unnecessary sentence in view of the requirement of implementing the CSS spec, fullstops…

### Comments on the proposed authoring changes

@clapierre, I have added a text in the note, I hope that is what you meant. Also:

> Ok, but I am not sure that last one is needed. Reading Systems have to deal with all the old content that uses style attributes, so I am not sure avoiding them now makes any difference. (@bduga)

The question is whether the claim in the sentence is correct (ie, that, in some cases, RS-s will cannot properly handle the effects of style attributes). I must admit I am not sure myself. If the statement is correct, then it is o.k. to nudge creators to use css files, otherwise the sentence may indeed be unnecessary.

epub34/authoring/index.html
```diff
@@ -6215,6 +6215,22 @@ <h4>Introduction</h4>
 									<p>Some types of screens will render animations and transitions poorly (e.g., those
 										with high latency).</p>
 								</li>
+								<li>
+									<p>
+										Creators should be aware that a reading systems may override some aspects of 
+										the creator's style settings; creators may want to consult the reading systems' 
+										user agent style sheets, if publicly available. Furthermore, reading systems may
```
llemeurfr: Following Brady's previous comment, as we are in a NOTE section, it would be fair to replace it with a "will". Reading systems **will** definitely modify some rendering aspects as a result of user interaction.

epub34/authoring/index.html
```diff
@@ -6215,6 +6215,22 @@ <h4>Introduction</h4>
...
+										also modify some rendering aspects as a result of user interaction (e.g., choice 
+										of fonts, text justification, or foreground and background colors).
+									</p>
+								</li>
+								<li>
+									<p>
+										Creators should avoid using [[html]] [^html-global/style^] attributes in content documents.
```
llemeurfr: "html attributes" is not as understandable as the previous "inline styles".

epub34/rs/index.html
```diff
@@ -1205,22 +1205,24 @@ <h3>Cascading Style Sheets (CSS)</h3>
 							[=EPUB content documents=].</p>
 					</li>
 					<li>
-						<p id="confreq-css-overrides">SHOULD NOT override the EPUB creator's style sheets, but SHOULD do
-							so in a way that preserves the Cascade when necessary: through a user agent style sheet, or
-							[[html]] [^html-global/style^] attributes.</p>
+						<p id="confreq-css-overrides">MAY override parts of the creator's style sheets (for, e.g., ergonomic or
```
llemeurfr: Because RSes will override the author stylesheet at the demand of the user, I propose an alternative wording: 
"SHOULD preserve as much as possible the cascade model of CSS when overriding some rendering aspects as a result of user interaction."
The mention of user agent stylesheets and html attributes is irrelevant in such a set of requirements (this is a technical solution).

I'd like it followed by a new one:
"It MAY reset the author's stylesheet in case accessible display and interactions require such action". 
I'm thinking here about the equivalent of the "reader mode" modern browsers support, and that is also a feature of top RSes.

epub34/rs/index.html
```diff
@@ -1205,22 +1205,24 @@ <h3>Cascading Style Sheets (CSS)</h3>
 							[=EPUB content documents=].</p>
 					</li>
 					<li>
-						<p id="confreq-css-overrides">SHOULD NOT override the EPUB creator's style sheets, but SHOULD do
-							so in a way that preserves the Cascade when necessary: through a user agent style sheet, or
-							[[html]] [^html-global/style^] attributes.</p>
+						<p id="confreq-css-overrides">MAY override parts of the creator's style sheets (for, e.g., ergonomic or 
+							user interface reasons). If it does, it SHOULD do so in a way that preserves the cascade model
+							of CSS: through a user agent style sheet, or [[html]] [^html-global/style^] attributes.
+							Reading system developers should also publicly document the user agent style sheets and how they
```
llemeurfr: Only open-source RSes do this. This cannot be a requirement IMO, but it could be a Note.

epub34/rs/index.html
```diff
@@ -1205,22 +1205,24 @@ <h3>Cascading Style Sheets (CSS)</h3>
...
+							interact with EPUB creator's style settings.</p>
 					</li>
 					<li>
-						<p id="confreq-css-user-styles">It MAY override parts of the EPUB creator's style sheet because
-							of user interaction.</p>
+						<p id="confreq-css-user-styles">MAY override parts of the EPUB creator's style sheet as a result
```
llemeurfr: See above, this is a repetition of a previous sentence I propose to modify. 
Regarding the second sentence, we have kept a requirement above, expressed as "SHOULD apply EPUB creator style sheets as written to EPUB content documents."
Here, it would make sense to write:
"SHOULD provide the user with an affordance that reverts the display to the author's stylesheet". 
This expresses a feature requirement, but does not insist on technical solutions.

epub34/authoring/index.html
bduga: It seems like there are two options considered here, styles applied via the `style` attribute and styles contained in standalone files. What about style in a `style` element? Is that easier to deal with than the attribute? Presumably it is harder to handle styles in an element in the document vs styles in a style sheet. 

We also seem to be placing an extra burden on authors when it comes to styling, since they can no longer take advantage of the [highest specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascade/Specificity#inline_styles) styles. 

Without more clarity and some pretty specific case studies, I would move to delete this entire paragraph.

epub34/authoring/index.html
```diff
@@ -6215,6 +6215,22 @@ <h4>Introduction</h4>
...
+										user agent style sheets, if publicly available. Furthermore, reading systems may
+										also modify some rendering aspects as a result of user interaction (e.g., choice 
+										of fonts, text justification, or foreground and background colors).
+									</p>
+								</li>
+								<li>
+									<p>
+										Creators should avoid using [[html]] [^html-global/style^] attributes in content documents. 
+										Styling should always be done via CSS files, otherwise a reading system will 
+										not be able to easily override them when interacting with the user.
```
bduga: If we don't strike this, we need some clarity here. What does "not be able to easily override" mean? What is easy? How much added work will it be to support them? Should RSes do it anyway, since content already exists with them? If so, what is gained by discouraging them?

epub34/rs/index.html
mattgarrish: It's back to a non-normative lowercase "should" so it's akin to a note. (Relying on case is a bad idea even if it's technically allowed, as it's not terribly accessible and generally confusing to readers, so if there's a way to rephrase and keep the emphasis it would be better.)

epub34/authoring/index.html
mattgarrish: The second bracketed reference is respec shorthand for the html style attribute, so the text will come out as "[html] style attributes", with "style" linked to the html definition.

epub34/rs/index.html
bduga: I really struggle with what we are trying to say here. Is it that user adjustment should be done by creating a style sheet and inserting it in a reasonable place (e.g. before or after author style sheets)? That seems naive - a lot of the changes we need to make have to be done by modifying existing styles, and it seems to be an oddly intrusive requirement on reading system implementations.

epub34/authoring/index.html
iherman: Are we sure? It is obviously true for mainstream reading systems, but I do not know about epub readers that are implemented as a browser extension: they may "just" make the browser display the content without further ado.

To meet in the middle, I will use the word "usually".

epub34/authoring/index.html
iherman: I must admit I am just a go-between here, reflecting on @clapierre's comment. I let him reflect on this further.

epub34/authoring/index.html
iherman: See above.

epub34/rs/index.html
iherman: I change the sentence using "it is expected" to avoid misunderstanding around the "should" word...

epub34/rs/index.html
iherman: Doh. The first sentence is indeed superfluous, just a repetition; I will simply remove it.

I like the more concise text of yours, so I will adopt it _except_ for the _MAY_ vs. _SHOULD_. We do not have a consensus on this I believe.

epub34/rs/index.html
iherman: - @llemeurfr, I agree with the more concise text on the RS' action; I am not sure that we can really say _all_ RS-s override the user style sheet. I would prefer to leave the _MAY_ clause there.
  
  Note that if we do not have the _MAY_ the a cursory read of the text may be interpreted by having somewhere a text that _REQUIRES_ such action. This is not the case, but it would read like that, while maintaining the _MAY_ clause emphasizes the non-obligatory nature of this.

  With the first clause remaining I also shortened a bit your text to avoid repeating things.
- @llemeurfr, If we maintain the first clause ("_MAY_ override parts of...") then your second sentence becomes unnecessary, it would just repeat what is there.
- @bduga, I believe that the formulation of @llemeurfr removes your concern. How it is done is an implementation detail, but the model (ie, the cascade) should remain valid

iherman:
I am struggling a bit with the last sentence of the (new) text (which has been kept verbatim).

> In addition to supporting CSS properties as defined above, a reading system's user agent style sheet SHOULD support the [html] [suggested default rendering](https://html.spec.whatwg.org/multipage/#rendering).

I am not sure how this sentence relates to the user agent style sheet, mentioned in the bullet points above.

Isn't it better to add a new penultimate bullet item, saying:

> _SHOULD_ support the [html] [suggested default rendering](https://html.spec.whatwg.org/multipage/#rendering) when defining its own, user agent style sheet.

epub34/authoring/index.html
```diff
@@ -6215,6 +6215,22 @@ <h4>Introduction</h4>
 									<p>Some types of screens will render animations and transitions poorly (e.g., those
 										with high latency).</p>
 								</li>
+								<li>
+									<p>
+										Creators should be aware that a reading systems usually override some aspects of 
+										the creator's style settings; creators may want to consult the reading systems' 
+										user agent style sheets, if publicly available. Furthermore, reading systems may
```
mattgarrish: Does this bullet really belong in this note?

The note is to draw attention to poor support for CSS features but this new bullet goes on a tangent into cautioning authors about reading systems overriding their styles. That's not a poor support issue.

epub34/authoring/index.html
mattgarrish: This could maybe be solved by adding more context: "EPUB creators should be aware than any reading system that allows users to modify the display will override some aspects..."

epub34/authoring/index.html
```diff
@@ -6215,6 +6215,22 @@ <h4>Introduction</h4>
...
+										also modify some rendering aspects as a result of user interaction (e.g., choice 
+										of fonts, text justification, or foreground and background colors).
+									</p>
+								</li>
+								<li>
+									<p>
+										Creators should avoid using [[html]] [^html-global/style^] attributes in content documents.
```
mattgarrish: This really sounds like normative guidance. In any case, similar to my previous comment it's out of place in a note about poor support for CSS features.

It's also probably more of a consideration for the accessibility specification or techniques, although it would really help to know more about how reading systems modify the display. Inline styles (particularly author !important ones) can be problematic, but it's historically been more of an issue with applying user style sheets in browsers.

epub34/authoring/index.html
sueneu: I see your point. I’d hate to see it dropped completely. It is important for people new to ebook production to know that reading systems don’t behave like web browsers. 
Where might this caution be a better fit?

epub34/authoring/index.html
clapierre: This is something we would fail a GCA certification if we see inline style=... within an EPUB as then this style cannot be overridden. Letting publishers know this is not the correct way to add styling to an EPUB is very important.

mattgarrish: > within an EPUB as then this style cannot be overridden

Unless the reading system is changing styles by modifying the dom, or the epub is preprocessed to remove inline styles, or...? That's why it would help to know how reading system styling is applied.

What is the process for changing the display when a user picks custom display properties?

Inline styling, especially with !important, is bad on more levels than just accessibility, so I'm not disagreeing with you in principle. I'm just not sure this is an authoring spec-level normative restriction, which is what it currently sounds like but masked in a note.

bduga: > inline style=... within an EPUB as then this style cannot be overridden.

FWIW, Play Books does override styles in `style` attributes, which is one of the reasons I am against recommending only using the cascade to modify styling.

epub34/authoring/index.html
mattgarrish: I'm not sure the note, independent of these changes, is well placed at the moment, to be honest. It might work better as its own informative section on reading system implementation considerations.

Instead of having this text as a bullet, it could be moved to a separate paragraph following the list so that it's not linked to the problematic css features. I'll make a pull request tomorrow to demonstrate.

epub34/rs/index.html
```diff
@@ -1205,22 +1205,22 @@ <h3>Cascading Style Sheets (CSS)</h3>
...
+						<p id="confreq-css-overrides">MAY override parts of the creator's style sheets (for, e.g., ergonomic or 
+							user interface reasons). If it does, it SHOULD preserve as much as possible the cascade model of CSS.
+							Reading system developers are expected to also publicly document the user agent style sheets and how they
+							interact with EPUB creator's style settings.</p>
 					</li>
 					<li>
-						<p id="confreq-css-user-styles">It MAY override parts of the EPUB creator's style sheet because
-							of user interaction.</p>
+						<p id="confreq-css-user-styles">
+							It MAY provide the user with an affordance that reverts the display to the author's stylesheet.</p>
```
mattgarrish: Similar to the comment I made about the authoring spec, I don't see how these two new bullets fit with the rest of the list. Here we're talking about what reading systems have to do to support css and then veer off into how to override the publication's css.

I'd take these bullets and make a new subsection on overriding the publication's css.

I'd try to make some changes, but the more I look at the sentence about preserving the cascade the less I understand what it's trying to achieve. I also have reservations about getting into reading system affordances.

epub34/authoring/index.html
iherman: (For reference in the discussion: https://github.com/w3c/epub-specs/pull/2778 is the PR @mattgarrish is referring to)

epub34/authoring/index.html
iherman: Admin: if https://github.com/w3c/epub-specs/pull/2778 is accepted and merged, this thread of comments may become obsolete.

epub34/rs/index.html
iherman: > I'd try to make some changes, but the more I look at the sentence about preserving the cascade the less I understand what it's trying to achieve. 

I presume (the reference to cascade comes from the existing text) it tries to emphasize that, conceptually, an RS should not simply ignore an attribute setting in, say, the style element in the XHTML if the RS style sheet is defined for the publication as a whole. Which may be wishful thinking as of today; RS systems may change things all over the place already by simply ignoring some settings. Which means this whole notion might indeed be moot and better remove it. Alas!, because the author's mental model of CSS may go down the drain…

> I also have reservations about getting into reading system affordances.

That was one of the problems starting the whole thread: can I, as a user, have at least a way to set to my own styling, or am I doomed? How else could we express this?

sueneu: Is it users or creators who may or may not have the ability to set styling?

iherman: Setting styling baking into the EPUB: creators. Users/readers may have the ability to virtually change styling via the RS' user interface.

mattgarrish: But what does it mean to preserve as much as possible the cascade model of CSS? That's what was throwing me. It sounds like it's suggesting that reading systems are defining their own styling systems outside of CSS.

Are we intending to say that the reading system should preserve the existing cascade of styling defined by the author as much as possible? And does that in turn mean the reading system should only change what it absolutely has to, or does it mean something more technical like not pre-processing the files to remove styling such that it can never be restored?

Knowing what test(s) would prove this statement might help.

bduga: > But what does it mean to preserve as much as possible the cascade model of CSS?

I have taken it to mean that styling applied by the RS, either at the behest of the user or for other reasons, should be done using the cascade. But that is more "using the cascade model", not "preserving". Trying to wrap my mind around it now, and it is really unclear - how would a RS degrade the cascade model? Maybe if they implemented a custom CSS handler that calculated specificity in a non-standard way?

>  It sounds like it's suggesting that reading systems are defining their own styling systems outside of CSS.

Some do of course, and I suppose they could break the cascade.

iherman:
This was discussed during the [pmwg meeting on 04 September 2025](https://w3c.github.io/pm-wg/minutes/2025-09-04.html#f0a7).

<h3>PR 2771 - <a href="https://github.com/w3c/epub-specs/pull/2771" rel="noopener noreferrer">w3c/<wbr>epub-specs#2771</a></h3>
<p><cite>&lt;ivan&gt;</cite> <a href="https://github.com/w3c/epub-specs/issues/2762" rel="noopener noreferrer">w3c/<wbr>epub-specs#2762</a></p>
<p><cite>ivan:</cite> the reason this is started is because of me - there is a common thread in a number of issues e.g. in the dark mode thread<br>
<span>… this is about the relationship between the author and reading system - author spending time on css and rs invalidate what they do</span><br>
<span>… we use html5 and css so styling is better than epub2 had</span><br>
<span>… I opened a pull request to start to put in some hints for reading systems in both authoring and rs specs about the way they would/should/might handle author css</span><br>
<span>… the ship has sailed on changing reading systems - rses have taken the lead on rendering and control over things like margins, fonts, etc.</span><br>
<span>… the pull request is still open but in the authoring spec we have added some statements warning authors some of their styling will be overwritten</span><br>
<span>… and the hope is that at least some reading systems will publish their styling so that authors can find out what happens without experimentation</span><br>
<span>… the other aspect is to ask reading system devs to give the option to the user to have the author's styling dominate</span><br>
<span>… apple reader offers a choice for an original view but I haven't tested it too much</span><br>
<span>… I don't think we can do what I'd prefer to give whole power to the authors</span></p>
<p><cite>mgarrish:</cite> I think that mostly covers where we are, the disagreement is on how the affordance works, is it specced out well enough<br>
<span>… it's somewhat implemented, but is it repeatable?</span><br>
<span>… that is as far as we've gotten now, is it realistic and how does it work</span></p>
<p><cite>ivan:</cite> we looked at both the css sections in both specs and the text came from earlier versions of epub and the meaning is not really clear<br>
<span>… the pr fixes some of those ambiguities</span></p>
<p><cite>LaurentLM:</cite> when reading systems add some css to the author it's because they have to normalize all the peculiarities<br>
<span>… it is difficult to remove the rs css - it will break the view if you take out too much</span><br>
<span>… most reading systems try to make a minimal changes as possible</span><br>
<span>… apple is not the only one to offer to go back to author styling</span><br>
<span>… thorium offers an option to get back to author css but still with the required optimizations done</span><br>
<span>… having a public reading system css would be good but I expect only open source ones will do so</span></p>
<p><cite>Brady:</cite> +1000 to laurent - the css reading systems add is to take the book from looking like garbage - without margins text would be a complete mess<br>
<span>… you keep adding css until the content looks good on your system</span><br>
<span>… the author doesn't know the problems of a device because the rs makes it look legible</span><br>
<span>… if we give the author complete control then the content will not render as they expect</span><br>
<span>… I do see where it does make sense like on fonts although even then publishers don't care too much and users tend to not want fonts changed from their preference</span><br>
<span>… publisher intent for font size is not always clear so google, for example, will figure out an optimal 100% size based on the device</span><br>
<span>… play books has a true original mode where you can see the content down to the page breaks and no one uses it because it's unreadable on their phone</span><br>
<span>… people who have found this feature have called support because they can't figure out how to get out the rendering</span><br>
<span>… adding a flag to get rid of the effort to make the content readable is strange - but if we really feel it is important as an optional feature okay as it is vague enough</span></p>
<p><cite>sueneu:</cite> I agree with brady about some aspects of author intent - author's don't always care about font<br>
<span>… margins and text size should be calculated to the device</span><br>
<span>… but sometimes authors want indented text, text to look like it's handwritten, especially in YA text</span><br>
<span>… sometimes they want a background to make it look like it was typed in a phone</span><br>
<span>… would it be possible to separate out the parts the reading system takes responsibility for and what authors can use css to customize</span></p>
<p><cite>ivan:</cite> this looks and sounds like a cultural difference - everything that Brady says could have been said by browser vendors to justify changing web sites<br>
<span>… no browser does this, outside perhaps of reading mode</span><br>
<span>… pagination is special to epub and I understand that part</span><br>
<span>… but why such a huge difference from the web for everything else - could be cultural</span></p>
<p><cite>ivan:</cite> how do fixed layouts get published if reading systems override their settings?</p>
<p><cite>&lt;LaurentLM&gt;</cite> FXL is totallty different. The RS does not normalizes.</p>
<p><cite>wendyreid:</cite> reading systems pay far more attention to FXL than reflow<br>
<span>… I agree with most of what Brady said but kobo respects much more of the original rendering</span><br>
<span>… one of the big challenges is making content work for so many devices</span><br>
<span>… because publishers convert from print they think about print pagination and don't think about phone displays, for example</span><br>
<span>… publishers aren't always considering every device their content will render on</span><br>
<span>… but there must be a balance we can strike between publisher and rs</span><br>
<span>… fonts is a case where sometimes it doesn't matter but sometimes it does and you need to see the differences</span><br>
<span>… but how do we sort through all the properties to determine what is important and what is not</span><br>
<span>… maybe this should be best practice styling recommendations that we know work well</span></p>
<p><cite>Brady:</cite> I think font is one of the more important ones that get changed and why reading systems differ from browsers<br>
<span>… it is problematic to overwrite math fonts or handwritten fonts but we don't want to use fonts like roboto</span><br>
<span>… what play books does is parses all the font rules and inserts a better font before a generic one</span><br>
<span>… you can't do that with the css cascade so you have to edit the css to do that</span><br>
<span>… a browser is a user agent but a reading system has a user agent - if you don't intercept you can't make the fix</span><br>
<span>… I have read books with different fonts and indents but we spent a lot of time making the outer margins work so the inner ones look good</span><br>
<span>… in terms of fixed layout we don't do anything - we don't let you change fonts or alter the styling</span><br>
<span>… we have an entire task force on fxl accessibility and it's still hard to read even for sighted people</span><br>
<span>… our culture is different because we have to show books created twenty years ago and make them look like they were created today</span><br>
<span>… you also don't go to the chrome store and buy a web site and be okay with it looking like crap</span></p>
<p><cite>&lt;sueneu&gt;</cite> +1 Brady on business model differences between the web and the RS</p>
<p><cite>Brady:</cite> if your competitor's product looks better than you lose business</p>
<p><cite>DaleRogers:</cite> I know content creators might create epubs with pages or indesign and never get under the hood to see the css<br>
<span>… so I understand what brady is saying about having to deal with content that was created by these programs from years ago</span><br>
<span>… the validator will tell me if it's conforming but I don't know all of what it checks</span><br>
<span>… but how does the author know what is the best practice for different reading systems - we follow the spec and hope for the best</span><br>
<span>… is there a css validator that checks if it will be problematic with different reading systems?</span><br>
<span>… I wonder if it would be helpful if we knew ahead of time how reading systems work and what css will work - is that a possibility?</span></p>
<p><cite>gautierchomel:</cite> ebooks are not updated as often as web sites - it is easier to update a web site<br>
<span>… as a publisher I want my css to be as simple as possible</span><br>
<span>… on the web designers have media queries to target different displays</span><br>
<span>… not sure I want to have to do that for reading systems</span><br>
<span>… I have not heard any complaints about this topic from french publishers but I will ask them</span><br>
<span>… we need more industry voices in this group to understand whether this is an individual or industry issue</span></p>
<p><cite>sueneu:</cite> I'm interested in whether this is an industry issue - reading systems seem to have taken the issue in stride - more voices would be better</p>
<p><cite>LaurentLM:</cite> we should have some authoring recommendations like choosing fonts wisely</p>
<p><cite>&lt;sueneu&gt;</cite> +1 Laurent</p>
<p><cite>LaurentLM:</cite> authors also don't follow some rules to make epubs - make we need more rules in the spec or a supplemental document</p>
<p><cite>&lt;gautierchomel&gt;</cite> <a href="https://readium.org/css/docs/" rel="noopener noreferrer">https://<wbr>readium.org/<wbr>css/<wbr>docs/</a></p>
<p><cite>gautierchomel:</cite> adding a link to the readium css<br>
<span>… ten years ago it was more problematic but now if I don't do strange things the display is good</span><br>
<span>… I don't even update books yearly - I want my content to display well for years to come</span></p>
<p><cite>Brady:</cite> it's great that readium publishes this but how many authors go through the css?<br>
<span>… if we find problems we typically go back to the publisher and request that they fix it</span><br>
<span>… the problem is they often come back and say it works on ibooks and then we have to recreate someone else's display to make the author content look right</span><br>
<span>… if I still worked at play books it would have fallen on me to publish the css but would anyone have looked or cared?</span><br>
<span>… the response will likely remain it works elsewhere to make it work - but I might be pessimistic</span></p>
<p><cite>wendyreid:</cite> I've had the same experience as brady - but as gautier said things have gotten better<br>
<span>… books made twenty years ago were really poor quality but today they are quite good</span><br>
<span>… if we tell people what to do now they may already have figured these things out</span><br>
<span>… it would be great if authoring tools could warn users about display choices that are really going to make a mess</span><br>
<span>… but we're not quite there and that may be where best practices come in</span></p>
<p><cite>sueneu:</cite> agree with directing tool makers to the reading system css than authors<br>
<span>… flightdeck is an online tool you subscribe to monthly and they will tell you if a book will work</span></p>
<p><cite>&lt;sueneu&gt;</cite> <a href="https://ebookflightdeck.com/" rel="noopener noreferrer">https://<wbr>ebookflightdeck.com/</a></p>
<p><cite>sueneu:</cite> they have a supplier-compliant matrix for checking</p>
<p><cite>LaurentLM:</cite> some publishers do test with thorium css and not just ibooks</p>
<p><cite>DaleRogers:</cite> I'm wondering if authors did not include css and did everything in markup would that show up fine or do reading systems expect some css?</p>
<p><cite>wendyreid:</cite> we generally expect at least some css but reading systems will handle a book without any - some things like asides will need styling to differentiate</p>
<p><cite>&lt;Zakim&gt;</cite> ivan, you wanted to where we go from here?</p>
<p><cite>Brady:</cite> if you look back at books from 20 years ago they didn't have much css and they display fine</p>
<p><cite>ivan:</cite> so where do we go from here?<br>
<span>… I think matt and I are at the last round of discussion with the PR</span><br>
<span>… I don't think we should merge it as-is but get more people to look at it</span><br>
<span>… it is much better than it used to be but in light of the discussion today we should get more review</span><br>
<span>… to respond to gauitier I was wondering if media queries would cover all the cases we are concerned about</span><br>
<span>… maybe we should discuss with the css people in Kobe whether there are media queries that are reading system specific</span></p>
<p><cite>Brady:</cite> the biggest thing missing is the knowledge of a page and from media queries is page size<br>
<span>… it's probably not solvable from the css side because they have pages but no one uses them</span></p>
<p><cite>wendyreid:</cite> when I asked about colour issues they directed me to their issue tracker - maybe we could ask about this as they have more pull with browsers<br>
<span>… in terms of next steps the PR is not the direction we want to take</span><br>
<span>… the text improvements are okay but not recommendations</span></p>
<p><cite>ivan:</cite> there are no new normative requirements in the text</p>
<p><a href="https://github.com/w3c/epub-specs/issues/2762" rel="noopener noreferrer">w3c/<wbr>epub-specs#2762</a></p>
<p><cite>wendyreid:</cite> people can use the issue to discuss further</p>

<hr /></details>

mattgarrish:
Merging the non-contentious parts of the pull request so we can start over on whether we need an affordance or better guidance for the styling conflicts.

PR #2772, opened 2025-08-06
Adjusted text color for the 'support' class in dark mode

iherman:
The color was a light gray on yellowish background, which was unreadable. I set the text color to dark gray and left the background as is.

Best seen on its effect on RS:

See:

    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/support-class-in-dark-mode/epub34/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/support-class-in-dark-mode/epub34/rs/index.html)

### DIFF
--- a/common/css/common.css
+++ b/common/css/common.css
@@ -355,6 +355,11 @@ body.darkmode p.caution, body.darkmode div.caution {
     background-color: hsl(180, 40%, 20%);
 }
 
+body.darkmode p.support, body.darkmode div.support {
+    background-color: rgb(255, 255, 230);
+    color: rgb(30, 30, 30)
+}
+
 body.darkmode div.caution-title > span {
     color: rgb(255,140,0)
 }


mattgarrish:
It's kind of shockingly bright with everything else black.

What about using var(--note-bg) and var(--note-text), instead, and extend a 1px yellow border around the top, right, and bottom to make it more like the note boxes in dark mode?

iherman:
This is a purely stylistic change, does not change the text. Merging it, to get it out of the way, we can always refine it later if we wish. At this moment, it will make things more readable.

iherman:
> It's kind of shockingly bright with everything else black.
> 
> What about using var(--note-bg) and var(--note-text), instead, and extend a 1px yellow border around the top, right, and bottom to make it more like the note boxes in dark mode?

Oops, sorry, my bad, I completely missed this remark. I should not have merged this. 

There is no "unmerge", but I will do a separate one with your proposals. Actually, there may be a need to completely rework the common.css file using variables (with a special variable setting for dark mode)...

PR #2784, opened 2025-08-28
Remove xmlns and xml:lang from root html elements

mattgarrish:
As the title says. We don't need to pretend these are xhtml documents when they're just html using some xhtml conventions.

I'm tempted to globally replace "/>" with ">", but it's not strictly necessary. Self-closing tags are valid html.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2784.html" title="Last updated on Aug 28, 2025, 12:44 PM UTC (4c83ae8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2784/bce520b...4c83ae8.html" title="Last updated on Aug 28, 2025, 12:44 PM UTC (4c83ae8)">Diff</a>

### DIFF
--- a/epub34/a11y-tech/index.html
+++ b/epub34/a11y-tech/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html xmlns="http://www.w3.org/1999/xhtml" lang="en-US" xml:lang="en-US">
+<html lang="en-US">
 	<head>
 		<meta charset="utf-8" />
 		<title>EPUB Accessibility Techniques 1.1.1</title>
--- a/epub34/a11y/index.html
+++ b/epub34/a11y/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html xmlns="http://www.w3.org/1999/xhtml" lang="en-US" xml:lang="en-US">
+<html lang="en-US">
 	<head>
 		<meta charset="utf-8" />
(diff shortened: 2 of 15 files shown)


mattgarrish:
> I'm tempted to globally replace "/>" with ">"

Scratch that idea. It'd probably wreak havoc on examples.

GeorgeKerscher:
I want to keep my validator, HTML-validate, happy. If I can simply replace /> with >, I would be happy. Currently I did not feel comfortable making that change, because I though Respec needed it.
Perhaps a note in Matt's editoring guide.the

mattgarrish:
That's fine for the document's you're creating, George. There's just no easy way to fix all the existing specification documents. We'd have to search on the possible self-closing tags and fix them manually, otherwise we could turn xml and xhtml examples non-conforming.

iherman:
Any idea why the check failed on A11y?

The action script says:

```
      [INFO] [Timeout: 19181ms] Processing ReSpec document...
  Error: ROR] Plugin core/biblio took too long.
  Error: ROR] Plugin core/biblio took too long.
```

and then it aborted.

iherman (approved):

mattgarrish:
No idea, other than it looks like some kind of backend timeout. I'll rerun the action and hopefully it just goes away.

PR #2785, opened 2025-08-29
Update the reading system requirements for css in 2771

mattgarrish:
This is an attempted update to the reading system requirements in #2771. I split off the two bullets that were dealing with overriding author styles and adding the affordance to view an altered version of the publication and created a new section for them.

Let me know if makes things better or worse...

* * *

[Preview](https://raw.githack.com/w3c/epub-specs/update/rs-req/epub34/rs/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/css-sections-review-(issue-2762)/epub34/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/update/rs-req/epub34/rs/index.html)

### DIFF
--- a/epub34/rs/index.html
+++ b/epub34/rs/index.html
@@ -1219,13 +1219,16 @@
 					<h4>Author styling overrides</h4>
 
 					<p id="confreq-css-overrides">A reading system MAY override parts of an [=EPUB publication|EPUB
-						publication's] styling (e.g., for ergonomic or user interface reasons), but SHOULD only do so in
-						a way that allow the author's original design to be restored.</p>
+						publication's] styling (e.g., for ergonomic or user interface reasons). Morevoer, they MAY allow
+						users to set default themes and style preferences (e.g., for more accessible reading) that
+						further alter the original design of a publication.</p>
 
-					<p id="confreq-css-user-styles">Reading systems MAY provide users with an affordance that removes
-						any reading system-applied styling, thereby allowing users to view a publication as designed by
-						the author (e.g., so users can check the original design if a publication refers to location,
-						color, etc. that gets overriden by the reading system styling).</p>
+					<p id="confreq-css-user-styles">These changes often make it impractical to render an EPUB
+						publication exactly as designed by an author, even if that is what the user wants. To assist
+						users in viewing an EPUB publication as closely as possible to the author intent, however,
+						reading systems MAY provide an affordance that reverts as much reading system-applied styling as
+						feasible. For example, it could reset any user-applied styling and themes back to the reading
+						system defaults.</p>
 				</section>
 			</section>
 
--- a/epub34/rs/index.html
+++ b/epub34/rs/index.html
@@ -1223,12 +1223,11 @@
 						users to set default themes and style preferences (e.g., for more accessible reading) that
 						further alter the original design of a publication.</p>
 
-					<p id="confreq-css-user-styles">These changes often make it impractical to render an EPUB
+					<!-- <p id="confreq-css-user-styles">These changes often make it impractical to render an EPUB
 						publication exactly as designed by an author, even if that is what the user wants. To assist
 						users in viewing an EPUB publication as closely as possible to the author intent, however,
 						reading systems MAY provide an affordance that reverts as much reading system-applied styling as
-						feasible. For example, it could reset any user-applied styling and themes back to the reading
-						system defaults.</p>
+						feasible.</p> -->
 				</section>
 			</section>
 
--- a/epub34/rs/index.html
+++ b/epub34/rs/index.html
@@ -1038,14 +1038,14 @@
 						<p>Reading system support for the <a data-cite="rdfa-core#s_model">attribute processing
 								model</a> [[rdfa-core]] is OPTIONAL.</p>
 					</section>
-					
+
 					<section id="sec-xhtml-its">
 						<h5>Internationalization Tag Set (ITS)</h5>
-						
+
 						<p>Reading system support for <a data-cite="its20#conformance-product-html5-its">processing ITS
 								markup</a> [[its20]] is OPTIONAL.</p>
 					</section>
-					
+
 					<section id="sec-xhtml-custom-attributes">
 						<h5>Custom attributes</h5>
 
@@ -1205,22 +1205,29 @@
 							[=EPUB content documents=].</p>
 					</li>
 					<li>
-						<p id="confreq-css-overrides">MAY override parts of the creator's style sheets (for, e.g., ergonomic or 
-							user interface reasons). If it does, it SHOULD preserve as much as possible the cascade model of CSS.
-							Reading system developers are expected to also publicly document the user agent style sheets and how they
-							interact with EPUB creator's style settings.</p>
-					</li>
-					<li>
-						<p id="confreq-css-user-styles">
-							It MAY provide the user with an affordance that reverts the display to the author's stylesheet.</p>
+						<p id="confreq-css-rs-html-default">SHOULD support the [[html]] <a data-cite="html#rendering"
+								>suggested default rendering</a> in their user agent style sheet(s).</p>
 					</li>
 				</ul>
 
-				<p id="confreq-css-rs-html-default">In addition to supporting CSS properties as defined above, a reading
-					system's user agent style sheet SHOULD support the [[html]] <a data-cite="html#rendering">suggested
-						default rendering</a>.</p>
+				<div class="note">
+					<p>Reading system developers are expected to publicly document their user agent style sheet(s) and
+						how they interact with EPUB creator's style settings. </p>
+				</div>
 
-				<!-- <p>Reading system developers should implement CSS support at the level of major browsers.</p> -->
+				<section id="sec-css-overrides">
+					<h4>Author styling overrides</h4>
+
+					<p id="confreq-css-overrides">A reading system MAY override parts of an [=EPUB publication|EPUB
+						publication's] styling (e.g., for ergonomic or user interface reasons), but SHOULD use
+						CSS-compatible methods (user agent style sheet, or [[html]] [^style^] attributes) that allow the
+						author's original design to be restored.</p>
+
+					<p id="confreq-css-user-styles">Reading systems MAY provide users with an affordance that removes
+						any reading system-applied styling, thereby allowing users to view a publication as designed by
+						the author (e.g., so users can check the original design if a publication refers to location,
+						color, etc. that gets overriden by the reading system styling).</p>
+				</section>
 			</section>
 
 			<section id="sec-scripted-content">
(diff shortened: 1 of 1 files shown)


epub34/rs/index.html
```diff
@@ -1205,22 +1205,29 @@ <h3>Cascading Style Sheets (CSS)</h3>
...
+					<p>Reading system developers are expected to publicly document their user agent style sheet(s) and
+						how they interact with EPUB creator's style settings. </p>
+				</div>
+
+				<section id="sec-css-overrides">
+					<h4>Author styling overrides</h4>
+
+					<p id="confreq-css-overrides">A reading system MAY override parts of an [=EPUB publication|EPUB
+						publication's] styling (e.g., for ergonomic or user interface reasons), but SHOULD use
+						CSS-compatible methods (user agent style sheet, or [[html]] [^style^] attributes) that allow the
```
mattgarrish: To question my own pull request... do we really need to say anything about css here? Could this recommendation just be to use methods that allow the author's original design to be rendered? In other words, just don't be destructive in restyling however you choose to do it.

epub34/rs/index.html
```diff
@@ -1205,22 +1205,29 @@ <h3>Cascading Style Sheets (CSS)</h3>
...
+						author's original design to be restored.</p>
 
-				<!-- <p>Reading system developers should implement CSS support at the level of major browsers.</p> -->
+					<p id="confreq-css-user-styles">Reading systems MAY provide users with an affordance that removes
```
bduga: This feels weird in a spec. What are the implications of this? Reading Systems may also provide search functionality, but why would we mention that? They might have some feature - who have we helped by saying that?

epub34/rs/index.html
bduga: I think removing CSS comments is a good idea, and just leave it with your vague "don't be destructive". I don't really know what CSS-compatible methods are - for instance if I rewrite the author styles, is that "CSS compatible"? In theory, the most destructive change is revertible if you keep the source content around, so do we even need to say anything? I am not sure what I would do differently due to this SHOULD.

bduga (commented):
I am still not sure what these statements are trying to accomplish. As a RS developer I am not sure what I should take away from them. They seem almost more directed at a product manager.

epub34/rs/index.html
mattgarrish: My take is that this is mostly aspirational guidance to raise some awareness that author design gets lost in the highly customizable nature of epub reading systems.

It feels like we need a basic default style sheet that's implemented by all reading systems first. That would be the more natural starting point from where author styling diverges from whatever common look and feel, theming, user preference styling gets applied. But I'm not optimistic about the prospects of such a thing at this point.

epub34/rs/index.html
mattgarrish: I took that part out in the last commits I made, but I share the same concern that it's kind of saying nothing.

I guess what I'm struggling with is why SHOULD a reading system be bound in this way when nothing is dependent on it? If the reading system doesn't provide the optional affordance to revert styling, then who cares how it applies styling to the publication.

And if it provides that affordance, then why do we need to tell devs to apply their styling in a way that isn't destructive? How else do you get back?

It might be simpler to say that reading systems may override the publication's styling and they may provide an affordance to revert any reading system styles. I know mentioning optional features is kind of pointless, but it's also harmless advice.

epub34/rs/index.html
```diff
@@ -1205,22 +1205,28 @@ <h3>Cascading Style Sheets (CSS)</h3>
...
+				<div class="note">
+					<p>Reading system developers are expected to publicly document their user agent style sheet(s) and
+						how they interact with EPUB creator's style settings. </p>
+				</div>
+
+				<section id="sec-css-overrides">
+					<h4>Author styling overrides</h4>
+
+					<p id="confreq-css-overrides">A reading system MAY override parts of an [=EPUB publication|EPUB
+						publication's] styling (e.g., for ergonomic or user interface reasons), but SHOULD only do so in
```
```suggestion
					<p id="confreq-css-overrides">A reading system MAY override parts of an [=EPUB publication|EPUB
						publication's=] styling (e.g., for ergonomic or user interface reasons), but SHOULD only do so in
```

epub34/rs/index.html
iherman: > My take is that this is mostly aspirational guidance to raise some awareness that author design gets lost in the highly customizable nature of epub reading systems.

Indeed. The comparison with search is not, in my opinion, fair. Authors do not have to really prepare for search; it just happens. And there is not much they can do to influence it. (This may actually change with annotations.) On the other hand, authors put time and energy into styling their pages, and they want to be in control.

To push things to an extreme: why do we allow for such a hugely powerful tool as CSS if authors cannot trust that whatever they do will really materialize in an RS?

> It feels like we need a basic default style sheet that's implemented by all reading systems first. That would be the more natural starting point from where author styling diverges from whatever common look and feel, theming, user preference styling gets applied. But I'm not optimistic about the prospects of such a thing at this point.

Me neither. That will not happen, with the low participation (and that is a euphemism) of RS implementers this will not happen... Alas!

epub34/rs/index.html
```diff
@@ -1205,22 +1205,28 @@ <h3>Cascading Style Sheets (CSS)</h3>
...
+								>suggested default rendering</a> in their user agent style sheet(s).</p>
 					</li>
 				</ul>
 
-				<p id="confreq-css-rs-html-default">In addition to supporting CSS properties as defined above, a reading
-					system's user agent style sheet SHOULD support the [[html]] <a data-cite="html#rendering">suggested
-						default rendering</a>.</p>
+				<div class="note">
+					<p>Reading system developers are expected to publicly document their user agent style sheet(s) and
+						how they interact with EPUB creator's style settings. </p>
```
iherman, line 1215: Just wondering. The user agent style sheet has, I presume, several parts: some general styling that are always active (e.g., related to pagination), and those aspects that are related to user interface choices. E.g., when I change to sepia, then certain CSS properties related to background color change. Would it be possible for RS to provide a documentation that make those type of adaptations also public? That might help authors to adapt their content...

epub34/rs/index.html
llemeurfr: This discussion misses the point that users, not RSes, require to have the last word on the presentation of the publication. This is about accessibility and comfort. 
Is there a basic default stylesheet for authors? no.  How could there be a basic default stylesheet for RSes after 10 years, now that the dust has settled on EPUB 3?
OK to state that RSes have the duty to allow the user, *as much as possible*, to see the ebook as the author has created it; until the user decides to customize the layout *as much as he needs*.

mattgarrish: I don't think anyone is saying reading systems, or by extension users, can't modify the publication styles, or even have it done by default. The desire here, as you say, is to allow users to opt to start as close to what the author wanted them to see as possible. That's where epub reading systems diverge from browsers and what seems to be the perennial point of annoyance for authors.

Browsers have a default style sheet. You know what you're working with when you begin styling your site. EPUB doesn't have a similar concept of a minimalist starting point from which all authors can build, and so where does this "revert to author's styles" button take you? I don't know the answer to that and why I find this proposal problematic. Take out all the reading system-applied styling and does the book become a visual disaster because margins and other necessary styling is gone?

I don't think time is the issue, since html didn't have a common default style sheet for browsers for a long time and now it does, but I'm sceptical of the feasibility of figuring this out because we likely don't have enough representation to get such a thing done. The html default style sheet, I would imagine, is probably close to what an epub default style sheet would be, but maybe not.

But for argument's sake, say we get a few reading systems that try to implement this feature with no common idea of what reverting all the styling means. Is every one going to have a different take on what reverting to the author presentation means and are we just going to have another feature that makes neither authors nor users happy?

mattgarrish: See if the last commit improves things at all. I added an additional statement that reading systems may allow users to change the styling, as reading system-applied styling is only one part. I then use that to explain what this affordance might do - i.e., revert just the themes and styling the user has applied.

Does that make this proposal more palatable?

iherman: > Take out all the reading system-applied styling and does the book become a visual disaster because margins and other necessary styling is gone?

I do not think this would be the case. My original scenario is something like :

1. The author, possibly based on the RS's published default style sheet, creates and designs a document, which has a justified text and a red background color; both of these are set via CSS on the `<body>` element.
2. The RS may decide to overwrite the styling, based on its default theme, which displays an unjustified text on white background. Notes that by doing so, it overwrites the style element on `<body>`. The reader may also change theme, which may turn the background into dark brown, say, still ignoring what is on `<body>`

The question is: is there a way to tell the browser to use the justified text and red background? Or is it completely lost, i.e., the author did something superfluous?

To put it another way: the author's CSS is ignored by using the user agent style sheet (or elements thereof) as the top of the css stack (top of the cascade), so to say, instead of the bottom one (which, in my view, should be the case). And when I say it is time related, what I meant is that we cannot change that anymore because RS's have been doing that for decades now and, as a suboptimal choice what I am looking for is the possibility to remove those extra things from the top of the stack following user interaction. 

Put it another way: why don't we trust the authors/designers? The browsers seem to trust website designers after all (and yes, there are lots of absolute rubbish on the web, but the market works on pushing them aside).

epub34/rs/index.html
```diff
@@ -1205,22 +1205,31 @@ <h3>Cascading Style Sheets (CSS)</h3>
...
+						users to set default themes and style preferences (e.g., for more accessible reading) that
+						further alter the original design of a publication.</p>
 
-				<!-- <p>Reading system developers should implement CSS support at the level of major browsers.</p> -->
+					<p id="confreq-css-user-styles">These changes often make it impractical to render an EPUB
+						publication exactly as designed by an author, even if that is what the user wants. To assist
+						users in viewing an EPUB publication as closely as possible to the author intent, however,
+						reading systems MAY provide an affordance that reverts as much reading system-applied styling as
+						feasible. For example, it could reset any user-applied styling and themes back to the reading
+						system defaults.</p>
```
```suggestion
						users in viewing an EPUB publication as closely as possible to the author intent, however,
						reading systems MAY provide an affordance that reverts as much reading system-applied styling as
						feasible. For example, it could reset any user-applied styling and themes back to the reading
						system defaults.</p>
```

- The term 'user', though precise, is misleading me; maybe 'reader' is better.
- The example is not the one I would use; what I am looking for to the author's defaults, and not the reading system defaults.

epub34/rs/index.html
mattgarrish: I don't think epub was ever conceived as an author-primary display format, at least not until fixed layouts came along. The idea was that the content could be fit to any reflowable device, down to a reading system that only might display a line of text at a time (there was one like that but I don't know what became of it). Common look and feel (plus themes) was a useful enhancement as many ebooks have minimal styling and would look pretty ugly otherwise. Users also don't generally want to move from book to book and have radically different appearances.

So is the problem here really that users need to somehow revert to some select author styles that maybe make the publication truer to the original intent, or is the problem that what styles reading systems are likely to change, and what authors should avoid making significant to their content, isn't well documented?

epub34/rs/index.html
sueneu: > The term 'user', though precise, is misleading me; maybe 'reader' is better.

I’m not generally a fan of the word ‘user.’ People sometimes refer to their devices (i.e. a Kindle) as a ‘reader.’  In this case, ‘user’ is the less ambiguous term. Unless we want to add an adjective like “human reader.”

>I am looking for to the author's defaults, and not the reading system defaults.

Apple Books software has a theme called “Original” that will display some of the css in the epub package. 

1230-1231 Perhaps “author original settings” instead of “reading system defaults”

epub34/rs/index.html
iherman: > 
> So is the problem here really that users need to somehow revert to some select author styles that maybe make the publication truer to the original intent, or is the problem that what styles reading systems are likely to change, and what authors should avoid making significant to their content, isn't well documented?

Actually, both are problems, but what motivated me so far is primarily the former.

epub34/rs/index.html
iherman: > Apple Books software has a theme called “Original” that will display some of the css in the epub package.

I must admit I did not know that, but that is something that I am looking for (modulo what "some" means). 

I had a look with a very simple epub book which does, essentially, one thing with the content via CSS, namely it requires justification. That feature works: none of the other Apple themes justify by default (and they also change background color and font), and pushing that option switches back to the original background color and text justification.

(The book does not set the background color which seems to set to black, which ~~probably~~ corresponds to the fact that my system is set to be in dark mode.)

mattgarrish:
Do you think we can merge this back into 2771 so we only have one PR to review?

Another thought would be to remove the affordance paragraph, since we still haven't settled on it, just to be able to merge everything else back into main. We could then reopen a new pull request to pick up on whether we need an affordance and/or additional best practices somewhere, or continue that discussion in the original issue.

iherman:
> 
> Another thought would be to remove the affordance paragraph, since we still haven't settled on it, just to be able to merge everything else back into main. We could then reopen a new pull request to pick up on whether we need an affordance and/or additional best practices somewhere, or continue that discussion in the original issue.

Let us do that.

Issue #2754, opened 2025-07-01
Deprecating flow as a metadata


HadrienGardeur:
### Section

https://www.w3.org/TR/epub-33/#flow

### Describe the problem

As mentioned in https://github.com/w3c/epub-specs/issues/2750, letting content creators hint how reflowable EPUB publications should be laid out feels at odds with the philosophy of EPUB.

From an accessibility point of view, if this hint was followed by RS it could be problematic. If a user prefers scrolling through EPUB publications because it makes it easier to use a reading ruler, forcing them to use a different layout when opening a book that contains `rendition:flow` set to `paginated` isn't the best idea.

On the RS side, it's also worth pointing out that `scrolled-doc` and `scrolled-continuous` are often implementation differences, rather than two options available to users.

### Describe the fix or new feature you propose

I'm proposing that we should deprecate most of the values for `rendition:flow`: `paginated`, `scrolled-doc` and `auto`.

For `scrolled-continuous`, I have mixed feelings due to the fact that it's relevant strictly for maximizing backward compatibility with continuously scrolled comics in Japan (where it's combined with FXL). 

One option could be to mark all values as deprecated but keep this in the spec in a future 3.5 revision, unlike other properties marked for deprecation.

iherman:
> One option could be to mark all values as deprecated but keep this in the spec in a future 3.5 revision, unlike other properties marked for deprecation.

Purely administratively, I would prefer not to do that. If we define a feature as "deprecated", it should mean something clear. 

In [EPUB 3.3](https://www.w3.org/TR/epub-33/#deprecated) we said:

>> A deprecated feature is one the Working Group no longer recommends for use in this version of the specification. Deprecated features typically have limited or no support in [reading systems](https://www.w3.org/TR/epub-33/#dfn-epub-reading-system) and/or usage in [EPUB publications](https://www.w3.org/TR/epub-33/#dfn-epub-publication).

Obviously, these features were sort of "we would have preferred to remove it from the spec but our hands are tied by the charter". We seem to go in the direction of saying that "deprecated features will be removed from next release in case the charter allows it". Which is also fine as a definition. But it should be one or the other.

We may have to be inventive and come up with an alternative terminology for these features (or invent a term for those that we propose for removal in 3.5).

HadrienGardeur:
I hear your arguments @iherman and they make a lot of sense.

[...]

Issue #2751, opened 2025-06-30
Deprecating controls over orientation


HadrienGardeur:
### Section

https://www.w3.org/TR/epub-33/#orientation

### Describe the problem

The current spec allows content creators to provide an orientation in which the content creator intends the content to be rendered in.

If followed, this could force users to rotate their device in order to see part (spine overrides) or all (metadata) of the content, which clearly goes against the idea of EPUB being a natively accessible publication format (not everyone can easily rotate a device at will).

This also goes against the general philosophy of EPUB (even FXL), where users are generally free to consume content however they want.
Pushed to the extreme, this property allows content creators to alternate between `portrait` and `landscape` every other resource, making things even worse than forcing it once.

In practice, reading systems mostly ignore this property, allowing the user to decide how they prefer to read their content.

### Describe the fix or new feature you propose

Controls over orientation (metadata and spine overrides) should be deprecated in EPUB 3.4 and marked for removal in a future 3.5 revision.

sueneu:
I know publishers and creatives who feel strongly about the orientation of their work in Ebooks.  In some cases, changing orientation would degrade the creator’s intended message.

The use case that comes to mind is an illustrated children’s title. 

-  Often, artwork is created across a two-page print spread. Orienting that in portrait mode would result in a very small presentation of the artwork. Or splitting the artwork page by page and cropping the artwork in a way that could obscure meaning.
- It is not unusual for text to be part of the illustration. Text shown in a different orientation than intended would be difficult to read.

Another use case— tabular data, charts and graphs rendered as images

- Tables don’t always render well in EPUB, a publisher may believe that an table is more likely to be presented correctly across multiple rs if it is an image.
- a complex chart with labels designed for a landscape presentation could be difficult to see and read

Would it satisfy this concern if only one orientation was allowed per EPUB? In practical terms, it is already so. I have not yet found an rs that allows for multiple orientations within a single EPUB.

HadrienGardeur:
You can't block a user or a reading system from changing the orientation, which is why I don't think that we should give the illusion that this is possible.

> - Often, artwork is created across a two-page print spread. Orienting that in portrait mode would result in a very small presentation of the artwork. Or splitting the artwork page by page and cropping the artwork in a way that could obscure meaning.
> - It is not unusual for text to be part of the illustration. Text shown in a different orientation than intended would be difficult to read.
> - A complex chart with labels designed for a landscape presentation could be difficult to see and read

[...]

Issue #2749, opened 2025-06-30
Deprecating layout overrides


HadrienGardeur:
### Section

https://www.w3.org/TR/epub-33/#layout-overrides

### Describe the problem

Most reading system provide a completely different user experience for reflowable vs fixed-layout EPUB publications, with a wide array of options for visual adjustments on reflowable publications vs almost none for fixed layout.

Content producers are aware of these UX differences and target fixed layout exclusively for a subset of their publications.

The current specification allows EPUB files that can mix both with the use of layout overrides but this isn't used in practice and could lead to a potentially very confusing UX for the user.
Some reading systems are also unable to switch between layouts in the middle of the publication since they use distinct rendering techniques in their codebase per layout type.

### Describe the fix or new feature you propose

Layout overrides should be deprecated in EPUB 3.4 and marked for removal in a future 3.5 revision.

iherman:
This was discussed during the [pmwg meeting on 11 November 2025](https://w3c.github.io/pm-wg/minutes/2025-11-11-f2f.html#465b).

<details><summary><i>View the transcript</i></summary>

<h4>Layout Overrides - <a href="https://github.com/w3c/epub-specs/issues/2749" rel="noopener noreferrer">w3c/<wbr>epub-specs#2749</a></h4>
<p><cite>wendyreid:</cite> Next is layout override<br>
<span>… this is how we can have mixed reflow and prepaginated</span><br>
<span>… this is a tricky one, because we have a solid set of use cases and idea</span><br>
<span>… There are reading systems that support it</span><br>
<span>… there are a lot of RSes that can't/don't implement it</span><br>
<span>… there are probably better solutions</span></p>
<p><cite>Hadrien:</cite> I think layout override is problematic as it can create complex UX problems, and it doesn't solve all the problems<br>
<span>… I would keep this at the resource level, but not in the spine, and I would limit it to things outside of the spine</span><br>
<span>… so I could have a link to a reflowable from fixed or vice versa</span><br>
<span>… I don't think we need mixed fixed and reflowable when going cover to cover</span><br>
<span>… this is what we need for magazines and textbooks</span><br>
<span>… what we have no is problematic and we should explore other options</span></p>
<p><cite>shiestyle:</cite> We use this feature in e.g. light novels<br>
<span>… There are often some pages that need to be fixed for precise layout</span><br>
<span>… at least kadokawa uses this feature</span></p>
[...]

PR #2788, opened 2025-09-07
Support for `roll` in `layout`

HadrienGardeur:
- Bumped layout and content dimensions out of the FXL section as they're shared across document types
- Introduced the definition of a "scrolled document"
- Added a new value for `rendition:layout`: `roll`
- Added a new example for a "scrolled document"
- Re-used illustration that was initially created for https://github.com/w3c/epub-specs/pull/2602

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2788.html" title="Last updated on Dec 5, 2025, 3:25 PM UTC (0858df8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2788/ba46f10...0858df8.html" title="Last updated on Dec 5, 2025, 3:25 PM UTC (0858df8)">Diff</a>


HadrienGardeur:
I've opened a draft PR for now since this is my first attempt in a long time to work with ReSpec and I'm still missing at least the EPUB RS side of things.

I'm tempted to follow what I've done in this first commit and bring the `rendition:layout` property out of the FXL section: https://w3c.github.io/epub-specs/epub34/rs/#layout

sueneu (commented):
Looks good! I am delighted to see item level overrides. They will be useful for front and backmatter. I do have questions about how the item-level overrides will work and what the fall backs could be. 

Could this PR be approved and we refined the document as we move forward?

epub34/authoring/images/example_rendering_scrolled.svg
bduga, line 1: While I love this cartoon, I am pretty sure we will need permission from Randall Munroe (the cartoonist) before we include it. I am not sure if this qualifies as fair use, or if that cartoon is now available for general use, but we might want to avoid issues and just make a simple/an even simpler drawing to show this off.

mattgarrish:
I'm worried about the structure with these edits. If you go directly to the fixed layout section, for example, you'll miss the section about setting the content dimensions, and without a parallel "scrolled layouts" section you seem to have to piece together that you need to set dimensions by reading through the section to check where scrolled documents has been added to the text.

I think pulling the layout property out of FXL is fine, but the section name might be useful to change. When it's not under the property settings heading it loses some context (e.g., maybe change it to "Layout control" or "Layout preferences").

But is a scrolled document even a new thing that needs defining if a scrolled layout is just a different means of reading a set of fixed layout documents? It's not really a new layout of the content but a different type of flow setting for FXLs.

If the only difference is that you can't set some fxl properties when setting scrolled, is there a way of defining scrolled within the existing fxl section, or of having a scrolled layouts section that links across to how to set the dimensions for FXLs while excluding the properties? I expect presenting "scrolled document" as something new when they're just FXLs in a different context is going to be confusing.

HadrienGardeur:
@mattgarrish 

> If you go directly to the fixed layout section, for example, you'll miss the section about setting the content dimensions, and without a parallel "scrolled layouts" section you seem to have to piece together that you need to set dimensions by reading through the section to check where scrolled documents has been added to the text.

My goal was to avoid repeating this information by moving it to the top. We could add a section that references this in FXL and add a section for "scrolled publications" as well, but it would end up mostly empty.

> I think pulling the layout property out of FXL is fine, but the section name might be useful to change. When it's not under the property settings heading it loses some context (e.g., maybe change it to "Layout control" or "Layout preferences").

I'm fine with any suggesting that you might have for renaming these sections.

> But is a scrolled document even a new thing that needs defining if a scrolled layout is just a different means of reading a set of fixed layout documents? It's not really a new layout of the content but a different type of flow setting for FXLs.

It's a new thing with different constraints from Fixed Layout:

- the resource is fitted to the width of the viewport instead of having both dimensions contained in the viewport
- the publication only makes sense if scrolled, which means that pagination isn't an option (the content is divided into arbitrary tiles, not anything meaningful that can be displayed on its own)
- we very much want to advertise to any software processing these files that they're something different from reflowables/fixed layout

If you look at it from another angle:

- reflowable publications are whatever you want them to be
- fixed layout publications are pre-paginated, each resource is a page, these pages are usually fitted to the viewport of your device and can often be presented in a spread (some RS can scroll FXL publications, but it's less common)
- scrolled publications must be scrolled, each resource is a tile in that scroll, where the tile is fitted to the width of your device with no visible gap between tiles

HadrienGardeur:
@sueneu 

> Looks good! I am delighted to see item level overrides. They will be useful for front and backmatter. I do have questions about how the item-level overrides will work and what the fall backs could be.

I didn't change anything for item level overrides and I'm not a big fan of them, they're not well-supported in RS and can provide a very inconsistent experience. They're also less useful in the spine than if we could provide the same info in the manifest.

https://github.com/w3c/epub-specs/issues/2749

mattgarrish:
> * fixed layout publications are pre-paginated, each resource is a page, these pages are usually fitted to the viewport of your device and can often be presented in a spread (some RS can scroll FXL publications, but it's less common)
> * scrolled publications must be scrolled, each resource is a tile in that scroll, where the tile is fitted to the width of your device with no visible gap between tiles

Sure, the presentation of the fixed dimension documents is different, I'm not arguing that, but you're not defining anything new about the layout of a content document. We have documents with reflowable text and we have documents with fixed dimensions. Those fixed dimension documents in turn are either used within fixed layout displays or scrolled displays. That's the similarity I want to try and capture better if we can.

I see the problem in having fixed layout document be the container for both definitions, but what I was trying to get at is whether there is a way to define a common root for both type of documents. For example, something like:

1. Layout control
2. Fixed-dimension layouts
    1. Introduction
    2. Content document dimensions
    3. Fixed Layouts
    4. Scrolled Layouts
3. Reflowable layouts

Then we don't have the content document dimensions floating at the same level as both fixed and reflowable layouts and we don't have to reference out to apply this info to both.

And the scrolled section might not need a lot of explanation, but I still think it's better to have something than rely solely on a property value to introduce these. Assume readers know nothing and are going to want to understand when and how to apply scrolled layouts.

HadrienGardeur:
> That's the similarity I want to try and capture better if we can.

I see, so it's mostly about the way we organize the spec then.

I'm fine with your suggestion, we could group them under a shared section where they would share the content document dimensions section.

For scrolled layouts, there's something else that I haven't tackled yet: we want to include a paragraph dedicated to backward compatibility with what some of the publishers in Japan are currently doing.

epub34/authoring/index.html
```diff
@@ -7180,28 +7163,175 @@ <h5>Layout</h5>
...
+									href="#viewport.ebnf.value">value</a> a positive <a
+									data-cite="css2/syndata.html#numbers">number</a> [[css2]] or the keyword
+									<code>device-height</code>; and
+							</li>
+							<li>the <code>width</code> property MUST have as its value a positive <a
+								data-cite="css2/syndata.html#numbers">number</a> [[css2]] or the keyword
+								<code>device-width</code>.
+							</li>
+						</ul>
+						<p>The <code>device-width</code> and <code>device-height</code> values refer to the 100%
```
bduga: s/to the 100%/to 100%/

Pre-existing, but might as well fix now.

epub34/authoring/index.html
```diff
@@ -7050,98 +7059,72 @@ <h3>Introduction</h3>
...
-
-					<section id="layout" data-epubcheck="true"
-						data-tests="https://w3c.github.io/epub-structural-tests/#08-layout_layout.feature_L22,https://w3c.github.io/epub-structural-tests/#08-layout_layout.feature_L27,https://w3c.github.io/epub-structural-tests/#08-layout_layout.feature_L36,https://w3c.github.io/epub-structural-tests/#08-layout_layout.feature_L43,https://w3c.github.io/epub-structural-tests/#08-layout_layout.feature_L50">
-						<h5>Layout</h5>
-
-						<p>The <code>rendition:layout</code> property specifies whether the content is reflowable or
-							pre-paginated.</p>
+				<p id="property-layout-global">When the <a href="#layout"><code>rendition:layout</code>
+					property</a> is specified on a <code>meta</code> element, it indicates that the
+					paginated or reflowable layout style applies globally (i.e., for all spine items).</p>
```
bduga: Should this also list scrolled?

bduga:
Overall looks good, but +1 to the structure Matt proposed. Ideally the top level would be "Fixed layouts", with "Prepaginated" and "Scrolled" subsections, but fixed layout as a term likely has too much baggage to be used that way.

epub34/authoring/index.html
HadrienGardeur: Yeah, I think so.

epub34/authoring/images/example_rendering_scrolled.svg
HadrienGardeur: We're already using this cartoon extensively in the spec, for example: https://www.w3.org/TR/epub-33/#spread-none-example

I also re-used it because:

- Wendy and Ivan had already created this illustration for a previous PR
- and I was working on this PR in a train, not the ideal place for drawing new illustrations

Given the Creative Commons license used here, I think we should be fine since we handled attribution for this work.

bduga: Good points! I somehow glossed over the attribution text.

mattgarrish:
> I see, so it's mostly about the way we organize the spec then.

Right, sorry for confusingly using fixed layouts in my first post. It's the fixed dimension documents that are the commonality between the two and that I think we can leverage to make a section fitting for both.

HadrienGardeur:
I've re-organized things based on @mattgarrish feedback.

I would personally drop the reflowable layout section entirely (see https://github.com/w3c/epub-specs/issues/2754) and only keep `scrolled-continuous` for `rendition:flow` under a section dedicated to legacy support for scrolled comics in Japan, somehow similar to what we currently have in EPUB 3.3 for viewport dimensions: https://www.w3.org/TR/epub-33/#viewport

iherman:
I must admit I am confused by the terminology. The confusion comes with the term "scrolled". At first read, the way I would understand `rendition:layout = scrolled` is that the full document is scrolled with an automatic adaptation to the screen size. Essentially, the way a browser would display the whole EPUB content as if the files were, conceptually, concatenated. But this is ***not*** what is proposed here, because `rendition:layout = scrolled` is defined (in the current structure) as a property applicable to fixed dimension documents only, ie, where the dimensions are set and "frozen". But this is not reflected by the property term.

Then comes the confusion: how does this relate to `rendition:flow = scrolled-doc`? After all, that feature is specified for non-fixed dimension publication; otherwise the behavior is similar. 

Actually, it is not clear to me why having `scrolled` appears as a third value of `rendition:layout` in the first place. The way I understand the text is that we have two major categories: fixed dimension documents and the ones without this (which we call `reflowable` although that term is misleading in this respect). Both categories have a "scrolled" version. Wouldn't it be more logical to use `rendition:layout` as a dividing line between the two major categories (roughly as it is now) and use the `rendition:flow` property values of `scrolled-doc` for both, defines as what it means now for reflowables, and the new meaning for fixed dimension ones?

(I understand the current terms are not ideal for this approach, but that is history. We will have to live with that.)

I claim no expertise in this area, so I may miss something essential. But then it must be clear in the spec what I miss, because I find the current proposal confusing...

HadrienGardeur:
@iherman if we're talking about continuously scrolled comics specifically, the use of HTML as a container and the requirement to set content dimensions in HTML are mostly things that get in the way but that we need to use because EPUB is designed that way.

Ideally, we would:

- skip the HTML container for individual bitmaps
- references images directly in the spine
- and maybe provide the dimension of each images in the manifest

Support for comics is very painful in EPUB because our requirements often get in the way of what authors, publishers and reading systems would really like to do.

> Then comes the confusion: how does this relate to `rendition:flow = scrolled-doc`?

We've already had this discussion a few times now:

- `rendition:flow` only applies to reflowable documents
- it's a hint and not a widely supported one
- IMO it's also a very misguided thing to have in our spec in the first place because reflowable publications are all about freedom for the user to do whatever they want (I know some people feel differently, but this is has been the reality of the market for the last 18 years)
- for continuously scrolled comics, we need something much stronger than a hint (RS shouldn't paginate this content at all, but it's fine moving one viewport at a time into the content) and we also need a property that indicates that the viewport should be fixed to the width of the resource (unlike `pre-paginated` where both dimensions are contained in the viewport by default)
- with current RS, using `rendition:flow = scrolled-doc` together with `rendition:layout = pre-paginated` (as it's currently done in Japan) we mostly end up with each tile displayed in a spread, where the tile is very difficult to read since it's zoomed out. Only Apple Books is capable of supporting these properties together in a way that works as expected.
- in our previous discussions, we agreed that it's better to have a property that would very clearly identify the distinct behavior that we're looking for (hence a new value for `rendition:layout`) than fail silently

We can have this whole discussion all over again (we've been going at it for a few years by now), but we had somehow reached consensus that it's better to have continuously scrolled publications as a separate thing from reflowable/pre-paginated and include a note about backward compatibility with what has been used in the past in Japan.

HadrienGardeur:
As a point of reference, here's what I think would be the ideal solution: https://readium.org/webpub-manifest/profiles/divina#:~:text=Example%201%3A%20A%20scrolled%20publication%20that%20contains%20two%20episodes

We can't reach that point with EPUB because of our restrictions with bitmaps in spine.

FYI, this is what South Korea has adopted as a national (TTA) standard (I can share their spec if needed).

mattgarrish:
Going somewhat on an aside, but it would help to be clearer that properties apply only to the categories they're listed under. Having two fixed dimension layouts might get people thinking they could set the orientation, for example. (We don't have to fix that all in this PR, though.)

As it is, you can set any rendering property because you can mix layouts in a publication, but only the properties specific to a layout apply to documents in that rendering. That's why the existing flow properties can get confusing.

And can we ban spine overrides for scrolled layouts in the interest of simplicity, or is it necessary to mix layouts?

HadrienGardeur:
> And can we ban spine overrides for scrolled layouts in the interest of simplicity, or is it necessary to mix layouts?

You'll notice that I skipped them in this PR and that's on purpose.

I think that spine overrides for layouts are harmful and should be dropped from the spec completely: https://github.com/w3c/epub-specs/issues/2749

The only use case where they _might_ be useful would be in fallbacks (as we discussed two weeks ago I think), but we can't use them in fallbacks because they're in the spine instead of the manifest.

> Having two fixed dimension layouts might get people thinking they could set the orientation, for example.

I don't think that authors should be able to set an orientation at all, it's potentially harmful and goes against EAA requirements: https://github.com/w3c/epub-specs/issues/2751

mattgarrish:
> You'll notice that I skipped them in this PR and that's on purpose.

Yes, I spotted that there wasn't a spine override for scrolled and am happy to live without one, but I believe you can still override a global scrolled setting to put reflowable or pre-paginated documents into a scrolled layout, right? (Or did I miss a restriction?) Seems really messy.

I can see the use for having a few fixed layout documents in a reflowable publication, but I never really understood the need for reflowable in a fixed layout publication. It might be too late to deprecate them, but let's at least avoid proliferating it as a practice.

HadrienGardeur:
> […] but I believe you can still override a global scrolled setting to put reflowable or pre-paginated documents into a scrolled layout, right? (Or did I miss a restriction?) Seems really messy.

Yeah it's a mess and we really want to avoid that. I wasn't sure how to introduce a restriction since my preference would be to just deprecate the whole thing.

mattgarrish:
I think this could be done in the layout overrides section. It might be as simple as changing the first sentence to:

> Except in scrolled layouts, EPUB creators MAY specify the following properties...

iherman:
This was discussed during the [pmwg meeting on 11 September 2025](https://w3c.github.io/pm-wg/minutes/2025-09-11.html#1d5b).

<h3>PR 2788 - <a href="https://github.com/w3c/epub-specs/pull/2788" rel="noopener noreferrer">w3c/<wbr>epub-specs#2788</a></h3>
<p><cite>wendyreid:</cite> Discussing scrolled in rendition:layout<br>
<span>… over to Hadrien</span></p>
<p><cite>Hadrien:</cite> This is a followup to continuous scrolled comics<br>
<span>… we had an agreement that a new value was a good idea, and we would add a note to explain what has happened to date</span><br>
<span>… Working on it, there was a lot in fixed layout that aren't really fixed layout</span><br>
<span>… for instance, the property for layout was a subsection, even though it can specify flowing</span><br>
<span>… so i bumped it up a level, and then after some more feedback from mgarrish I reorganized again</span><br>
<span>… I also added an example from a previous doc</span><br>
<span>… this is really a draft, since I have yet to add the RS side, and I still don't have the note</span><br>
<span>… I am a bit concerned about revisiting again</span><br>
<span>… since we already have agreement</span></p>
<p><cite>duga:</cite> I was reading the comments this morning<br>
<span>… I understand not wanting to re litigate this, it works, I'm not interested in revisiting this</span><br>
<span>… we shouldn't discount the fact that ivan is confused, so I'm certain others are also</span><br>
<span>… I'm concerned we haven't made this clear enough</span><br>
<span>… perhaps it is because we are missing the reading system piece</span><br>
<span>… perhaps "scrolled" is the wrong word, it is a behavior, not a format</span><br>
<span>… and we have used it somewhere else for something different</span><br>
<span>… perhaps we should find a name that expressed the format and not the reading system behavior</span><br>
<span>… like contiguous vertical</span></p>
<p><cite>ivan:</cite> Forget about my comment about alternatives<br>
<span>… I officially revoke that comment</span></p>
<p><cite>ivan:</cite> but the term itself is what I find confusing, since it is already used<br>
<span>… so if we find another term I am happy</span></p>
<p><cite>CharlesL:</cite> Are there horizontal scroll comics<br>
<span>… ?</span></p>
<p><cite>Hadrien:</cite> That is a thing, but uncommon<br>
<span>… We did cover a possible direction field</span></p>
<p><cite>SueNeu:</cite> Do we feel that horizontal scroll can be specified with FL?</p>
<p><cite>Hadrien:</cite> Not really, same problem as vertical</p>
<p><cite>mgarrish:</cite> Are comics the sole use case here?<br>
<span>… Rather than scrolled</span></p>
<p><cite>&lt;LaurentLM&gt;</cite> comics is a US term</p>
<p><cite>duga:</cite> I did consider that, but it seems that is also a layout term<br>
<span>… what if someone wants to use if for something else</span><br>
<span>… I don't like the term vertical that I suggested,</span><br>
<span>… because then we can't pivot to a horizontal format</span></p>
<p><cite>Hadrien:</cite> I do think the only real use case are comics<br>
<span>… having an html container is less than ideal for this use</span><br>
<span>… Industry would prefer a list of images</span></p>
<p><cite>SueNeu:</cite> Amazon calls everything comics, even children's books<br>
<span>… I have a lot of questions what the document would look like</span></p>
<p><cite>Hadrien:</cite> What do you mean?</p>
<p><cite>SueNeu:</cite> Is it one html with multiple images? Can this apply for panel by panel navigation?</p>
<p><cite>Hadrien:</cite> for this content, there is no concept of a page. Think of the files as if they were tiles<br>
<span>… the comics themselves are designed to be read in a continuous scroll</span></p>
<p><cite>Dale:</cite> I wonder if this is the way it is, because people approach epubs coming from the web, and that is the convention there<br>
<span>… Do we want to lock people into a convention or do we want people to have a chance to experiment?</span></p>
<p><cite>LaurentLM:</cite> Today there is no question, the webtoon is tiled vertically, period. If there is one case where the RS must follow the author intent, this is it<br>
<span>… We have seen some horizontal, but they did not have much success</span><br>
<span>… If scrolled is not a layout, how about pageless? It is very literal</span></p>
<p><cite>CharlesL:</cite> According to some LLM, I got 'axis' or 'single axis', or 'ribbon'.<br>
<span>… also continuous</span></p>
<p><cite>wendyreid:</cite> To Dale, the reason these exist is due to mobile phones<br>
<span>… It is the most natural way to read comics on the phone</span><br>
<span>… traditional omics are not as good on the phone (zoom, etc)</span><br>
<span>… the challenge is more and more RSes support scrolled for flowing as well</span><br>
<span>… we do need to be clear that we are discussing fixed width image format and not other content</span></p>
<p><cite>shiestyle:</cite> I just proposed unpaginated or pagefree</p>
<p><cite>shiestyle:</cite> we can define vertical is the default</p>
<p><cite>duga:</cite> I'm enjoying the "pageless" and "page-free" we could add "contiguous"<br>
<span>… I don't like "unpaginated" since it is negative</span><br>
<span>… we could take this offline and find a way for ranked choice voting</span><br>
<span>… we could talk about this name for the rest of the meeting</span></p>
<p><cite>wendyreid:</cite> That makes sense, let's go back to content</p>
<p><cite>SueNeu:</cite> Hadrien mentioned tiles, what are the tiles, are they svg, jpeg, etc?</p>
<p><cite>LaurentLM:</cite> Currently they are jpeg pictures</p>
<p><cite>Hadrien:</cite> It is still common to send a single file, and let the publisher cut it down to size<br>
<span>… but the reality is no one in this industry cares about html</span><br>
<span>… the industry today is just bitmaps</span><br>
<span>… and html isn't really helping, even for a11y</span></p>
<p><cite>ivan:</cite> I think we said we should look at the a11y issues again, with fallbacks for images<br>
<span>… and maybe this is something we should address in the WG, but not today</span></p>
<p><cite>wendyreid:</cite> Aside from the name and a11y, what else is there?</p>
<p><cite>ivan:</cite> I am in favor with going forward, and the new structure works for me<br>
<span>… for the RS, you have a sentence that characterizes what you expect will happen</span><br>
<span>… the one thing missing is the log entry, that we need</span><br>
<span>… Big question: will we have enough implementations?</span></p>
<p><cite>SueNeu:</cite> I was also thinking that scrolling was an option for flowing, but wasn't supported by RSes</p>
<p><cite>wendyreid:</cite> Totally different question, probably for authoring spec:<br>
<span>… we do have to tell authors how to construct a scrolled document, and we need that to be consistent enough that implementers can do stuff with it</span><br>
<span>… traditionally, these are released as chapters, but you may also get a season, is it one html file with a single chapter, or do we have something to tell us where the boundaries are? How do we address that?</span><br>
<span>… One hack might be a single file per chapter</span></p>
<p><cite>Hadrien:</cite> There is a technical reason for multi files, mainly performance<br>
<span>… ideally we would have a bunch of images, and then something else that would define a structure</span><br>
<span>… I think it will have to be one html per tile, and structure can be in the nav document</span><br>
<span>… I do agree that full seasons make sense for some retailers who can't handle a subscription</span></p>
<p><cite>mgarrish:</cite> The whole 1to1 html to image is an old topic, but you can add images directly to the spine, with a fallback to a single html document<br>
<span>… So there are workarounds to the spine level image documents</span></p>
<p><cite>wendyreid:</cite> On the html file per tile question, we used to require this for performance, but do we really still need this? Performance is now much better<br>
<span>… And does having a local file change anything, since you don't have to stream?</span></p>
<p><cite>Hadrien:</cite> For a good RS, they usually have a few files in memory and discard the ones that won't be needed soon<br>
<span>… Usually this is done with images directly</span><br>
<span>… on the web, I have seen canvas, iframe, etc</span><br>
<span>… but you always have this caching to make it work</span></p>
<p><cite>duga:</cite> both of these work, you could put all your content into a single HTML and then put images in the spine<br>
<span>… or you could split up your images to individual HTML files</span><br>
<span>… the industry will decide which one</span><br>
<span>… if the giant HTML causes performance issues people will stop</span><br>
<span>… if publishers don't like making one HTML per panel they will not do that</span></p>
<p><cite>Hadrien:</cite> I think what mgarrish suggested was that we can do images in spine, we can just point at a "this is broken" html<br>
<span>… which is different from what you said</span><br>
<span>… if people are comfortable with that then it is fine</span><br>
<span>… so I very much agree with mgarrish</span></p>
<p><cite>Dale:</cite> as a web designer, I was taught that as I develop a design, I am always in the viewport<br>
<span>… so I am always in the browser, and I adjust what I am doing to the browser</span><br>
<span>… and I am learning not all RSes handle content in the same way</span><br>
<span>… so I am wondering, as we discuss the spec, how much does the spec say this is how it should be and the RS must adjust, vs how much are we explaining to authors what will work</span></p>
<p><cite>duga:</cite> I understand we can have a single file that says "this is broken"<br>
<span>… or we can have a single HTML file that contains all images with single line of CSS to make width 100%</span><br>
<span>… and you don't get that the book is broken</span><br>
<span>… it might not be ideal, but it could be a fall back</span><br>
<span>… it would retain the content for RS that require HTML</span><br>
<span>… then it is up to the reading system to choose to use the HTML or the image isn the spine</span><br>
<span>… this is consistent with the current spec</span></p>
<p><cite>shiestyle:</cite> In Japan we had many comic epubs, we had one image in one html in the spine<br>
<span>… and we use the same for scrolled</span><br>
<span>… we have many RSes that support that, and they also support vertical scroll</span><br>
<span>… even for traditional comics</span><br>
<span>… So in Japan, the only difference is RS behavior. EPUB behavior is the same</span><br>
<span>… with just the metadata on the EPUB</span><br>
<span>… we don't need the additional metadata for episodes, etc</span></p>
<p><cite>ivan:</cite> The approach from duga is fine, with the minimal html file<br>
<span>… But images in the spite with fallback to single html doesn't work because the image size isn't available from just the image</span></p>
<p><cite>Hadrien:</cite> The PR is designed this way due to the spec. If we want to do images in spine, then we need different requirements<br>
<span>… on a spec level I think it is hard to say what to put in these documents (can't require images, vs broken comment)</span><br>
<span>… The benefit to images in spine is it removes the guesswork</span><br>
<span>… but if we do it here, we should also consider it for comics</span></p>
<p><cite>Hadrien:</cite> the ability to know vs guess is a huge benefit</p>
<p><cite>CharlesL:</cite> I would be uncomfortable if it cannot be made accessible<br>
<span>… we would need an image in the spine with some way to specify alt text</span></p>
<p><cite>mgarrish:</cite> if you do all the images in a single html document, you would need to only reference it once, then add blank documents</p>

<hr /></details>

bduga:
> but I never really understood the need for reflowable in a fixed layout publication

The case where I have personally wanted this is in the foreword of a collection of comics, which can be very painful to read as fixed layout. Of course, I also concede that the market has yet to produce a business case to justify the expense of implementation. In general, to avoid the mess, I think we either need to allow spine overrides for this value, or deprecate spine overrides altogether. I am fine with the latter, but I think it is actually a bit tricky as it will require us to codify fixed layout publications vs flowing text publications, which until now have been a hallucination of the industry. And it might be better to take that up in a new issue. Maybe for now add an editorial comment to revisit the issue before publication?

sueneu:
> Then comes the confusion: how does this relate to rendition:flow = scrolled-doc?

Does anyone use, and does any reading system support `rendition:flow = scrolled-doc`

If this isn’t in use, could we deprecate the scrolling option for reflowable books to eliminate confusion?

sueneu:
> I can see the use for having a few fixed layout documents in a reflowable publication, but I never really understood the need for reflowable in a fixed layout publication. It might be too late to deprecate them, but let's at least avoid proliferating it as a practice.

I do have some cases where a book would be best presented with some pages fixed and some reflowable. 

Case one:
A middle grade book follows the story of two archeologists. Comics pages highlighting their dinosaur discoveries are interspersed with chapters of biographical text.  example: https://www.leeandlow.com/books/the-doomsday-detectives/

Case Two: 
A poetry anthology. Each chapter is about a different poet, and there is text followed by poems. Since the book is specifically about disabilities, it would be ideal to offer a customizable reading experience for the text, but honor the poet’s choice of line breaks and indents for the poems. https://www.leeandlow.com/books/beauty-is-a-verb/

Either
> reflowable in a fixed layout publication

or

> fixed layout documents in a reflowable publication

would work just fine.

Reading System support is important issue here. If it would be easier for reading systems to display fixed pages within a reflowable title, that would be fine.

iherman:
Per the [current spec](https://www.w3.org/TR/epub-34/#deprecated) (but also the previous), 

> [EPUB conformance checkers](https://www.w3.org/TR/epub-34/#dfn-epub-conformance-checker) should alert EPUB creators to the presence of deprecated features when encountered in EPUB publications.

I am not sure what epubcheck really does, but that may mean that EPUB Publications that pass epubcheck today will be flagged with a warning tomorrow. Technically, it would not make the EPUB Publications "invalid", but close. Would that be acceptable for publishers? Not sure...

HadrienGardeur:
@iherman @mattgarrish based on our discussions in the latest call, we're a bit at a crossroads right now:

- if we decide to use bitmaps in the spine, it doesn't make much sense to group "fixed-dimensions layout" together
- if we keep things as they are, we'll need to look into all sorts of edge cases (layout overrides, having more than one image per resource in the spine…)

I think that it will be hard to keep track of these discussions in a single PR, so I suggest opening up two different issues:

- one dedicated to bikeshedding (rotulus FTW)
- and another one related to bitmaps vs HTML in spine specifically for continuously scrolled comics

iherman:
@HadrienGardeur I agree we should separate the two questions along the lines you propose. The image in the spine issue would have further consequences which are beyond to a rotulus layout.

iherman:
Re bike shedding: I myself love the term "rotulus" (see [Wikipedia](https://en.wikipedia.org/wiki/Rotulus)), it expresses quite well what we are talking about. However, we have to acknowledge that it may create problems insofar as it is a fairly mysterious word for non-English speakers (I must admit I did not remember this word either; my spell checker considers it as an unknown word, too).

I looked for synonyms, the ones I could gather were "scroll", "roll" (sometimes with adjectives like "vertical roll", "document roll", "parchment roll"), "codex", "volumen". Out of these, "roll", or possibly "document roll" may be a good compromise.

bduga:
I am not sure if this is the best place to list out replacement ideas, but
I am not really sure what the best place is. I made a Google Doc with some
items that came from the meeting, and I have just added yours. Doc with
comments enabled is here
<https://docs.google.com/document/d/1FgC98D6O7T_lW_eb541jOKoAxP5XsKLoqtB7eQV6Lfo/edit?usp=sharing>,
but I have no idea how long that will survive now that it is a publicly
scrapable link.

HadrienGardeur:
> I agree we should separate the two questions along the lines you propose. The image in the spine issue would have further consequences which are beyond to a rotulus layout.

I've opened both of these issues. I'll wait until we make progress on them before updating this PR any further.
- https://github.com/w3c/epub-specs/issues/2791
- https://github.com/w3c/epub-specs/issues/2792

HadrienGardeur:
I've updated the PR based on our resolution for https://github.com/w3c/epub-specs/issues/2791

HadrienGardeur:
As suggested by @iherman, I'm happy to have this merged after a review from @mattgarrish as long as we keep in mind that our on-going discussion about images in spine (https://github.com/w3c/epub-specs/issues/2792) could heavily impact this PR.

You can look at the examples that I've shared in https://github.com/w3c/epub-specs/issues/2792#issuecomment-3316087401 to compare it with this PR.

mattgarrish:
I haven't had a chance yet to review this again, but I noticed a couple of things while resolving the merge conflicts:

- I wouldn't use the term "rolled document(s)". It sounds like each document is its own rolled up thing (and we could be accused of smoking our own documents 😄). I'd call them "roll document(s)" to emphasize more that they're part of the roll and to avoid people mistaking the layout value for "rolled".
- I noticed pre-paginated and rolled documents were often mentioned together. If it's common to need to talk about these as a pair, it might be helpful to define an umbrella term to reduce the repetition. That's why "epub content documents" exists, so we don't always have to say "XHTML content documents and SVG content documents". A term like "fixed-dimension documents" could cover both. But feel free to ignore if it's less prevalent of a problem than it looked like in the conflicts.

HadrienGardeur:
> I'd call them "roll document(s)" to emphasize more that they're part of the roll and to avoid people mistaking the layout value for "rolled".

Works for me, I'll update the PR accordingly.

> I noticed pre-paginated and rolled documents were often mentioned together. If it's common to need to talk about these as a pair, it might be helpful to define an umbrella term to reduce the repetition.

I wouldn't work on that one yet because it might be temporary. If we accept images in spines, I don't think that we'll need to group them at all. Since we're meeting in a month at TPAC and there seems to be no strong objection so far to images in spine, it's worth being patient.

iherman:
This was discussed during the [pmwg meeting on 09 October 2025](https://w3c.github.io/pm-wg/minutes/2025-10-09-dc.html#7bfa).

<h3>PR for Support for roll in layout</h3>
<p><cite>&lt;shiestyle&gt;</cite> <a href="https://github.com/w3c/epub-specs/pull/2788" rel="noopener noreferrer">w3c/<wbr>epub-specs#2788</a></p>
<p><cite>CharlesL1:</cite> One question, we have a new rendition layout role, comics used to be fixed, right?<br>
<span>… what do we now expect with this change?</span><br>
<span>… what will reading systems do differently?</span></p>
<p><cite>Hadrien:</cite> pre-paginated is the current value, the expectation is you may find one or the other<br>
<span>… in pre-paginated every resource is a page, maybe with layout hints</span><br>
<span>… similar to print comics</span><br>
<span>… with roll, you will fit to the width of the viewport, and you can scroll through with no gaps</span><br>
<span>… looks like a single image to the user</span><br>
<span>… There are some open issues, like direction</span><br>
<span>… and direction may change how fit works (e.g. left-to-right might fit height)</span><br>
<span>… There may be some publications that have multiple rolls (e.g. a collection of web comics)</span><br>
<span>… this is an open issue</span><br>
<span>… the idea is instead of rendering controls, we would provide semantics that identify chapters or episodes, then leave it the RS to decide display</span></p>
<p><cite>shiestyle:</cite> Roll will not have spread</p>
<p><cite>LaurentLM:</cite> In the PR, it says "fixed dimensions", but in our case it is 1 fixed dimension</p>
<p><cite>Hadrien:</cite> But you are confusing viewport and images</p>
<p><cite>LaurentLM:</cite> I am not sure if I am speaking about images</p>
<p><cite>Hadrien:</cite> Which is why I am not pushing the PR at the moment, since a lot changes with images in spine</p>
<p><cite>LaurentLM:</cite> So why is it fixed dimensions?</p>
<p><cite>Hadrien:</cite> It doesn't matter, we will probably discard a lot of it<br>
<span>… we originally grouped FL and roll, but we may not need to if we allow images in spine</span><br>
<span>… I don't think this will get much work or discussion until TPAC where we decide on images in spine</span></p>
<p><cite>shiestyle:</cite> We still have time before TPAC, we should finalize there<br>
<span>… but we can still discuss now</span></p>
<p><cite>Hadrien:</cite> Looks like we only have the week before TPAC given schedules, do you really want to discuss images in spine then?</p>
<p><cite>shiestyle:</cite> we will have another meeting before tpac</p>
<p><cite>Hadrien:</cite> So you want to discuss images in spine during that call? It will take the entire call</p>
<p><cite>shiestyle:</cite> We can arrange it</p>

<hr /></details>

iherman:
This was discussed during the [pmwg meeting on 10 November 2025](https://w3c.github.io/pm-wg/minutes/2025-11-10-f2f.html#1fea).

<h4>Images in spine</h4>
<p><cite>&lt;shiestyle&gt;</cite> <a href="https://github.com/w3c/epub-specs/issues/2792," rel="noopener noreferrer">https://<wbr>github.com/<wbr>w3c/<wbr>epub-specs/<wbr>issues/<wbr>2792,</a> <a href="https://github.com/w3c/epub-specs/discussions/2825," rel="noopener noreferrer">https://<wbr>github.com/<wbr>w3c/<wbr>epub-specs/<wbr>discussions/<wbr>2825,</a> <a href="https://github.com/w3c/epub-specs/pull/2788" rel="noopener noreferrer">w3c/<wbr>epub-specs#2788</a></p>
<p><cite>shiestyle:</cite> The first topic is images in spine</p>
<p><cite>Hadrien:</cite> I think this has been contentious, we've discussed for a long time, it came up when we first discussed FXL<br>
<span>… we know that images in spine are used, we also know that people have issues with them, mainly for accessibility, I kind of feel that we're stuck in the same discussion loop</span><br>
<span>… we have the same arguments for or against</span><br>
<span>… we never manage to move forward</span><br>
<span>… we're stuck as a result with the status quo, my questions I hope can move us towards a solution for this</span><br>
<span>… we have some options within the spec, other options might require changes to the spec</span><br>
<span>… we should avoid the status quo, I don't think it's good to have a difference between the spec and the industry</span></p>
<p><cite>duga:</cite> So you often point out that there aren't really accessible comics, or they are hard to do<br>
<span>… I agree for comics, it doesn't make sense for XHTML, but for FXL we do have accessible content</span><br>
<span>… in that perspective, it is an accessibility issue for FXL, this subset of content is huge, comics is a big industry</span><br>
<span>… for children's books, it's good that that content is accessible today</span><br>
<span>… people are making use of these features</span><br>
<span>… it's not fair to disregard the accessibility aspect</span><br>
<span>… but if we do something for comics, that is weird, but we have rolled content, not sure if we should forget about HTML in this respect either</span><br>
<span>… the other question, you listed two options, image in spine or image in spine with fallback HTML</span><br>
<span>… would HTML in spine be illegal for roll comics?</span></p>
<p><cite>Hadrien:</cite> Not saying it should be illegal, it would be disappointing to those producing that content<br>
<span>… we shouldn't force people to put images in spine, but shouldn't force HTML either</span><br>
<span>… would negatively impact adoption</span></p>
<p><cite>LaurentLM:</cite> In every specification, certain things are triggered by certain metadata, roll comics are triggered by specific metadata<br>
<span>… we could say accessibility of scrolled comics is different, if we say we want to do the same for paginated comics, or do something different</span><br>
<span>… we could extend the time of solving the accessibility problem</span><br>
<span>… we don't have that content on the market, we don't know what an accessible comic needs to look like</span><br>
<span>… more research needs to be done</span></p>
<p><cite>Hadrien:</cite> There is a difference between partially accessible comics vs fully accessible comics, we don't know what makes a fully accessible comic</p>
<p><cite>LaurentLM:</cite> Just adding HTML to comics doesn't make it accessible</p>
<p><cite>rdeltour:</cite> I have a simple question, does it have to be EPUB<br>
<span>… images in spine, it creates hurdles in interoperability, could it be a separate standard, a profile, something that is different?</span><br>
<span>… another question about CBZ, basically a list of images in a zip, could be we somehow have a hybrid CBZepub?</span><br>
<span>… could an epub be a CBZ or vice versa</span></p>
<p><cite>Hadrien:</cite> First question is good, we see some places saying EPUB is not the solution. But one option to discuss, do we do something more like web publications, SK is doing JSON with metadata and some controls<br>
<span>… it goes to Ivan's question, has the boat sailed</span><br>
<span>… but for some markets, EPUB is huge, like Japan, we can't pretend people don't use it</span><br>
<span>… for CBZ, the order of images matter, the .epub may not be recognized by CBZ readers</span><br>
<span>… you could name images in a way that works for CBZ</span><br>
<span>… you could make something that works for both</span><br>
<span>… or a packaged web publication that could also be an EPUB at the same time</span><br>
<span>… maybe a little too creative there</span></p>
<p><cite>wendyreid:</cite> I think I share Romain's question, does this have to be EPUB?</p>
<p><cite>wendyreid:</cite> One of the underlying parts of this discussion is, philosophically, what we think EPUB is?</p>
<p><cite>wendyreid:</cite> Do we think that, as the group that maintains EPUB, who do we serve and who do we follow?</p>
<p><cite>wendyreid:</cite> We try to strike balance between what industry wants and what readers need</p>
<p><cite>wendyreid:</cite> And lately also government regulations</p>
<p><cite>wendyreid:</cite> On the surface, about images in spine, questions what do we want EPUB to be</p>
<p><cite>wendyreid:</cite> Many of us fought for EPUB to be the accessible standard, accessible by default.</p>
<p><cite>wendyreid:</cite> Images in spine is where we need to make decision, in order of priority, where do we put things?</p>
<p><cite>wendyreid:</cite> Does accessibility trump everything?</p>
<p><cite>wendyreid:</cite> Disappointing that some markets didn't choose EPUB as the standard, for variety of reasons</p>
<p><cite>wendyreid:</cite> But companies releasing apps with PDF or CBZ are not thinking about accessibility</p>
<p><cite>wendyreid:</cite> in their application or business model</p>
<p><cite>wendyreid:</cite> That's a choice that they made in their business</p>
<p><cite>wendyreid:</cite> But do we want to make that choice for EPUB</p>
<p><cite>wendyreid:</cite> We know there are many inaccessible EPUBs out there, because businesses made that decision. But for EPUB, you have to make a *choice* to make inaccessible. We give all the tools, they choose not use.</p>
<p><cite>wendyreid:</cite> But here we would give them a tool to make inaccessible content.</p>
<p><cite>wendyreid:</cite> Make that possible.</p>
<p><cite>fantasai:</cite> What are the implementation challenges with having images in spine vs in HTML</p>
<p><cite>Hadrien:</cite> Apps that implement this, they don't use webviews, they use platform APIs, with HTML they need to find ways to render those images, if its SVG they need to rasterize the images<br>
<span>… not saying they should do that, but that is what they do</span></p>
<p><cite>fantasai:</cite> Is there or has their been an effort to create a profile for HTML that prioritizes images?</p>
<p><cite>Hadrien:</cite> Japan has it, it's not used by everyone even in Japan, it's something you know if you're in industry, but it has no official status<br>
<span>… the closest would be the EBPAJ profile</span></p>
<p><cite>fantasai:</cite> Would it make sense to standardize on that?</p>
<p><cite>Hadrien:</cite> Possibly, it falls on the line of having an alternate profile or best practices document</p>
<p><cite>fantasai:</cite> The HTML file would need to conform in a spec, it would validate, for reading systems would know to parse as an image, but we have the additional structure, the framework for creating accessible content is still there</p>
<p><cite>Hadrien:</cite> The reality would be it might be fragmented, but would move people in the right direction possibly</p>
<p><cite>ivan:</cite> To put things in perspective, the approach Romain suggested is how we published the audiobook standard<br>
<span>… it follows the rules of Publication Manifest, it's an analogy</span><br>
<span>… a very different point, at some point in your slides, image in the spine fallback to HTML as a solution, there were two other approaches discussed, one was suggested by Matt</span><br>
<span>… referring to the fact you could put the alt into the package, it's legal in the spec,</span><br>
<span>… other possibility was to include the accessibility info in the image itself through metadata</span></p>
<p><cite>Hadrien:</cite> The first is to use meta:refines, refine the image using it's id to provide description<br>
<span>… one way to do it</span><br>
<span>… those have been discussed before</span></p>
<p><cite>gpellegrino:</cite> We can have images in spine with fallbacks, I suggest we leave the spec as is, we advise on how to make comics</p>
<p><cite>rdeltour:</cite> For the sake of completeness, in one of the slides you mention was the objective was to link a page to an image, maybe it's also possible to use the nav doc, may also be used to link the pages to the document<br>
<span>… not sure if that works, maybe worth looking into</span></p>
<p><cite>seth:</cite> I have missed some of the previous meetings, we've heard a lot from reading systems people, I never hear from publishers<br>
<span>… I want to know what the issues are with tooling to add the necessary accessibility info to documents</span><br>
<span>… it can't just be there's no interest</span><br>
<span>… there must be challenges</span><br>
<span>… is there a gap in the EPUB production process that blocks this being possible</span></p>
<p><cite>Hadrien:</cite> A page in a comic can have a lot of information, you'd need to describe each panel, and even within panels<br>
<span>… comics are visual by nature, for screen readers, you'd need to produce text in the right order</span><br>
<span>… almost a transformation of the content, like a novelization</span><br>
<span>… there's a huge complexity in how to produce the thing</span><br>
<span>… do you need to get the author involved, are you following the author's intent?</span><br>
<span>… lots of open questions and cost associated</span><br>
<span>… if the resulting text is just a huge block of text, it's not functional either, it would need to be a full HTML document with everything you'd need</span></p>
<p><cite>Hadrien:</cite> Some of it is cost, some of it is technical, there's needs beyond just screenreader users<br>
<span>… one panel at a time plus text to speech, not only do you need equivalent text for a given fragment, then we have tactile images</span><br>
<span>… we can do some of that today, we can't do what say EAA expects</span></p>
<p><cite>ivan:</cite> I want to come back to what Gregorio said, and the spec of today allows what you need as allowing an HTML file as fallback, I may have misunderstood the second part, this may only be valid for roll comics<br>
<span>… I would be uneasy to add something to the spec for just one type of content</span></p>
<p><cite>&lt;fantasai&gt;</cite> +1 ivan</p>
<p><cite>ivan:</cite> it relates to what Hadrien said, falling back to an HTML file, seems to be only a rudimentary way to handle accessibility for an image<br>
<span>… not sure we can handle it today</span></p>
<p><cite>gpellegrino:</cite> Ivan, my proposal was not to allow this way only for roll content, since it's already allowed for any content in the spine<br>
<span>… my proposal was to have guidelines for producing comics</span></p>
<p><cite>fantasai:</cite> I wanted to scope the question down to sighted users who need accessibility support?<br>
<span>… to make that work, you need to have a format to support that additional content, you need a reader to support that, and you need tools to produce that content</span><br>
<span>… does any of that exist?</span></p>
<p><cite>Laurent:</cite> No</p>
<p><cite>fantasai:</cite> If we want EPUB to be able to do this, is images in spine getting us closer or further to that?</p>
<p><cite>Hadrien:</cite> We're years away from being able to do what you describe, its too early to say, I don't think it would be a step in the wrong direction to add images in spine<br>
<span>… if we really want to achieve comics accessibility we can say its someone else's problem, or we can work towards it</span></p>
<p><cite>wendyreid:</cite> Discussed in fixed layout accessibility task force about that. How do we make this possible?</p>
<p><cite>wendyreid:</cite> We have some proposals, to solve what Elika described</p>
<p><cite>wendyreid:</cite> Discussing side-by-side content</p>
<p><cite>wendyreid:</cite> parallel content that's related to each other</p>
<p><cite>wendyreid:</cite> I don't think we're as far away from this as we think we are</p>
<p><cite>wendyreid:</cite> There are reading systems already that do side-by-side content (forgot name)</p>
<p><cite>wendyreid:</cite> One that allows fixed layout and relayout together</p>
<p><cite>wendyreid:</cite> A lot of it, unfortunately, it's about tooling.</p>
<p><cite>wendyreid:</cite> Lack of will to make this happen</p>
<p><cite>wendyreid:</cite> Especially nowadays, with AI, offers us a lot of ways to automate this that previously would have been fully manual</p>
<p><cite>wendyreid:</cite> Only a few years ago, alt text for comics would have been fully manual, now can automate much of it</p>
<p><cite>wendyreid:</cite> Image detection of panels</p>
<p><cite>wendyreid:</cite> Etc.</p>
<p><cite>wendyreid:</cite> I'm reluctant to say that it's not possible. It's increasingly possible.</p>
<p><cite>wendyreid:</cite> Alternate format providers are already doing this work, just mainstream publishers are not</p>
<p><cite>wendyreid:</cite> So I'm reluctant to give in on this.</p>
<p><cite>wendyreid:</cite> We have the technology. Some of the tools already exist.</p>
<p><cite>wendyreid:</cite> A lot more publishers can do in content development that they're not doing.</p>
<p><cite>wendyreid:</cite> And more on reading system side that could make it possible</p>
<p><cite>wendyreid:</cite> I've seen scrolled fixed layout content in EPUB readers. It's not impossible.</p>
<p><cite>wendyreid:</cite> Doesn't mean we should make provisions in the spec to sacrifice our principles</p>
<p><cite>shiestyle:</cite> We have 10 minutes left, we want to conclude this topic in TPAC, maybe gpellegrino's proposal, using the spec as is<br>
<span>… we will not make consensus to make images in spine without fallbacks</span><br>
<span>… should we vote on this?</span></p>
<p><cite>Hadrien:</cite> I think some people don't like image in spine with HTML fallback but its the current spec, having a best practice document that describes how to do this, that is doable<br>
<span>… it solves what I was mentioning, that helps to drive things toward clearer guidance</span></p>
<p>[Seth asks question about tooling and alternate text in Japanese]</p>
<p><cite>seth:</cite> Can someone from the Japanese publishing community share why they don't produce accessible content, is it tooling or something else?</p>
<p><cite>toshiakikoike:</cite> When preparing alt text for manga, just the speech bubble text is not sufficient, but people make prepare the description of the story, we can't prepare that content without the author<br>
<span>… it's difficult to prepare that content</span></p>
<p>ここには日本の出版社の人がいると思うので、聞きたい。漫画はなぜアクセシビリティをサポートしないのか。どういう理由があるのか？</p>
<p><cite>Ito-san:</cite> Japanese publishers don't have the right to produce that type of content, only authors can</p>
<p><cite>shiestyle:</cite> I'll explain this a bit more tomorrow, manga content is generally out of scope</p>
<p>漫画のアクセシビリティというのは、何を情報として含めるのかが難しい。ただセリフをテキストに変えさえすれば、漫画のコマに含まれている情報を正確に伝えることができたというのかというと、そうではないとも思っていまる。</p>
<p><cite>Hadrien:</cite> Just wanted to reply to Wendy's comment, and it leads into the next discussion<br>
<span>… there's a number of documents that can't be made fully accessible, we have children's books, textbooks, comics, the solutions for all those can be slightly different</span><br>
<span>… there's options for magazines, newspapers, there's content out there where you can view articles</span><br>
<span>… we don't have a way to support those in EPUB</span><br>
<span>… I've talked to many people, but they are missing the format, they don't know how to distribute it</span><br>
<span>… I don't think comics fall into that, comics is a lot about storytelling, comics have more in comics with media overlays or SMIL than textbooks or newspapers</span><br>
<span>… there's AI solutions out there, but they haven't been tested, and they don't have a way to represent that</span><br>
<span>… there's prototypes, but it's still early</span><br>
<span>… we don't have an easy way out of the hole we've dug ourselves into, is it now or in the future? Up to us</span></p>
<p><cite>ivan:</cite> The way I read the standard now, for each image you'd have an HTML fallback, ideally one HTML with the alt text, making the publisher's lives easier, but you can't refer to a fragment in a fallback<br>
<span>… if this is our solution, we force the author to produce one HTML per image</span></p>
<p><cite>Tomikura-san:</cite> What accessibility features do people want?</p>
<p><cite>Hadrien:</cite> We hear from people that there is interest, people want access to what their friends are reading<br>
<span>… also with EAA, in theory, most books on the market will be natively accessible, orgs will re-orient to focus on different types of content, usually education or comics</span><br>
<span>… orgs want to produce the content</span><br>
<span>… maybe publishers won't do it, but alternate format providers will</span><br>
<span>… with Marrakesh treaty, people can do this, we might be in a world where speciality providers can do this work</span><br>
<span>… it doesn't put burden on publishers, but it does create challenges for the providers</span></p>

<hr /></details>

iherman:
This was discussed during the [pmwg meeting on 10 November 2025](https://w3c.github.io/pm-wg/minutes/2025-11-10-f2f.html#0255).

<h3>Image in Spine (cont.)</h3>
<p><cite>&lt;shiestyle&gt;</cite> <a href="https://github.com/w3c/epub-specs/issues/2792," rel="noopener noreferrer">https://<wbr>github.com/<wbr>w3c/<wbr>epub-specs/<wbr>issues/<wbr>2792,</a> <a href="https://github.com/w3c/epub-specs/discussions/2825," rel="noopener noreferrer">https://<wbr>github.com/<wbr>w3c/<wbr>epub-specs/<wbr>discussions/<wbr>2825,</a> <a href="https://github.com/w3c/epub-specs/pull/2788" rel="noopener noreferrer">w3c/<wbr>epub-specs#2788</a></p>
<p><cite>fantasai:</cite> Did we conclude on the images-in-spine issue? Should there be a resolution?</p>
<p><cite>ivan:</cite> Agree that having a resolution is what we need<br>
<span>… but I wonder whether we can make the formal references to fallbacks a bit more flexible than they are today, to make it easier to collect all the hypertext into one file instead of one per image for 200 images</span><br>
<span>… Might require adjusting how fallback is identified in the package document</span><br>
<span>… Maybe an additional entry which would allow referencing a fragment in the fallback file</span><br>
<span>… Because right now we don't have that</span></p>
<p><cite>Hadrien:</cite> So, saying you'd have a single HTML with all the text, and you'd have the fallback point to a fragmentID</p>
<p><cite>ivan:</cite> Yes</p>
<p><cite>Hadrien:</cite> If we do that, that opens a lot of other questions</p>
<p><cite>duga:</cite> That fundamentally changes fallbacks and the spine and how this works.<br>
<span>… Maybe a great change, but much bigger change.</span><br>
<span>… Fallback in the spine is meant to be, this is the content document you can use instead of this other document.</span><br>
<span>… You can just swap the files out</span><br>
<span>… But we don't have fragments of documents in the spine. In fact we have specific restriction on re-using a document in the spine.</span><br>
<span>… Avoids complications with navigation and paging etc.</span><br>
<span>… So this is a giant can of worms.</span><br>
<span>… If we do fallbacks, we should do fallbacks as we do now</span></p>
<p><cite>ivan:</cite> Ok, but we have a problem.<br>
<span>… image in the spine with HTML fallback is technically true, but very impractical</span></p>
<p><cite>Hadrien:</cite> Yes, it is impractical</p>
<p><cite>wendyreid:</cite> I would agree it was impractical if 1) it wasn't already how things happen and 2) if person who is making epub had to hand-write every HTML document<br>
<span>… but this is automated. It's not difficult to produce the files.</span><br>
<span>… The tooling is already there to do it, workflow is there to do it.</span><br>
<span>… If there was a size concern here (and the size here comes from the images, realistically speaking), maybe</span><br>
<span>… It does seem a bit silly to have 100 HTML files that aren't used, but in terms of effort it's not a big deal.</span></p>
<p><cite>Hadrien:</cite> My suggestion comes from pragmatic viewpoint which is what's out there<br>
<span>… I see 2 things</span><br>
<span>… Images in spine with HTML fallback, and HTML with image fallback.</span><br>
<span>… We should recognize these as valid ways to deal with this problem.</span><br>
<span>… not have it be word-of-mouth</span><br>
<span>… So we need to acknowledge what's out there.</span><br>
<span>… Secondly, one of those two should be how we recommend authoring content.</span><br>
<span>… "this is the way you should do it"</span><br>
<span>… Not necessarily in the spec, but that could cover other comic-related issues</span><br>
<span>… We can also recommend other things like metadata, semantics etc</span><br>
<span>… With that approach, we won't break anything, won't ask anything new, would just say "this is what is used, and this is what should be used"</span></p>
<p><cite>rdeltour:</cite> If we knew how well fallbacks are supported in reading systems<br>
<span>… we do have some tests for manifest fallbacks</span><br>
<span>… but those are about one single document having a fallback</span><br>
<span>… but I wonder if someone has tested when all the documents have fallbacks, how does that work, how to transition from one document to another</span><br>
<span>… whether this is sufficiently tested to be a viable solution</span></p>
<p><cite>gpellegrino:</cite> I think the time is not right for thinking about solutions for comics etc. So opening the possibility to just images in spine maybe problematic. So suggest we continue as we have now, postponing the discussion, and we can open a task force thinking about a11y and comics</p>
<p><cite>gpellegrino:</cite> Maybe inside the a11y task force we already have</p>
<p><cite>gpellegrino:</cite> but I wouldn't take it up right now</p>
<p><cite>ivan:</cite> I agree</p>
<p><cite>ivan:</cite> Laurent, is it possible for you to take a typical example which has fallbacks with HTML and has a relatively large number of images, and turn it into a test case for our test suite?</p>
<p><cite>LaurentLM:</cite> Yes</p>
<p><cite>ivan:</cite> It would be a valuable addition to our test suite. Romain is right, we have a basic test, but we don't have a good stress test.</p>
<p><cite>ivan:</cite> With that, I think repeating what Elika said, I think we should have a vote and have a resolution on what we put for advice for this type of documents</p>
<p><cite>LaurentLM:</cite> I totally agree, and disagree with Gregorio</p>
<p><cite>LaurentLM:</cite> We don't validate 3.4 today. So we have time to make a pre-recommendation, and ask reading systems to support what's in the spec</p>
<p><cite>&lt;gpellegrino&gt;</cite> +1 to Laurent</p>
<p><cite>LaurentLM:</cite> We can easily make the evolution to support fallbacks if we know this is the right direction.</p>
<p><cite>LaurentLM:</cite> So let's make a pre-recommendation that is a resolution of this</p>
<p><cite>LaurentLM:</cite> And then when we make 3.4 we will be ready, and reading systems will be ready</p>
<p><strong>ACTION:</strong> LaurentLM to add a images-with-fallbacks stress test to test suite</p>
<p><cite>wendyreid:</cite> Something to consider is images with XHTML fallbacks, those happen to have alt text which is great, how do I signal to reading system that I want the fallback instead of the images?<br>
<span>… Most reading systems don't open the OPF file to see what's there.</span><br>
<span>… Maybe we now need to make fallbacks more obvious to users?</span></p>
<p><cite>LaurentLM:</cite> There's a simple way. If there is a version of a comic with some accessibility description, then the publisher should put it as the first choice. Image as a fallback.</p>
<p><cite>LaurentLM:</cite> Favorite of the publisher should be first.</p>
<p><cite>Hadrien:</cite> That doesn't work. Many have put HTML first and images as fallback because it works in more systems, e.g. Apple Books can't work with images in spine.<br>
<span>… So I don't think HTML in spine signals an intent from the producer</span><br>
<span>… It has more to do with the state of the industry than anything else</span><br>
<span>… There's no easy way to signal this preference.</span><br>
<span>… We have a problem in general with fallback. We use spine to convey information instead of using manifest.</span><br>
<span>… If we go in that direction, we're uncover a number of problems with how we do information in spine</span></p>
<p><cite>duga:</cite> Would go a step farther and say, if you have an accessible HTML document, you've made an accessible comic, don't put an image in the spine<br>
<span>… Might not work in some crummy reader, they should update.</span></p>
<p><cite>&lt;Hadrien&gt;</cite> +1 for what brady just said</p>
<p><cite>duga:</cite> You spent a lot of effort making that.<br>
<span>… So should only do this for comics that are otherwise inaccessible.</span><br>
<span>… If there is useful HTML accessibility info, put it in the spine, don't put images there.</span><br>
<span>… Similarly, for childrens books. Put the HTML in the spine. And that can be covered in the note.</span></p>
<p><cite>wendyreid:</cite> So maybe better resolution is, we're keeping the spec as-is, but we are in agreement that comics deserves better guidance, so we want to write that guidance.</p>
<p><cite>Hadrien:</cite> And want to acknowledge what has been done for last 10 years, in millions of files.</p>
<p><cite>fantasai:</cite> Brady covered what I wanted to say, if we are writing best practices, if you do it clearly enough, but if its clear enough readers that use images can extract them, enabling that functionality<br>
<span>… while making the content more accessible for readers with better capabilities</span></p>
<p>PROPOSED: Write a best practices document for comics content as a separate working group note. No change to the specification for images in spine.</p>
<p><cite>&lt;fantasai&gt;</cite> +1</p>
<p><cite>&lt;ikkwong&gt;</cite> +1</p>
<p><cite>&lt;gpellegrino&gt;</cite> +1</p>
<p><cite>&lt;LaurentLM&gt;</cite> +1</p>
<p><strong>RESOLUTION:</strong> Write a best practices document for comics content as a separate working group note. No change to the specification for images in spine.</p>
<p><cite>ivan:</cite> Should I close the issue? discussion?</p>
<p><cite>Hadrien:</cite> Connected to this resolution is roll documents.<br>
<span>… I was blocked for this PR because unsure about direction</span><br>
<span>… But we need to decide more wrt roll documents</span><br>
<span>… We had some remaining complexity with how to describe them in the specification, that is resolved with this, we can have an example with image with HTML fallback</span><br>
<span>… We can get rid of many complexities.</span></p>
<p><cite>ivan:</cite> I will mark the issue as needing to be closed, and chairs can figure it out</p>
<p>[discussion of what to discuss]</p>

<hr /></details>

HadrienGardeur:
@mattgarrish I know that you wanted to merge that PR to avoid further conflicts.

Based on our discussions in Kobe, I moved back to a draft where FXL and roll documents are not covered under the same section ("Fixed dimension documents") anymore.

I can add an example in that section as a placeholder or we can merge things as-is and work in a new PR, whatever works best for you. Since we'd like to deprecate a number of `rendition:*` properties, it's best to get this out of the way first.

mattgarrish:
Ya, if you want to merge this that's fine with me. I don't have an issue with keeping the layouts separate, but I still find that section (not specifically what you've done) kind of scattershot so I'd like to propose some more changes to it, too (like we have an intro that doesn't mention any of the layouts, and then an intro to fixed layouts but not one for reflowable; the content dimensions might fit under the layout property definition; etc.). I'll make a branch off this and see what I can do to make it read better. Shouldn't take me long and I wouldn't expect it would make major merge conflicts for any properties you want to deprecate.

epub34/authoring/index.html
```diff
@@ -7984,6 +7981,10 @@ <h4>The <code>rendition:align-x-center</code> property</h4>
 					</div>
 				</section>
 			</section>
+
+			<section id="sec-roll-layout">
+				<h3>Roll layouts</h4>
```
sueneu: `<h3>Roll layouts</h4>`

s/b
`<h3>Roll layouts</h3>`
fix typo

HadrienGardeur:
I've extended the example a bit more (to illustrate the typical use case for roll documents) and fixed a typo identified by @sueneu.

There's a conflict with the `main` branch though, would you mind taking a look before we merge @mattgarrish ?

Once merged, I'll create a separate PR to tackle back compatibility with what some Japanese publishers are currently doing.

HadrienGardeur:
@mattgarrish given the massive work that you're doing in your new PR, should I still merge this one in `main`?

mattgarrish:
> should I still merge this one in `main`?

You probably don't need to since I have your roll additions in the other pull request. If you merge this one it'll probably just cause unnecessary merge conflicts. I'd close it off and we can continue to fix up #2844 until we're good with it.

HadrienGardeur:
I'm closing this PR then, we'll just need to bring back my updated example to your new PR.

mattgarrish:
> we'll just need to bring back my updated example to your new PR.

Ya, feel free to push any changes you need to #2844

PR #2789, opened 2025-09-08
Add an RS affordance on resetting CSS to the author's settings (as much as possible)

iherman:
After the [last WG discussion](https://w3c.github.io/pm-wg/minutes/2025-09-04.html#f0a7), and the merge of #2771 (which covered the non-controversial changes on the CSS related sections as discussed in #2762), this PR contains a more controversial text discussed originally: a _MAY_ statement for the RS-s, providing an affordance to re-set the author's CSS as close to his/her originals as possible. (Note that this _MAY_ statement is already covered in Apple Books and Thorium.)

Note that this PR does ***not*** cover another controversial matter that came up in #2762: providing author guidance on what properties to avoid due to the possible interference by the Reading System. If someone wants to discuss/propose such an authoring guidance, this should be done in a separate issue/PR (to keep discussions manageable).

See:

    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/RS-css-author-affordance/epub34/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/RS-css-author-affordance/epub34/rs/index.html)

### DIFF
--- a/epub34/rs/index.html
+++ b/epub34/rs/index.html
@@ -1219,15 +1219,15 @@
 					<h4>Author styling overrides</h4>
 
 					<p id="confreq-css-overrides">A reading system MAY override parts of an [=EPUB publication|EPUB
-						publication's] styling (e.g., for ergonomic or user interface reasons). Morevoer, they MAY allow
+						publication's] styling (e.g., for ergonomic or user interface reasons). Moreover, it MAY allow
 						users to set default themes and style preferences (e.g., for more accessible reading) that
 						further alter the original design of a publication.</p>
 
-					<!-- <p id="confreq-css-user-styles">These changes often make it impractical to render an EPUB
+					<p id="confreq-css-user-styles">These changes often make it impractical to render an EPUB
 						publication exactly as designed by an author, even if that is what the user wants. To assist
 						users in viewing an EPUB publication as closely as possible to the author intent, however,
-						reading systems MAY provide an affordance that reverts as much reading system-applied styling as
-						feasible.</p> -->
+						a reading system MAY provide an affordance that reverts as much reading system-applied styling as
+						feasible.</p>
 				</section>
 			</section>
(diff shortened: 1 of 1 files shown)


bduga (commented):
I am ok with this change, but it still feels odd to have in the spec. Basically, it is ok for a RS to change, but it is also ok for a RS to not change things.  So I guess +0 from me.

iherman:
I did some minimal testing with the following CSS in the content file

```css
body {
	background-color: teal;
	color: white;
	text-align: justify;
}
```
### Apple books behaviour 

Apple's setting are on two dimensions: there is a switch for dark/light mode (with a separate switch to follow system setting) and there are a number of “themes” (“quiet”, “calm”, etc.). There is a special theme called “original”.

- In light mode:
	- The teal background is kept in all themes. Justification is changed per themes; switching to “original” means to justify.
- In dark mode:
	- The teal background is _always_ ignored. Justification is changed per themes; switching to “original” means to justify.

(Note the inconsistency between dark and light mode...)

### Thorium

There is no separate dark/light mode, there are just themes for that effect. There is also a theme called “neutral”. They work as one would expect: switching to, say, “night” ignores the teal background (set to black), the “neutral” switches to the original teal.

Interestingly, for text justification there is a separate switch, which is not bound to a theme; it has a default that, in this case, seems to keep to the content of the file (i.e., with justification)

iherman:
It very much looks like we have no consensus on this; furthermore, too much time has gone by, and this PR has zillions of merge conflicts. Closing without merge.

PR #2795, opened 2025-09-19
Add Annotation usecase about sharing outside of a file or an app

gautierchomel:
Existing use cases occur in controlled environments, such as ePubs sold with annotations or annotations shared within the app. This adds a new use case: the annotation file is made available somewhere, and someone downloads it. This would also probably cover sending annotations through email.

[Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fepub-specs%2Fwg-notes%2Fannotations-ucr%2F&doc2=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fepub-specs%2Fcbc791b6f83a5b3f9596eefe94313e4cc3c0bf4c%2Fwg-notes%2Fannotations-ucr%2Findex.html)

### DIFF
--- a/wg-notes/annotations-ucr/index.html
+++ b/wg-notes/annotations-ucr/index.html
@@ -217,6 +217,20 @@
                     Their name or nickname is also attached to their annotations.
 	            </p>
             </section>
+			<section>
+				<h3>Annotations shared through social media</h3>
+            	<p>
+	                A user reads an ebook and prepares some annotations, which they feel are particularly insightful or useful. 
+                    They decide to share these annotations with their followers or a broader community on a social media platform. 
+                    They export their annotation file, which includes their name or nickname attached to each annotation.
+                    They then post a link to the annotation file on their social media page, perhaps with a short description or a teaser of the content. 
+                    They may offer the file for free or charge a small fee using an integrated payment system. 
+                    Another user, who is reading the same ebook, sees the post and decides to download the annotation file. 
+                    They import the annotations into their own reading system. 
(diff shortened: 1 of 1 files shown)


iherman:
For the export case, there is a major overlap with [§5.3 Annotation used in the publishing workflow](https://w3c.github.io/epub-specs/wg-notes/annotations-ucr/#annotations-used-in-the-publishing-workflow). From the point of view of this document, only the action of export is relevant, all other parts (like the fact of paying or not for the content) is irrelevant; it does not seem to add new requirements.

§5.3 has an ambiguous formulation for what is, actually, and import:

> The EPUB creator associates the annotation set with the EPUB file in an editing tool that supports this specification, goes through the annotations, and corrects the EPUB.

My proposal would be (to avoid overlaps) is to expand §5.3 referring to other possible usage of the main actions, namely exporting/importing the annotations. It is also a good idea to emphasize that the annotations can include metadata, like the creator's name and address (something which is part of this proposal, but is not in the current document).

iherman:
Actually, I also realized that there are also lots of overlaps with §5.4 (Classroom example.)

gautierchomel:
Monetising could imply the need for a protection method. But it may also be necessary in a classroom context.

llemeurfr:
We updated the "Annotations used in the publishing workflow" section based on Ivan's proposal. The possible addition of a username or nickname is already present in the "Annotation used in the classroom" section. 
We can therefore close this PR without adding a redondant use case.

PR #2801, opened 2025-09-25
Remove "epub creator" from the accessibility documents

mattgarrish:
This PR removes all use of "epub creator" from the accessibility 1.2 spec and techniques.

While not the most enjoyable thing to do, it wasn't as migraine inducing as I expected. It's also about half as many instances as there are in the authoring spec, so I'll keep going with that one.

It does lead to passive voice having to be used in some cases, but the specs are already full of that. And I had to use "publisher" in place of "epub creator" in the distribution section since there's no way to write out the responsibility on the person distributing the content.

Otherwise, it was kind of a helpful exercise in reviewing some of the statements we were making, as there were some spots that definitely needed some improvement.

(Note to self: don't try to merge this until the new 1.2 docs are published next week.)

***

EPUB Accessibility 1.2:
* [Preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2216/epub34/a11y/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2216/epub34/a11y/index.html)

EPUB Accessibility Techniques 1.2:
* [Preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2216/epub34/a11y-tech/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/a11y-tech/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2216/epub34/a11y-tech/index.html)

### DIFF
--- a/epub34/a11y-tech/index.html
+++ b/epub34/a11y-tech/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en-US">
+<html lang="en-US" xmlns="http://www.w3.org/1999/xhtml">
 	<head>
 		<meta charset="utf-8" />
 		<title>EPUB Accessibility Techniques 1.2</title>
@@ -129,8 +129,9 @@
 
 				<p>This document is not intended to be read in isolation, in other words, as it does not define
 					conformance requirements for making accessibility claims or cover every method for producing
-					accessible content. An EPUB publication must meet all the requirements of EPUB Accessibility 1.2 to
-					make a claim of accessibility; verifying only the techniques in this document is not sufficient.</p>
+					accessible content. An EPUB publication has to meet all the requirements of EPUB Accessibility 1.2
+					to make a claim of accessibility; verifying only the techniques in this document is not
+					sufficient.</p>
 			</section>
 
 			<section id="sec-terminology">
@@ -149,9 +150,9 @@
 
 			<p>The accessibility techniques described in this document are advisory in nature. They are intended to help
 				create <a data-cite="epub/#dfn-epub-publication">EPUB publications</a> that conform to the requirements
-				in [[epub-a11y-12]], but they are not all applicable in all situations and there may be other ways to
-				meet the requirements of that specification. As a result, this document should not be read as providing
-				prescriptive requirements.</p>
+				in [[epub-a11y-12]], but they are not all applicable in all situations and there can be other ways to
+				meet the requirements of that specification. As a result, this document is not intended to be read as
+				providing prescriptive requirements.</p>
 
 			<p>These techniques also do not address issues in digital publishing for which no universally accessible
 				solutions exist. The W3C's Digital Publishing Interest Group has published a note that outlines many of
@@ -230,8 +231,8 @@
 				<section id="sec-wcag-general-res">
 					<h4>Helpful resources</h4>
 
-					<p>Anyone unfamiliar with [[wcag2]] may find the number of techniques daunting, as they are intended
-						to provide broad coverage of possible solutions.</p>
+					<p>Anyone unfamiliar with [[wcag2]] might find the number of techniques daunting, as they are
+						intended to provide broad coverage of possible solutions.</p>
 
 					<p>Assistance applying these techniques to EPUB content documents is available from the following
 						sources:</p>
@@ -264,7 +265,7 @@
 
 					<p>As EPUB allows two <a data-cite="epub/#dfn-epub-content-document">EPUB content documents</a> to
 						be rendered together in a <a data-cite="epub/#spread">synthetic spread</a> [[epub]], the order
-						of content within a single document cannot always be evaluated in isolation. Content may span
+						of content within a single document cannot always be evaluated in isolation. Content might span
 						visually from one document to the next. For example, a sidebar might span the bottom of two
 						pages.</p>
 
@@ -327,8 +328,8 @@
 						<h5>Note about the table of contents</h5>
 
 						<p>A common question about the EPUB table of contents is what completeness it needs to have with
-							respect to the headings of the publication. Although the obvious answer seems like it should
-							be a simple aggregation of all headings for all sections, practically there are several
+							respect to the headings of the publication. Although the obvious answer is to create a
+							simple aggregation of all the headings for all the sections, practically there are several
 							usability challenges to this approach.</p>
 
 						<p>Factors such as device screen sizes can make the table of contents for publications with a
@@ -369,12 +370,12 @@
 
 					<p>The table of contents provides users more than just links into the content. It is also a means to
 						understand the structure and ordering of an <a data-cite="epub/#dfn-epub-publication">EPUB
-							publication</a>. Consequently, users may have difficulty locating where they are in a
+							publication</a>. Consequently, users might have difficulty locating where they are in a
 						publication, where they want to go, and also how to return to previous locations when the order
 						of entries in the table of contents does not match the linear reading order.</p>
 
 					<p>To ensure that the table of contents matches the linear order of the content, the order of its
-						entries should reflect both:</p>
+						entries needs to reflect both:</p>
 
 					<ul>
 						<li>the order of <a data-cite="epub/#dfn-epub-content-document">EPUB content documents</a> in
@@ -382,15 +383,15 @@
 						<li>the order of each referenced section within its respective EPUB content document.</li>
 					</ul>
 
-					<p>Only if there is a logical case for an alternative arrangement of entries should the ordering
-						differ. Such scenarios typically only occur when the content does not have to be read linearly
-						or when additional information is included at the end of a table of contents. For example, the
-						table of contents for a magazine might be ordered to list all the major articles first, followed
-						by features, etc.</p>
+					<p>Only if there is a logical case for an alternative arrangement of entries is it advised that the
+						ordering differ. Such scenarios typically only occur when the content does not have to be read
+						linearly or when additional information is included at the end of a table of contents. For
+						example, the table of contents for a magazine might be ordered to list all the major articles
+						first, followed by features, etc.</p>
 
 					<p>When the ordering of the table of contents does not match the content, the <a
 							href="https://w3c.github.io/publ-a11y/epub-a11y-meta-guide/1.0/draft/#accessibilitySummary"
-							>accessibility summary</a> should explain why.</p>
+							>accessibility summary</a> needs to explain why.</p>
 
 					<p>Avoid linking directly to supplementary content from of the table of contents. Instead, define
 						navigation lists to group together related links, such as a table of figures or a table of
@@ -423,11 +424,12 @@
 						This success criterion does not apply to typical EPUB publications, however, as EPUB content
 						documents do not repeat content in the same way that web sites do.</p>
 
-					<p>Each new content document may begin with similar content, such as learning objectives or key
+					<p>Each new content document might begin with similar content, such as learning objectives or key
 						terms, but this content is part of the body of the publication and not identical to what came
-						before. Consequently, it is not required to add a link to skip it. (Secondary content should be
-						identified in accordance with <a href="https://www.w3.org/TR/WCAG/#info-and-relationships"
-							>success criterion 1.3.1</a>, however.)</p>
+						before. Consequently, it is not required to add a link to skip it. (Secondary content still
+						needs to be identified in accordance with <a
+							href="https://www.w3.org/TR/WCAG/#info-and-relationships">success criterion 1.3.1</a>,
+						however.)</p>
 
 					<div class="note">
 						<p>If an EPUB publication were to reproduce a set of web pages with their full site trappings,
@@ -452,7 +454,7 @@
 								document</a> to identify key structures.</p>
 					</div>
 
-					<p>Although the <code>role</code> attribute may seem similar in nature to the <a
+					<p>Although the <code>role</code> attribute might seem similar in nature to the <a
 							data-cite="epub/#sec-epub-type-attribute"><code>type</code> attribute</a> [[epub]], their
 						target uses in EPUB content documents do not overlap.</p>
 
@@ -539,7 +541,7 @@
 &lt;/html></pre>
 
 					<div class="note">
-						<p>When more than one instance of a role is included in a document, each must be uniquely
+						<p>When more than one instance of a role is included in a document, each has to be uniquely
 							identified. The <code>aria-labelledby</code> attribute provides the name of each landmark in
 							the preceding example. The attribute is not required if only one instance is present, so it
 							is omitted from the following examples.</p>
@@ -609,9 +611,9 @@
 						navigation in its own way.</p>
 
 					<p>The EPUB specification does not require that EPUB publications include a specific set of
-						landmarks, but it is recommended that they include a link to the start of the body matter as
-						well as to any major reference sections (e.g., table of contents, endnotes, bibliography,
-						glossary, index).</p>
+						landmarks; it only recommends to include a link to the start of the body matter as well as to
+						any major reference sections (e.g., table of contents, endnotes, bibliography, glossary,
+						index).</p>
 
 					<aside class="example" title="Landmarks expressed in the EPUB 3 navigation document">
 						<pre>&lt;nav epub:type="landmarks">
@@ -783,12 +785,12 @@
 						needs to have a numbered heading element that reflects its actual position in the
 						publication.</p>
 
-					<p>EPUB publications should also be chunked so that the first heading in a document always has the
-						highest number. For example, if a document starts with an <code>h3</code> heading, there should
-						not be an <code>h2</code> heading later in the document (e.g., do not include the start of a new
-						section with the trailing subsections of the previous). It is acceptable for there to be
-						subsequent headings at the same level as the first (e.g., multiple subsections in one document
-						could all have <code>h3</code> headings).</p>
+					<p>It is advised to chunk EPUB publications so that the first heading in a document always has the
+						highest number. For example, if a document starts with an <code>h3</code> heading,
+							an <code>h1</code> or <code>h2</code> heading does not appear later in the document (e.g.,
+						do not include the start of a new section with the trailing subsections of the previous). It is
+						acceptable for there to be subsequent headings at the same level as the first (e.g., multiple
+						subsections in one document could all have <code>h3</code> headings).</p>
 
 					<aside class="example" title="Heading order across documents">
 						<p>In this example, there are two consecutive EPUB content documents in a textbook. The first
@@ -844,8 +846,8 @@
 					<h4>Heading topic or purpose</h4>
 
 					<p><a data-cite="wcag2#non-text-content">Success Criterion 2.4.6</a> [[wcag2]] currently states that
-						all headings must describe their topic or purpose. The implication of this wording is that all
-						chapters in a novel, for example, have a topic or purpose and that the topic or purpose is
+						all headings have to describe their topic or purpose. The implication of this wording is that
+						all chapters in a novel, for example, have a topic or purpose and that the topic or purpose is
 						always clearly reflected by the title of the chapter. Not only is this not always the case, but
 						this success criterion also complicates the use of chapter numbers as headings since these do
 						not establish a topic.</p>
@@ -857,8 +859,7 @@
 						wording &#8212; namely, that the requirement for headings is only that they establish a unique
 						context for the content.</p>
 
-					<p>By this interpretation, the headings in publications should always pass provided they are
-						unique.</p>
+					<p>By this interpretation, the headings in publications always pass provided they are unique.</p>
 
 					<p>It is expected that the wording of the success criterion will be updated to better reflect the
 						uniqueness requirement, likely in the future WCAG 3 due the complexities of changing the wording
@@ -876,7 +877,7 @@
 					<p>The first version of these techniques only required alternative text for images regardless of
 						their complexity. This exception is no longer valid.</p>
 
-					<p>Image-based content in <a data-cite="epub/#dfn-epub-publication">EPUB publications</a> must now
+					<p>Image-based content in <a data-cite="epub/#dfn-epub-publication">EPUB publications</a> now has to
 						meet [[wcag2]] requirements for alternative text and extended descriptions to conform with
 						[[epub-a11y-12]].</p>
 
@@ -942,7 +943,7 @@
 						proper language for users.</p>
 
 					<p class="note">The languages specified in the package document have no effect on individual EPUB
-						content documents (i.e., the language of each document must be specified using the language
+						content documents (i.e., the language of each document has to be specified using the language
 						expression mechanisms it provides).</p>
 				</section>
 
@@ -968,9 +969,9 @@
 
 					<p>Although it is not strictly required to set this information to meet <a
 							data-cite="wcag2#language-of-page">Success Criterion 3.1.1</a> [[wcag2]], as it is
-						non-normative, it should be considered a best practice to always set this field with the proper
-						language information. (Note that EPUB3 requires the language always be specified, so omitting
-						will fail validation requirements.)</p>
+						non-normative, it is best practice to always set this field with the proper language
+						information. (Note that EPUB3 requires the language always be specified, so omitting will fail
+						validation requirements.)</p>
 
 					<p>Although <a data-cite="epub/#dfn-epub-reading-system">reading systems</a> do not use this
 						language information to render the text content of the EPUB publication, they do use it to
@@ -979,7 +980,7 @@
 
 					<p class="note">The languages specified in the package document have no effect on individual <a
 							data-cite="epub/#dfn-epub-content-document">EPUB content documents</a> (i.e., the language
-						of each document must be specified using the language expression mechanisms it provides).</p>
+						of each document has to be specified using the language expression mechanisms it provides).</p>
 				</section>
 			</section>
 
@@ -1036,8 +1037,8 @@
 					contains more than one rendition will only have access to the default. Unless this rendition is the
 					accessible one, the EPUB publication might not be readable by them.</p>
 
-				<p>Using multiple renditions to meet accessibility requirements should only be done when there is no
-					other alternative. EPUB publications that contain multiple renditions are conformant to the
+				<p>Using multiple renditions to meet accessibility requirements is only advised when there is no other
+					alternative. EPUB publications that contain multiple renditions are conformant to the
 					[[epub-a11y-12]] specification if at least one rendition meets all the content requirements, but
 					such publications, at a minimum, need to note that a reading system that supports multiple
 					renditions is required in their <a
@@ -1068,9 +1069,9 @@
 					<p>For accessibility purposes, it is most important to specify the <code>role</code> attribute value
 						as this is what <a data-cite="epub-a11y-11#dfn-assistive-technology">assistive technologies</a>
 						will process. The <code>epub:type</code> attribute is not recognized by these devices, but it
-						may be used by <a data-cite="epub/#dfn-epub-reading-system">epub reading systems</a> to enhance
-						the user experience. For compatibility with both types of devices, it is recommended to apply
-						both attributes, although only the <code>role</code> attribute is strictly required for
+						could be used by <a data-cite="epub/#dfn-epub-reading-system">epub reading systems</a> to
+						enhance the user experience. For compatibility with both types of devices, it is advised to
+						apply both attributes, although only the <code>role</code> attribute is strictly required for
 						accessibility conformance.</p>
 
 					<aside class="example" title="Expressing a page break">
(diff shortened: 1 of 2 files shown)


epub34/a11y/index.html
```diff
@@ -788,18 +775,20 @@ <h6>Page breaks</h6>
...
 									<ul>
-										<li>they SHOULD include page break markers for all pages reproduced from the
-											source (i.e., blank pages and content not reproduced in the digital edition
-											do not require markers).</li>
-										<li>they should include page break markers for all pages in the source (whether
-											reproduced or not), but this is not a requirement.</li>
+										<li>the page break markers SHOULD identify all pages reproduced from the source
+											(i.e., blank pages and content not reproduced in the digital edition do not
+											require markers).</li>
+										<li>the page break markers should identify all pages in the source, whether
```
clapierre: SHOULD capitalized here too?

epub34/a11y/index.html
```diff
@@ -1229,20 +1217,20 @@ <h5>Other conformance claims</h5>
...
+							<p>As [[wcag2]] does not define how to specify the conformance level in the URL, an
+								alternative means of relating this information will be necessary (e.g., it could be
+								expressed in an accessibility summary).</p>
 						</div>
 
-						<p>Conversely, EPUB creators might need to indicate that an EPUB publication does not meet any
-							accessibility standards (e.g., in order to claim an exemption). In these cases, they can use
-							a <code>dcterms:conformsTo</code> property with value "<code>none</code>".</p>
+						<p>Conversely, to indicate that an EPUB publication does not meet any accessibility standards
+							(e.g., in order to claim an exemption), the metadata could include a
```
clapierre, line 1221: COULD all caps?

epub34/a11y/index.html
```diff
@@ -1613,13 +1600,12 @@ <h2>Optimized publications</h2>
 				<li>The EPUB publication should identify the standard or guidelines it follows in a
 						<code>conformsTo</code> property in accordance with [[dcterms]] so this information can be made
 					available to users.</li>
-				<li>If the standard does not define conformance values for the <code>conformsTo</code> property, EPUB
-					creators should use a URL [[url]] to where the standard is publicly available so users can look up
-					the specific details of the standard.</li>
+				<li>If the standard does not define conformance values for use in the <code>conformsTo</code> property,
+					the <code>conformsTo</code> property should include a URL [[url]] to where the standard is publicly
```
clapierre: SHOULD caps?

epub34/a11y/index.html
```diff
@@ -1613,13 +1600,12 @@ <h2>Optimized publications</h2>
...
+					available so users can look up the specific details of the optimization.</li>
 				<li>If the identifier is not sufficient for a user to understand conformance (e.g., the guidelines are
-					not publicly available), EPUB creators should provide additional information about they have
-					optimized the content in the <a href="#confreq-schema-accessibilitySummary">accessibility
-						summary</a>.</li>
+					not publicly available), the metadata should provide additional information about the content has
```
clapierre: SHOULD caps? (not sure)

clapierre (approved):

epub34/a11y/index.html
mattgarrish: No, because this sentence ends with "but this is not a requirement". We can't make it normative and then cancel it out like that. We explicitly put that qualifier at the end so no one would mistake it for a normative requirement, but it still feels wrong. Let me see what else we can do with these.

epub34/a11y/index.html
mattgarrish: No, could isn't a normative keyword.

You should probably open an issue if you want to make this a normative requirement. It might be out of place to require how to indicate a publication is non-conforming. That's why it's only ever been a suggestion to use "none".

(I also don't think anyone will actually report that their publication is non-conforming.)

epub34/a11y/index.html
mattgarrish: No, the whole section on optimized publications is informative so we can't have normative keywords here.

I wonder if this should really be an appendix, or even split off into its own note? Unlike the informative distribution, which still affects accessible publications, this classification of optimized publications is kind of an aside.

epub34/a11y/index.html
mattgarrish: Same as my previous comment about this being informative.

epub34/a11y-tech/index.html
```diff
@@ -357,9 +351,9 @@ <h5>Note about the table of contents</h5>
 							accessibility support for EPUB evolves — making complete tables of contents usable by
 							everyone — there are legitimate usability reasons why they are not provided now.</p>
 
-						<p>When EPUB creators choose not to provide links to all the headings, however, they should
-							optimize the linking they do provide for the best overall reading experience. Some
-							considerations on how to achieve this include:</p>
+						<p>When opting not to provide links to all the headings, it is best to optimize the links that
+							are provide to improve the overall reading experience. Some considerations on how to achieve
+							this include:</p>
```
iherman:
```suggestion
						<p>When opting not to provide links to all the headings, it is best to optimize the links that
							are provided to improve the overall reading experience. Some considerations on how to achieve
							this include:</p>
```

epub34/a11y-tech/index.html
```diff
@@ -505,11 +498,10 @@ <h4>Do not repeat roles across chunked content</h4>
...
 
-					<p>When content is chunked in this way, it often requires the <a
-							href="https://www.w3.org/TR/epub/#dfn-epub-creator">EPUB creator</a> to make decisions about
-						how best to restructure the information. A part, for example, will typically not include all the
-						chapters that belong to it. The EPUB creator will instead separate the part heading from each
-						chapter, putting each into a separate document.</p>
+					<p>When content is chunked in this way, it often requires tough decisions about how best to
+						restructure the information. A part, for example, will typically not include all the chapters
+						that belong to it. Instead, the part heading might be separated from each chapter, leaving each
+						chapter in a separate document.</p>
```
iherman:
```suggestion
					<p>When content is chunked in this way, it often requires tough decisions about how best to
						restructure the information. A part, for example, will typically not include all the chapters
						that belong to it. Instead, the part heading might be separated from each chapter, leaving each
						chapter in a separate document.</p>
```

I do not have a problem with this text, but I just remark that the word "tough" adds an emphasis to the sentence that the original does not really have. But I am o.k. keeping this if you think it is appropriate.

epub34/a11y-tech/index.html
```diff
@@ -1128,8 +1113,8 @@ <h4>Provide page break markers</h4>
 						list</a> is the only way a user can jump to the locations.</p>
 
 					<aside class="example" title="Adding a hyperlink destination in EPUB 2">
-						<p>In this example, the EPUB creator adds an XHTML 1.1 <code>span</code> element to use as a
-							hyperlink destination.</p>
+						<p>In this example, an XHTML 1.1 <code>span</code> element is used as a hyperlink
+							destination.</p>
```
iherman, line 1108:
```suggestion
						<p>In this example, an XHTML <code>span</code> element is used as a hyperlink
							destination.</p>
```

With an eye on the discussion about non-XML syntax to HTML, is it wise to refer to XHTML 1.1 in this document? Isn't better to use the XHTML term only? (That may be true elsewhere in the document, not affected by this PR.)

epub34/a11y-tech/index.html
```diff
@@ -1163,10 +1148,9 @@ <h4>Identify pages in audio playback</h4>
 						in the audio playback of a publication it is not only distracting, but can be confusing, as well
 						(e.g., the number could be read out in the middle of a sentence).</p>
 
-					<p>To mitigate this potential annoyance to readers, <a
-							href="https://www.w3.org/TR/epub/#dfn-epub-creator">EPUB creators</a> need to identify page
-						announcements in <a href="https://www.w3.org/TR/epub/#dfn-media-overlay-document">media overlay
-							documents</a> when they are included. Identification allows a <a
+					<p>To mitigate this potential annoyance to readers, identify page announcements in <a
+							href="https://www.w3.org/TR/epub/#dfn-media-overlay-document">media overlay documents</a>
+						when they are included. Identification allows a <a
```
iherman:
```suggestion
					<p>To mitigate this potential annoyance to readers, page announcements in <a
							href="https://www.w3.org/TR/epub/#dfn-media-overlay-document">media overlay documents</a>
						should be identified when they are included. Identification allows a <a
```

Isn't it more consistent to put this sentence in passive mode, too, just as it was done elsewhere? I've made an attempt to it, but you are the final arbiter...

epub34/a11y/index.html
```diff
@@ -231,20 +229,19 @@ <h4>Internationalization</h4>
...
-						href="https://www.w3.org/TR/epub/#attrdef-refines"><code>refines</code> attribute</a>
-					[[epub-3]]. If EPUB creators cannot avoid expressions that require this attribute, they will have to
-					accept a certain amount of ambiguity in their statements (i.e., relationships between expression may
-					only be apparent by their placement in the package document metadata).</p>
+						>EPUB publication</a>, regardless of what version of EPUB it conforms to.</p>
+
+				<p>For example, even though EPUB 2 [[opf-201]] was produced before this specification, and so makes no
+					mention of it, it is possible to create EPUB 2 publications that conform its the accessibility and
+					discoverability requirements. Ideally, though, EPUB 2 publications will be upgraded to the latest
+					version of EPUB 3 to get access to the most advanced accessibility features and techniques.</p>
```
iherman, line 232:
```suggestion
				<p>For example, even though EPUB 2 [[opf-201]] was produced before this specification, and so makes no
					mention of it, it is possible to create EPUB 2 publications that conform to the accessibility and
					discoverability requirements. Ideally, though, EPUB 2 publications will be upgraded to the latest
					version of EPUB 3 to get access to the most advanced accessibility features and techniques.</p>
```

epub34/a11y/index.html
```diff
@@ -520,18 +513,17 @@ <h4>WCAG conformance requirements</h4>
...
 					<div class="note">
 						<p>The <a href="https://www.w3.org/WAI/about/groups/agwg/">W3C Accessibility Guidelines Working
 								Group</a> is currently developing WCAG 3. As this version potentially represents a
 							significant departure from WCAG 2, a future version of this specification will address
-							conformance requirements related to it. EPUB creators are encouraged to adopt WCAG 3 once it
-							is stable and widely recognized, but conformance to the new version is not a requirement of
-							this standard.</p>
+							conformance requirements related to it. Adoption of WCAG 3 is encouraged once it is stable
+							and widely recognized, but conformance to the new version is not a requirement of this
+							standard.</p>
```
iherman, line 521:
```suggestion
							conformance requirements related to it. Adoption of WCAG 3 is encouraged, but conformance to the that version is not a requirement of this
							standard.</p>
```

(This goes beyond this PR). I think we should consider EPUB 3 as stable and widely recognized by now...

epub34/a11y/index.html
```diff
@@ -745,14 +733,13 @@ <h6>Page list</h6>
...
-											markers</a> in the content.</p>
-									<p>EPUB creators should include links for all pages in the source whether they are
-										reproduced or not, but this is not a requirement.</p>
+									<p>The page list SHOULD include links to all pages of content reproduced from the
+										source (i.e., it does not have to include links for blank pages or content not
+										reproduced in the digital edition).</p>
+									<p>The page list MUST include links to all the <a href="#sec-page-breaks-obj">page
+											break markers</a> in the content.</p>
+									<p>The page list should include links for all pages in the source, whether they are
+										reproduced or not, but this is not a conformance requirement.</p>
```
iherman:
```suggestion
									<p>The page list MUST include links to all the <a href="#sec-page-breaks-obj">page
											break markers</a> in the content.</p>
									<p>The page list should include links to all pages in the source, whether they are
										reproduced or not, but this is not a conformance requirement.</p>
```

Just being consistent...

epub34/a11y/index.html
```diff
@@ -1020,20 +1008,21 @@ <h6>Escapability</h6>
...
-										be able to escape from lists, sidebars, figures, and other highly structured
-										content, unless EPUB creators encode the structural semantics of those
-										elements.</p>
-									<p>When EPUB creators provide this information, <a
-											href="https://www.w3.org/TR/epub/#dfn-epub-reading-system">reading
+											href="https://www.w3.org/TR/epub/#dfn-epub-publication">EPUB
+											publications</a> include structural semantics in the synchronized text/audio
+										format. Users may not be able to escape from lists, sidebars, figures, and other
+										highly structured content, without the structural semantics of those elements
+										available.</p>
```
iherman:
```suggestion
											publications</a> include structural semantics in the synchronized text/audio
										format. Users may not be able to escape from lists, sidebars, figures, and other
										highly structured content, without the structural semantics of those elements
										being available.</p>
```

iherman (approved):
Sorry for the headache...

epub34/a11y-tech/index.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en-US">
+<html lang="en-US" xmlns="http://www.w3.org/1999/xhtml">
```
mattgarrish:
```suggestion
<html lang="en-US">
```

epub34/a11y-tech/index.html
mattgarrish: Ya, but if it was an easy decision then there'd be nothing to document... 😄 

We could change the focus away from the decision making with: "it often requires restructuring the information to fit within the new file structure". That's more what the technique is about.

```suggestion
					<p>When content is chunked in this way, it often requires restructuring the information to fit
						within the new file structure. A part, for example, will typically not include all the chapters
						that belong to it. Instead, the part heading might be separated from each chapter, leaving each
						chapter in a separate document.</p>
```

epub34/a11y-tech/index.html
mattgarrish: This is explicitly an example of EPUB 2 markup which is why xhtml 1.1 is called out. It's not how you would add a page break to an epub 3 file, as you have both the epub:type and role attributes at your disposal.

epub34/a11y-tech/index.html
mattgarrish: Ya, that's the balancing act. Passive voice is typically shunned for writing but then one of the problems we've always had is who is the actor. I'm fine either way.

epub34/a11y/index.html
mattgarrish: You've flagged a reference to WCAG 3. Did you mean to apply this to a different change or is it just a misread? (WCAG 3 is far from being ready for use.)

epub34/a11y-tech/index.html
iherman: Ah o.k., that is fine then.

epub34/a11y/index.html
iherman: Oh sht... I completely misread, and I mentally did an WCAG 3 -> EPUB 3 change. 

Sorry, I am tired....

epub34/a11y-tech/index.html
```diff
@@ -388,9 +382,8 @@ <h4>Ensure the order of table of contents entries matches linear order</h4>
 						also how to return to previous locations when the order of entries in the table of contents does
 						not match the linear reading order.</p>
 
-					<p><a href="https://www.w3.org/TR/epub/#dfn-epub-creator">EPUB creators</a> should therefore ensure
-						that the entries in the table of contents always match the linear order of the content.
-						Specifically, the order of entries should reflect both:</p>
+					<p>It is therefore recommended to ensure that the entries in the table of contents always match the
+						linear order of the content. Specifically, the order of entries should reflect both:</p>
```
sueneu: Can we simplify this?

> Therefore, it is best that the entries in the table of contents always match the linear order of the content. For example, the order of entries should reflect both:

epub34/a11y/index.html
mattgarrish: > Sorry, I am tired....

Lol, I feel your pain! Editing while fighting off a cold isn't much fun...

epub34/a11y-tech/index.html
```diff
@@ -405,15 +398,16 @@ <h4>Ensure the order of table of contents entries matches linear order</h4>
...
 						by features, etc.</p>
 
-					<p>When the ordering of the table of contents does not match the content, EPUB creators should
-						include an explanation why in the <a href="#accessibilitySummary">accessibility summary</a>.</p>
+					<p>When the ordering of the table of contents does not match the content, the <a
+							href="#accessibilitySummary">accessibility summary</a> should explain why.</p>
 
-					<p>EPUB creators should avoid including links to supplementary content at the end of the table of
-						contents. Links to figure, tables, illustrations and similar content is better included as a
+					<p>It is also recommended to avoid including links to supplementary content at the end of the table
```
sueneu: > It is also recommended to avoid including links to supplementary content

Simplify to
Avoid including links to supplementary content

sueneu:
Matt,
  I added some comments in the document. Is that the best way to do this?

epub34/a11y-tech/index.html
```diff
@@ -405,15 +398,16 @@ <h4>Ensure the order of table of contents entries matches linear order</h4>
...
+						of contents. Links to figure, tables, illustrations and similar content is better included as a
 						separate navigation elements (either in the <a
 							href="https://www.w3.org/TR/epub/#dfn-epub-navigation-document">EPUB navigation document</a>
-						or in the spine). EPUB creators can include links to these additional navigation lists in the
-						table of contents.</p>
+						or in the spine). The table of content can then provide links to these navigation elements to
+						ensure that users can access them (not all reading systems support the rendering of additional
+						navigation elements).</p>
```
sueneu: > The table of content can then provide links to these navigation elements to ensure that users can access them
 
let’s add an “s” to contents and call these “navigation lists” so the phrasing is the same from one paragraph to the next.

The table of contents can link to these navigation lists to ensure that users can access them

epub34/a11y-tech/index.html
```diff
@@ -477,16 +471,15 @@ <h4>ARIA roles and <code>epub:type</code></h4>
 						footnotes or special presentations of the content).</p>
 
 					<p>Since each attribute offers different advantages, it is not necessary that they be used together.
-						Due to the lack of restrictions on where <a href="https://www.w3.org/TR/epub/#dfn-epub-creator"
-							>EPUB creators</a> can use the <code>type</code> attribute, pairing the attributes may cause
-						accessibility issues (e.g., putting roles on the [[html]] <a data-lt="body"><code>body</code>
-							element</a>).</p>
+						Due to the lack of restrictions on where the <code>type</code> attribute can be used, pairing
+						the attributes may cause accessibility issues (e.g., if it leads to putting roles on the
+						[[html]] <a data-lt="body"><code>body</code> element</a>).</p>
```
sueneu: > Due to the lack of restrictions on where the <code>type</code> attribute can be used, pairing the attributes may cause accessibility issues (e.g., if it leads to putting roles on the

Pairing the attributes may cause accessibility issues. For example, it could lead to putting roles on the

mattgarrish: Actually, the paragraph may need rewriting or even just deletion. I think this text comes from the days before ARIA in HTML and support for validating the role attribute in epubcheck and ace. You can't pair roles and types in invalid contexts like on the body element anymore without failing validation.

epub34/a11y-tech/index.html
mattgarrish:
```suggestion
					<p>Avoid including links to supplementary content at the end of the table
```

gautierchomel (approved):
Great improvement!

mattgarrish:
> Is that the best way to do this?

You can make suggestions by clicking on the document icon with the +- in it at the start of menu bar and then editing the text directly. That makes the text into a commit I can accept right away (or I can always tweak the comment before committing).

It just makes things a little faster than going back to my local copy and editing it there.

epub34/a11y-tech/index.html
mattgarrish: Two cases are getting mixed here. One would be navigation lists in the spine, but the other is custom nav elements in the navigation document. I think we need to distinguish these better as the ending parenthetical makes it sound like a reading system might not be able to render in-spine lists when it's talking specifically about access to the custom elements in the nav document. I'll see what I can do to clarify that better.

epub34/a11y-tech/index.html
mattgarrish: It's not so much an example as the requirement, but I agree this is a mouthful. Maybe:

> To ensure the table of contents matches the linear order of the content, the order of its entries should reflect both:

iherman:
> Is that the best way to do this?

You can select the lines with a '+' in
the review part (those are the 'new' parts), which then pops up a usual comment box. Except that you also have a 

Screenshot 2025-09-26 at 17 36 18
 
in the header part. Pushing that button will copy the lines you selected into the comment; whatever you change on those lines in the comment will become the proposed changes that can be committed easily.

You can also add comments for the change proposal at the end of the box (after the ``` line)

epub34/a11y-tech/index.html
```diff
@@ -167,10 +164,10 @@ <h2>About the techniques</h2>
 				these issues [[dpub-accessibility]]. As solutions become available, they will be incorporated into the
 				appropriate document, whether this one or one it refers to.</p>
 
-			<p>If EPUB creators encounter issues that are not covered in these or related techniques, they are
-				encouraged to report the issue to the appropriate community for guidance on how to meet accessibility
-				standards. The <a href="https://www.w3.org/WAI/IG/">W3C Web Accessibility Interest Group</a> has a
-				public mailing list where issues meeting [[wcag2]] and [[wai-aria]] requirements can be raised. The <a
+			<p>When encountering issues that are not covered in these or related techniques, please report an issue to
```
bduga, line 162: Reading this over, the change seems to have really helped. In hindsight, it was silly to limit reporting issues to EPUB creators - RS implementers shouldn't report issues?! This makes a lot more sense, even if it doesn't really change anything.

epub34/a11y-tech/index.html
bduga: But, it should be table of contents no matter what, right? Is table of content ever correct?

epub34/a11y-tech/index.html
```diff
@@ -1098,8 +1084,7 @@ <h4>Provide page break markers</h4>
 						strictly required for accessibility conformance.</p>
 
 					<aside class="example" title="Expressing a page break">
-						<p>In this example, the EPUB creator identifies an HTML <code>span</code> element as a page
-							break.</p>
+						<p>In this example, the an HTML <code>span</code> element is identified as a page break.</p>
```
bduga:
```suggestion
						<p>In this example, the HTML <code>span</code> element is identified as a page break.</p>
```

epub34/a11y/index.html
```diff
@@ -1460,17 +1447,17 @@ <h4>Re-evaluating conformance</h4>
 					</ul>
 
 					<p>If the EPUB publication includes substantively the same markup and content as the previous
-						release, the EPUB creator may only need to evaluate the new modifications to re-confirm
-						conformance.</p>
+						release, only the new modifications will likely need evaluation to re-confirm conformance, so
+						long as the conformance standard has not changed.</p>
```
bduga: > so long as the conformance standard has not changed

This looks new - do we need it, since standard changes are covered in the text below?

epub34/a11y/index.html
```diff
@@ -1488,7 +1475,7 @@ <h4>Re-evaluating conformance</h4>
 					<p>Even in the case of non-substantive changes, this specification recommends an updated evaluation
 						(full or partial) if the accessibility standards have changed.</p>
 
-					<p>EPUB creators should consider even more progressive approaches than those described here. Waiting
+					<p>Publishers should consider even more progressive approaches than those described here. Waiting
```
bduga: Oh, that's ugly. Publishers/content creators/authors/conversion houses/tooling developers/content farms/AI churn mills. This is the first time I have missed the use of EPUB creator in this change. Is "Even more more progressive approaches than those described here should be considered" too horrific? #IAmNotAWriter

epub34/a11y/index.html
```diff
@@ -1499,15 +1486,15 @@ <h4>Re-evaluating conformance</h4>
 			<section id="sec-feedback">
 				<h3>Feedback</h3>
 
-				<p>Although EPUB creators are expected to assess their publications prior to certifying them accessible,
+				<p>Although publishers are expected to assess their publications prior to certifying them accessible,
```
bduga: Not again! And so soon. Maybe something like:

```suggestion
				<p>While it is expected that publications will be assessed prior to certifying them accessible, 
```

epub34/a11y/index.html
```diff
@@ -1499,15 +1486,15 @@ <h4>Re-evaluating conformance</h4>
...
 					due to the sheer volumes of content produced and the complexity of checking all the markup it is
 					sometimes the case that issues will slip through to the final product. Accessibility reporting can
 					also not be detailed enough for every user to be able to determine the suitability of the content
 					for their needs.</p>
 
 				<p>To assist users in reporting any accessibility issues they encounter, or to discover more about the
-					accessibility of an EPUB publication, EPUB creators MAY include a contact email using the <a
-						href="#contactEmail"><code>a11y:contactEmail</code> property</a>.</p>
+					accessibility of an EPUB publication, EPUB publications MAY include a accessibility contact email
```
bduga:
```suggestion
					accessibility of an EPUB publication, EPUB publications MAY include an accessibility contact email
```

epub34/a11y/index.html
```diff
@@ -1677,31 +1663,31 @@ <h2>Optimized publications</h2>
 			<h2>Distribution</h2>
 
 			<div class="note">
-				<p>Although <a href="https://www.w3.org/TR/epub/#dfn-epub-creator">EPUB creators</a> do not have to
-					follow the recommendations in this section to conform to this specification, some jurisdictions
-					require EPUB creators to follow similar practices. <a
+				<p>Although <a href="https://www.w3.org/TR/epub/#dfn-epub-publication">EPUB publications</a> do not have
+					to meet the recommendations in this section to conform to this specification, some jurisdictions
+					require similar distribution practices. <a
```
bduga: This seems less clear to me. The previous text made it clear that some jurisdictions require creators to do certain things, this seems less clear that there may be additional legal requirements I have to consider. Maybe
```suggestion
					require EPUB publications to follow similar practices. <a
```

I dunno. Maybe that isn't better, but it seems clearer to me.

epub34/a11y/index.html
```diff
@@ -1677,31 +1663,31 @@ <h2>Optimized publications</h2>
...
+			<p>While many aspects of the distribution process are outside the scope of this specification, there are
+				factors that publishers can control. For example, while a publisher typically does not control the
+				accessibility of the digital rights management (DRM) scheme applied to their EPUB publications, they do
+				control what usage rights to apply to their EPUB publications. So even though a DRM scheme may allow a
+				publisher to block access to the text of the publication, the publisher needs to take care not to apply
+				such a restriction as it could block the ability for <a>assistive technologies</a> to read the text
+				aloud.</p>
+
+			<p>To minimize the effects of distribution on accessibility, this specification advises that publishers
+				adhere to the following distribution practices:</p>
```
bduga: Ouch! Ok, let me take a shot.
```suggestion
			<p>While many aspects of the distribution process are outside the scope of this specification, there are
				factors that can be controlled. For example, while a publisher typically does not control the
				accessibility of the digital rights management (DRM) scheme applied to their EPUB publications, they do
				control what usage rights to apply to their EPUB publications. So even though a DRM scheme may allow a
				publisher to block access to the text of the publication, that publisher needs to take care not to apply
				such a restriction as it could block the ability for <a>assistive technologies</a> to read the text
				aloud.</p>

			<p>To minimize the effects of distribution on accessibility, this specification advises adherence
			    to the following distribution practices:</p>
```

That at least confines "publisher" to the example, which seems ok.

epub34/a11y/index.html
```diff
@@ -1711,8 +1697,8 @@ <h2>Distribution</h2>
 
 			<div class="note">
 				<p>A distributor may implement a digital rights management scheme that inherently impairs accessibility
-					through no fault of the EPUB creator. Following the guidance in this section does not restrict EPUB
-					creators from using such distributors. The intent is only that the EPUB creator not impair
+					through no fault of the publisher. Following the guidance in this section does not restrict
```
bduga: This note breaks my brain. I think maybe we need to rethink it - are we basically saying "the stuff that isn't your fault isn't your fault"? Can we say that without the word publisher? I don't even know who "you" is here - apparently it is not distributors, but what happens when a distributor creates or modifies your epub? It really seems like we are saying what happens downstream (by anyone) is not your fault, unless it was, in which case don't do that.

bduga (commented):
I am really happy with this change, I think it makes things clearer. I guess my only real issues are towards the end with the distribution section.

epub34/a11y/index.html
mattgarrish: Ya, I think this can be boiled down to:

> The guidance in this section does not restrict EPUB publications from being distributed through inaccessible distribution channels. For example, where the distributor applies DRM that impairs access to the content.

bduga: Yeah, that actually makes sense to me. 👍

epub34/a11y-tech/index.html
mattgarrish: I rewrote this yesterday to focus on when they would be paired. I also worked in a reference to the aria in html doc.

epub34/a11y-tech/index.html
mattgarrish: > But, it should be table of contents no matter what, right?

Ya, that's a straight up typo. I revised this paragraph to try and make it clearer that we're talking about tables of figures, etc. and that they can be created using custom nav elements in the nav doc or as html in an epub content document. The stuff about reading system support I moved to a note.

epub34/a11y/index.html
mattgarrish: What I did here is remove the list so it doesn't look like two requirements. The first paragraph now cover the normative requirement and the second recommends including page breaks for every print page. I've kept that it's not a requirement for conformance, but using a different lowercase keyword and separating the statements should improve that distinction.

PR #2804, opened 2025-09-30
Remove epub creator from authoring and rs specs

### LINKED ISSUE #2216: "EPUB Creators MUST" and "EPUB Creators SHOULD" in EPUB A11Y 1.1
EPUB A11Y 1.1 has several sentences beginning with "EPUB Creators MUST" or "EPUB Creators SHOULD". I propose to reformulate them as requirements on EPUB publications rather than those on humans.

First, the scope of EPUB A11Y 1.1 is the accessibility and discoverability of EPUB Publications.  Humans are outside the scope.

Second, it is not possible to create tests for these sentences.  Ace does not check EPUB creators either.

Third, these sentences do not impose any requirements on EPUB publications, since programs or other users are not prohibited from changing what EPUB creators did.

3.4.1.3.1 PAGINATION SOURCE

The 1st para, "Meeting this Objective"

> EPUB Creators MUST identify that source in the Package Document metadata.

Reformulate the above sentence.

3.4.1.3.2 PAGE LIST

The 2nd para, in "Meeting this Objective"

> EPUB Creators SHOULD ....

Ditto.

The 3rd para, in "Meeting this Objective"

>EPUB Creators should

Is this just a non-normative sentence?  (Note: "should" rather than "SHOULD" here).

3.4.1.3.3 Page Breaks

The 2nd para, "Meeting this Objective"

>they SHOULD include page break markers for ....

What is the subject ("they")?  It cannot be "an EPUB Creator" since it is singular.  Reformulate this sentence without using "EPUB creator" as the subject.

> they should include

Is this just a non-normative sentence?  (Note: "should" rather than "SHOULD" here).

The last para, "Meeting this Objective"

> EPUB Creators MUST identify the page numbers in the markup that controls the playback.

Reformulate the above sentence.

3.4.2.3.1 COMPLETENESS

1st para, "Meeting this Objective"

>  EPUB Creators MUST provide synchronized audio playback via Media Overlays for all textual content.

Reformulate the above sentence.

3.4.2.3.2 Reading Order

Ths 1st para, "Meeting this Objective"

 > EPUB Creators SHOULD ...

Reformulate the above sentence.

3.4.2.3.3 Skippability

The 1st para, "Meeting this Objective"

>  EPUB Creators SHOULD ...

Reformulate the above sentence.

3.4.2.3.4 Escapability

The 1st para, "Meeting this Objective"

> EPUB Creators SHOULD ...

Reformulate the above sentence.

3.4.2.3.5 Navigation Document

The 1st para, "Meeting this Objective"

> EPUB Creators SHOULD ...

Reformulate the above sentence.

3.3.2.1 Page and Publication

> EPUB Creators MUST evaluate their accessibility as part of the larger work.

Reformulate the above sentence.

> EPUB Creators MUST evaluate the WCAG guidelines for content to be ...

Reformulate the above sentence.

3.3.2.2 Applying the Conformance Criteria

> EPUB Creators MUST NOT use EPUB's fallback mechanisms to provide a conforming alternate version [WCAG2], ...

Here is my proposal.

  Every EPUB and Foreign Content Document in the spine plane (i.e., either in the spine element or in the fallbackchain) MUST meet the conformance requirements of the level claimed.

3.5.2 Publication Conformance

There are four occurrences of "EPUB Creators MUST".  All of them should be reformulated.

Here is my proposed rewrite for the first one.

Values matching the pattern A11Y-VER indicate the version number of the EPUB Accessibility specification the publication conforms to, not including the decimal points.  The value "11" indicates conformance to this version of this specification.

mattgarrish:
In addition to finishing up the removal of epub creator, I also rewrote all the normative keywords that were masking as non-normative by being lowercase. I know there's an RFC that allows this but every time I come across them I have to try and figure out if they're mistakenly lowercase and really normative, even in informative sections (and that's not even getting into how accessible the practice really is). Just as we don't need epub creator we don't need to be using these keywords outside their intended use. The only place I left them is in the iana registrations since those are technically external documents we're reproducing. (I'm going to go back and do the same for the accessibility documents next.)

Fixes #2216 

***

Reading Systems 1.2:
* [Preview](https://cdn.statically.io/gh/w3c/epub-specs/editorial/issue-2216/epub34/rs/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/editorial/issue-2216/epub34/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2804.html" title="Last updated on Oct 2, 2025, 12:47 PM UTC (41578d9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2804/599d0f4...41578d9.html" title="Last updated on Oct 2, 2025, 12:47 PM UTC (41578d9)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -5390,8 +5390,8 @@ No Entry</pre>
 						foreign content documents hyperlinked to from hyperlinked documents have to be listed, and so
 						on.).</p>
 
-					<p>All EPUB and foreign content documents hyperlinked to from the [=EPUB navigation document=] also
-						MUST be listed in the <code>spine</code>, regardless of whether the EPUB navigation document is
+					<p>All EPUB and foreign content documents hyperlinked to from the [=EPUB navigation document=] 
+						MUST also be listed in the <code>spine</code>, regardless of whether the EPUB navigation document is
 						included in the <code>spine</code>.</p>
 
 					<div class="note">
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -930,12 +930,11 @@
 							of whether XHTML and SVG documents are listed in the spine or embedded in other EPUB content
 							documents they the same requirements for authoring and reading system support.</p>
 
-						<p>In practice, it means that EPUB creators can put XHTML and SVG core media type resources in
-							the spine without any modification or fallback as they are also conforming XHTML and SVG
-							content documents. But, this is a unique case &#8212; all other core media type resources
-							become foreign content documents when used in the spine (i.e., foreign content documents
-							include all foreign resources and all core media type resources except for XHTML and
-							SVG).</p>
+						<p>In practice, it means that XHTML and SVG core media type resources are allowed in the spine
+							without any modification or fallback as they are also conforming XHTML and SVG content
+							documents. But, this is a unique case &#8212; all other core media type resources become
+							foreign content documents when used in the spine (i.e., foreign content documents include
+							all foreign resources and all core media type resources except for XHTML and SVG).</p>
 					</div>
 				</section>
 			</section>
@@ -3272,9 +3271,8 @@
 					<div class="note">
 						<p>The <code>dir</code> attribute is marked <a href="#under-implemented">under-implemented</a>
 							as [=reading systems=] often only support a single default directionality for text display.
-							[=EPUB creators=] are still strongly encouraged to set the proper directionality of text
-							values in the [=package document=] to ensure proper rendering once this situation
-							improves.</p>
+							Regardless, it is still strongly encouraged to set the proper directionality of text values
+							in the [=package document=] to ensure proper rendering once this situation improves.</p>
 					</div>
 
 					<p
@@ -3292,7 +3290,7 @@
 					</ul>
 
 					<p data-tests="#pkg-dir-auto_root-rtl,#pkg-dir-auto_root-unset">Reading systems will assume the
-						value <code>auto</code> when EPUB creators omit the attribute or use an invalid value.</p>
+						value <code>auto</code> when attribute is not present or has an invalid value.</p>
 
 					<div class="note">
 						<p>The base direction specified in the <code>dir</code> attribute does not affect the ordering
@@ -3407,8 +3405,8 @@
 					<h4>The <code>refines</code> attribute</h4>
 
 					<p>Establishes an association between the current expression and the element or resource identified
-						by its value. [=EPUB creators=] MUST use as the value a [=path-relative-scheme-less-URL
-						string=], optionally followed by <code>U+0023&#160;(#)</code> and a [=URL-fragment string=] that
+						by its value. The value of the attribute MUST be a [=path-relative-scheme-less-URL string=],
+						optionally followed by <code>U+0023&#160;(#)</code> and a [=URL-fragment string=] that
 						references the resource or element they are describing.</p>
 
 					<aside class="example" title="Specifying that a creator is the illustrator">
@@ -3696,12 +3694,12 @@
 						</li>
 					</ol>
 
-					<p>The package document does not provide complex metadata encoding capabilities. If [=EPUB
-						creators=] need to provide more detailed information, they can associate metadata records (e.g.,
-						that conform to an international standard such as [[onix]] or are created for custom purposes)
-						using the [^link^] element. This approach allows reading systems to process the metadata in its
-						native form, avoiding the potential problems and information loss caused by translating to use
-						the minimal package document structure.</p>
+					<p>The package document does not provide complex metadata encoding capabilities. If more detailed
+						information needs to be provided, metadata records (e.g., that conform to an international
+						standard such as [[onix]] or are created for custom purposes) can be associated with the EPUB
+						publication using the [^link^] element. This approach allows reading systems to process the
+						metadata in its native form, avoiding the potential problems and information loss caused by
+						translating to use the minimal package document structure.</p>
 
 					<p id="core-metadata-reqs">In keeping with this philosophy, the package document only has the
 						following minimal metadata requirements: it MUST contain the [[dcterms]] [^dc:title^],
@@ -3733,9 +3731,9 @@
 					</aside>
 
 					<p>The [^meta^] element provides a generic mechanism for including <a href="#sec-vocab-assoc"
-							>metadata properties from any vocabulary</a>. Although EPUB creators MAY use this mechanism
-						for any metadata purposes, they will typically use it to include rendering metadata defined in
-						EPUB specifications.</p>
+							>metadata properties from any vocabulary</a>. Although this mechanism can be used for any
+						metadata purposes, it is typically used to include rendering and accessibility metadata defined
+						in EPUB specifications.</p>
 
 					<div class="note">
 						<p>See [[epub-a11y-111]] for accessibility metadata recommendations.</p>
@@ -3815,9 +3813,9 @@
 								</dd>
 							</dl>
 
-							<p>The [=EPUB creator=] MUST provide an identifier that is unique to one and only one [=EPUB
-								publication=] — its [=unique identifier=] — in an <code>dc:identifier</code> element.
-								This <code>dc:identifier</code> element MUST specify an <code>id</code> attribute whose
+							<p>The <code>dc:identifier</code> element MUST specify an identifier that is unique to one
+								and only one [=EPUB publication=] — its [=unique identifier=]. This
+									<code>dc:identifier</code> element MUST specify an <code>id</code> attribute whose
 								value is referenced from the [^package^] element's <a
 									href="#attrdef-package-unique-identifier"><code>unique-identifier</code>
 									attribute</a>.</p>
@@ -3833,27 +3831,25 @@
 &lt;/package&gt;</pre>
 							</aside>
 
-							<p>Although not static, EPUB creators should make changes to the unique identifier for an
-								EPUB publication as infrequently as possible. Unique Identifiers should have maximal
-								persistence both for referencing and distribution purposes. EPUB creators should not
-								issue new identifiers when making minor revisions such as updating metadata, fixing
-								errata, or making similar minor changes.</p>
+							<p>Although not static, changes to the unique identifier should only be made as infrequently
+								as possible. Unique Identifiers should have maximal persistence both for referencing and
+								distribution purposes. Do not issue new identifiers when making minor revisions such as
+								updating metadata, fixing errata, or making similar minor changes.</p>
 
-							<p>EPUB creators MAY specify additional identifiers.</p>
+							<p>Additional identifiers MAY be specified.</p>
 
 							<div class="note">
-								<p>EPUB creators are advised to use <a
-										href="https://url.spec.whatwg.org/#absolute-url-string">absolute-URL strings</a>
-									[url] for identifiers whenever possible. The inclusion of a domain owned by the EPUB
-									creator can improve the uniqueness of the identifier, for example, while the use of
-									a URN with a <a href="https://datatracker.ietf.org/doc/html/rfc8141#section-2.1"
+								<p>It is advised to use <a href="https://url.spec.whatwg.org/#absolute-url-string"
+										>absolute-URL strings</a> [url] for identifiers whenever possible. The inclusion
+									of a domain can improve the uniqueness of the identifier, for example, while the use
+									of a URN with a <a href="https://datatracker.ietf.org/doc/html/rfc8141#section-2.1"
 										>namespace identifier</a> [[rfc8141]] improves processing by reading
 									systems.</p>
 							</div>
 
-							<p>EPUB creators MAY use the <a href="#identifier-type"><code>identifier-type</code>
-									property</a> to indicate that the value of a <code>dc:identifier</code> element
-								conforms to an established system or an issuing authority granted it.</p>
+							<p>The <a href="#identifier-type"><code>identifier-type</code> property</a> MAY be used to
+								indicate that the value of a <code>dc:identifier</code> element conforms to an
+								established system or an issuing authority granted it.</p>
 
 							<aside class="example" title="Specifying the type of the identifier">
 								<p>In this example, the <code>identifier-type</code> property is used with the <a
@@ -3955,8 +3951,8 @@
 </pre>
 							</aside>
 
-							<p>[=EPUB creators=] should use only a single <code>dc:title</code> element to ensure
-								consistent rendering of the title in [=reading systems=].</p>
+							<p>It is advised to use only a single <code>dc:title</code> element to ensure consistent
+								rendering of the title in [=reading systems=].</p>
 
 							<div class="note">
 								<p>Although it is possible to include more than one <code>dc:title</code> element for
@@ -4056,15 +4052,14 @@
 &lt;/metadata&gt;</pre>
 							</aside>
 
-							<p>Although [=EPUB creators=] MAY specify additional <code>dc:language</code> elements for
-								multilingual Publications, [=reading systems=] will treat the first
-									<code>dc:language</code> element in document order as the primary language of the
-								EPUB publication.</p>
+							<p>Although additional <code>dc:language</code> elements MAY be specified for multilingual
+								[=EPUB publications=], [=reading systems=] will treat the first <code>dc:language</code>
+								element in document order as the primary language.</p>
 
 							<div class="note">
 								<p>[=Publication resources=] do not inherit their language from the
-										<code>dc:language</code> element(s). EPUB creators must set the language of a
-									resource using the intrinsic methods of the format.</p>
+										<code>dc:language</code> element(s). The language of each resource has to be set
+									using the intrinsic methods of the format.</p>
 							</div>
 						</section>
 					</section>
@@ -4147,10 +4142,9 @@
 
 							<p>The <dfn class="export" data-lt-no-plural="" data-dfn-type="element"
 										><code>dc:creator</code></dfn> element [[dcterms]] represents the name of a
-								person, organization, etc. responsible for the creation of the content. [=EPUB
-								creators=] MAY <a href="#subexpression">associate</a> a <a href="#role"
-										><code>role</code> property</a> with the element to indicate the function the
-								creator played.</p>
+								person, organization, etc. responsible for the creation of the content. A <a
+									href="#role"><code>role</code> property</a> MAY be <a href="#subexpression"
+									>associated</a> with the element to indicate the function the creator played.</p>
 
 							<aside class="example" title="Specifying that a creator is an author">
 								<p>In this example, the <a href="https://id.loc.gov/vocabulary/relators.html">MARC
@@ -4174,11 +4168,11 @@
 &lt;/metadata&gt;</pre>
 							</aside>
 
-							<p>The <code>dc:creator</code> element should contain the name of the creator as EPUB
-								creators intend [=reading systems=] to display it to users.</p>
+							<p>The <code>dc:creator</code> element should contain the name of the creator as [=reading
+								systems=] are expected to display it to users.</p>
 
-							<p>EPUB creators MAY use the <a href="#file-as"><code>file-as</code> property</a>
-								<a href="#subexpression">to associate</a> a normalized form of the creator's name, and
+							<p>The <a href="#file-as"><code>file-as</code> property</a> MAY be used <a
+									href="#subexpression">to associate</a> a normalized form of the creator's name, and
 								the <a href="#alternate-script"><code>alternate-script</code> property</a> to represent
 								the creator's name in another language or script.</p>
 
@@ -4204,8 +4198,8 @@
 &lt;/metadata&gt;</pre>
 							</aside>
 
-							<p>If an [=EPUB publication=] has more than one creator, EPUB creators should specify each
-								in a separate <code>dc:creator</code> element.</p>
+							<p>If an [=EPUB publication=] has more than one creator, specify each in a separate
+									<code>dc:creator</code> element.</p>
 
 							<p>The document order of <code>dc:creator</code> elements in the <code>metadata</code>
 								section determines the display priority, where the first <code>dc:creator</code> element
@@ -4228,8 +4222,7 @@
 &lt;/metadata&gt;</pre>
 							</aside>
 
-							<p>EPUB creators should represent secondary contributors using the [^dc:contributor^]
-								element.</p>
+							<p>Secondary contributors are represented using the [^dc:contributor^] element.</p>
 						</section>
 
 						<section id="sec-opf-dcdate" data-epubcheck="true"
@@ -4239,8 +4232,8 @@
 							<p>The <dfn class="export" data-lt-no-plural="" data-dfn-type="element"
 									><code>dc:date</code></dfn> element [[dcterms]] defines the publication date of the
 								[=EPUB publication=]. The publication date is not the same as the <a
-									href="#last-modified-date">last modified date</a> (the last time the [=EPUB
-								creator=] changed the EPUB publication).</p>
+									href="#last-modified-date">last modified date</a> (the last time the EPUB
+								publication was changed).</p>
 
 							<p>It is RECOMMENDED that the date string conform to [[iso8601-1]], particularly the subset
 								expressed in W3C Date and Time Formats [[datetime]], as such strings are both human and
@@ -4256,8 +4249,8 @@
 &lt;/metadata&gt;</pre>
 							</aside>
 
-							<p>EPUB creators should express additional dates using the specialized date properties
-								available in the [[dcterms]] vocabulary, or similar.</p>
+							<p>Additional dates can be expressed using the specialized date properties available in the
+								[[dcterms]] vocabulary, or similar.</p>
 
 							<p>EPUB publications MUST NOT contain more than one <code>dc:date</code> element.</p>
 						</section>
@@ -4267,15 +4260,16 @@
 
 							<p>The <dfn class="export" data-lt-no-plural="" data-dfn-type="element"
 										><code>dc:subject</code></dfn> element [[dcterms]] identifies the subject of the
-								[=EPUB publication=]. [=EPUB creators=] should set the [=value=] of the element to the
-								human-readable heading or label, but may use a code value if the subject taxonomy does
-								not provide a separate descriptive label.</p>
+								[=EPUB publication=]. It is advised to set the [=value=] of the element to the
+								human-readable heading or label, but a code value can be used if the subject taxonomy
+								does not provide a separate descriptive label.</p>
 
-							<p>EPUB creators MAY identify the system or scheme they drew the element's [=value=] from
-								using the <a href="#authority"><code>authority</code> property</a>.</p>
+							<p>The system or scheme the element's [=value=] is drawn from can be identified using the <a
+									href="#authority"><code>authority</code> property</a>.</p>
 
-							<p>When a scheme is identified, EPUB creators MUST <a href="#subexpression">associate</a> a
-								subject code using the <a href="#term"><code>term</code> property</a>.</p>
+							<p>When a scheme is identified, a subject code MUST be <a href="#subexpression"
+									>associated</a> with the element using the <a href="#term"><code>term</code>
+									property</a>.</p>
 
 							<aside class="example" title="Specifying a BISAC code and heading">
 								<pre>&lt;metadata …&gt;
@@ -4328,7 +4322,7 @@
 								publication=] is of a specialized type (e.g., annotations or a dictionary packaged in
 								EPUB format).</p>
 
-							<p>[=EPUB creators=] MAY use any text string as a [=value=].</p>
+							<p>The element's [=value=] MAY be any text string.</p>
 
 							<div class="note">
 								<p>The former <abbr title="International Digital Publishing Forum">IDPF</abbr> EPUB 3
@@ -4426,8 +4420,8 @@
 						of the element represents the assertion. (Refer to <a href="#sec-vocab-assoc"></a> for more
 						information.)</p>
 
-					<p id="meta-expr-types">This specification defines two types of metadata expressions that [=EPUB
-						creators=] can define using the <code>meta</code> element:</p>
+					<p id="meta-expr-types">This specification defines two types of metadata expressions that can be
+						defined using the <code>meta</code> element:</p>
 
 					<ul>
 						<li id="primary-expression">A <em>primary expression</em> is one in which the expression defined
@@ -4441,8 +4435,8 @@
 							expression by defining the role of the person.</li>
 					</ul>
 
-					<p>EPUB creators MAY use subexpressions to refine the meaning of other subexpressions, thereby
-						creating chains of information.</p>
+					<p>Subexpressions MAY be used to refine the meaning of other subexpressions, thereby creating chains
+						of information.</p>
 
 					<p class="note">All the [[dcterms]] elements represent primary expressions, and permit refinement by
 						meta element subexpressions.</p>
@@ -4451,8 +4445,7 @@
 							href="#sec-default-vocab">default vocabulary</a> for use with the <code>property</code>
 						attribute.</p>
 
-					<p>EPUB creators MAY add terms from other vocabularies as defined in <a href="#sec-vocab-assoc"
-						></a>.</p>
+					<p>Terms from other vocabularies MAY be added as defined in <a href="#sec-vocab-assoc"></a>.</p>
 
 					<aside class="example" title="Using properties with reserved prefixes">
 						<p>For the full list of reserved prefixes, refer to <a href="#sec-reserved-prefixes"></a>.</p>
@@ -4475,8 +4468,8 @@
 &lt;/metadata&gt;</pre>
 					</aside>
 
-					<p id="attrdef-scheme">The <code>scheme</code> attribute identifies the system or scheme the EPUB
-						creator obtained the element's [=value=] from. The value of the attribute MUST be a <a
+					<p id="attrdef-scheme">The <code>scheme</code> attribute identifies the system or scheme the
+						element's [=value=] was obtained from. The value of the attribute MUST be a <a
 							href="#sec-property-datatype"><var>property</var> data type value</a> that resolves to the
 						resource that defines the scheme. The <code>scheme</code> attribute does not have a <a
 							href="#sec-default-vocab">default vocabulary</a> (i.e., all values require a <a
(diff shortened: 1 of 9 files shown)


epub34/authoring/vocab/rendering.html
```diff
@@ -97,17 +97,15 @@ <h3>Package rendering vocabulary</h3>
...
 		
-		<p>[=Reading system=] developers may introduce functionality not defined in this specification to address 
-			reading system-specific issues rendering [=EPUB content documents=].</p>
-		
-		<p>To facilitate this experimentation, [=EPUB creators=] MAY include custom properties and [=EPUB spine | spine=]
-			overrides for use in the [=package document=] provided they do not use the <code>rendition:</code>
-			prefix.</p>
+		<p>To address reading system-specific issues rendering [=EPUB content documents=], 
+			[=reading system=] developers MAY define custom properties and [=EPUB spine | spine=]
+			overrides for use in the [=package document=]. These properties MUST NOT use the
```
bduga: This MUST NOT seems correct, but also wasn't mentioned here before. Does it already exist elsewhere, or is this new?

epub34/authoring/vocab/rendering.html
```diff
@@ -97,17 +97,15 @@ <h3>Package rendering vocabulary</h3>
...
 		<h4>Custom rendering properties</h4>
 		
-		<p>[=Reading system=] developers may introduce functionality not defined in this specification to address 
-			reading system-specific issues rendering [=EPUB content documents=].</p>
-		
-		<p>To facilitate this experimentation, [=EPUB creators=] MAY include custom properties and [=EPUB spine | spine=]
-			overrides for use in the [=package document=] provided they do not use the <code>rendition:</code>
-			prefix.</p>
+		<p>To address reading system-specific issues rendering [=EPUB content documents=], 
+			[=reading system=] developers MAY define custom properties and [=EPUB spine | spine=]
```
bduga: This is an interesting rewrite - previously it was allowing epub creators to use custom properties, but now it is allowing reading system developers to define custom properties. Should it be more explicit that these custom properties and spine overrides are allowed to be in the epub?

bduga (approved):

epub34/authoring/vocab/rendering.html
mattgarrish: It was an implied "must not" in the old wording:

> ... provided they do not use the rendition: prefix.

Better to surface the restriction in normative terms, imo.

epub34/authoring/vocab/rendering.html
mattgarrish: This is a good example of why having lowercase non-normative keywords gets confusing. If we're allowing reading system developers to extend the specification, that is clearly a normative allowance to me. I don't know how we get to allowing custom properties without saying who can create them and for what purpose. That's why I kept the "may" keyword for who can define the properties but made it officially normative.

As for stripping out that epub creators can use them, or that epub publications can allow them, nothing in this section changes that that is already true. It's a redundant allowance since you can include any metadata properties and overrides you want in a publication by defining a prefix for them. That's the extensibility model of epub.

That was my thinking behind the change, anyway. Feel free to disagree.

mattgarrish: Although, I'm kind of wondering now why we have this section at all. It sort of seems to mirror the custom attributes section, but custom attributes would otherwise be invalid without our allowing them. This section is really a whole lot of nothing since we can't stop a reading system from using custom properties if we tried.

epub34/authoring/vocab/rendering.html
bduga: Oh, yeah, I guess my eyes glazed over for that paragraph.

epub34/authoring/vocab/rendering.html
mattgarrish: Thinking on this some more, the only meaningful statement is that you can't define custom properties in the rendition: vocabulary. And oddly, the RS spec says to ignore properties that conflict behaviorally with rendition: properties but we don't say anything about not creating such properties here. 

Maybe we could improve this section with text like:

>Custom properties and spine overrides can be included in the package document to address rendering issues specific to particular reading systems, as defined by the developers of those reading systems.
> The only restrictions are that such custom properties MUST NOT be defined with a rendition: prefix and MUST NOT conflict behaviorally with properties in the package rendering vocabulary.

> If extensions are needed that multiple independent reading systems can use, the preferred method is to extend the package rendering vocabulary through a revision to the standard.

epub34/rs/index.html
```diff
@@ -287,12 +287,11 @@ <h3>Error reporting</h3>
...
 					are the developer tools that web browsers provide for debugging HTML pages and applications.</p>
 
-				<p>[=EPUB creators=], for example, can greatly benefit from having access to such processing information
-					(e.g., to efficiently debug their EPUB publications). For maximum use in debugging, it is
-					recommended that reading systems not only report issues they directly encounter but also issues
-					reported by any applications they use (e.g., HTML, CSS, and JavaScript errors reported from a
-					browser core used to render the content, or validation issues reported by <a
-						href="https://www.w3.org/publishing/epubcheck/">EPUBCheck</a>).</p>
+				<p>Access to such processing information allows, for example, more efficient debugging of EPUB
+					publications. For maximum use in debugging, it is recommended that reading systems not only report
```
```suggestion
					publications. For maximum use in debugging, it is recommended that reading systems not only report
```

I believe "RECOMMENDED" is also an RFC term, isn't it? I am not sure what to replace it with, though. "it is preferable"?

epub34/rs/index.html
```diff
@@ -1089,7 +1088,7 @@ <h3>MathML</h3>
 							</li>
 						</ul>
 
-						<p class="note">Reading systems may choose to use third-party libraries such as <a
+						<p class="note">Reading systems can opt to use third-party libraries such as <a
```
```suggestion
						<p class="note">Reading systems can opt for third-party libraries such as <a
```

May be wrong, but this sounds better to my ears...

epub34/rs/index.html
```diff
@@ -2162,10 +2159,10 @@ <h2>Processing <code>property</code> values</h2>
 						<summary>Explanation</summary>
 						<p>The path-relative-scheme-less-URL string definition has a number of restrictions that apply
 							to the reference whether it has a prefix or not.</p>
-						<p>For example, the reference must not begin with a [=URL-scheme string=] followed by a colon.
+						<p>For example, the reference cannot begin with a [=URL-scheme string=] followed by a colon.
```
```suggestion
						<p>For example, better to avoid a reference beginning with a [=URL-scheme string=] followed by a colon.
```

The word "cannot" does not sound equivalent to the "must not" that it is supposed to replace. "Cannot" sounds like "not able to" to me.

(There may be a better replacement than mine, though.)

epub34/rs/index.html
```diff
@@ -2325,8 +2322,8 @@ <h2>Accessibility</h2>
 				<li>Document Object Model (DOM) &#8212; Provide access to the [[dom]] of EPUB content documents so that
 					users can investigate the semantics and structure expressed in the source.</li>
 				<li>Table of Contents &#8212; Ensure that users can access the link text, including <a
-						data-cite="epub-34#confreq-nav-a-title">text alternatives for visual content</a>. Activating the
-					table of contents links should be possible using different inputs.</li>
+						data-cite="epub-34#confreq-nav-a-title">text alternatives for visual content</a>. Ensure users
+					can activate the table of contents links using different inputs.</li>
```
```suggestion
						data-cite="epub-34#confreq-nav-a-title">text alternatives for visual content</a>, and that users
					can activate the table of contents links using different inputs.</li>
```

(It avoids repeating the word "Ensure" in two consecutive sentences...)

epub34/rs/index.html
```diff
@@ -2460,17 +2457,17 @@ <h3>Recommendations</h3>
...
+					choose to inform users by other means, such as when a user creates an account on their web site).
 					Anonymization of any collected data is RECOMMENDED for the privacy and the security of the user and
 					reading system.</p>
 
-				<p>It is also understood that user data may be required or helpful for some reading system affordances.
-					In these cases, anonymization is also RECOMMENDED. Reading systems also SHOULD inform users of what
-					data is needed, what it is to be used for, and to provide methods to opt-out.</p>
+				<p>It is also understood that user data could be required or helpful for some reading system
+					affordances. In these cases, anonymization is also RECOMMENDED. Reading systems also SHOULD inform
+					users of what data is needed, what it is to be used for, and to provide methods to opt-out.</p>
```
```suggestion
					affordances. In these cases, anonymization is also RECOMMENDED. Reading systems SHOULD also inform
					users of what data is needed, what it is to be used for, and to provide methods to opt-out.</p>
```

Isn't that a better word order? (I note that the order was similar in the original, too.)

epub34/authoring/index.html
```diff
@@ -592,14 +571,14 @@ <h3>Terminology</h3>
...
-							may locate resources listed in <a href="#sec-resource-locations"></a> outside the EPUB
-							container).</p>
+							of an [=EPUB publication=]. In the absence of this resource, [=reading systems=] might not
+							render the EPUB publication as intended. Examples of publication resources include the
+							[=package document=], [=EPUB content documents=], CSS Style Sheets, audio, video, images,
+							embedded fonts, and scripts.</p>
+						<p>The package document [=EPUB manifest | manifest=] has to include a list of all publication
+							resources and the resources typically have to be bundled in the [=EPUB container=] (the
+							exception being that resources listed in <a href="#sec-resource-locations"></a> can be
+							located outside the EPUB container).</p>
```
```suggestion
						<p>The package document [=EPUB manifest | manifest=] has to include a list of all publication
							resources that typically have to be bundled in the [=EPUB container=] (the
							exception being that resources listed in <a href="#sec-resource-locations"></a> can be
							located outside the EPUB container).</p>
```

Avoiding word repetition...

epub34/authoring/index.html
```diff
@@ -950,14 +925,14 @@ <h4>The content plane</h4>
...
-							authoring and reading system support.</p>
-
-						<p>In practice, it means that EPUB creators can put XHTML and SVG core media type resources in
-							the spine without any modification or fallback (they are also conforming XHTML and SVG
-							content documents), but this is a unique case. All other core media type resources become
+							and SVG as core media type resources with the requirement that the markup conform to their
+							respective EPUB content document definitions. This common definition ensures that regardless
+							of whether XHTML and SVG documents are listed in the spine or embedded in other EPUB content
+							documents they have the same requirements for authoring and reading system support.</p>
+
```
iherman, line 933:
```suggestion
							and SVG as core media type resources with the requirement that the markup conform to their
							respective EPUB content document definitions. This common definition ensures that regardless
							of whether XHTML and SVG documents are listed in the spine, or embedded in other EPUB content
							documents, they have the same requirements for authoring and reading system support.</p>

```

epub34/authoring/index.html
```diff
@@ -1769,15 +1734,15 @@ <h4>File and directory structure</h4>
...
+						the root directory provided they are not within the <code>META-INF</code> directory. [=EPUB
+						publications=] MUST NOT contain references to files in the <code>META-INF</code> directory.</p>
 
 					<div class="note">
 						<p>Some [=reading systems=] do not provide access to resources outside the directory where the
-							[=package document=] is stored. EPUB creators should therefore place all resources at or
-							below the directory containing the package document to avoid interoperability issues.</p>
+							[=package document=] is stored even though this is not a restriction defined in
+							[[epub-rs-34]]. To avoid interoperability issues with these reading systems, place all
+							resources at or below the directory containing the package document.</p>
```
```suggestion
							[=package document=] is stored even though this is not a restriction defined in
							[[epub-rs-34]]. To avoid interoperability issues with these reading systems, it is advised to place all
							resources at or below the directory containing the package document.</p>
```

epub34/authoring/index.html
```diff
@@ -2637,13 +2598,13 @@ <h6>Order of compression and encryption</h6>
...
-										<code>Compression</code> XML element defined below. When EPUB creators do not
-									compress streams of data before encrypting, they MAY provide the additional
-										<code>EncryptionProperties</code> metadata to specify the size of the initial
-									resource (i.e., before encryption).</p>
+								<p>When streams of data are compressed before encrypting, additional
+										<code>EncryptionProperties</code> metadata SHOULD be provided to specify the
+									size of the initial resource (i.e., before compression and encryption), as per the
+										<code>Compression</code> XML element defined below. When streams of data are not
+									compresed before encrypting, additional <code>EncryptionProperties</code> metadata
+									MAY be provided to specify the size of the initial resource (i.e., before
```
```suggestion
									compressed before encrypting, additional <code>EncryptionProperties</code> metadata
									MAY be provided to specify the size of the initial resource (i.e., before
```

epub34/authoring/index.html
```diff
@@ -2811,55 +2772,59 @@ <h6>The <code>signatures</code> element</h6>
 								</dl>
 
 								<p>The <code>signature</code> element contains child elements of type
-										<code>Signature</code>, as defined by [[xmldsig-core1]]. [=EPUB creators=] can
-									apply signatures to an EPUB publication as a whole or to its parts, and can specify
-									the signing of any kind of data (i.e., not just XML).</p>
+										<code>Signature</code>, as defined by [[xmldsig-core1]]. Signatures can be
+									applied to an EPUB publication as a whole or to its parts. They can also be used to
+									sign of any kind of data (i.e., not just XML).</p>
```
```suggestion
										<code>Signature</code>, as defined by [[xmldsig-core1]]. Signatures can be
									applied to an EPUB publication as a whole or to its parts. They can also be used to
									sign any kind of data (i.e., not just XML).</p>
```

epub34/authoring/index.html
```diff
@@ -2811,55 +2772,59 @@ <h6>The <code>signatures</code> element</h6>
...
+								<p>When a <code>signatures.xml</code> file is not included, no part of the [=OCF
+									abstract container=] is signed at the container level but digital signing might
+									exist within the EPUB publication.</p>
 
-								<p id="sig-container">When an EPUB creator creates a data signature for the OCF abstract
-									container, they SHOULD add the signature as the last child <code>Signature</code>
-									element of the <code>signatures</code> element.</p>
+								<p id="sig-container">When a data signature is created for the OCF abstract container,
+									the signature SHOULD be stored in the last child <code>Signature</code> element of
+									the <code>signatures</code> element.</p>
```
iherman, line 2789:
```suggestion
								<p id="sig-container">When a data signature is created for the OCF abstract container,
									the signature SHOULD be stored as the last child <code>Signature</code> element of
									the <code>signatures</code> element.</p>
```

epub34/authoring/index.html
```diff
@@ -5428,18 +5381,18 @@ <h4>The <code>spine</code> element</h4>
...
-							<code>spine</code>, where hyperlinking encompasses any linking mechanism that requires the
-						user to navigate away from the current resource. Common hyperlinking mechanisms include the
-						[^a/href^] attribute of the [[html]] [^a^] and [^area^] elements and scripted links (e.g., using
-						DOM Events and/or form elements). The requirement to list hyperlinked resources applies
-						recursively (i.e., EPUB creators must list all EPUB and foreign content documents hyperlinked to
-						from hyperlinked documents, and so on.).</p>
+					<p id="spine-inclusion-req">All EPUB and foreign content documents that are hyperlinked to from
+						publication resources in the <code>spine</code> MUST be list in the <code>spine</code>, where
+						hyperlinking encompasses any linking mechanism that requires the user to navigate away from the
+						current resource. Common hyperlinking mechanisms include the [^a/href^] attribute of the
```
iherman, line 5386:
```suggestion
					<p id="spine-inclusion-req">All EPUB and foreign content documents that are hyperlinked to from
						publication resources in the <code>spine</code> MUST be listed in the <code>spine</code>, where
						hyperlinking encompasses any linking mechanism that requires the user to navigate away from the
						current resource. Common hyperlinking mechanisms include the [^a/href^] attribute of the
```

epub34/authoring/index.html
```diff
@@ -5428,18 +5381,18 @@ <h4>The <code>spine</code> element</h4>
...
+						elements). The requirement to list hyperlinked resources applies recursively (i.e., all EPUB and
+						foreign content documents hyperlinked to from hyperlinked documents have to be listed, and so
+						on.).</p>
 
-					<p>EPUB creators also MUST list in the <code>spine</code> all EPUB and foreign content documents
-						hyperlinked to from the [=EPUB navigation document=], regardless of whether EPUB creators
-						include the EPUB navigation document in the <code>spine</code>.</p>
+					<p>All EPUB and foreign content documents hyperlinked to from the [=EPUB navigation document=] also
+						MUST be listed in the <code>spine</code>, regardless of whether the EPUB navigation document is
+						included in the <code>spine</code>.</p>
```
```suggestion
					<p>All EPUB and foreign content documents hyperlinked to from the [=EPUB navigation document=] 
						MUST also be listed in the <code>spine</code>, regardless of whether the EPUB navigation document is
						included in the <code>spine</code>.</p>
```

epub34/authoring/index.html
```diff
@@ -5906,32 +5858,31 @@ <h5>Structural semantics</h5>
...
 						<div class="note">
 							<p>The listing of RDFa does not express a preference on the part of the Working Group, only
-								that these attributes represent an extension of the HTML grammar. EPUB creators can also
-								specify <a data-cite="html#microdata">microdata attributes</a> [[html]] and <a
-									data-cite="json-ld11#">linked data</a> [[json-ld11]] in XHTML content documents as
-								both are natively supported.</p>
+								that these attributes represent an extension of the HTML grammar. <a
+									data-cite="html#microdata">Microdata attributes</a> [[html]] and <a
+									data-cite="json-ld11#">linked data</a> [[json-ld11]] are natively supported in XHTML
+								content documents.</p>
```
```suggestion
								that these attributes represent an extension of the HTML grammar. <a
									data-cite="html#microdata">Microdata attributes</a> [[html]] and <a
									data-cite="json-ld11#">linked data</a> [[json-ld11]] are natively supported in XHTML
								content documents and can be used as well.</p>
```

epub34/authoring/index.html
```diff
@@ -5906,32 +5858,31 @@ <h5>Structural semantics</h5>
...
-							localization.</p>
+						<p>The [[its20]] specification defines a set of attributes that MAY be used in [=XHTML content
+							documents=] to add support for internationalization, translation, and localization.</p>
 
-						<p>EPUB creators MUST only use ITS attributes as they are defined in <a
-								data-cite="its20#html5-markup">Using ITS markup in HTML</a> [[its20]] (i.e., EPUB 3 does
-							not support the namespaced attributes).</p>
+						<p>ITS attributes MUST only be used as they are defined in <a data-cite="its20#html5-markup"
+								>Using ITS markup in HTML</a> [[its20]] (i.e., EPUB 3 does not support the namespaced
+							attributes).</p>
```
```suggestion
						<p>ITS attributes MUST only be used as defined in <a data-cite="its20#html5-markup"
								>Using ITS markup in HTML</a> [[its20]] (i.e., EPUB 3 does not support the namespaced
							attributes).</p>
```

epub34/authoring/index.html
```diff
@@ -5955,8 +5906,8 @@ <h5>Custom attributes</h5>
 
 						<div class="note">
 							<p>Custom attributes are usually defined in a reading system-specific manner and are not
-								intended for use by other reading systems. This specification should be extended to
-								provide extensions that multiple independent reading systems can use.</p>
+								intended for use by other reading systems. The preferred method to add extensions for
+								multiple indenpendent reading system to use is to extend this specification.</p>
```
```suggestion
								intended for use by other reading systems. The preferred method to add extensions for
								multiple independent reading systems to use is to extend this specification.</p>
```

epub34/authoring/index.html
```diff
@@ -6347,8 +6292,8 @@ <h4>Script inclusion</h4>
 								if a future update adds the concept.</p>
 						</div>
 
-						<p>[=EPUB creators=] should note that [=reading systems=] are required to behave as though a
-							unique <a data-cite="html#origin">origin</a> [[html]] has been assigned to each [=EPUB
+						<p>Note that [=reading systems=] are required to behave as though a unique <a
+								data-cite="html#origin">origin</a> [[html]] has been assigned to each [=EPUB
```
```suggestion
						<p>Note that [=reading systems=] are required to behave as though a unique <a
								data-cite="html#origin">origin</a> [[html]] had been assigned to each [=EPUB
```

Isn't that the correct grammatical form? But I may be wrong.

epub34/authoring/index.html
```diff
@@ -6501,30 +6442,30 @@ <h2>EPUB navigation document</h2>
...
-					navigational interfaces from information found in the EPUB navigation document, and this may make
-					the result difficult to read. If EPUB creators require such formatting and functionality, then they
-					should also include the EPUB navigation document in the [=EPUB spine | spine=]. The use of
-					progressive enhancement techniques for scripting and styling of the navigation document will help
-					ensure the content will retain its integrity when rendered in a non-browser context.</p>
+					navigation elements (i.e., the rest of the document can be marked up like any other XHTML content
+					document). As a result, it can also be part of the linear reading order, avoiding the need for
+					duplicate tables of contents. Navigation elements that are only for machine processing (e.g., the
+					page list) can be hidden using the <a href="#sec-nav-doc-use-spine"><code>hidden</code>
+						attribute</a>.</p>
```
```suggestion
					duplicate tables of contents. Navigation elements that are only meant for machine processing (e.g., the
					page list) can be hidden using the <a href="#sec-nav-doc-use-spine"><code>hidden</code>
						attribute</a>.</p>
```

epub34/authoring/index.html
```diff
@@ -6945,33 +6885,33 @@ <h4>Other <code>nav</code> elements</h4>
...
+					levels of subsections.</p>
 
-				<p>In these cases, EPUB creators are advised use the [[html]] [^html-global/hidden^] attribute to
-					indicate which (if any) portions of the navigation data are excluded from rendering in the content
-					flow. The <a data-cite="epub-rs-34#confreq-nav-hidden">the attribute has no effect</a>
-					[[epub-rs-34]] on how reading systems render the navigation document outside of the spine.</p>
+				<p>In these cases, it is advised to use the [[html]] [^html-global/hidden^] attribute to indicate which
+					(if any) portions of the navigation data are excluded from rendering in the content flow. The <a
+						data-cite="epub-rs-34#confreq-nav-hidden">the attribute has no effect</a> [[epub-rs-34]] on how
+					reading systems render the navigation document outside of the spine.</p>
```
iherman, line 6905:
```suggestion
					(if any) portions of the navigation data are excluded from rendering in the content flow. The <a
						data-cite="epub-rs-34#confreq-nav-hidden">attribute has no effect</a> [[epub-rs-34]] on how
					reading systems render the navigation document outside of the spine.</p>
```

epub34/authoring/index.html
```diff
@@ -7322,16 +7255,14 @@ <h5>Synthetic spreads</h5>
 
 							<dt>auto</dt>
 							<dd>
-								<p>The EPUB creator is not defining an explicit synthetic spread behavior. Default
-									value.</p>
+								<p>No synthetic spread behavior preference. Default value.</p>
```
```suggestion
								<p>No synthetic spread behavior preference is defined. Default value.</p>
```

epub34/authoring/index.html
```diff
@@ -7479,9 +7410,9 @@ <h5>Synthetic spreads</h5>
 
 						<aside class="example" id="spread-both-with-intro-example"
 							title="Overriding the global spread behavior">
-							<p>In this example, the EPUB creator overrides the global reflowable setting in the spine
-								for the introductory page. The intention is for reading systems to render it as a
-								reflowable document.</p>
+							<p>In this example, the global reflowable setting in the spine is overridden for the
+								introductory page. The intention is for reading systems to render it as a reflowable
+								document.</p>
```
```suggestion
							<p>In this example, the global reflowable setting is overridden in the spine for the
								introductory page. The intention is for reading systems to render it as a reflowable
								document.</p>
```

epub34/authoring/index.html
```diff
@@ -7602,30 +7531,29 @@ <h5>Spread placement</h5>
...
-								<code>rendition:page-spread-left</code> and <code>rendition:page-spread-right</code>
-							properties on the [=EPUB spine | spine=] items for the two adjacent EPUB content documents,
-							and omit the properties on spine items where one-up or two-up presentation is equally
-							acceptable.</p>
-
-						<p>EPUB creators MUST NOT declare more than one <code>rendition:page-spread-*</code> property,
-							and/or their unprefixed equivalents, on any given spine item (e.g., it is valid to specify
-							both "<code>rendition:page-spread-left page-spread-left</code>" in case reading systems only
+						<p>Although it is common practice to specify to use a spread in certain device orientations, the
+							content itself does not represent a true spread &#8212; two consecutive pages that have to
```
```suggestion
						<p>Although it is common practice to specify the use a spread in certain device orientations, the
							content itself does not represent a true spread &#8212; two consecutive pages that have to
```

epub34/authoring/index.html
```diff
@@ -7602,30 +7531,29 @@ <h5>Spread placement</h5>
...
+							be rendered side-by-side for readability, such as a two-page map. To indicate that two
+							consecutive pages represent a true spread, the <code>rendition:page-spread-left</code> and
+								<code>rendition:page-spread-right</code> properties SHOULD be set on the [=EPUB spine |
+							spine=] items for the two adjacent EPUB content documents and the properties omitted on
+							spine items where one-up or two-up presentation is equally acceptable.</p>
```
```suggestion
							spine=] items for the two adjacent EPUB content documents, and the properties omitted on
							spine items where one-up or two-up presentation is equally acceptable.</p>
```

epub34/authoring/index.html
```diff
@@ -8625,7 +8546,7 @@ <h4>Introduction</h4>
 					<p>The <code>text</code> element <code>src</code> attribute references the associated phrase,
 						sentence, or other segment of the EPUB content document by its URL [[url]] reference. The
 							<code>audio</code> element <code>src</code> attribute similarly references the location of
-						the corresponding audio clip and adds the OPTIONAL <a href="#attrdef-smil-clipBegin"
+						the corresponding audio clip and adds the optional <a href="#attrdef-smil-clipBegin"
```
```suggestion
						the corresponding audio clip and adds the optional <a href="#attrdef-smil-clipBegin"
```

"optional" is also a keyword in rfc2119; I guess it must be replaced, too. But I am not sure with what...

epub34/authoring/index.html
```diff
@@ -8881,24 +8799,24 @@ <h5>Overlay structure</h5>
...
 
 					<section id="sec-media-overlays-granularity" class="informative">
 						<h5>Overlay granularity</h5>
 
-						<p>The granularity level of the media overlay depends on how [=EPUB creators=] mark up the
-							[=EPUB content document=] and the type of fragment identifier they use in the [^text^]
-							elements' <code>src</code> attributes and the [^seq^] elements' <code>epub:textref</code>
+						<p>The granularity level of the media overlay depends on how the [=EPUB content document=] are
+							marked up and the type of fragment identifier used in the [^text^] elements'
+								<code>src</code> attributes and the [^seq^] elements' <code>epub:textref</code>
```
```suggestion
						<p>The granularity level of the media overlay depends on how the [=EPUB content document=] is
							marked up and the type of fragment identifier used in the [^text^] elements'
								<code>src</code> attributes and the [^seq^] elements' <code>epub:textref</code>
```

epub34/authoring/index.html
```diff
@@ -9732,42 +9646,41 @@ <h3>Recommendations</h3>
 						implementations.</li>
 				</ul>
 
-				<p>EPUB creators also need to consider the privacy rights of users and avoid situations where they are
-					intentionally collecting data. Ideally, EPUB creators SHOULD NOT track their users, but this is not
-					realistic for all types of publishing.</p>
+				<p>Also consider the privacy rights of users and avoid intentionally collecting data. Ideally, users
+					SHOULD NOT be tracked, but this is not realistic for all types of publishing.</p>
```
iherman, line 9646: There is an issue with normative vs. not. Is this section normative? Yes, it has _not_ been marked as non-normative, ie, it is by default. But... this means that all normative statements should be tested, right?

Or... isn't this a bug and shouldn't that section be also marked as informative (which means that it has to be reworked, too)?

(I just checked: the VC specification, for example, marked the whole of the privacy/security section as informative...)

cc @reidmore-online

epub34/authoring/index.html
```diff
@@ -10467,13 +10376,11 @@ <h4>The <code>prefix</code> attribute</h4>
...
-					<p>EPUB creators MUST declare the attribute in the namespace
-							<code>http://www.idpf.org/2007/ops</code> in [=EPUB content documents=] and [=media overlay
-						documents=].</p>
+					<p>The attribute MUST be declared in the namespace <code>http://www.idpf.org/2007/ops</code> when
+						used in [=EPUB content documents=] and [=media overlay documents=].</p>
 
 					<aside class="example" title="Declaring prefixes in an XHTML content document">
-						<p>In this example, the EPUB creator declares a prefix for the Z39.98 Structural Semantics
-							Vocabulary.</p>
+						<p>In this example, the a prefix is declared for the Z39.98 Structural Semantics Vocabulary.</p>
```
```suggestion
						<p>In this example, a prefix is declared for the Z39.98 Structural Semantics Vocabulary.</p>
```

epub34/authoring/index.html
```diff
@@ -10500,44 +10407,42 @@ <h4>The <code>prefix</code> attribute</h4>
...
 					<h4>Reserved prefixes</h4>
 
 					<div class="caution">
-						<p>Although reserved prefixes are an authoring convenience, [=EPUB creators=] should avoid
-							relying on them as they may cause interoperability issues. [=EPUB conformance checkers=]
-							will often reject new prefixes until their developers update the tools to the latest version
-							of the specification, for example. EPUB creators should declare all prefixes they use to
-							avoid such issues.</p>
+						<p>Although reserved prefixes are an authoring convenience, they can cause issues in the supply
+							chain. Vendors, for example, will often reject new prefixes until they update their [=EPUB
```
```suggestion
						<p>Although reserved prefixes are an authoring convenience, they can cause interoperability issues. Vendors, for example, will often reject new prefixes until they update their [=EPUB
```

Trying to be closer to the original text...

iherman (approved):
We are lucky that the English language has so many different words expressing _almost_ the same thing (may – might, should – have to) :-)

epub34/rs/index.html
mattgarrish: Ya, when I was going through the a11y docs I realized I forgot to check for recommended in these two. I was only looking for must, should, and may. Planning to do another pass later today as there are a few of these to write out. It's a synonym for "should" so I think "advised" also works well for these and makes the guidance more consistent.

epub34/authoring/index.html
mattgarrish: Vendor rejection isn't what I would call an interop problem, though. There is a disconnect between the claim and the example. The usual problem was exactly what the example cites, which is that vendors would reject epubs with new reserved prefixes because their versions of epubcheck were out of date and gave warnings.

If we want to keep the claim that it causes interop issues, we should either find a broader way to state the problem that includes supply chain problems or cite an example of a problem that reading systems have had with reserved prefixes. (I honestly can't think of a case off the top of my head, or if there was, it hasn't been an issue in years since we haven't added new reserved prefixes in a long time.)

epub34/authoring/index.html
mattgarrish: This is the problem we ran into last revision by initially making our privacy and security section informative and then scattering all kinds of normative sounding keywords throughout them. I believe this was the compromise to make some of these normative even though, as you say, they can't be tested.

epub34/authoring/index.html
mattgarrish: Does it even need saying that these are optional in the introduction? It's just generally explaining that they exist and how to use them. I'd just take the word out.

mattgarrish:
```suggestion
						the corresponding audio clip and adds the <a href="#attrdef-smil-clipBegin"
```

epub34/rs/index.html
mattgarrish: I think "better to avoid" makes it more permissive than it is. That would be guidance for a "should not". We could say:
```suggestion
						<p>For example, the reference is not allowed to begin with a [=URL-scheme string=] followed by a colon.
```

epub34/authoring/index.html
mattgarrish: Needs an "of" to be grammatically correct
```suggestion
						<p>Although it is common practice to specify the use of a spread in certain device orientations, the
							content itself does not represent a true spread &#8212; two consecutive pages that have to
```

epub34/authoring/vocab/meta-property.html
```diff
@@ -147,9 +147,9 @@ <h5>belongs-to-collection</h5>
 							itself a member of another collection.</p>
 						<p>To allow [=reading systems=] to organize collections and avoid naming collisions (e.g.,
 							unrelated collections might share a similar name, or different editions of a
-							collection could be released), [=EPUB creators=] SHOULD provide an identifier that uniquely
-							identifies the instance of the collection. The <code>dcterms:identifier</code>
-							property must carry this identifier.</p>
+							collection could be released), an identifier that uniquely
+							identifies the instance of the collection SHOULD be provided using a 
+							<code>dcterms:identifier</code> property.</p>
```
sueneu: There is a long parenthetical in the middle of the advice here. That can impact clarity. We could avoid this if we reorder the paragraph. 

> A unique identifier SHOULD be provided for each instance of a collection. This will allow [=reading systems=] to organize collections and avoid naming collisions. (e.g. unrelated collections might share a similar name, or different editions of a collection could be released).

epub34/authoring/index.html
mattgarrish: The behaviour continues into the present, so "has been" sounds more accurate here. "had been" makes it sound like the assignment of an origin happened but now there's a new post-assignment state.

epub34/rs/index.html
mattgarrish: I've updated these references in the last commit. There are a small handful that I think are fine to leave in lowercase as they're being used as adjectives that can't be substituted (but don't make normative requirements) or describing support levels in another specification. These can't be confused with requirements.

epub34/authoring/vocab/meta-property.html
mattgarrish: Fixed in the last commit.

epub34/rs/index.html
iherman: Sounds good to me.

epub34/authoring/index.html
iherman: I yield to your English knowledge :-)

epub34/authoring/index.html
iherman: I think we should (a) ignore this issue for the current PR and (2) open another issue. This approach may open up some attacks against the spec imho.

epub34/authoring/index.html
iherman: I do not have a strong opinion on the issue, and I personally can live with the change. I was just noting that the change goes slightly beyond the editorial change of the PR.

It may be a bit of an overkill to raise this as a separate issue, though. It would be good to get another approval for the change before we go ahead, though. @reidmore-online ? @bduga ?

mattgarrish: How about as a compromise we strip the text about where the issues might occur and leave the example to explain, so only say:

> 
```suggestion
						<p>Although reserved prefixes are an authoring convenience, they can cause issues. Vendors, for example, will often reject new prefixes until they update their [=EPUB
```

If someone knows of another example, we can always add it.

epub34/authoring/index.html
mattgarrish: > (a) ignore this issue for the current PR and (2) open another issue

Ya, this goes far beyond the scope of what we're trying to solve here.

epub34/authoring/index.html
iherman: Works for me. Committing.

mattgarrish:
How long do we want to keep these two pull requests open for review? There are maybe a couple of open PRs that will need to be merged and resolved, but until these are merged it poses a problem for creating any future PRs.

iherman:
Raise the question on our meeting that starts in 3 minutes :-)

mattgarrish:
Since we didn't get to this on the call, how about we leave them open until end of day tomorrow and then merge?

PR #2810, opened 2025-10-02
Restructure optimized publications section

### LINKED ISSUE #2802: Create a separate note for optimized publications
### Section

https://www.w3.org/TR/epub-a11y-111/#sec-optimized-pubs

### Describe the problem

Optimized publications aren't in the scope of the epub accessibility document but yet we have a whole informative section on them.

I think we added the section at least in part to avoid any offence that republishers targeting specific user groups are making inaccessible content because they don't conform to the standard, but we could do that without making it a section of a normative document.

Should we consider moving this section to its own note and reference it from the accessibility specification?

We could also choose to make it an appendix to get it out of the body where it doesn't belong, but that still leaves it as an informative aside in a normative document.

mattgarrish:
As we discussed on the TF call today, this pull request breaks the optimized publications section in two.

The text that explains how publications that don't meet wcag conformance level but are targeted to specific user needs are still accessible is now included under section 3 on accessible publications. I've named that subsection "Targeted accessibility" to avoid the perception that we're defining optimizations.

The rest of the text that recommends adding accessibility and discoverability metadata based on this document is now in an appendix called "guidelines for optimization standards".

The old section was informative and the new sections are informative so I haven't listed this in the change log as a substantive change. I'm open to listing it anyway if it might confuse people what happened to the old section.

One question: do we want to keep all three examples showing theoretical ways of describing conformance or maybe just keep the one using the link to the daisy audio guidelines since that's rooted in a real standard?

When we figure out hosting for a new page detailing optimization standards on the daisy site we can update the link. I've left an editor's note for now. I'll also update the existing registry document to redirect to the new page.

Fixes #2802 

***

EPUB Accessibility 1.2:
* [Preview](https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2802/epub34/a11y/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2216/epub34/a11y/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/a11y/issue-2802/epub34/a11y/index.html)

### DIFF
--- a/epub34/a11y/index.html
+++ b/epub34/a11y/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en-US" xmlns="http://www.w3.org/1999/xhtml">
+<html lang="en-US">
 	<head>
 		<meta charset="utf-8" />
 		<title>EPUB Accessibility 1.2</title>
--- a/epub34/a11y/index.html
+++ b/epub34/a11y/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en-US">
+<html lang="en-US" xmlns="http://www.w3.org/1999/xhtml">
 	<head>
 		<meta charset="utf-8" />
 		<title>EPUB Accessibility 1.2</title>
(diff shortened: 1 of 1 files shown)


clapierre (approved):
looks great Matt.

mattgarrish:
If there's no other feedback on this change, I'll merge it tomorrow.

GeorgeKerscher (approved):

Issue #2807, opened 2025-10-02
Retiring SHA-1 in favor of SHA-256 for font obfuscation


iherman:
### Section

https://www.w3.org/TR/epub-34/#sec-font-obfuscation

### Describe the problem

The [Font obfuscation](https://www.w3.org/TR/epub-34/#sec-font-obfuscation) algorithm starts by creating an obfuscation key. This key is, essentially, the hash of the publication's unique ID. The specification refers to SHA-1 as the hashing algorithm to be used. However, SHA-1 is, by now, considered to be outdated, insecure, vulnerable to hash collision attacks, see, e.g., [the relevant Wikipedia page](https://en.wikipedia.org/wiki/Secure_Hash_Algorithms). The current alternative is to use a member of the SHA-2 family (SHA-256, SHA-384, or SHA-512).

We may say we do not really care because we do not claim to be very secure when performing obfuscation. I.e., SHA-1 works for us. Which is correct, but the practical problem is that NIST has [officially retired SHA1](https://www.nist.gov/news-events/news/2022/12/nist-retires-sha-1-cryptographic-algorithm):

> "NIST is announcing that SHA-1 should be phased out by Dec. 31, 2030, in favor of the more secure SHA-2 and SHA-3 groups of algorithms.

This means that some organizations may not be allowed to use SHA-1 after that date. I.e., we cannot be idle on this.

### Describe the fix or new feature you propose

My proposal is to:

- declare the usage of SHA-1 deprecated, and replaced by SHA-256 (this is the most widely used version of SHA-2; other, more secure versions like SHA-384 or SHA-512 are used for highly security-sensitive applications, which we are not)
- reading systems are required to understand SHA-1 for backward compatibility reasons.

Note that the details of the obfuscation algorithm's will have to be modified, too: the current wording relies on the fact that the output of SHA-1, i.e., the obfuscation key, has exactly 20 bytes as opposed to SHA-256, which produces 32 bytes.

mattgarrish:
> because we do not claim to be very secure when performing obfuscation

Ya, all any thief has to do is read the algorithm section and it tells them exactly how to reverse the encryption. That's not going to change no matter what algorithm we use.

> This means that some organizations may not be allowed to use SHA-1

That's the better argument for a change, imo.

sueneu:
> This means that some organizations may not be allowed to use SHA-1

[...]

PR #2814, opened 2025-10-07
Add section for obsolete but conforming features

mattgarrish:
We don't have to use this, but it's easier to see what we'd be getting into by mocking up what a new section would look like with the legacy features and font obfuscation.

(If this is okay, we'd still need to update the RS specification.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2814.html" title="Last updated on Oct 13, 2025, 12:25 AM UTC (763db48)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2814/660f70d...763db48.html" title="Last updated on Oct 13, 2025, 12:25 AM UTC (763db48)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -3148,8 +3148,7 @@
 &lt;/package&gt;</pre>
 					</aside>
 
-					<p>Allowed on: [^collection^], <a href="#sec-opf-dcmes">Dublin Core elements</a>, [^meta^], and
-						[^package^].</p>
+					<p>Allowed on: <a href="#sec-opf-dcmes">Dublin Core elements</a>, [^meta^], and [^package^].</p>
 				</section>
 
 
@@ -3190,8 +3189,9 @@
 						<pre>&lt;dc:title id="pub-title"&gt;The Lord of the Rings&lt;/dc:title&gt;</pre>
 					</aside>
 
-					<p>Allowed on: [^collection^], <a href="#sec-opf-dcmes">Dublin Core elements</a>, [^item^],
-						[^itemref^], [^link^], [^manifest^], [^meta^], [^package^], and [^spine^].</p>
+					<p>Allowed on: <a href="#sec-collection-elem"><code>collection</code></a>, <a href="#sec-opf-dcmes"
+							>Dublin Core elements</a>, [^item^], [^itemref^], [^link^], [^manifest^], [^meta^],
+						[^package^], and [^spine^].</p>
 				</section>
 
 				<section id="attrdef-media-type">
@@ -3322,7 +3322,7 @@
 &lt;/package&gt;</pre>
 					</aside>
 
-					<p>Allowed on: [^collection^], <a href="#sec-opf-dcmes">Dublin Core elements</a>, [^meta^], and
+					<p>Allowed on: <a href="#sec-collection-elem"><code>collection</code></a>, <a href="#sec-opf-dcmes">Dublin Core elements</a>, [^meta^], and
 						[^package^].</p>
 				</section>
 			</section>
@@ -3429,7 +3429,8 @@
 								</p>
 							</li>
 							<li>
-								<p> [^collection^] <code>[0 or more]</code>
+								<p> <a href="#sec-collection-elem"><code>collection</code></a> <code>[0 or more]</code>
+									<span class="legacy">([=obsolete but conforming=])</span>
 								</p>
 							</li>
 						</ul>
@@ -9435,8 +9436,8 @@ html.my-document-playing * {
 						</ul>
 					</dd>
 
-					<dt>Collections</dt>
-					<dd id="sec-pkg-collections">
+					<dt id="sec-pkg-collections">Collections</dt>
+					<dd>
 						<ul>
 							<li id="sec-collection-elem">
 								<a data-cite="epub-33#sec-collection-elem"><code>collection</code> element</a>
@@ -9450,7 +9451,7 @@ html.my-document-playing * {
 						</div>
 					</dd>
 
-					<dt>Legacy features</dt>
+					<dt id="sec-pkg-legacy">Legacy features</dt>
 					<dd>
 						<ul>
 							<li id="sec-opf2-meta">
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en-US" xmlns="http://www.w3.org/1999/xhtml">
+<html lang="en-US">
 	<head>
 		<meta charset="utf-8" />
 		<title>EPUB 3.4</title>
@@ -76,6 +76,10 @@
 		<style>
 			pre {
 				white-space: break-spaces !important;
+			}
+			#app-unsupported dt {
+				margin-top: 2rem;
+				margin-bottom: 1rem;
 			}</style>
 	</head>
 	<body>
@@ -1160,7 +1164,7 @@
 								<code>application/x-dtbncx+xml</code>
 							</td>
 							<td> [[opf-201]] </td>
-							<td>The <a href="#obsolete-but-conforming">obsolete but conforming</a> NCX</td>
+							<td>The [=obsolete but conforming=] NCX</td>
 						</tr>
 						<tr>
 							<td id="cmt-smil">
@@ -1669,8 +1673,8 @@
 					archive: the "physical container". The rules for ZIP physical containers build upon the ZIP
 					technologies used by [[odf]].</p>
 
-				<p>OCF also retains an <a href="#sec-font-obfuscation">obsolete but conforming algorithm for obfuscating
-						embedded fonts</a> but this functionality is no longer advised.</p>
+				<p>OCF also retains an [=obsolete but conforming=] algorithm for <a href="#sec-font-obfuscation"
+						>obfuscating embedded fonts</a> but this functionality is no longer advised.</p>
 			</section>
 
 			<section id="sec-container-abstract">
@@ -1709,8 +1713,8 @@
 						</dt>
 						<dd>
 							<p>Contains information about the encryption of [=publication resources=]. This file is
-								mandatory when using the <a href="#sec-font-obfuscation">obsolete but conforming font
-									obfuscation feature</a>.</p>
+								mandatory when using the [=obsolete but conforming=] <a href="#sec-font-obfuscation"
+									>font obfuscation feature</a>.</p>
 						</dd>
 
 						<dt>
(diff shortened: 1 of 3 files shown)


iherman (approved):
Thanks for this @mattgarrish, it is clearer now what you propose.

I am fine with it, actually. It separates things from the core spec, and that is a good thing, let us concentrate on what really is part of the recommendation. I would say let is go for it.

I realize this is only a mock-up, but I have a comment on the text itself. There is a difference in the way we structure the individual features between §A2 and §A3. The latter is more feature oriented, whereas the former refers to the section that originally hosted the feature. I must admit, it took me a minute to realize that the font obfuscation part is under the "OCF" heading. I think the approach taken in §A3 is better.

I also wonder whether it should be appendix A, ie, the first one. A number of appendices are actually "positive" ones, ie, adding features of making things more detailed (structural semantics, external identifiers, etc). I think putting this "negative" one is better at the end, right before the examples. (I realize the order is the same in this PR as in the current spec, though.). But this is really minor.

mattgarrish:
> There is a difference in the way we structure the individual features between §A2 and §A3.

Ya, I kind of stripped down the text for A.2 but didn't spend a lot of time on it. I was mostly trying to get the structure down. I also have a bunch of errors because I linked the uses of "obsolete but conforming" to the section instead of the dfn. I'll give it another pass to try and improve it.


> I also wonder whether it should be appendix A, ie, the first one.

I wonder if we can better arrange all the appendixes. The viewport meta definition is squashed in after the prefixed properties, for example, but the prefixed properties have lost most of their usefulness at this point. The schemas can probably go right before the IANA registrations as both are technical details.

If we order by most used features, it probably should go:

A. Expressing structural semantics
B. Vocabularies
C. The viewport meta tag
D. Allowed external identifiers
E. Prefixed CSS properties
F. Unsupported features
G. Detailed examples
H. Schemas
I. Media type registrations

But I could also see putting the viewport meta tag definition and the allowed external identifiers as A. and B. so that they aren't pushed down below all the vocabularies.

iherman:
> 
> If we order by most used features, it probably should go:
> 
> A. Expressing structural semantics
> B. Vocabularies 
> C. The viewport meta tag 
> D. Allowed external identifiers 
> E. Prefixed CSS properties 
> F. Unsupported features 
> G. Detailed examples
> H. Schemas 
> I. Media type registrations
> 
> But I could also see putting the viewport meta tag definition and the allowed external identifiers as A. and B. so that they aren't pushed down below all the vocabularies.

Yeah, the viewport tag might be the most important one with all these fixed layout thingies, not sure how widely structural semantics is used in practice. So pushing C as the first makes probably sense. Otherwise the first four could really be in any order. 

Isn't prefixed css properties the same case as font obfuscation? Can't it be just a subsection of F?

mattgarrish:
> Yeah, the viewport tag might be the most important one with all these fixed layout thingies

But the appendix is just a copy of a definition that we probably shouldn't even have.  I doubt anyone reads past the actual section on setting the dimensions to read an ebnf for the syntax. I kind of like it buried the more I think about it...

> Isn't prefixed css properties the same case as font obfuscation? Can't it be just a subsection of F?

Ya, I think it also fits in the category of obsolete but conforming. We allow them but we'd prefer they go away.

iherman:
This was discussed during the [pmwg meeting on 09 October 2025](https://w3c.github.io/pm-wg/minutes/2025-10-09.html#0c3f).

<h3>Font Obfuscation - <a href="https://github.com/w3c/epub-specs/issues/2807" rel="noopener noreferrer">w3c/<wbr>epub-specs#2807</a></h3>
<p><cite>&lt;ivan&gt;</cite> <a href="https://github.com/w3c/epub-specs/pull/2814" rel="noopener noreferrer">w3c/<wbr>epub-specs#2814</a></p>
<p><cite>ivan:</cite> I started this discussion and it took an unexpected turn<br>
<span>… I was reviewing another PR and realized that the obfuscation section uses a SHA1 for hashing</span><br>
<span>… I realized that SHA is one of the algorithm that is considered out of date</span><br>
<span>… because of security issues and lack of unique hashes</span><br>
<span>… [1] will retire SHA1 soon which leads to governments and other orgs</span><br>
<span>… not being able to use it</span><br>
<span>… I proposed that we update the version of SHA</span><br>
<span>… Brady saw this as an opportunity to retire obfuscation from the spec</span><br>
<span>… not 100% because we have to be backward compatible</span><br>
<span>… we can add text saying that people should avoid using obfuscation</span><br>
<span>… and use WOFF fonts</span><br>
<span>… in the discussion there was agreement that this is the way to go</span><br>
<span>… mattgarish has already made a PR</span><br>
<span>… that includes other obsolete but still usable section</span><br>
<span>… these are not specifically deprecated</span><br>
<span>… we ask that people not use these technologies but reading systems with continue to understand them</span><br>
<span>… do we want to remove obfuscation in 3.4, and include the new section?</span><br>
<span>… it makes sense to move older features to a separate section</span><br>
<span>… it cleans up the spec, we don't have to keep all the font obfuscation specs for ever for example</span></p>
<p><cite>duga:</cite> I thinks its better to use WOFF fonts than to rewrite the obfuscation spec<br>
<span>… the reality is that SHA1 isn't broken, there is no maintenance required</span><br>
<span>… just some won't use it</span></p>
<p>'…it is not the end of the world if we choose to do nothing<br>
<span>… it is still used. So it is a little weird if one of the major tools, InDesign, still uses obfuscation</span><br>
<span>… it still makes sense to move it into older features</span></p>
<p><cite>&lt;wendyreid&gt;</cite> s/'.../...</p>
<p><cite>SueNeu:</cite> Can confirm, Adobe Indesign still produces font obfuscation</p>
<p><cite>ivan:</cite> leonard did chime in on the thread and said he didn't see any reason we couldn't go ahead with WOFF fonts<br>
<span>… perhaps this is an incentive for InDesign to change</span></p>
<p><cite>wendreid:</cite> are we all in agreement that we are deprecating font obfuscation?</p>
<p><cite>duga:</cite> is deprecate the right term?</p>
<p><cite>mattgarrish:</cite> obsolete but conforming is what is used now</p>
<p><cite>ivan:</cite> the question we also have to answer is what do we expect epub check to do if it sees obfuscation?</p>
<p><cite>mattgarrish:</cite> right now, we expect it to say nothing as long as it is used correctly</p>
<p>Proposed: Make font obfuscation an obsolete but conforming EPUB feature in EPUB 3.4.</p>
<p><cite>&lt;CharlesL1&gt;</cite> +1</p>
<p><cite>&lt;SueNeu&gt;</cite> s/mattgarrish /mgarrish</p>
<p><strong>RESOLUTION:</strong> Make font obfuscation an obsolete but conforming EPUB feature in EPUB 3.4.</p>
<p><cite>wendyreid:</cite> resolved</p>
<p><cite>CharlesL:</cite> do we have anything that lists all the features that are OK but can't really be used anymore from EPUB check?<br>
<span>… can it give a publisher a list of things they shouldn't be doing the future</span></p>
<p><cite>rdetour:</cite> ordinarily we issue a note that they should not use these features</p>

<hr /></details>

mattgarrish:
The collection element probably belongs in here, too.

mattgarrish:
I'll do a separate PR on top of this one to rearrange sections so it doesn't interfere with reviewing this change.

mattgarrish:
My last concern is the title now that we've added the obsolete features, specifically since font obfuscation is generally supported and technically we want xml:dir to be supported.

I can't think of a better name, though. "Unsupported and unmaintained features" gets a little closer but still misses the mark on xml:dir.

Maybe we could avoid trying to describe all the cases by naming it "Feature support exceptions"?

sueneu:
>Feature support exceptions

Someone may think this means that RS don’t support those features
maybe something like this?

>**Obsolete Features**
These features are no longer in common use. The specifications are not maintained or updated. Reading systems will support obsolete features to ensure that older titles are displayed correctly. However, new EPUBs should refrain from including the following.

mattgarrish:
Calling them all obsolete features has the same problem. If we split off the obsolete but conforming and deprecated sections, and renamed deprecated to "obsolete and non-conforming", then that would work as a heading. But we also have "under-implemented" features in there which defines xml:dir.

xml:dir not an obsolete feature but one that was requested for the package document metadata during 3.3 but couldn't be shown to have multiple implementations due to its late addition. Rather than drop it, we had to shuffle it off to the appendix and warn about it to keep it in the specification.

I guess that begs the question of whether we can show two implementations of directionality in reading system metadata display so we could remove the label and subsection. (I expect that's still unlikely.) Or can we live with two appendixes to keep obsolete and under-implemented from overlapping, since under-implemented should ideally go away?

sueneu:
Totally agree we shouldn’t mix the two. 
I’m not sure what under-implemented means here, and why we are calling this out in the spec? How is this determined? 

Are there other under-implemented features that we can or plan to call out? If so, it makes sense to have its own section. 

Otherwise, could we keep xml:dir in the regular spec but put a warning that it is under-implemented?

mattgarrish:
> I’m not sure what under-implemented means here, and why we are calling this out in the spec?

It means it lacks the two or more implementations necessary for W3C process to be kept in a specification. The 3.3 revision was the first time that the EPUB 3 spec required proof of implementation so we were concerned that there would be other features that had to be kept but couldn't meet this requirement. That's where the label came from.

It's sort of like a one-time grandfathering of features because epub 3 had already existed for over a decade before going through w3c standardization, but oddly the only feature that ended up in that bucket was the brand new xml:dir.

Since it only applies to one attribute, and will never be used again, it does seem odd to give it its own label. But I don't know if it will cause issues trying to get back to REC if it doesn't have a clear under-implemented label explaining why it's being allowed. It also might be problematic to try and define what we mean by labelling it under-implemented in the same section where we define the attribute.

This is where I lean on @iherman to tell us which approach will work best for w3c.

iherman:
(To make things clearer to @sueneu: per the W3C process, if a new feature is added to the Recommendation, but does not get at least two independent implementations, that features should be dropped from the final specification. This would have been very awkward for EPUB 3.3, because we were (and are) supposed to be strictly backward compatible. We were therefore forced to bend the rules a bit.)

> Since it only applies to one attribute, and will never be used again, it does seem odd to give it its own label. But I don't know if it will cause issues trying to get back to REC if it doesn't have a clear under-implemented label explaining why it's being allowed. It also might be problematic to try and define what we mean by labelling it under-implemented in the same section where we define the attribute.

I think that 

- We could/should remove that section from Appendix A. As the text says, we do not intend to use it again, it is over the top to have this as a separate section...
- I'm tempted to say that we should remove the reference from the title of [§5.3.1 The dir attribute (under-implemented)](https://www.w3.org/TR/epub-34/#attrdef-dir). The introductory note in the section makes the situation clear. On the other hand, I am a bit worried that W3C process people would make a fuss about it...

    B.t.w. [The EPUB 3.3 test results](https://w3c.github.io/epub-tests/history/epub33/results#sec-consolidated-internationalization-results) says that there _is_ one implementation for dir (Colibrio). We should talk to the Readium implementers whether they plan to implement this feature. If so, the whole issue become moot.

CC @llemeurfr, @hadriengardeur @gautierchomel

> This is where I lean on @iherman to tell us which approach will work best for w3c.

😊

mattgarrish:
I took out the under-implemented label and section and instead integrated some of the prose into a new note at the beginning of the definition for the dir attribute. Let me know if you think that will work.

I also renamed the appendix to "obsolete features" but have kept deprecated unchanged. Changing that label yet again is probably a bad idea and it's not fully accurate to say they aren't conforming when they are still allowed with only warnings.

iherman:
> I took out the under-implemented label and section and instead integrated some of the prose into a new note at the beginning of the definition for the dir attribute. Let me know if you think that will work.
> 
> I also renamed the appendix to "obsolete features" but have kept deprecated unchanged. Changing that label yet again is probably a bad idea and it's not fully accurate to say they aren't conforming when they are still allowed with only warnings.

This all works for me really well, @mattgarrish.

If Thorium (or another RS implements `dir`) then we can modify slightly the opening note for its definition section, although, I think, the bulk of the text will remain valid (I do not expect massive implementation of `dir`, alas!). But we can cross that bridge if and when we get there.

As far as I am concerned, we can merge this PR, and close the relevant issue...

PR #2818, opened 2025-10-14
Reorganization of appendixes

mattgarrish:
This pull request started from the discussion in https://github.com/w3c/epub-specs/pull/2814#pullrequestreview-3313313766 about where the obsolete features belong in the list of appendixes but I ended up trying to fix a couple of lost sections buried in the appendixes at the same time:

- The "vocabulary association mechanisms" section was buried under the vocabularies appendix even though it's central to the package document. We did this because it's also applies (to a much lesser extent in practice) to the epub:type attribute, but squirreling it away in an appendix means you end up reading the package document section without really understanding anything up front about prefix attribute, default vocabularies, etc. You have to follow links to make your way back to the appendix.
- The "expressing structural semantics" appendix is similar in that it covers epub:type for content documents, but again, because it can be used in media overlays it ended up in an appendix even though the use is the same for each.

To fix these issues, I moved vocabulary association mechanisms under the package document section and added a note to the introduction to highlight that it also applies to structural semantics. And for the type attribute I moved it under the shared technologies in the section on epub content documents (again adding a note to highlight that applies to media overlays). It's not perfect, but at least it gets the definitions into the body and as close to their primary use cases as possible.

The rest of the pull request is just a minor shuffling of the remaining appendixes:

A. Detailed examples
B. Allowed external identifiers
C. The viewport meta tag
D. Vocabularies
E. Obsolete features
F. Schemas
G. Media type registrations

I moved the examples to the front of the list as they get lost and are out of place when squashed between technical details.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2818.html" title="Last updated on Oct 15, 2025, 2:27 PM UTC (abf692e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2818/7d45572...abf692e.html" title="Last updated on Oct 15, 2025, 2:27 PM UTC (abf692e)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -11,7 +11,7 @@
 		<script src="../../common/js/add-caution-hd.js" class="remove"></script>
 		<script src="../../common/js/data-test-display.js" class="remove"></script>
 		<script src="../../common/js/copyright.js" class="remove"></script>
-		<script class="remove"> //<![CDATA[
+		<script class="remove">
             var respecConfig = {
                 group: "pm",
                 wgPublicList: "public-pm-wg",
@@ -70,7 +70,7 @@
 						]
 					}
 				] */
-            };//]]>
+            };
 		</script>
 		<style>
 			pre {
@@ -78,8 +78,7 @@
 			}
 			#app-unsupported dt {
 				margin-top: 2rem;
-				margin-bottom: 
-				1rem;
+				margin-bottom: 1rem;
 			}</style>
 	</head>
 	<body>
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -9394,1205 +9394,6 @@ html.my-document-playing * {
 					privacy and security.</p>
 			</section>
 		</section>
-		<section id="app-obsolete" class="appendix">
-			<h2>Obsolete features</h2>
-
-			<section id="sec-obs-conform">
-				<h3>Obsolete but conforming features</h3>
-
-				<p>An <dfn id="obsolete-but-conforming">obsolete but conforming</dfn> feature is one that is not
-					deprecated but that is also either not designed for use in EPUB 3 [=reading systems=] or that would
-					ideally be [=deprecated=] except that it would invalidate a significant base of existing [=EPUB
-					publications=].</p>
-
-				<p>[=EPUB publications=] MAY include the obsolete but conforming features defined in this section. Their
-					usage MUST conform to their referenced definitions.</p>
-
-				<dl>
-					<dt>Open Container Format (OCF)</dt>
-					<dd>
-						<ul>
-							<li id="sec-font-obfuscation">
-								<p><a data-cite="epub-33#sec-font-obfuscation">Font obfuscation</a> [[epub-33]]</p>
-								<div class="caution">
-									<p>NIST is advising that use of the SHA-1 algorithm [[fips-180-4]] be phased out by
-										the end of 2030. The Publishing Maintenance Working Group does not intend to
-										support font obfuscation in EPUB publications past that date due to its reliance
-										on SHA-1, although reading systems will have to continue to support
-										deobfuscation for existing EPUB publications.</p>
-									<p>Better methods of protecting fonts exist. Both [[woff]] and [[woff2]] fonts, for
-										example, allow the embedding of licensing information and provide some
-										protection through font table compression. The use of remotely hosted fonts also
-										allows for font subsetting.</p>
-								</div>
-							</li>
-						</ul>
-					</dd>
-
-					<dt id="sec-pkg-collections">Collections</dt>
-					<dd>
-						<ul>
-							<li id="sec-collection-elem">
-								<a data-cite="epub-33#sec-collection-elem"><code>collection</code> element</a>
-								[[epub-33]] </li>
-						</ul>
-						<div class="caution">
-							<p>When EPUB 3 was maintained by IDPF, a number of specifications were developed that relied
-								on the <code>collection</code> element. Due to a lack of adoption and implementation in
-								reading systems, these specifications are no longer maintained and use of the element is
-								no longer advised.</p>
-						</div>
-					</dd>
-
-					<dt id="sec-pkg-legacy">Legacy features</dt>
-					<dd>
-						<ul>
-							<li id="sec-opf2-meta">
-								<p><a href="https://idpf.org/epub/20/spec/OPF_2.0.1_draft.htm#Section2.2">OPF 2
-											<code>meta</code> element</a> [[opf-201]]</p>
-							</li>
-
-							<li id="sec-opf2-guide">
-								<p><a href="https://idpf.org/epub/20/spec/OPF_2.0.1_draft.htm#Section2.6">OPF 2
-											<code>guide</code> element</a>&#160;[[opf-201]]</p>
-							</li>
-
-							<li id="sec-opf2-ncx">
-								<p><a href="https://idpf.org/epub/20/spec/OPF_2.0.1_draft.htm#Section2.4.1">OPF 2
-										NCX</a> [[opf-201]]</p>
-							</li>
-						</ul>
-						<div class="note">
-							<p>EPUB 3 reading systems ignore these features. They are replaced by the [^meta^] element,
-								the <a href="#sec-nav-landmarks">landmarks nav</a>, and the <a href="#sec-nav-toc">toc
-									nav</a>, respectively.</p>
-							<p>The features are retained only to provide a measure of backwards compatibility with
-								reading systems that only support EPUB 2. As such reading systems are now rare,
-								including these features has limited value.</p>
-						</div>
-					</dd>
-
-					<dt id="css-prefixes">Prefixed CSS properties</dt>
-					<dd>
-						<ul>
-							<li id="sec-css-prefixed-writing-modes-text-orientation">
-								<a data-cite="epub-33#sec-css-prefixed-writing-modes-text-orientation"
-										><code>-epub-text-orientation</code></a> [[epub-33]]</li>
-							<li id="sec-css-prefixed-writing-modes-writing-mode">
-								<a data-cite="epub-33#sec-css-prefixed-writing-modes-writing-mode"
-										><code>-epub-writing-mode</code></a> [[epub-33]] </li>
-							<li id="sec-css-prefixed-writing-modes-text-combine">
-								<a data-cite="epub-33#sec-css-prefixed-writing-modes-text-combine"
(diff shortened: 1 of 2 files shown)


epub34/authoring/index.html
```diff
@@ -3109,854 +3108,921 @@ <h3>Parsing URLs in the package document</h3>
...
-							Consequently, it is strongly encouraged to set the proper directionality of text values in
-							the package document to ensure proper rendering once this situation improves.</p>
-						<p>The lessening of requirements for this attribute was only done to account for the different
-							process under which EPUB was developed prior to being brought into W3C. New features
-							developed under W3C processes will not receive a similar exemption.</p>
-					</div>
+					<p>EPUB defines a method of referencing terms and properties defined in metadata and semantic
+						vocabularies using the <a href="#sec-property-datatype"><var>property</var> data type</a>. The
+							<code>property</code> and <code>rel</code> attributes use the data type, for example, to
+						define properties and relationships in the [=package document=].</p>
```
iherman, line 3120: Just wondering... I would think that adding `epub:type` directly into this list (with a parenthetical note saying that `epub:type` is not used in the package document) is better than having a separate note at the end of the section. As it is now, it suggests that `epub:type` usage is, sort of, an afterthought or secondary feature (which is not).

epub34/authoring/index.html
```diff
@@ -3109,854 +3108,921 @@ <h3>Parsing URLs in the package document</h3>
...
-						<p>Although it lacks the necessary reading system support, it is integral to the content model
-							on which EPUB is built (i.e., for internationalization support in the [=package document=]).
-							Consequently, it is strongly encouraged to set the proper directionality of text values in
-							the package document to ensure proper rendering once this situation improves.</p>
-						<p>The lessening of requirements for this attribute was only done to account for the different
-							process under which EPUB was developed prior to being brought into W3C. New features
-							developed under W3C processes will not receive a similar exemption.</p>
-					</div>
+					<p>EPUB defines a method of referencing terms and properties defined in metadata and semantic
+						vocabularies using the <a href="#sec-property-datatype"><var>property</var> data type</a>. The
```
iherman, line 3118:
```suggestion
						vocabularies using the <a href="#sec-property-datatype"><var>property</var> data type</a>. The
```

Rereading all that tells me that the "property data type" is not the optimal term. We say that a "property data type" is used in the property attribute, but also in other attributes, which makes it confusing to me. Would it be possible to change the "property data type" to something else?

It could be `shortened URL`, `abbreviated URL`... The [RDF Turtle spec](https://www.w3.org/TR/rdf12-turtle/#dfn-prefixed-name) uses the term `prefixed name` which we could reuse after all, or use `prefixed value` for what it is. I am sure there are other alternatives.

This is a more complex editorial change if you agree with me... sorry about that!

epub34/authoring/index.html
```diff
@@ -3109,854 +3108,921 @@ <h3>Parsing URLs in the package document</h3>
...
 
-					<p
-						data-tests="#pkg-dir_creator-rtl,#pkg-dir_rtl-root-ltr,#pkg-dir_rtl-root-unset,#pkg-dir_unset-root-rtl,#pkg-dir_unset-root-unset,#pkg-dir-auto_root-rtl"
-						>Specifies the <a data-cite="bidi#BD5">base direction</a> [[bidi]] of the textual content and
-						attribute values of the carrying element and its descendants.</p>
+					<p>A <var>property</var> value is like a CURIE [[rdfa-core]] — it represents a URL [[url]] in
+						compact form. The expression consists of a prefix and a reference, where the prefix — whether
+						literal or implied — is a shorthand mapping of a URL that typically resolves to a term
+						vocabulary. When a [=reading system=] converts the prefix to its URL representation and combines
+						with the reference, the resulting URL normally resolves to a fragment within that vocabulary
```
```suggestion
						vocabulary. When a [=reading system=] converts the prefix to its URL representation and concatenates
						with the reference, the resulting URL normally resolves to a fragment within that vocabulary
```

I'm just worried that the reader would believe that we are talking about a combination in the URL sense. There is a difference: if my prefix is `http://a.b.c` and the referent is `D`, a URL join would yield `http://a.b.c/D` where a CURIE resolution would yield `http://a.b.cD`. Hence, the importance of using a `/`, `#`, or a `?` character...

iherman (approved):
I could not put this comment directly into the changes, GitHub did not let me. Looking at §5.7.2.1 should probably make it explicit that the `properties` uses the `property data type` (or whatever the final name is) as the datatype for its values. It does say in a roundabout way (it talks about the default vocabulary, and it also to the necessary section in the last sentence of the section), but it sounds cleaner if it is explicitly part of the definition of the attributes.

epub34/authoring/index.html
iherman: B.t.w., if we go down the HTML lane, then it should probably refer to `epub-type` as well. Although... it might become a bit awkward to use a mechanism which very much smells xml in a non-xml environment (where would we put the `prefix` definition, and how would we define it for HTML? Maybe something like `epub-prefix`).

But, depending on how the HTML discussion goes, this remark may be moot.

mattgarrish: > it might become a bit awkward to use a mechanism which very much smells xml in a non-xml environment

In other words, we never should have hacked up our own version of RDFa? 😉 

We don't have epub-prefix so epub-type is limited to the structural semantics vocabulary. I'd be surprised if more than a few people have ever used vocabulary association for structural semantics. What other vocabulary is there but the slightly more expansive DAISY structural semantics vocabulary that the EPUB version was culled from? It doesn't make a strong case for adding another epub- prefixed html attribute. If anything, the utility of prefixes and additional vocabularies in xhtml could be questioned.

epub34/authoring/index.html
mattgarrish: It was supposed to be a CURIE data type, but that would have been picking sides in the semantic wars between RDFa and microdata. We landed on property because it was primarily for expressing metadata properties in the package document.

Do you think we could just use CURIE now? The ebnf for the syntax was lifted from RDFa core, although we've since changed the reference to IRIs to use the URL spec.

epub34/authoring/index.html
iherman: In practice, this means that epub:type making use of vocabulary extension would be restricted to XHTML. I can live with that; as you say, it would concern very few (if any) users. But then this must be made explicit in the spec.

epub34/authoring/index.html
iherman: That war will never be over 😀… But both RDFa and microdata are hardly ever used, I am afraid, so it does not matter too much (using json-ld data embedded in the header of the html file won over both). This also means that I do not think people would particularly care, so we could use CURIE without too many problems. 

The downside is that I do not believe anybody will ever update the RDFa Core standard, so if some bugs come up, we are stuck. We may be better off by using either the term CURIE or "compact uri" in our spec, but not touching our own ebnf definitions.

epub34/authoring/index.html
sueneu: I’m having trouble understanding the Vocabulary association mechanisms section. 
Can we simplify the text? If I’m having trouble, others might also. Particularly if English is a second language. Something like this
>EPUB defines a formal method of referencing terms and properties defined in metadata and semantic vocabularies using the [property data type](https://w3c.github.io/epub-specs/epub34/authoring/#sec-property-datatype). The epub:type attribute uses this data type in [EPUB content documents](https://w3c.github.io/epub-specs/epub34/authoring/#dfn-epub-content-document) and [media overlay documents](https://w3c.github.io/epub-specs/epub34/authoring/#dfn-media-overlay-document) to add [structural semantics](https://w3c.github.io/epub-specs/epub34/authoring/#app-structural-semantics), for example, while the property and relattributes use the data type to define properties and relationships in the [package document](https://w3c.github.io/epub-specs/epub34/authoring/#dfn-package-document).

Could be written to include just one example and state that other uses are possible.
>Here are formal ways to reference external vocabularies used in metadata attributes. These are in addition to definitions for EPUB-specific attribute vocabularies. 

>The [property data type](https://w3c.github.io/epub-specs/epub34/authoring/#sec-property-datatype) references the metadata vocabulary in use. For example, you can reference Dublin Core metadata when identifying an author. You would use the property data type to point reading systems to that vocabulary. 

>The epub:type attribute can be used in [EPUB content documents](https://w3c.github.io/epub-specs/epub34/authoring/#dfn-epub-content-document) and [media overlay documents](https://w3c.github.io/epub-specs/epub34/authoring/#dfn-media-overlay-document) to add [structural semantics](https://w3c.github.io/epub-specs/epub34/authoring/#app-structural-semantics). The property and rel attributes use the data type to define properties and relationships in the [package document](https://w3c.github.io/epub-specs/epub34/authoring/#dfn-package-document).

epub34/authoring/index.html
mattgarrish: > We may be better off by using either the term CURIE or "compact uri" in our spec

Is that like generic v brand name? 😄 

I'd be fine with just calling it a compact url. (cool kids don't use IRIs!) I think somewhere we mention compatibility with rdfa and curies, so that gives us a little distance to justify our own definition.

epub34/authoring/index.html
mattgarrish: Do we still have a "formal" in there? I'm sure I wrote out one instance just because it was bugging me when I looked at an introduction. What would an informal method even be doing in a specification?

The most confusing thing about CURIEs is that they look a lot like xml namespace declarations, so people expect to use xmlns:abc="" to declare them. And most reading systems don't even bother expanding curies. They just string match against the reserved prefixes we've defined.

But the CURIE itself is a compact way of writing a URL. The prefix is usually a reference to a dereferenceable vocabulary, but could just be a unique identifier in the form of a URL so you can use your own custom terms. You have to expand that prefix in order to get the base URL and the value after is the specific property or term or semantic being used. Fun stuff... if you're a semantic web geek!

Let me see what I can do with the intro, though.

sueneu: >Do we still have a "formal" in there? 

About the word “formal”— does it have a specific meaning in this context? Is it synonymous with “official” or “defined”? Or like “hey, it's supposed to be this way”?

*edited to put in the quote mark

mattgarrish: > does it have a specific meaning in this context?

Nope. It's just excess verbiage. Everything we define in the standard is formal and there is no possibility of informal methods in the package document.

epub34/authoring/index.html
mattgarrish: To avoid overloading this pull request, I'm going to merge the current rearrange and open separate PRs to rename the property data type and tweak the introduction.

iherman: I just wanted to propose that:-)

PR #2821, opened 2025-10-16
Rename property data type to compact URL

### LINKED ISSUE #2824: Rename property data type
### Section

https://www.w3.org/TR/epub-34/#sec-property-datatype

### Describe the problem

Opening an issue for the change in https://github.com/w3c/epub-specs/pull/2821 so that we have a record of it in the list referenced from the change log.

First requested in https://github.com/w3c/epub-specs/pull/2818#discussion_r2432038395 :

> Rereading all that tells me that the "property data type" is not the optimal term. We say that a "property data type" is used in the property attribute, but also in other attributes, which makes it confusing to me. Would it be possible to change the "property data type" to something else?

> It could be shortened URL, abbreviated URL... The [RDF Turtle spec](https://www.w3.org/TR/rdf12-turtle/#dfn-prefixed-name) uses the term prefixed name which we could reuse after all, or use prefixed value for what it is. I am sure there are other alternatives.

Rename the data type to "Compact URL".

mattgarrish:
I got stuck on the name. I had it for a time as `compactURL`, then `compact url`, and ultimately went with plain old "compact URL" since most uses don't benefit from a more classic data type name. I'm easily convinced to go back to one of the earlier iterations if we want it to stand out more from short urls and the like, though.

I've also updated the intro to the vocabulary association mechanisms section since I started having to rewrite it anyway because it used the old term.

Fixes #2824 

***

Reading Systems: [Preview](https://cdn.statically.io/gh/w3c/epub-specs/editorial/rename-property-datatype/epub34/rs/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/editorial/rename-property-datatype/epub34/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2821.html" title="Last updated on Oct 20, 2025, 12:31 PM UTC (804ed04)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2821/fdee4e1...804ed04.html" title="Last updated on Oct 20, 2025, 12:31 PM UTC (804ed04)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -3145,7 +3145,7 @@
 						<pre>&lt;item href="images/cover.jpg" properties="cover-image" media-type="image/jpeg" /&gt;</pre>
 					</aside>
 
-					<p>When a [=reading system=] converts a compact URL to a full URL, the result normally resolves to a
+					<p>When a [=reading system=] converts a compact URL to a full URL, the result usually resolves to a
 						fragment within that vocabulary that contains human- and/or machine-readable information about
 						the term, but this is not required. Prefixes are ultimately just unique identifiers, similar to
 						how XML namespaces [[xml-names]] work.</p>
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en-US" xmlns="http://www.w3.org/1999/xhtml">
+<html lang="en-US">
 	<head>
 		<meta charset="utf-8" />
 		<title>EPUB 3.4</title>
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en-US">
+<html lang="en-US" xmlns="http://www.w3.org/1999/xhtml">
 	<head>
 		<meta charset="utf-8" />
 		<title>EPUB 3.4</title>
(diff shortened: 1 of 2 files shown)


epub34/authoring/index.html
```diff
@@ -3116,44 +3116,69 @@ <h3>Vocabulary association mechanisms</h3>
...
+
+						<p>In this example, the compact URL in the <code>properties</code> attribute is taken from its
+							default vocabulary so only the reference (<code>cover-image</code>) is present:</p>
+
+						<pre>&lt;item href="images/cover.jpg" properties="cover-image" media-type="image/jpeg" /&gt;</pre>
+					</aside>
+
+					<p>When a [=reading system=] converts a compact URL to a full URL, the result normally resolves to a
+						fragment within that vocabulary that contains human- and/or machine-readable information about
+						the term, but this is not required. Prefixes are ultimately just unique identifiers, similar to
```
```suggestion
					<p>When a [=reading system=] converts a compact URL to a full URL, the result usually resolves to a
						fragment within that vocabulary that contains human- and/or machine-readable information about
						the term, but this is not required. Prefixes are ultimately just unique identifiers, similar to
```

"normally" suggests (for me) that other usages are "abnormal". This isn't the case; there has been a long-standing battle whether terms should be fragments or URL paths. The best known (for us) case for the latter is dcterms, which is surely not "abnormal"...

iherman (approved):
I am perfectly fine with this naming. Thank you!

epub34/authoring/index.html
iherman: Alternatively, we could also say "normally resolves to fragment or a path within that vocabulary". In that case "normally" is fine: it is possible to end the prefix with a '?', for example, or without any of the '#', '/', or '?' characters, but that would indeed be "abnormal".

mattgarrish: Prefixes have to be valid to xsd:anyURI, so what you get when you combine that with the reference can't really be predicted. I think "usually" or "typically" is fine here, since all we're saying is that the most common result is a URL that points directly to the definition of the term or property being used. For an intro, we should probably avoid getting too deep into the weeds.

iherman: Ok. Let us leave it at "usually"!

PR #2831, opened 2025-11-10
deprecated the rendition:orientation property

### LINKED ISSUE #2751: Deprecating controls over orientation

iherman:
This implements the [F2F Resolution on 10th of November](https://w3c.github.io/pm-wg/minutes/2025-11-10-f2f.html#3397). In particular:

- In the authoring spec
    - the corresponding section has been removed, including the subsection on override properties
    - the item has been added to §E.2, referring back to the EPUB 3.3 as a reference definition of the spec
    - change log has been updated
- In RS spec
    - the corresponding section has been removed

See:

* For EPUB 3.4:
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/deprecate-orientation-control-issue-2751/epub34/authoring/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/authoring/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/deprecate-orientation-control-issue-2751/epub34/authoring/index.html)
    * [Preview](https://cdn.statically.io/gh/w3c/epub-specs/deprecate-orientation-control-issue-2751/epub34/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/deprecate-orientation-control-issue-2751/epub34/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2831.html" title="Last updated on Nov 11, 2025, 7:26 AM UTC (7838ee0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2831/9f042dc...7838ee0.html" title="Last updated on Nov 11, 2025, 7:26 AM UTC (7838ee0)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -7421,87 +7421,6 @@ No Entry</pre>
 						</section>
 					</section>
 
-					<section id="orientation" data-epubcheck="true"
-						data-tests="https://w3c.github.io/epub-structural-tests/#08-layout_layout.feature_L74,https://w3c.github.io/epub-structural-tests/#08-layout_layout.feature_L79,https://w3c.github.io/epub-structural-tests/#08-layout_layout.feature_L86,https://w3c.github.io/epub-structural-tests/#08-layout_layout.feature_L93">
-						<h5>Orientation</h5>
-
-						<p>The <code>rendition:orientation</code> property specifies which orientation the [=EPUB
-							publication=] is intended to be rendered in. </p>
-
-						<p id="property-orientation-global">When the <a href="#orientation"
-									><code>rendition:orientation</code> property</a> is specified on a [^meta^] element,
-							it indicates that the intended orientation applies globally (i.e., for all [=EPUB spine |
-							spine=] items).</p>
-
-						<p>One of the following values MUST be used with the <code>rendition:orientation</code>
-							property:</p>
-
-						<dl class="variablelist">
-							<dt>landscape</dt>
-							<dd>
-								<p>Render the content in landscape orientation.</p>
-							</dd>
-
-							<dt>portrait</dt>
-							<dd>
-								<p>Render the content in portrait orientation.</p>
-							</dd>
-
-							<dt>auto</dt>
-							<dd>
-								<p>The content is not orientation constrained. Default value.</p>
-							</dd>
-						</dl>
-
-						<p id="fxl-orientation-duplication">The <code>rendition:orientation</code> property MUST NOT be
-							declared more than once. In addition, it MUST NOT be declared using the <a
-								href="#attrdef-refines"><code>refines</code> attribute</a>. Refer to <a
-								href="#orientation-overrides"></a> for setting the property for individual [=EPUB
-							content documents=].</p>
-
-						<aside class="example" id="fxl-ex2" title="Specifying global landscape orientation">
-							<p>In this example, items in the spine are to be rendered in landscape mode.</p>
-
-							<pre>&lt;package …&gt;
-   &lt;metadata …&gt;
-      …
-      &lt;meta
-          property="rendition:layout"&gt;
-         pre-paginated
-      &lt;/meta&gt;
-
-      &lt;meta
-          property="rendition:orientation"&gt;
-         landscape
-      &lt;/meta&gt;
-   &lt;/metadata&gt;
-   …
-&lt;/package&gt;</pre>
-						</aside>
-
-						<section id="orientation-overrides" data-epubcheck="true"
-							data-tests="https://w3c.github.io/epub-structural-tests/#08-layout_layout.feature_L102,https://w3c.github.io/epub-structural-tests/#08-layout_layout.feature_L107">
-							<h6>Orientation overrides</h6>
-
-							<p id="property-orientation-local">The following properties MAY be specified on [=EPUB spine
-								| spine=] [^itemref^] elements to override the <a href="#property-orientation-global"
-									>global value</a>:</p>
-
-							<dl>
-								<dt id="orientation-auto">rendition:orientation-auto</dt>
-								<dd>The [=reading system=] determines the orientation to render the spine item in.</dd>
-
-								<dt id="orientation-landscape">rendition:orientation-landscape</dt>
-								<dd>Render the given spine item in landscape orientation.</dd>
-
-								<dt id="orientation-portrait">rendition:orientation-portrait</dt>
-								<dd>Render the given spine item in portrait orientation.</dd>
-							</dl>
-
-							<p>A spine item MUST NOT declare more than one of these overrides.</p>
-						</section>
-					</section>
-
 					<section id="spread" data-epubcheck="true"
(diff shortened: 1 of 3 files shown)


iherman:
The resolution says "update Reading Systems to ignore it." I am not sure what to do here: the RS spec simply refers to the authoring spec for all deprecated features, without listing them or telling them what to do if those features are met. I see no reason for this property to be an outlier in this respect... @mattgarrish ?

mattgarrish:
> I see no reason for this property to be an outlier in this respect...

Unfortunately, I don't think it's that easy. To date, we've deprecated stuff without saying that reading systems have to stop supporting it. We've left it to developers to decide if it's worth the effort, and have kept the processing sections for font obfuscation and epub prefixed css properties, for example, though these also refer to back to the rs 3.3 specification for details.

This change is contradictory if we're saying that reading systems must stop supporting rendition:flow. We can't say that they may support deprecated features and also that they must not support this one.

We might be able to work around it using SHOULDs, but then we'd still need to keep the processing section and also have it point back to 3.3. For example:

> Reading systems SHOULD ignore the [rendition:flow property and its overrides](https://www.w3.org/TR/epub-33/#flow) [epub-33] when specified in the package document. Reading systems SHOULD render overflow content in reflowable EPUB content documents using their default overflow handling method or a user preference.
> 
> If a reading system supports the rendition:flow property and its overrides, refer to [The `rendition:flow` property](https://www.w3.org/TR/epub-rs-33/#flow) [epub-rs-33] for the processing requirements.

If it's only a SHOULD to ignore the property, then the MAY support statement isn't so problematic. But if we're requiring all reading systems to ignore the property and use their own default going forward, then we'd have to look at changing the appendix wording to something like:

> Reading systems MAY support deprecated authoring features [epub-34] except where specified otherwise in this specification.

That would give more wiggle room to say in the rendition:flow section that they must not support the property and overrides anymore. But it still means retaining the section; I don't think we can get around that.

Do you know how strictly we want to stop support?

mattgarrish:
Heh, I just noticed I was responding to the wrong property. Not sure where I thought I saw rendition:flow being removed...

But I think the issue remains the same. We have processing instructions for rendition:orientation that we have to do something about if reading systems can continue to support the property. If the only instruction is that rendition:orientation can be supported as in the authoring spec, then the requirements on the rs get lost (even if they're pretty simplistic in this case).

mattgarrish:
Thinking out loud now, but a third option might be to list the deprecated reading system features first and include rendition:orientation in it. And then follow that by saying that reading systems may support deprecated authoring features except where their processing is also deprecated... ?

iherman:
Answering your question:

> Do you know how strictly we want to stop support?

There was a brief discussion to say that the RS should always behave as if the property was set to "auto". Which, I believe, essentially says "ignore the property and decide it for yourself" based, e.g., on the physical orientation of the device. But that is, I believe, obvious, so not sure how to put it as a requirement, so, during the discussion, the WG just said to drop it.

It feels to me as if this was the general approach towards all deprecated features, so maybe we can say, in that section, that RS should ignore those properties and, where applicable, behave according to what an "auto" value defines. After all, "auto" is present in many of these properties, it is not special to `rendition:orientation`...

mattgarrish:
> behave according to what an "auto" value defines

Ya, "auto" was created specifically so that all reading systems were compliant even if they didn't support fxl. It's basically a default to do whatever you do. The problem is you can't mention to default to auto without the old context.

But maybe it's okay to just take the section out and list it as deprecated in the RS appendix without any other changes. We say that reading systems may support deprecated features but assuming some developer really wants to support this property for some reason, when they come back to the RS specification they're going to find that the processing behaviours are also deprecated. If they still want to go ahead with support, they can follow the reference back to the 3.3 definition.

It was having it listed as deprecated in the authoring spec but made to disappear from the rs that struck me as odd.

(Feel free to change "reading" to "Reading" in the first paragraph of the appendix, and also move that appendix after the RSO appendix to match the reorg we did in the authoring spec.)

iherman:
> (Feel free to change "reading" to "Reading" in the first paragraph of the appendix, and also move that appendix after the RSO appendix to match the reorg we did in the authoring spec.)

ack.

Done.

iherman (commented):
> It was having it listed as deprecated in the authoring spec but made to disappear from the rs that struck me as odd.

Maybe, before publishing, we can decide to make an explicit list; we shall see. If we do it now, we will probably forget some...

mattgarrish:
> Maybe, before publishing, we can decide to make an explicit list; we shall see.

Not sure what you mean here. There already are deprecated processing aspects listed. It's not just a referral across to the authoring specification.

If we're going to remove the whole section and take the "it never existed" approach rather than deprecate, the minimum we need to do is log that support for the property is gone. It's a significant change from 3.3.

iherman:
> > Maybe, before publishing, we can decide to make an explicit list; we shall see.
> 
> Not sure what you mean here. There already are deprecated processing aspects listed. It's not just a referral across to the authoring specification.

What I meant is that, in the current form, the text says:

> rendition:orientation and [orientation overrides](https://www.w3.org/TR/epub-33/#orientation-overrides) properties [[epub-33](http://localhost:8001/LocalData/github/EPUB/epub-specs/epub34/authoring/#bib-epub-33)]

and other, similar cases will come; at some point we may want to make an explicit listing of "orientation overrides properties". But, thinking it again, I believe it is all right as it is.
> 
> If we're going to remove the whole section and take the "it never existed" approach rather than deprecate, the minimum we need to do is log that support for the property is gone. It's a significant change from 3.3.

I believe the "never existed" approach is not an alternative due to the backward compatibility requirement.

iherman:
In view of the decision taken [at the F2F](https://w3c.github.io/pm-wg/minutes/2025-11-11-f2f.html#465b) a number of properties will have the same fate as the orientation property. We have also agreed that (1) @HadrienGardeur will make a complete listing of which properties should be deprecated and (2) we would extend the current PR to include those, rather than having separate PR-s. Ie, at this moment, the PR should not be merged. I will mark it as "draft" to avoid accidental GitHub actions...

mattgarrish:
Closing this as it was superseded by #2844

PR #2832, opened 2025-11-11
Remove references to html syntax

### LINKED ISSUE #2715: Adopt the HTML syntax in addition to the XML syntax for content documents

### LINKED ISSUE #2761: [HTML Survey] Adopt some way to signify that an EPUB uses HTML instead of XHTML
### Section

https://www.w3.org/TR/epub-34/#sec-package-elem

### Describe the problem

Several responses to the survey have made reference to or requested a way to differentiate EPUB files with HTML from those using XHTML. Different methods have been proposed, including: 

- Separate version number (i.e. EPUB4)
- Separate file extension (i.e. .epub4)
- Metadata to identify use of HTML 

(Chair hat off) I understand the desire to differentiate the books in some way, but I don't think this change rises to the level of a complete version number change like EPUB4, without us bundling in many more features and changes that would make it worth such an advancement. 

I do think having some way to easily identify these books is compelling, and we should discuss different ways to do it. Some options off the top of my head: 

- `<package version="3.4"...>`
- `<meta property="rendition:serialization"...>html</meta>`

mattgarrish:
This should undo all the changes we've made.

I left the note about xhtml not being 1.1 with the xhtml content document definition, as I think it's helpful to have it where people are most likely to land (with some minor adjustments).

I also kept the "html" heading in the core media types section as having xhtml content documents under the "other" heading never seemed quite right.

I'm marking it as final review since I assume we should give anyone not at tpac a week to catch up on the resolutions before merging.

***

Overview: [Preview](https://cdn.statically.io/gh/w3c/epub-specs/chore/revert-html-syntax/epub34/overview/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/overview/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://cdn.statically.io/gh/w3c/epub-specs/chore/revert-html-syntax/epub34/overview/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2832.html" title="Last updated on Nov 26, 2025, 2:13 PM UTC (7c42529)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2832/9f042dc...7c42529.html" title="Last updated on Nov 26, 2025, 2:13 PM UTC (7c42529)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -5858,7 +5858,7 @@ No Entry</pre>
 							the use of the XML syntax. XML, however, is integral to many publishing workflows and
 							expected in many vendor ingestion systems and reading systems. It is also not expected that
 							support for XHTML will be dropped from browsers, even if there are additional support
-							differences between the syntax in the future. Support for the HTML syntax will be addressed
+							differences between the syntax in the future. Support for the HTML syntax may be addressed
 							in a successor format to EPUB 3.</p>
 					</div>
 
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -3030,10 +3030,10 @@
 			<section id="sec-package-intro" class="informative">
 				<h3>Introduction</h3>
 
-				<p>The [=package document=] is an XML document that consists of a set of elements that each encapsulate
-					information about a particular aspect of an [=EPUB publication=]. These elements serve to centralize
-					metadata, detail the individual resources, and provide the reading order and other information
-					necessary for its rendering.</p>
+				<p>The [=package document=] is an <a href="#sec-xml-constraints">XML document</a> that consists of a set
+					of elements that each encapsulate information about a particular aspect of an [=EPUB publication=].
+					These elements serve to centralize metadata, detail the individual resources, and provide the
+					reading order and other information necessary for its rendering.</p>
 
 				<p>The following list summarizes the information found in the package document:</p>
 
@@ -5838,8 +5838,39 @@ No Entry</pre>
 				<section id="sec-xhtml-intro" class="informative">
 					<h4>Introduction</h4>
 
-					<p>This section defines a profile of [[html]] for creating [=XHTML content documents=]. An instance
-						of an XML document that conforms to this profile is a [=core media type resource=].</p>
+					<p>An [=XHTML content document=] is an instance of an <a href="#sec-xml-constraints">XML
+							document</a> expressed using the <a data-cite="html#the-xml-syntax">XML syntax</a> of the
+						[[[html]]] [[html]].</p>
+
+					<p>XHTML content documents that conform to this profile are [=core media type resources=] and one of
+						the two formats that constitute [=EPUB content documents=] &#8212; that can be used without
+						fallbacks in the [^spine^].</p>
+
+					<div class="note">
+						<p>EPUB 3 does not support the <a data-cite="html#the-html-syntax">HTML syntax</a> of [[html]]
+							as a core media type. Although both serializations allow the same elements to be expressed,
+							not all JavaScript libraries and APIs are designed to handle the XML syntax, in particular
+							the shadow DOM. This means, for example, that although the [^template^] element [[html]] can
+							be authored in the XML syntax it may not be usable in the same way as in the HTML
+							syntax.</p>
+
+						<p>The Publishing Maintenance Working Group is aware that the HTML Standard no longer recommends
+							the use of the XML syntax. XML, however, is integral to many publishing workflows and
+							expected in many vendor ingestion systems and reading systems. It is also not expected that
+							support for XHTML will be dropped from browsers, even if there are additional support
+							differences between the syntax in the future. Support for the HTML syntax will be addressed
+							in a successor format to EPUB 3.</p>
+					</div>
+
+					<div class="note">
+
+						<p>Earlier versions of EPUB 3 referred to W3C's HTML5 standard to define EPUB content documents,
+							but that standard was largely a snapshot of the HTML Standard maintained by the WHATWG.
+							After reaching an agreement on the development of HTML in 2018, the W3C standard was retired
+							and EPUB 3 now exclusively refers to the HTML Standard. It does not mean that EPUB 3 does
+							not support HTML5 anymore in XHTML content documents, only that HTML5 is now a marketing
+							term that refers to the HTML Standard as the successor to HTML 4.01 and XHTML 1.1.</p>
+					</div>
 				</section>
 
 				<section id="sec-xhtml-req" data-epubcheck="true"
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -226,25 +226,15 @@
 
 					<p>The [[html]] standard defines a single content model with rules for expressing a tag set using
 						either the <a data-cite="html#syntax">HTML syntax</a> or the <a
-							data-cite="html#the-xhtml-syntax">XML syntax</a>. EPUB 3 allows authoring of content
-						documents using either of these syntaxes even though the [[html]] standard <a
-							data-cite="html/xhtml.html#the-xhtml-syntax">no longer recommends the use of the XML
-							syntax</a>. The Working Group recognizes that XML remains an integral technology in the
-						publishing ecosystem and will not remove support for the XML syntax from EPUB 3. Regardless,
-						publishers that prefer to keep using the XML syntax will need to monitor future support for it,
-						and might have to adapt to the HTML syntax to gain access to some features of [[html]].</p>
+							data-cite="html#the-xhtml-syntax">XML syntax</a>. EPUB 3, however, only allows authoring of
+						content documents using the XML syntax, in what are referred to as [=XHTML content
+						documents=].</p>
 
-					<div class="issue" data-number="2715" title="Adoption of the HTML syntax">
-						<p>The change to allow both syntaxes of HTML is an open issue in EPUB 3.4. The above paragraph
-							was added to help explain the change, but would be amended if the HTML syntax is not
-							adopted.</p>
-					</div>
-
-					<p>The <a href="#sec-xhtml">HTML profile defined by this specification</a> inherits all definitions
+					<p>The <a href="#sec-xhtml">XHTML profile defined by this specification</a> inherits all definitions
 						of semantics, structure and processing behaviors from [[html]] unless otherwise specified.</p>
 
 					<p>In addition, this specification <a href="#sec-xhtml-extensions">defines a set of extensions</a>
-						to the [[html]] document model that can be included in [=HTML content documents=].</p>
+						to the [[html]] document model that can be included in XHTML content documents.</p>
 				</section>
 
 				<section id="sec-overview-relations-svg">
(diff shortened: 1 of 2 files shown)


iherman (approved):
My only comment: shouldn't we keep `epub-type` as an _alternative_ to `epub:type`? It is certainly forward-looking. It may be a load on RS-s but maybe not that big.

But maybe we should let sleeping dogs lie...

cc @llemeurfr @bduga

bduga:
> It is certainly forward-looking.

Who knows what the future will bring. Without a clear reason to add it I don't think we should, especially since we can't simply replace epub:type. Just my ¥3.

epub34/authoring/index.html
```diff
@@ -5864,40 +5833,20 @@ <h4>The <code>itemref</code> element</h4>
...
-				</div>
+				<h3>XHTML content documents</h3>
 
 				<section id="sec-xhtml-intro" class="informative">
 					<h4>Introduction</h4>
 
-					<p>This section defines a profile of [[html]] for creating HTML content documents. An instance of an
-						HTML document that conforms to this profile is a [=core media type resource=].</p>
+					<p>This section defines a profile of [[html]] for creating [=XHTML content documents=]. An instance
+						of an XML document that conforms to this profile is a [=core media type resource=].</p>
```
sueneu:
```suggestion
						of an XHTML document
```

epub34/authoring/index.html
```diff
@@ -6541,93 +6482,38 @@ <h4>Structural semantics</h4>
...
-
-							<p>To provide feedback on this change, please open issues in the <a
-									href="https://github.com/w3c/epub-specs/issues">GitHub tracker for the EPUB
-									specifications</a>.</p>
-						</div>
-
-
 						<div class="caution">
 							<p>Although the <code>epub:type</code> attribute is similar in nature to the [^/role^]
 								attribute&#160;[[html]], the attributes serve different purposes. The values of the
```
sueneu, line 6550:
```suggestion
								attribute [[html]],
```
removing the entity for non-breaking space, since it is not used elsewhere

epub34/overview/index.html
```diff
@@ -293,30 +290,20 @@ <h3>Navigation document</h3>
...
+				These are HTML or SVG documents that describe the readable content and reference associated media
+				resources (e.g., images, audio, and video clips).</p>
+
+			<p>HTML documents for EPUB are formally defined as a <a data-cite="epub-34#sec-xhtml">profile of
+				HTML</a> [[html]]. In particular, that profile requires the usage of the <a
+					data-cite="html#writing-xhtml-documents">XML syntax</a> [[html]], referred to as [=XHTML content
+				documents=] in the EPUB specifications.</p>
+
+			<div class="note">
+				<p>The term "XHTML" should not be confused with an earlier, fully XML based version of HTML denoted as
```
sueneu, line 303: This is a helpful clarification, thanks Matt!

epub34/overview/index.html
```diff
@@ -650,14 +637,11 @@ <h3>Notes</h3>
...
-						has an exemption under the applicable jurisdiction's laws.
-					</li>
-					<li>
-						[[[epub-fxl-a11y]]] [[epub-fxl-a11y]]: provides guidance on how to make fixed-layout EPUB publications accessible.	
-					</li>
+						has an exemption under the applicable jurisdiction's laws. </li>
+					<li> [[[epub-fxl-a11y]]] [[epub-fxl-a11y]]: provides guidance on how to make fixed-layout EPUB
+						publications accessible. </li>
 					<li>EPUB Multiple-Rendition Publications 1.1 [[epub-multi-rend-11]]: defines the creation and
 						rendering of EPUB publications consisting of more than one Rendition. </li>
```
sueneu, line 646: Can we flag this as “generally unsupported” or something similar? People might think they can make this kind of EPUB and expect it to render in an RS.

epub34/overview/index.html
```diff
@@ -777,11 +755,10 @@ <h3>EPUB 3.3: 2022</h3>
 			<section id="epub34" class="ednote">
 				<h3>EPUB 3.4: 2027</h3>
 
-				<p>The work on EPUB 3.4 [[epub-34]] started in 2025. The goal was to add some new features 
-					to EPUB 3.3, and to clarify some existing features. The new features include the ability to
-					add standard annotating features to EPUB publications, and the ability to include 
-					"webtoons" in EPUB publications. The new features are expected to be published as
-					 a W3C Recommendation in 2027.</p>
+				<p>The work on EPUB 3.4 [[epub-34]] started in 2025. The goal was to add some new features to EPUB 3.3,
```
sueneu, line 758:
```suggestion
				<p>The work on EPUB 3.4 [[epub-34]] started in 2025. The goal was to add new features to EPUB 3.3,
```
removed “some” if we feel a modifier is needed, can we be more specific? “community-requested” or “useful” or ?

mattgarrish:
If we add epub-type for xhtml contents documents, we also increase the burden on reading systems to support things like pop-up footnotes. You'd have to check for epub:type, epub-type, or role setting the semantics.

epub34/overview/index.html
iherman: Note that this section is more of a placeholder and, at some point, must be made more precise. At this point, we have also removed things (well deprecated things) and not only added...

iherman:
> If we add epub-type for xhtml contents documents, we also increase the burden on reading systems to support things like pop-up footnotes. You'd have to check for epub:type, epub-type, or role setting the semantics.

True. I hereby withdraw my proposal :-)

epub34/authoring/index.html
mattgarrish: I almost made this change before spotting that we had "xml" in the original text. I think it's better to leave it as xml document because that ties it in with the requirements for xml documents that are specified earlier (for example, the encoding requirement which isn't enforced by html). There's also the problem that html doesn't define instances of xhtml documents anymore, so it starts to sound like we're talking about xhtml 1.1.

It might be helpful to link xml document to the section on xml processing, though, to make clearer why we're calling it that here.

epub34/authoring/index.html
mattgarrish: Ya, the editor in me is torn on this. I know I never add them even though I feel bad for the poor orphans that result... :smile:

I just find it a nuisance to add the entities every time we have a reference, and I find it makes it harder to read with the respec shorthand on the reference, but do you feel strongly about having them @iherman? (I know you often add them.)

The looming problem if we try to standardize, of course, is that there's no consistency about whether the reference is more strongly tied to the text before or after it, so we can't just search and replace the problem away.

epub34/overview/index.html
mattgarrish: If we make a support statement, it's not only multiple renditions that should get one (the tts stuff we know also isn't supported). It might be better if you open an issue on the overview to address this, as it could lead us far astray of just undoing the html syntax additions.

(I didn't know you could get respec to add a linked spec name using triple brackets. Maybe we should fix that up in a separate PR, too. Learned something new...)

epub34/overview/index.html
mattgarrish: Right, I'd leave this for now. The overview usually gets a final pass review to be more specific once we know what's definitely in and out.

epub34/authoring/index.html
sueneu: Makes sense, thanks Matt

epub34/authoring/index.html
iherman: I usually add them, but through Unicode and not with entities. I have a setup whereby an option+space (on my mac) puts it in the text. I do not know whether this is something you can set up in your environment...

And yes, the orphan references at the beginning of a line are a thorn in my eyes :-)

mattgarrish:
I've added a couple of notes to the xhtml content document intro section in the last commit.

One is to try and be a bit more upfront about the divergences between the syntaxes, without scaring people too much, and that we don't intend to change from xml in epub 3.

The other tries to explain that "html5" didn't go away with the switch from w3c's html5 spec to the whatwg one, only that "html5" is just a marketing term now. Maybe that helps some small bit with the people who think we're talking about xhtml 1.1 because of our generic looking [html] references?

But I can be convinced we don't need either of these or that they need changing. The intro just always looks kind of bare.

epub34/authoring/index.html
mattgarrish: No, I don't have anything like that. I have to insert the character from a menu or copy it in from elsewhere.

Now that I think of it, I was finding these as nbsp entities and then globalling them out periodically to 160 back when we had the xhtml ns declared. oxygen kept flagging them all as errors which made it harder to find real errors.

There's probably no real harm in adding a non-breaking space before all the references, as worst case it just pulls an extra word down. But we should do that in a separate pull request.

reidmore-online: There's some new CSS around preventing orphans, I wonder if that's a better solution than adding non-breaking spaces everywhere, but I imagine that's a respec change?

reidmore-online (approved):

epub34/authoring/index.html
mattgarrish: I think this one's on us. I wasn't using orphan in the technical sense of the opening line of a paragraph being separated from the rest, just that the specification reference gets separated from what it refers to. For example:

```
                                   ... the `audio` element
[html]. ...
```

(Probably closer to a widow, I suppose, but still not something I think css can address.)

It may not be worth the bother. I skimmed probably a few hundred and only found a couple of spots where the reference ended up out of place.

epub34/authoring/index.html
```diff
@@ -5864,40 +5833,51 @@ <h4>The <code>itemref</code> element</h4>
...
+							the shadow DOM. This means, for example, that although the [^template^] element [[html]] can
+							be authored in the XML syntax it may not be usable in the same way as in the HTML
+							syntax.</p>
+
+						<p>The Publishing Maintenance Working Group is aware that the HTML Standard no longer recommends
+							the use of the XML syntax. XML, however, is integral to many publishing workflows and
+							expected in many vendor ingestion systems and reading systems. It is also not expected that
+							support for XHTML will be dropped from browsers, even if there are additional support
+							differences between the syntax in the future. Support for the HTML syntax will be addressed
+							in a successor format to EPUB 3.</p>
```
```suggestion
							differences between the syntax in the future. Support for the HTML syntax may be addressed
							in a successor format to EPUB 3.</p>
```

Let us not take a commitment for the change...

epub34/authoring/index.html
```diff
@@ -5864,40 +5833,51 @@ <h4>The <code>itemref</code> element</h4>
...
+					</div>
+
+					<div class="note">
+
+						<p>Earlier versions of EPUB 3 referred to W3C's HTML5 standard to define EPUB content documents,
+							but that standard was largely a snapshot of the HTML Standard maintained by the WHATWG.
+							After reaching an agreement on the development of HTML in 2018, the W3C standard was retired
```
```suggestion
							After reaching an <a href="https://www.w3.org/2019/04/WHATWG-W3C-MOU.html">agreement</a> on the development of HTML in 2019, the W3C standard was retired
```

I presume that is the agreement you referred to...

iherman (approved):
I have the impression that the other comments (use of nbsp, or changes in the overview) should be subject of separate PR-s. 

Any reason why we do not merge this?

iherman:
> But I can be convinced we don't need either of these or that they need changing. The intro just always looks kind of bare.

I think it is good to have these, also in view of the survey discussions. See my minor comments.

mattgarrish:
> Any reason why we do not merge this?

No, it kind of slipped my mind after tpac week, but I think we've had this open long enough now for any feedback.

Issue #2841, opened 2025-12-03
Mark rendition properties as obsolete


HadrienGardeur:
### Section

https://w3c.github.io/epub-specs/epub34/authoring/#sec-fxl-package

### Describe the problem

During the 2025 TPAC in Kobe, [we voted on the following resolution](https://w3c.github.io/pm-wg/minutes/2025-11-11-f2f.html#:~:text=RESOLUTION%3A%20EPUB%20should%20prioritize%20user%20reading%20preferences%20for%20content%20appearance%20and%20reading%20experience):

> RESOLUTION: EPUB should prioritize user reading preferences for content appearance and reading experience

I previously opened a number of related issues earlier this year:

- https://github.com/w3c/epub-specs/issues/2751
- https://github.com/w3c/epub-specs/issues/2754
- https://github.com/w3c/epub-specs/issues/2750
- https://github.com/w3c/epub-specs/issues/2753

@iherman also started working on a PR post-TPAC: https://github.com/w3c/epub-specs/pull/2831

### Describe the fix or new feature you propose

Based on this resolution, the following properties should be ~~deprecated~~ marked as obsolete in our next revision:

- [`rendition:orientation`](https://w3c.github.io/epub-specs/epub34/authoring/#orientation) and [the equivalent values for spine overrides](https://w3c.github.io/epub-specs/epub34/authoring/#layout-overrides)
- [`rendition:flow`](https://w3c.github.io/epub-specs/epub34/authoring/#flow) and [the equivalent values for spine overrides](https://w3c.github.io/epub-specs/epub34/authoring/#flow-overrides)
- [`rendition:spread`](https://w3c.github.io/epub-specs/epub34/authoring/#spread) and [the equivalent values for spine overrides](https://w3c.github.io/epub-specs/epub34/authoring/#spread-overrides)

sueneu:
If we deprecate `rendition:spread` is there a way to show fxl children’s picture books as a spread?

There are cases where looking at a children’s picture book page by page would be a poorer experience, since art is created to be seen across two pages.
(edited to fix a typo)

HadrienGardeur:
> If we deprecate `rendition:spread` is there a way to show fxl children’s picture books as a spread?

Authors can place resources in the spread using [spine overrides](https://w3c.github.io/epub-specs/epub34/authoring/#page-spread) but they shouldn't be able to force the display of a spread.

Displaying a spread can be a complex decision based on multiple criteria, including:

- screen size and ratio
[...]

PR #2842, opened 2025-12-06
Reorganized layout section

mattgarrish:
Opening this pull request to give an idea of what I'd like to do with the layout rendering control section. As I mentioned in #2788, I find what we have a confusing mess of property definitions in the body with poor explanations of how they go together for the layouts (like spreads and placement which aren't strictly for pre-paginated content but are lumped under fxl).

What I've done (so far) is:

- moved the layout section up ahead of content documents as, pedagogically, you're typically picking your layout before getting into the details of content documents
- add a layout types section to explain each of three main layout types
- moved all the property definition text back into appendix D.5 and stopped trying to categorize their uses (we can do that in the layout sections). It'll also help to not have one vocabulary that is a mapping table pointing into the body, and should help avoid an oddly empty section on layout control if we do indeed end up deprecating almost all of the properties
- kept synthetic spreads as a separate subsection from the layout types since it applies to both reflowable and pre-paginated content (which maybe gets pared down furhte
- moved the section on fixed layout documents and setting their dimensions into the content documents section.
- dropped the "roll document" stuff as it perpetuates a difference that doesn't exist at the content document level (roll and pre-paginated are different ways of presenting fixed layout documents)

I'm only opening this as a draft as it's not ready for review of the text yet, and I would suggest no bothering to read too deeply into what is there right now. I'll do a more serious edit of the sections if people agree with this approach.

(One aside in making these changes is that it made clearer to me what a mess scrolled-continuous is no matter what we do. It's not solely an overflow handling property but mixes scrolled-doc into a new type of layout that spans spine items (and is really odd as a spine override). It's too bad we couldn't define roll as either requiring fxl document or reflowable ones and deprecate scrolled-continuous (make reading systems recognize it as a roll layout). As it is, you'll get one kind of "roll" from scrolled-continuous and another kind from layout=roll.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2842.html" title="Last updated on Dec 7, 2025, 2:43 AM UTC (344a16c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2842/d65abd7...344a16c.html" title="Last updated on Dec 7, 2025, 2:43 AM UTC (344a16c)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -3,7 +3,7 @@
 	<head>
 		<meta charset="utf-8" />
 		<title>EPUB 3.4</title>
-		<!--<meta name="color-scheme" content="light dark" />-->
+		<meta name="color-scheme" content="light dark" />
 		<script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
 		<script src="./biblio.js" class="remove"></script>
 		<script src="../../common/js/fix-errata.js" class="remove"></script>
@@ -6131,7 +6131,7 @@ No Entry</pre>
 
 			<section id="sec-spread-placement" data-epubcheck="true"
 				data-tests="https://w3c.github.io/epub-structural-tests/#08-layout_layout.feature_L166,https://w3c.github.io/epub-structural-tests/#08-layout_layout.feature_L171">
-				<h3>Spread placement</h3>
+				<h3>Synthetic spreads</h3>
 
 				<p>When a [=reading system=] renders a [=synthetic spread=], the default behavior is to populate the
 					spread by rendering the next [=EPUB content document=] in the next available unpopulated
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -175,7 +175,7 @@
 
 				<p>EPUB publications by default are intended to reflow to fit the available screen space. It is also
 					possible to create publications that have pixel-precise fixed layouts using images and/or CSS
-					positioning. The metadata to control layouts are defined in <a href="#sec-rendering"></a>.</p>
+					positioning. The metadata to control layouts are defined in <a href="#sec-layout"></a>.</p>
 
 				<p>[=Media overlay documents=] complement EPUB content documents. They provide declarative markup for
 					synchronizing the text in EPUB content documents with prerecorded audio. The result is the ability
@@ -3730,7 +3730,7 @@
 						</li>
 						<li>
 							<p>to provide access to all rendering metadata needed to control the layout and display of
-								the content (e.g., <a href="#sec-fxl-package">fixed-layout properties</a>).</p>
+								the content (e.g., <a href="#app-rendering-vocab">package rendering properties</a>).</p>
 						</li>
 					</ol>
 
@@ -5828,7 +5828,7 @@ No Entry</pre>
 				</section>
 			</section>
 		</section>
-		<section id="sec-rendering">
+		<section id="sec-layout">
 			<h2>Layout rendering</h2>
 
 			<section id="sec-general-rendering-intro" class="informative">
@@ -5930,16 +5930,16 @@ No Entry</pre>
 				<section id="sec-pre-paginated">
 					<h4>Pre-paginated layouts</h4>
 
-					<p>In a pre-paginated layout, each [=fixed-layout document=] referenced a the [=EPUB spine | spine=]
-						[^itemref^] represents one page of content. Reading systems are expected to scale the document
-						to fit the viewport or synthetic spread.</p>
-
 					<div class="note">
 						<p>Pre-paginated layouts were commonly referred to as fixed layouts prior to the introduction of
 								<a href="#sec-roll">roll layouts</a>. As both layouts make use of [=fixed-layout
 							documents=], the term "fixed layout" is now used exclusively for describing EPUB content
 							documents with fixed dimensions, not for the layout type they appear in.</p>
 					</div>
+
+					<p>In a pre-paginated layout, each [=fixed-layout document=] referenced by a [=EPUB spine | spine=]
+						[^itemref^] represents one page of content. Reading systems are expected to scale the document
+						to fit the viewport or synthetic spread.</p>
 
 					<p>To declare that an EPUB publication uses a pre-paginated layout, the <a href="#layout"
 								><code>rendition:layout</code> property</a> MUST be declared in a <code>meta</code>
@@ -5984,8 +5984,9 @@ No Entry</pre>
 					</div>
 
 					<p id="fxl-layout-duplication" data-tests="#lay-fxl-layout-duplication">When the property is set to
-							<code>pre-paginated</code> for a spine item, its content dimensions MUST be set as defined
-						in <a href="#sec-content-dimensions"></a>.</p>
+							<code>pre-paginated</code> for a spine item, the [^itemref^] element MUST reference a <a
+							href="#sec-fxl-docs">fixed layout document</a> or there MUST be one in the [=manifest
+						fallback chain=].</p>
 
 					<p>The following properties from the <a href="#app-rendering-vocab">Package rendering vocabulary</a>
 						can be applied to pre-paginated spine items to further control their rendering:</p>
@@ -5997,7 +5998,7 @@ No Entry</pre>
 						<li>the <a href="#spread"><code>rendition:spread</code> property</a> and its <a
 								href="#spread-overrides">spine overrides</a> are used to indicate a preference for how
 							reading systems ...</li>
-						<li>the <a href="#page-spread"><code>rendition:page-spread-*</code> spine override
+						<li>the <a href="#spread-placement"><code>rendition:page-spread-*</code> spine override
 								properties</a> are used to indicate a preference for how to place spine items in a
 							synthetic spread (on the right or left side of a two-page spread, or centered without a
 							spread).</li>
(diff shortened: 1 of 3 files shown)


HadrienGardeur:
> moved the layout section up ahead of content documents as, pedagogically, you're typically picking your layout before getting into the details of content documents

I think that's a good approach.

> kept synthetic spreads as a separate subsection from the layout types since it applies to both reflowable and pre-paginated content

Is this really supported though? 

Support for spreads feels completely at odds with the concept of reflowable documents: you can't guarantee how many columns and screens worth of content you'll be displaying.
Spreads only work with FXL because each resource is contained to either full screen or half screen. You can't do that with reflowable or roll documents.

> dropped the "roll document" stuff as it perpetuates a difference that doesn't exist at the content document level (roll and pre-paginated are different ways of presenting fixed layout documents)

In practice, most roll documents produced in the near future will use images in spine with an HTML fallback (unclear if this will be a reflowable or an FXL resource, we don't have the ability to convey that info anyway until we change our approach for layout overrides).

That's the reason why I updated the example in my PR, to better align with what publishers producing continuously scrolled comics will be using.

> One aside in making these changes is that it made clearer to me what a mess scrolled-continuous is no matter what we do.

This is getting deprecated based on our decision at TPAC: https://github.com/w3c/epub-specs/issues/2841

iherman:
I really like the approach.

Practical comments:

- it seems that the branch you used for this new branch was prior to the version that removed the HTML serialization: the content document section still refers to the pending issue regarding HTML vs. XHTML. This may lead to merge conflicts down the line...
- There is a [(draft) PR to deprecate the orientation layout](https://github.com/w3c/epub-specs/pull/2831); my intention was to update that PR based on the issue [deprecating a number of additional layout related properties](https://github.com/w3c/epub-specs/issues/2841). Again, doing so would create major merge conflicts later. I wonder whether this PR could not deprecate those properties, making that PR obsolete.

mattgarrish:
> * it seems that the branch you used for this new branch was prior to the version that removed the HTML serialization

Ya, I branched off Hadrien's open pull request to get those changes. But what I've done can be boiled down to adding an entirely new section after the package document, moving the fxl doc section, and replacing the vocabulary in appendix D. I can easily reproduce that in a new branch based off the current master, which probably makes the most sense.

> * Again, doing so would create major merge conflicts later. I wonder whether this PR could not deprecate those properties, making that PR obsolete.

Ya, that's largely why I'm trying to rush this and opened it in an incomplete stage. But if you want to do it that way it's fine with me. I'd rather not waste any more time cleaning up the appendix D additions if we're just turning around and removing almost all the properties.

Are they being deprecated or made obsolete but conforming? If we do the former we invalidate existing content. It feels like they belong in the latter category along with the -epub prefixed CSS properties since they're generally harmless and not worth having to remediate content to remove.

mattgarrish:
> Is this really supported though?

Right, I wasn't trying to get into non-editorial decision making so we can move to merge this quickly. Spreads and placement predate FXL and the spread placement section explicitly says the properties apply to both reflowable and fxl.

> In practice, most roll documents produced in the near future will use images in spine with an HTML fallback

Right, but this feels like more of a technical difference between pre-paginated and roll layouts that we can explain in the roll layout section, not a new class of documents. But nothing is set in stone yet. As I said, this is not a finished product and I still need to put more thought into the layout descriptions. I could well come back around to it being simpler to define a term for fxl documents + images in spine, or maybe we should still consider discussing images in spine and getting their dimensions in the fxl document section. Even if they require fallbacks, saying nothing about them when we know they're used doesn't seem helpful.

iherman:
> > * it seems that the branch you used for this new branch was prior to the version that removed the HTML serialization
> 
> Ya, I branched off Hadrien's open pull request to get those changes. But what I've done can be boiled down to adding an entirely new section after the package document, moving the fxl doc section, and replacing the vocabulary in appendix D. I can easily reproduce that in a new branch based off the current master, which probably makes the most sense.
> 

I trust you on that.

> > * Again, doing so would create major merge conflicts later. I wonder whether this PR could not deprecate those properties, making that PR obsolete.
> 
> Ya, that's largely why I'm trying to rush this and opened it in an incomplete stage. But if you want to do it that way it's fine with me. I'd rather not waste any more time cleaning up the appendix D additions if we're just turning around and removing almost all the properties.

Exactly. That is why I propose to keep the #2831 pending and to be closed or completed later, depending on how this PR goes. I think it would make more sense to do the structural change and the deprecation in one step.

> 
> Are they being deprecated or made obsolete but conforming? If we do the former we invalidate existing content. It feels like they belong in the latter category along with the -epub prefixed CSS properties since they're generally harmless and not worth having to remediate content to remove.

I chose deprecation because that is what the WG resolution said, although that detail was not really discussed. What probably influenced me is that the current list of deprecated features also includes, e.g., `spread-portrait`, `rendition:spread` with value `portrait`, or `rendition:viewport`. What is the difference between those and the orientation setting? My understanding from the F2F was that the orientation setting was meant more as a "hint" rather than anything else, an RS-s ignore them...

But I have no problem moving it to the obsolete but conforming category.

mattgarrish:
> What is the difference between those and the orientation setting?

Those ones were deprecated because they were either somewhat nonsensical (portrait spreads) or were replaced by other metadata (rendition:viewport with in-document dimension settings). 

In this case, we're not replacing the metadata with anything, we're just completely scaling back all the rendering hints. It feels like a violation of our principle of not making old content invalid, especially when there's little reason.

I think the deprecated spread value could be dropped if we make all of rendition:spread obsolete but conforming, but the rendition:viewport metadata probably does need to be deprecated because it was not just a hint. At the time, it was the way to set the ICB globally allowing you to skip per-file declarations.

iherman:
> In this case, we're not replacing the metadata with anything, we're just completely scaling back all the rendering hints. It feels like a violation of our principle of not making old content invalid, especially when there's little reason.

You're right.

> I think the deprecated spread value could be dropped if we make all of rendition:spread obsolete but conforming, but the rendition:viewport metadata probably does need to be deprecated because it was not just a hint. At the time, it was the way to set the ICB globally allowing you to skip per-file declarations.

I do not have a strong feeling about this (ie, whether we simply remove the spread value or leave it as it is). I would say: let us not bother now, and leave it.

mattgarrish:
> I would say: let us not bother now, and leave it.

So where are we on deprecating properties? It sounds like the idea is to drop almost everything based on my reading of https://github.com/w3c/epub-specs/issues/2841, but this is still just a proposal, right? The only resolution I saw in the F2F minutes was to open an issue, so I shouldn't move ahead with anything more than removing the one property in the other issue yet?

HadrienGardeur:
> Right, I wasn't trying to get into non-editorial decision making so we can move to merge this quickly. Spreads and placement predate FXL and the spread placement section explicitly says the properties apply to both reflowable and fxl.

It has a major impact on how the current PR is organized and IMO this should be discussed in an upcoming call (I won't be able to attend this week unfortunately).
Do we even test support for spreads on reflowable files? I'm not even sure how this could be done.

In the current draft for EPUB 3.4 RS everything is under FXL (as it should be IMO): https://w3c.github.io/epub-specs/epub34/rs/#sec-fxl-props

> Are they being deprecated or made obsolete but conforming? If we do the former we invalidate existing content. It feels like they belong in the latter category along with the -epub prefixed CSS properties since they're generally harmless and not worth having to remediate content to remove.

Thanks for bringing this up @mattgarrish.

I would say that overall they should be obsolete based on this clarification.

iherman:
> So where are we on deprecating properties? It sounds like the idea is to drop almost everything based on my reading of #2841, but this is still just a proposal, right? The only resolution I saw in the F2F minutes was to open an issue, so I shouldn't move ahead with anything more than removing the one property in the other issue yet?

Right, we were a bit sloppy at the meeting; it was one of the last items of the two days and everyone was a bit exhausted. _My_ understanding is that all items in #2841 should be obsoleted (in the case we agree on this, in contrast to deprecated) but it would indeed be better to have that put into a formal resolution. I think we can do it this Thursday, although I understand that the edrlab people will not be around (but we know the vote of @HadrienGardeur 😀)

@sueneu @shiestyle @reidmore-online can we get this on the agenda of this week's call, just to make it a bit more formal? We can also have a formal decision to use "obsolete but conforming" as opposed to "deprecated" for all those, including the orientation one.

mattgarrish:
One last question before I close off this PR: is it required, recommended or optional that roll layouts be supported?

I'm just rearranging the reading system spec to match the authoring layout section and am not sure what support statement to attach to the roll section.

For comparison, I recall we started FXL off without a strict support requirement but later made it required to support fxl docs, which in the era of mixing pre-paginated and fxl terminology would seem to mean that pre-paginated layouts are required to be supported. (Is that really what we wanted, though, as it would seem to invalidate any reading system designed, for example, only for reflowable docs on small screens.)

iherman:
> One last question before I close off this PR: is it required, recommended or optional that roll layouts be supported?

My instinctive reaction is that 'roll' is aimed at a very special market, which may not be relevant for a number of specialized areas (eg, a specialized RS for higher education). I would therefore opt for "recommended". Depending on how the market reacts, we may make it required at a later stage.

As an aside: why closing off this PR?)

mattgarrish:
> As an aside: why closing off this PR?)

I've created a new branch off the current main and ported the changes here. There are 22 merge conflicts with this one that I don't want to spend time sorting through because this comes from an old branch. I'll open it shortly.

HadrienGardeur:
> My instinctive reaction is that 'roll' is aimed at a very special market, which may not be relevant for a number of specialized areas (eg, a specialized RS for higher education). I would therefore opt for "recommended". Depending on how the market reacts, we may make it required at a later stage.

I agree that this is the right approach for now.

Issue #2792, opened 2025-09-16
Using images in the spine for comics


HadrienGardeur:
### Describe the problem

Our current specification allows bitmaps in the spine as long as they also include a fallback to XHTML or SVG, but unfortunately that's not well-supported by reading systems.

We've seen them used anyway:

- a major publisher in Japan has produced EPUB files with bitmaps in the spine, without including a fallback
- while many others (not just in Japan) have used HTML in the spine, with a fallback to the bitmap (weird semantically, but an easy way to find the bitmap equivalent of a spine item)

From an author, publisher and reading system point of view, using bitmaps without the need for XHTML in spine would be the better option since many reading systems either use a special viewing mode for comics or only support comics (and cannot work with HTML).

While working on continuously scrolled comics, we see that once again, having an HTML container for these image tiles adds complexity and makes the distinction between FXL and this new layout less obvious than it should be.

This has been discussed in a recent PR and WG call at: https://github.com/w3c/epub-specs/pull/2788

### Describe the fix or new feature you propose

For continuously scrolled comics, we should use bitmaps in the spine with a fallback to an HTML document.

We MAY recommend authors to:

- group together images that belong to the same scroll/roll (tiles from the same episode) in a single reflowable HTML document and use it as a fallback for the first image of every scroll/roll
- and use another HTML document as a fallback for other images of that scroll/roll

It's also very tempting to address paginated comics as well. We could either:

- recommend bitmaps in spine with an HTML equivalent (limited use and support right now, aside from one major publisher)
- or HTML in spine with a bitmap fallback (wider spread use from publishers and support in RS)

iherman:
Just for my understanding: by "bitmap" you mean one of the image format listed as allowed media types in EPUB, i.e., JPEG, PNG, or GIF formats, right?

HadrienGardeur:
Yes.

It would be interesting to support AVIF though, it's clearly the future (and present on some websites already) of images on the Web.

iherman:
> It would be interesting to support AVIF though, it's clearly the future (and present on some websites already) of images on the Web.

That is a different issue: should we add AVIF to the list of accepted media types in general. Let us consider it at another time.

HadrienGardeur:
@iherman I've opened a separate issue: https://github.com/w3c/epub-specs/issues/2794

reidmore-online:
[...]

PR #2844, opened 2025-12-09
Reorganized layout section and properties

### LINKED ISSUE #2749: Deprecating layout overrides

### LINKED ISSUE #2750: Deprecating flow overrides
### Section

https://www.w3.org/TR/epub-33/#flow-overrides

### Describe the problem

Reflowable EPUB publications are flexible by design, allowing the user to decide how they would prefer reading them.

In addition to visual adjustment options, it also means that the user is always free to decide if they'd rather read their publications:

- paginated
- scrolled per resource
- scrolled through the entire publication

The spec allows content creators to provide a preference for this information per resource in the spine.

This is fundamentally at odds with the philosophy of reflowable EPUB publications and could lead to a very confusing UX if implemented.

Flow overrides should be deprecated in EPUB 3.4 and marked for removal in a future 3.5 revision.

### LINKED ISSUE #2751: Deprecating controls over orientation

### LINKED ISSUE #2753: Deprecating controls over the use of spreads
### Section

https://www.w3.org/TR/epub-33/#spread

### Describe the problem

While the ability to control spread placement is very useful, the same can't be said of the `spread` metadata and its spine overrides equivalents (`rendition:spread-auto`, `rendition:spread-both`, `rendition:spread-landscape`, `rendition:spread-none`, `rendition:spread-portrait`).

Many FXL publications use this metadata to either force the display of spreads at all times (`both`) or to never display them (`none`), with no good reason for doing so.

In practice, we see that many reading systems either:

- decide to display spreads based on device size and orientation, rather than what's indicated in this metadata
- or let the user decide what they'd like to see

Like most metadata available in the "Layout rendering control" section of the specification, this is fundamentally at odds with the philosophy of EPUB to allow such controls at all.

`portrait` and `rendition:spread-portrait` are already deprecated, but the remaining properties should be deprecated in EPUB 3.4 as well and marked for removal in a future 3.5 revision.

### LINKED ISSUE #2754: Deprecating flow as a metadata

### LINKED ISSUE #2841: Mark rendition properties as obsolete

### LINKED ISSUE #2847: Does align-x-center apply to both reflowable and fxl layouts?
### Section

https://www.w3.org/TR/epub-34/#align-x-center

### Describe the problem

The align-x-center property is currently classified under the reflowable layouts section of the authoring specification but in the reading system specification the processing requirements say:

> When the [rendition:align-x-center property](https://www.w3.org/TR/epub-rs-34/#align-x-center) is set on a spine item, reading systems SHOULD render the content centered horizontally within the [viewport](https://www.w3.org/TR/epub-34/#dfn-viewport) or spread, as applicable. This property does not affect the rendering of the spine item, only the placement of the resulting content box.
> 
> For reflowable content, reading systems that support this property MUST center each virtual page.

The second paragraph seems to suggest that the first one applies to both reflowable and pre-paginated layouts. Is that true or does it only apply to reflowable and the first paragraph is only about reflowable documents in spreads?

If it is true that the property applies to fxl documents, can you also apply this property to roll documents? (I would assume not since the roll display already has requirements on centering and aligning the documents, but just checking.)

mattgarrish:
This pull request is a continuation of #2842 but rebased on the current draft to avoid merge conflicts.

A brief recap from #2842:

> - moved the layout section up ahead of content documents
> - added a layout types section to explain each of three main layout types
> - moved all the property definition text back into appendix D.5
> - kept synthetic spreads as a separate subsection
> - moved the section on fixed layout documents and setting their dimensions into the content documents section.
> - dropped the "roll document"

New to this pull request:

- I've integrated the changes to remove rendition:orientation and its spine overrides from #2831 (adding it as obsolete but conforming and leaving rendition:viewport as deprecated)
- I've added a layout section to the reading systems specification and put the requirements I could find for each layout type together
- I added a couple of requirements for rolls based off the existing authoring spec prose and our discussions about no layout control properties applying to it

A bit more controversially, perhaps, I've noted that images in spine are a type of fixed-layout document even if they require fallbacks, and added additional caveats about vendor acceptance and accessibility in the existing note about fxl support. I also added a section to the authoring spec that the icb dimensions will be obtained from the image metadata and to refer to the image format definition for how that's done. I similarly stated for reading systems that if they opt to support image formats in the spine that's where the dimensions are obtained. That should be all we need to say so that we don't have to classify images in spine as separate from fxl documents for either pre-paginated or roll layouts.

Still to do before we look to merge is to resolve on the other properties and overrides we're deprecating so I can take them out without worrying that they'll have to be re-added. This will not only affect the vocabulary, but could mean making spread placement specific to pre-paginated layouts if we decide they aren't useful for reflowable documents.

Fixes #2841 
Fixes #2754 
Fixes #2753 
Fixes #2751 
Fixes #2750 
Fixes #2749 
Fixes #2847 

See:

* For EPUB 3.4:
    * [Preview](https://raw.githack.com/w3c/epub-specs/fix/reorg-layouts/epub34/authoring/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/authoring/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fix/reorg-layouts/epub34/authoring/index.html)
    * [Preview](https://raw.githack.com/w3c/epub-specs/fix/reorg-layouts/epub34/rs/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fix/reorg-layouts/epub34/rs/index.html)

(Added preview links for RS, too, IH)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2844.html" title="Last updated on Dec 18, 2025, 2:14 PM UTC (2445a58)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2844/ba46f10...2445a58.html" title="Last updated on Dec 18, 2025, 2:14 PM UTC (2445a58)">Diff</a>


HadrienGardeur:
As I've already said in https://github.com/w3c/epub-specs/pull/2842, this is a good step forward to make the spec easier to read, nice work @mattgarrish !

Looking at this PR, section "[6.3 Synthetic spreads](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2844.html#page-spread)" remains somehow problematic for me:

- I think that this should be under "6.2.2 Pre-paginated layout" since the concept of spreads only makes any sense for FXL (cf https://github.com/w3c/epub-specs/pull/2842#issuecomment-3627470597)
- while in Kobe, [we voted on a resolution](https://w3c.github.io/pm-wg/minutes/2025-11-10-f2f.html#1a21:~:text=RESOLUTION%3A%20Adapt%20spec%20language%20in%20Core%20and%20Reading%20Systems%20to%20say%20that%20if%20no%20information%20about%20placement%20in%20spreads%2C%20the%20first%20resource%20should%20be%20a%20page%20on%20its%20own.) to: "Adapt spec language in Core and Reading Systems to say that if no information about placement in spreads, the first resource should be a page on its own."

This second item was an action item for me but since we're re-working the whole section right now, I think we could tackle it in this PR as well.

sueneu:
>since the concept of spreads only makes any sense for FXL

Is there a current use case for spreads in reflowable?

If we limit spreads to FXL are we limiting future uses of something like Regions that allows for both fixed and reflowable elements? (and is that a problem we want to worry about right now?)

mattgarrish:
Don't we have spread placement for reflowable to handle the case of mixing pre-paginated documents or having "reflowable" SVGs in the spine? (what is a reflowable svg?)

The reading system won't know if you have a pre-paginated two-page layout in the middle of a chapter, so you need to hint whether to keep flowing the next document after the spread into the left-hand slot or add a blank page so that a new chapter, for example, starts in the right-hand slot.

The same is likely true when reflowable documents are added to pre-paginated layouts, even though support for this is poor.

HadrienGardeur:
> Don't we have spread placement for reflowable to handle the case of mixing pre-paginated documents or having "reflowable" SVGs in the spine? (what is a reflowable svg?)

If you mix pre-paginated and reflow together, they won't be displayed side by side in a spread since the concept of synthetic spread is completely foreign to reflowable documents.

I think that there's confusion about two separate concepts:
- when a reflowable document is paginated, this is usually done with columns where the RS treats each reflowable resource as a long horizontal roll where we move the viewport to view one or two columns at a time
- whereas a synthetic spread is usually handled using two webviews or iframes side by side in a container

> […] even though support for this is poor.

This is poorly supported for several reasons:

- Use cases are poorly documented, if documented at all (full-bleed images in the middle of a reflowable document seems to be the most common one).
- Reading systems often treat reflowable and FXL EPUB as two completely different formats, with different affordances and UX. Mixing them up in a single publication adds a lot of confusion (for example: should I allow the user to change the font size if this feature only works for a single ressource in a FXL publications with hundreds of resources?).

HadrienGardeur:
To double down on why the concept of synthetic spread is foreign to reflowable documents:

- when the user or the RS decides to display more than a single column of text at a time, the number of columns displayed is tied to the user settings and the viewport (screen size and/or window size, even on mobile we can have windows now)
- this means that we never know if the last screen/viewport will display one, two or even more than two columns
- reading systems often end up filling the rest of the screen with empty columns to deal with this use case

Let's say that we have a reflowable publication with full bleed images in the middle of the spine (which seems to be the main use case): some of them are displayed on their own, some of them are displayed in a spread.

The spine looks something like this:

- chapter 1
- full bleed image placed in the "left" of a synthetic spread
- chapter 2
- two full bleed images placed in the "left" and the "right" of a synthetic spread

The user doesn't like using columns when reading reflowable documents so they've disabled them:

- The reading system will use the entire viewport for chapter 1 until the viewport has been moved to the end of it
- Then it will draw the first full bleed image either on the left or in the center (based on user preferences, screen size and/or device orientation)
- Then it will use the entire viewport for chapter 2 until the viewport has been moved to the end of it
- Finally it will either draw the two full bleed images together in a spread or each of them separately fullscreen (based on user preferences, screen size and/or device orientation)

Any spread placement for these reflowable resources (chapter 1 and chapter 2) would be ignored, since there's nothing that the RS can do with them.

From a design/product perspective, I can also tell you that this is a bit of a nightmare.
Most RS will usually let the user decides whether they want columns or spreads as two separate options that can be set globally or per publication (this depends on the RS). Displaying these two settings side by side or changing the display options in the middle of the publication would be really bad in terms of UX. My best guess is that most RS will look at the `rendition:layout` value specified in `metadata` and won't adapt their user settings for these spine overrides at all.

mattgarrish:
> If you mix pre-paginated and reflow together, they won't be displayed side by side in a spread since the concept of synthetic spread is completely foreign to reflowable documents.

Ya, that's true. We do say the properties are only for synthetic spread placement, not simulated two-page reading. But we still have to allow for synthetic spread placement in reflowable layouts because of overrides to contain the documents in the spread. 

So, yes, spreads may only be created for pre-paginated documents, but they aren't exclusive to a global pre-paginated layout. We can, of course, define them under pre-paginated layouts and then have the reflowable refer back to the section for embedded spreads.

It just seems like a missed opportunity that you couldn't have a fxl document in one slot and a reflowable in another, but perhaps there's no practical use for it, and it undoubtedly would complicate rendering. But think of an art book, for example, where you could have a painting continuously in view in one slot with a scrolling reflowable document describing it in the other.

epub34/authoring/index.html
```diff
@@ -5829,6 +5828,410 @@ <h4>The <code>itemref</code> element</h4>
...
+						can be applied to reflowable spine items to further control their rendering:</p>
+
+					<ul>
+						<li>the <a href="#align-x-center"><code>rendition:align-x-center</code> spine override</a> is
+							used to indicate a preference for having the specified spine item's content be
+							center-aligned by reading systems.</li>
+					</ul>
+
+					<p>All other properties in the Package rendering vocabulary are ignored for reflowable spine
+						items.</p>
```
```suggestion
					<p>All other properties in the Package rendering vocabulary are ignored for reflowable spine
						items.</p>
```

Shouldn't we use a more normative language for this type of statement (it is repeated in the other layout subsection). Something like "Other properties in the Package rendering vocabulary _MUST NOT_ be used in spine items". If we are afraid of making some valid documents suddenly invalid, we could used _SHOULD NOT_.

HadrienGardeur:
> It just seems like a missed opportunity that you couldn't have a fxl document in one slot and a reflowable in another, but perhaps there's no practical use for it, and it undoubtedly would complicate rendering. But think of an art book, for example, where you could have a painting continuously in view in one slot with a scrolling reflowable document describing it in the other.

This would create its own set of problems related to reading progression. 

What would happen if I swipe or tap to go forward/backward? Would it provide progression within the spread like you're suggesting, or would I move to the next synthetic spread? 

This would be even worse if chapter 1 is displayed in a synthetic spread on the left while chapter 2 is displayed on the right. I'm not sure how you would handle progression at all in this case, it would be extremely confusing.

> So, yes, spreads may only be created for pre-paginated documents, but they aren't exclusive to a global pre-paginated layout. We can, of course, define them under pre-paginated layouts and then have the reflowable refer back to the section for embedded spreads.

For layout overrides, we define `rendition:layout-reflowable` under "6.2.1 Reflowable layouts" and `rendition:layout-pre-paginated` under "6.2.2 Pre-paginated layouts".

We don't refer to `rendition:layout-pre-paginated` in "6.2.1 Reflowable layouts". Why should we treat spread placements differently and refer them in "6.2.1 Reflowable layouts" ?

iherman:
For someone not well versed in FXL (sorry, pre-paginated or roll), the section on spreads is really difficult to read. I wonder whether it is possible to reuse, or at least refer to, some of the diagrams we created in 3.3. I realize they are present in the vocabulary definition, but they should, somehow, be present in this section as well. Or, if necessary, create some versions of the diagrams.

(Yes, this could be done in a later PR, once this one is finalized.)

epub34/authoring/index.html
mattgarrish: Those lists are a placeholder for now until we figure out what properties are even surviving this revision. I had all the other properties listed but then took them out to match what looks like will be left after we cull the vocabulary.

But going back to when the list had global properties in it, not just the remaining spine overrides, the problem is that you can mix layouts using spine overrides. So say I want to globally set a value for all the pre-paginated overrides I've added to my globally reflowable layout. A restriction that you must not or should not set pre-paginated properties in reflowable layouts would make it invalid and you'd have to repeat spine overrides everywhere. They're fine to set but they're ignored for reflowable documents by reading systems... hence the statement.

There's also the problem that warnings are as bad as errors for vendor ingestion, so it's probably better to leave the requirement to ignore other properties on reading systems than restrict their presence.

mattgarrish:
> We don't refer to `rendition:layout-pre-paginated` in "6.2.1 Reflowable layouts".

Right, I'm forgetting where I wrote things. I was thinking that I mentioned pre-paginated overrides under reflowable, but I mention the overrides within their respective layout sections.

The tpac discussions were too late at night for me, but if you're both confident that the group meant to resolve to remove all these properties and limit spreads to fxl, I can go ahead and make more changes now. I have a commit history I can backtrack to now, at least.

sueneu:
> For someone not well versed in FXL (sorry, pre-paginated or roll), the section on spreads is really difficult to read. I wonder whether it is possible to reuse, or at least refer to, some of the diagrams we created in 3.3. I realize they are present in the vocabulary definition, but they should, somehow, be present in this section as well. Or, if necessary, create some versions of the diagrams.
> 
> (Yes, this could be done in a later PR, once this one is finalized.)

Readability is a worthwhile goal but maybe too many edits for this PR. Would it be worth making a new issue about this so we can bring it up after this PR is merged?

iherman:
> Readability is a worthwhile goal but maybe too many edits for this PR. Would it be worth making a new issue about this so we can bring it up after this PR is merged?

Yes but, at this moment, that issue would not make sense, it is related to this PR. Once the PR is merged, I am happy to turn that into an issue.

HadrienGardeur:
> The tpac discussions were too late at night for me, but if you're both confident that the group meant to resolve to remove all these properties and limit spreads to fxl, I can go ahead and make more changes now. I have a commit history I can backtrack to now, at least.

We didn't discuss this at TPAC, I'm bringing it up because the reorg moves "Synthetic spreads" to a higher level section of its own: "6.3 Synthetic spreads".

In our current draft, it's under "8.2 Fixed layouts > 8.2.2 Fixed-layout package settings > 8.2.2.4 Spread placement".

bduga:
Wow, it is hard to keep up with this conversation! Just a few points:

1. I disagree with @HadrienGardeur's reasoning for why spreads shouldn't be allowed in mixed reflow/fixed content as multi col is an implementation detail and isn't used universally. I know Play Books doesn't, I don't think Apple does and I am pretty sure Amazon doesn't (but they have so many products I may well be both right and wrong).

2. I agree with @HadrienGardeur that this is likely too complicated to really worry about now. In fact, I expect if we attempt to find 2 implementations of synthetic spread properties working in an epub with flowing text, we would fail. Given that, it is probably prudent to accept reality and limit these to fixed layout, though I would be pleased to learn I am wrong.

3. @mattgarrish's idea for spreads is interesting (cool even!), but also not how I would expect that to work, and feels like a misuse of these properties. I would expect the first chapter to display, then the pre-paginated content to display. Say I want a chapter to display as paginated text, followed by the fixed layout pages - e.g. a forward to a comic, followed by the comic itself, but I need it to start in the correct page slot. I would  be surprised if a RS placed the forward in a single scrollable page, with the first comic page facing it, and since we just removed the last signal for publishers to tell us if a reflow document should scroll there is no way for the publisher to specify the treatment they want. I expect this feature would require more spec work, plus some industry backing. I would also be pleased if that happened, but I am not holding be breath.

iherman:
> The tpac discussions were too late at night for me, but if you're both confident that the group meant to resolve to remove all these properties and limit spreads to fxl, I can go ahead and make more changes now. I have a commit history I can backtrack to now, at least.

The issue of @HadrienGardeur (https://github.com/w3c/epub-specs/issues/2841) does list those properties that we have talked about at TPAC. As we have discussed elsewhere, it would formally be cleaner if we had a formal resolution for the proposal in that issue (which replaces a number of other issues). @sueneu @shiestyle @reidmore-online let us put that on the agenda for this week.

However, I do not see any resolution in the [minutes of tpac](https://w3c.github.io/pm-wg/minutes/2025-11-11-f2f.html#f730) on restricting the spread properties to pre-paginated layout. This should be discussed and decided upon, hopefully this week again. There is enough arguments around in this discussion, but to have a separate issue raised for that purpose may make it cleaner. Pinging again @sueneu @shiestyle @reidmore-online

HadrienGardeur:
@mattgarrish I tweaked some of the language related to spread placement in a previous commit but I can see that you're still adjusting that section so I'll comment instead.

> When a [reading system](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2844.html#dfn-epub-reading-system) renders a [synthetic spread](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2844.html#dfn-synthetic-spread), the default behavior is to populate the spread by rendering the next [EPUB content document](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2844.html#dfn-epub-content-document) in the next available unpopulated [viewport](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2844.html#dfn-viewport), where the next available viewport is determined by the given [page progression direction](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2844.html#attrdef-spine-page-progression-direction) or by local declarations within [EPUB content documents](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2844.html#dfn-epub-content-document).

This might be a good place to also indicate that without spread placement information, there's an expectation that the first resource should be displayed on its own.

We have [a resolution](https://w3c.github.io/pm-wg/minutes/2025-11-10-f2f.html#1a21:~:text=RESOLUTION%3A%20Adapt%20spec%20language%20in%20Core%20and%20Reading%20Systems%20to%20say%20that%20if%20no%20information%20about%20placement%20in%20spreads%2C%20the%20first%20resource%20should%20be%20a%20page%20on%20its%20own.) on day 1 of TPAC specifically about that:

> RESOLUTION: Adapt spec language in Core and Reading Systems to say that if no information about placement in spreads, the first resource should be a page on its own.

This was also an action item for me to open a PR to cover this change, let me know if you'd prefer that.

> Although it is common practice to specify the use of a spread in certain device orientations, the content itself does not represent a true spread — two consecutive pages that have to be rendered side-by-side for readability, such as a two-page map.

Since we're making `rendition:spread` obsolete, talking about a "common practice to specify the use of a spread in certain device orientations" feels weird.

> To indicate that two consecutive pages represent a true spread, the rendition:page-spread-left and rendition:page-spread-right properties SHOULD be set on the [spine](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2844.html#dfn-epub-spine) items for the two adjacent EPUB content documents, and the properties omitted on spine items where one-up or two-up presentation is equally acceptable.

We've agreed to drop that concept from the specification at TPAC, there was [a resolution](https://w3c.github.io/pm-wg/minutes/2025-11-10-f2f.html#1a21:~:text=RESOLUTION%3A%20Remove%20text%20saying%20that%20presence%20of%20spread%20information%20implies%20that%20the%20pages%20SHOULD%20be%20laid%20out%20as%20a%20spread%2C%20clarify%20that%20either%20presentation%20equally%20acceptable.) for this on day 1:

> RESOLUTION: Remove text saying that presence of spread information implies that the pages SHOULD be laid out as a spread, clarify that either presentation equally acceptable.

mattgarrish:
Sorry, ya, I had a few things from this afternoon that I wanted to finish up but I'm done now, at least until after we discuss this Thursday if you want to take over and make more edits.

It looks like we both tried to write out that bit about spreads applying to reflowable docs, but I made a bit more extensive changes there to mention spreads in reflowable layouts. See if it still makes sense to you. I'm not wedded to anything I've written at this point.

The other changes I made were pretty straightforward removals (which can be undone if necessary), but I also added a section to the reading system spec about handling obsolete but conforming features as we were only talking about deprecated ones. It's basically the same option to support these features but the difference for obsolete but conforming is that the legacy features must be ignored.

HadrienGardeur:
@mattgarrish your changes look good and I'm happy that you updated these examples that relied on `rendition:spread` before.

To comply with the resolutions that I referenced above, I would personally:

- drop [this entire paragraph](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2844.html#sec-pre-paginated:~:text=Although%20it%20is%20common,presentation%20is%20equally%20acceptable.)
- and add something related to the first resource of a publication being displayed on its own (the equivalent of `rendition:page-spread-center`) when there's no spread placement information in the first paragraph of that section

We might also need to add this behavior in the RS document, which would require a test as well.

mattgarrish:
I've made some additional fixes now, including removing that paragraph. I did some rewriting to the first few paragraphs of the section, too. The reading system spec still had some mixing of requirements for reflowable documents in spreads, and I noticed we said nothing at all about how to handle rendition:page-spread-center, so I tried to clean it up as best I could, as well. Again, feel free to directly edit anything if you think it can stand for improvement.

iherman:
This was discussed during the [pmwg meeting on 11 December 2025](https://w3c.github.io/pm-wg/minutes/2025-12-11.html#5a1f).

<h3>Reorganized Layout Section - <a href="https://github.com/w3c/epub-specs/pull/2844" rel="noopener noreferrer">w3c/<wbr>epub-specs#2844</a></h3>
<p><cite>wendyreid:</cite> In working on the reorg PR, some questions have come up about moving some things to the fxl section<br>
<span>… like synthetic spreads</span><br>
<span>… mgarrish are there others?</span></p>
<p><cite>mgarrish:</cite> The biggest one is the one we just resolved<br>
<span>… our previous big problem was it being spread over two specs, 3.3 glommed them together</span><br>
<span>… I cleaned that up a lot</span><br>
<span>… and Hadrien's roll is in it, a few other things</span><br>
<span>… some issues like where are spine overrides allowed</span><br>
<span>… not sure if we want to get into that now</span></p>
<p><cite>ivan:</cite> The big question does spreads ever apply to flowing?<br>
<span>… strictly speaking today it is allowed, but the should we continue with that? Or reorg the doc to make that better</span></p>
<p><cite>mgarrish:</cite> I introduced images in spine simply to acknowledge that they are fxl<br>
<span>… just to make it clear</span><br>
<span>… if those are ok, and if we are ok with spreads and reflow not going together?</span></p>
<p><cite>SueNeu:</cite> I haven't seen a used case where it makes sense for spreads in reflow<br>
<span>… but there is a need for mixed documents that have reflow and hybrid</span><br>
<span>… I just want to make sure if we disallow spreads we aren't taking options away from some creators?</span></p>
<p><cite>DaleRogers:</cite> I want to agree with SueNeu. In the epubs I create, I have mixed fxl/reflow<br>
<span>… I am learning that sometimes documents are treated as reflow or fxl as an entire epub</span><br>
<span>… it would be nice to have that capability</span></p>
<p><cite>Hadrien:</cite> I don't think there is any impact<br>
<span>… the other thing is, no one knows how to implement this</span><br>
<span>… or how we would even test this</span><br>
<span>… and it is unused</span><br>
<span>… and it is hinted that it is for fxl</span><br>
<span>… this just makes sense for clarity</span><br>
<span>… as a top level section I think it causes confusion</span></p>
<p><cite>wendyreid:</cite> If we move spread placement to fxl, it will be clearer since in practice it only works for fxl<br>
<span>… You could use it on a reflow, but it doesn't really make sense</span></p>
<p><cite>&lt;shiestyle&gt;</cite> +1 to Wendy</p>
<p><cite>wendyreid:</cite> so we should say spread placement is only for fxl documents, whether they are complete epubs or hybrid</p>
<p><cite>SueNeu:</cite> Thanks, that is helpful.<br>
<span>… I agree now, it is better to focus on fxl</span></p>
<p><cite>DaleRogers:</cite> For me it is more semantics. Does document mean an xhtml page or the epub?</p>
<p><cite>wendyreid:</cite> We mean content document</p>
<p><cite>rdeltour:</cite> Just checking that the PR stays open for a few days, I have some comments<br>
<span>… one is about identifying fxl documents</span><br>
<span>… we have some statements that are hard to test</span><br>
<span>… so we say when the creator has this intent, but we can't test intent</span><br>
<span>… if instead, we say a document is fxl BECAUSE it has a viewport (e.g.) then we can test</span></p>
<p><cite>mgarrish:</cite> It will definitely be open for a while<br>
<span>… I need to do a proper clean up</span><br>
<span>… so please comment on the PR</span></p>
<p><cite>ivan:</cite> we should do a proper resolution before we move on</p>
<p><cite>&lt;wendyreid&gt;</cite> PROPOSED: Move the spread placement section into the pre-paginated section of the specification.</p>
<p><cite>&lt;DaleRogers&gt;</cite> +1</p>
<p><cite>&lt;kersc&gt;</cite> +1</p>
<p><cite>&lt;CharlesL1&gt;</cite> +1</p>
<p><strong>RESOLUTION:</strong> Move the spread placement section into the pre-paginated section of the specification.</p>

<hr /></details>

sueneu (commented):
“outdated” is much simpler and works well here. Is there a place where our use of it is defined for people who might want that? 

Maybe in Authoring Shorthands?
`obsolete: features or attributes that are no longer in common use but are still considered conforming for backwards compatibility`

We should probably make it clear in the RS spec that obsolete features should still be supported.

iherman:
I am late to the party, and maybe what I hit was a RS error. But I have just had a problem with an fxl file that did not display properly without the `<meta property="rendition:spread">none</meta>`.

I have indeed prepared an EPUB with a bunch of images in the spine (all images are of the same dimension), each having a fallback to an html file, so the EPUB itself is 100% ok per epubcheck. If I display the book in Thorium, it only displays every second image, as if it tried to put each image on top of one another. If I set the meta above, it works just fine.

Is it a Thorium bug, or are we maybe to hasty obsoleting that particular value for that property?

(My experience of fxl is close to nil, so it may well be on me.)

@HadrienGardeur

HadrienGardeur:
> Is it a Thorium bug, or are we maybe to hasty obsoleting that particular value for that property?

This is probably a Thorium Desktop bug for publications with images in spine.

@iherman could you share this EPUB with me by email? I can open up an issue for Thorium Desktop and we'll test it out with Thorium Mobile as well.

iherman:
The book is at https://e.pcloud.link/publink/show?code=XZC0oAZ5CFtwfY4i9QFPeQTu0EL4Fq2zH87, if it helps (with the meta added).

mattgarrish:
> “outdated” is much simpler and works well here. Is there a place where our use of it is defined for people who might want that?

Yes, we have an appendix that explains the meanings and validation expectations for both outdated and deprecated. If you're reading the github diff and see the syntax `[= ... =]` around words that's a respec shorthand for linking terms to their definitions, so every instance of deprecated and outdated will get turned into hyperlinks in the rendered document.


> We should probably make it clear in the RS spec that obsolete features should still be supported.

I'm updating the reading system appendix in this PR, as well. We used to only say that they may support deprecated features, but I've added separate subsections for each category. The outdated features still have that subset of EPUB 2 legacy features that reading systems must not support that needs to be called out.

iherman (approved):
Apart from https://github.com/w3c/epub-specs/pull/2844#issuecomment-3646840923 (which may become irrelevant) I believe our best way forward is to merge this asap. There may be some details here and there, but we all seem to agree about the broad change, and it may become more efficient to discuss the details through dedicated issues.

Huge thanks to @mattgarrish

iherman:
Re https://github.com/w3c/epub-specs/pull/2844#issuecomment-3646840923 @HadrienGardeur, it works with setting the `rendition:page-spread-center` on each spine item. But it is much more convenient to have to set it at one place.

Where is it specified in the specification that, for images in the spine, the spread should be `none` by default. Because if it is not specified, then it is not, strictly speaking, a bug of Thorium...

HadrienGardeur:
> But it is much more convenient to have to set it at one place.

… and much more dangerous as well. I've seen many publications that set this property to `none` for no good reason, which negatively impacts the UX.

> Where is it specified in the specification that, for images in the spine, the spread should be none by default. Because if it is not specified, then it is not, strictly speaking, a bug of Thorium...

That's not the problem IMO.

Thorium Desktop most likely defaults to "auto" but I suspect a bug when laying out spreads with images in spine.

mattgarrish:
Here's a side-effect of removing the flow property that's not as easy as deleting the reading system requirement:

> Reading systems SHOULD support spine-level scripting in reflowable EPUB content documents that use the ["scrolled-doc"](https://www.w3.org/TR/epub-34/#flow-scrolled-doc) or ["scrolled-continuous"](https://www.w3.org/TR/epub-34/#flow-scrolled-continuous) [[epub-34](https://www.w3.org/TR/epub-rs-34/#bib-epub-34)] presentation modes defined by [the rendition:flow property](https://www.w3.org/TR/epub-rs-34/#flow). Similarly, if it supports spine-level scripting in reflowable EPUB content documents, it MUST implement the "scrolled-doc" presentation mode and SHOULD implement the "scrolled-continuous" presentation mode.

Replacing the flow values with their old rendering requirements, I'm going to change this to:

> Reading systems SHOULD support spine-level scripting in reflowable EPUB content documents when rendering the content such that users can scroll overflow content. Similarly, if a reading system supports spine-level scripting in reflowable EPUB content documents, it MUST provide users the option to scroll overflow content in reflowable EPUB content documents and SHOULD provide the option to present the publication as one continuous scroll from spine item to spine item <strike>(except where locally overridden)</strike>.

But let me know if anyone has a problem with this. I'd forgotten we enforced support for the flow types for scripting.

(Edit: removing the parenthetical about overriding values, since that can't be done anymore.)

mattgarrish:
> I believe our best way forward is to merge this asap

I'm getting down to minor cleanup stuff, but I still want to make sure everything is reasonably consistent. I'm hopping between documents today but I'll definitely flip this out of draft status before the end of the weekend and see if there's any final feedback.

mattgarrish:
Opening this for review now as there's not much left I can do.

I do have one last question about rendition:align-x-center, though. From the reading system processing section, it sounds like applies to both reflowable and pre-paginated content, since it has separate requirements for positioning in the viewport or spread and positioning in virtual pages.

Going back to 3.0.1 it was classified as a general property, meaning it was intended for both layouts, but somewhere along the line it got reclassified as a reflowable property.

@toshiakikoike @shiestyle if you can confirm what it applies to, I'll do a last bit of editing to fix either the authoring or rs specs.

iherman:
(Editorial: I believe the comment at https://github.com/w3c/epub-specs/pull/2844#issuecomment-3646840923 and the resulting thread, should be discussed as part of #2844, and not here. It was my mistake, apologies. 

See https://github.com/w3c/epub-specs/issues/2841#issuecomment-3649067501

iherman:
> > Reading systems SHOULD support spine-level scripting in reflowable EPUB content documents when rendering the content such that users can scroll overflow content. Similarly, if a reading system supports spine-level scripting in reflowable EPUB content documents, it MUST provide users the option to scroll overflow content in reflowable EPUB content documents and SHOULD provide the option to present the publication as one continuous scroll from spine item to spine item (except where locally overridden).
> 
> But let me know if anyone has a problem with this. I'd forgotten we enforced support for the flow types for scripting.
> 

I have no problem with the technical change.

However, I had some difficulties parsing the new text, so I tried to make it a bit more concise.  Here is my take:

> Reading systems SHOULD support spine-level scripting in reflowable EPUB content documents when rendered with scrollable overflow content. Similarly, if a reading system supports spine-level scripting in reflowable EPUB content documents, it MUST provide users the option to scroll overflow content and SHOULD provide the option to present the publication as one continuous scroll from spine item to spine item.

But I may have missed some nuances...

sueneu:
>Reading systems SHOULD support spine-level scripting in reflowable EPUB content documents when rendered with scrollable overflow content. Similarly, if a reading system supports spine-level scripting in reflowable EPUB content documents, it MUST provide users the option to scroll overflow content and SHOULD provide the option to present the publication as one continuous scroll from spine item to spine item. 

This is wonderful. I spot two places where we might use a simpler word, and its missing the caveat about local overrides. 

In both versions, I am confused by the first sentence. Is spine-level scripting in the “EPUB content documents”? Wouldn’t spine-level scripting be in the spine but affect the content documents? 
It is possible I don’t understand this correctly, I don’t add scripting.

>Reading systems SHOULD support spine-level scripting in reflowable EPUBs where content is rendered with scrollable overflow. Similarly, if a reading system supports spine-level scripting in reflowable EPUBs, it MUST give users the option to scroll overflow content and SHOULD provide the option to view the publication as one continuous scroll from spine item to spine item. Local overrides are permitted.

mattgarrish:
The edits look fine to me. It was getting late so I just spliced in the text that was in the processing instructions for scrolled-doc and scrolled-continuous.

I've deleted the local overrides part, though. As I mentioned in my edit to the comment above, when I copied the text I noticed that the hyperlink went to the flow property spine overrides in the authoring spec which are now gone. Support for overrides is only a thing if the reading system developer chooses to support the flow property, but in that case they will already know they have to account for the local overrides.

HadrienGardeur:
> Reading systems SHOULD support spine-level scripting in reflowable EPUB content documents that use the ["scrolled-doc"](https://www.w3.org/TR/epub-34/#flow-scrolled-doc) or ["scrolled-continuous"](https://www.w3.org/TR/epub-34/#flow-scrolled-continuous) [[epub-34](https://www.w3.org/TR/epub-rs-34/#bib-epub-34)] presentation modes defined by [the rendition:flow property](https://www.w3.org/TR/epub-rs-34/#flow). Similarly, if it supports spine-level scripting in reflowable EPUB content documents, it MUST implement the "scrolled-doc" presentation mode and SHOULD implement the "scrolled-continuous" presentation mode.

I'm not a fan of anything in that section, it feels very misguided.

Scripting is poorly supported for reflowable EPUB for many reasons, including:

- security concerns
- user preferences (styles injected by the RS could clash with the scripted content)
- progression affordances (taps and swipes)
- other affordances (for example popup footnotes or a dedicated view for images)

Pagination is part of the problem but it's not the only reason why scripting is poorly supported.

Convoluted recommandations (I don't know a single RS that can support both `scrolled-doc` and `scrolled-continuous`) just make our spec worse.

I recommend dropping the entire paragraph, a rewrite isn't worth the effort.

mattgarrish:
> I recommend dropping the entire paragraph, a rewrite isn't worth the effort.

The problem is we added those recommendations to try and push reading systems to support scripting for reflowable, so if we drop it we're back where we started.

It's also a bit ugly to have users have a working scripted publication if they prefer one overflow handling and not working if they prefer another. We may understand why support might be less flaky when the content isn't dynamically paginated, but how many users are ever going to appreciate that and not just leap to the conclusion that epub is hopelessly broken?

I'd suggest opening a new issue for this, though, as we should probably discuss what is realistic and/or if keeping the requirements as aspirational is worth it.

HadrienGardeur:
> I'd suggest opening a new issue for this, though, as we should probably discuss what is realistic and/or if keeping the requirements as aspirational is worth it.

Right, I agree that this should be a separate discussion.

I would put this one side by side with the HTML serialization: both goals can only be achieved if we set different expectations, this is not a format issue per se.

iherman:
> 
> I'd suggest opening a new issue for this, though, as we should probably discuss what is realistic and/or if keeping the requirements as aspirational is worth it.

+1

mattgarrish:
I'll open issues for this and the align-x-center property so we can close this off. I'll leave this PR open until tomorrow and then merge because I didn't get it open until late on Friday..

HadrienGardeur:
@mattgarrish I've finally caught up the last edits that you've made and have two comments related to spreads.

> To indicate to reading systems that a two-page spread is the optimal way to display pre-paginated spine items, or when only a single viewport is best […]

I think that this could be misunderstood. The presence of `page-spread-*` doesn't mean that either a two-page spread of a single viewport is best, it just tells the RS where to place a given resource in a synthetic spread.

To go back to my very first comment (https://github.com/w3c/epub-specs/pull/2844#issuecomment-3632673767) in this PR, we haven't tackled yet one of our resolution from Kobe:

> while in Kobe, [we voted on a resolution](https://w3c.github.io/pm-wg/minutes/2025-11-10-f2f.html#1a21:~:text=RESOLUTION%3A%20Adapt%20spec%20language%20in%20Core%20and%20Reading%20Systems%20to%20say%20that%20if%20no%20information%20about%20placement%20in%20spreads%2C%20the%20first%20resource%20should%20be%20a%20page%20on%20its%20own.) to: "Adapt spec language in Core and Reading Systems to say that if no information about placement in spreads, the first resource should be a page on its own."

Our current spec language talks about alternating between left/right but says nothing about the first resource:

> In this case, the reading system will populate the spread by rendering the next [EPUB content document](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2844.html#dfn-epub-content-document) in the next available unpopulated viewport, where the next available viewport is determined by the given [page progression direction](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2844.html#attrdef-spine-page-progression-direction) or by local declarations within [EPUB content documents](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2844.html#dfn-epub-content-document).

mattgarrish:
> > To indicate to reading systems that a two-page spread is the optimal way to display pre-paginated spine items, or when only a single viewport is best […]
> 
> I think that this could be misunderstood.

Okay, I've taken that part out and left starting from "The automatic population behavior MAY be overridden..." on.

> Our current spec language talks about alternating between left/right but says nothing about the first resource:

For the authoring spec, I changed the paragraph to:

> In this case, the reading system will render the first document on its own when using a synthetic spread (which slot the document is rendered in is determined by the given [page progression direction](http://localhost:120/epub34/authoring/#attrdef-spine-page-progression-direction) or by local declarations within [EPUB content documents](http://localhost:120/epub34/authoring/#dfn-epub-content-document)). It will then populate the spread by rendering the next [EPUB content document](http://localhost:120/epub34/authoring/#dfn-epub-content-document) in the next available unpopulated viewport. Both documents in the spread are replaced each time the user progresses forward or backward through the publication.

For the reading system specification, do we need to qualify the whole "is non-linear content presented or not" issue that arises with the "first" spine item? Here's what I have so far:

> If no spread placement information is specified, the resource referenced from the first [itemref element](https://www.w3.org/TR/epub-34/#dfn-itemref) [[epub-34](http://localhost:120/epub34/rs/#bib-epub-34)] in the spine SHOULD be presented by itself (i.e., with one slot empty). Which slot to use SHOULD be obtained from the [page-progression-direction attribute](https://www.w3.org/TR/epub-34/#attrdef-spine-page-progression-direction) [[epub-34](http://localhost:120/epub34/rs/#bib-epub-34)] value. For example, in an EPUB publication with a left-to-right page progression direction, the left-hand slot will be unpopulated.

> 
> Note
> If a reading system does not render [non-linear](https://www.w3.org/TR/epub-34/#attrdef-itemref-linear) [[epub-34](http://localhost:120/epub34/rs/#bib-epub-34)] content by default, the first spine item to place on its own is the first linear item in the spine.

Let me know if that works or not.

mattgarrish:
One other thing: the authoring spec mentions that the slot placement can be determined from info in the fxl doc. But we've tried enumerating "possible" sources of info in the past without it ever working so I've left it that that the ppd attribute should be used and if a reading system dev wants to use something else that's still fine since it's not a must. If there's only one source and it's just never been mentioned before we can alter the requirement, but otherwise I'm not going to try and dive into that problem.

epub34/authoring/index.html
```diff
@@ -5829,6 +5827,429 @@ <h4>The <code>itemref</code> element</h4>
...
+				<section id="sec-pre-paginated">
+					<h4>Pre-paginated</h4>
+
+					<p>In a pre-paginated fixed layout, each fixed-layout document referenced by a [=EPUB spine |
+						spine=] [^itemref^] represents one page of content. Reading systems are expected to scale the
+						document to fit the viewport or synthetic spread.</p>
+
+					<p>To declare that an EPUB publication uses a pre-paginated fixed layout, the <a href="#layout"
+								><code>rendition:layout</code> property</a> MUST be declared in a <code>meta</code>
+						element that specifies the value <dfn id="def-layout-pre-paginated">pre-paginated</dfn>.</p>
```
rdeltour: I think this should be rephrased to remove the MUST statement, as this MUST here is not testable.

In fact, there are two exclusive options:
- either a fixed-layout document is defined as such independently, and then we can keep a MUST here. For instance if we define a fixed-layout document as any HTML content document that has a viewport with dimensions, then we know what document is fixed-layout and what is not, and we can say the fixed-layout ones MUST be declared as such in the package document.
- or a fixed-layout document is such **because of** that very package document declaration. And then we cannot really use a MUST here: the `pre-paginated` property declares the intent to consider the document a fixed-layout, and we cannot test intentions.

Does that make sense?
I think that a pragmatic objective is to make all RFC2119 statements testable.

Note that this comment should not prevent the PR to be merged, I'm happy to transfer it to its own issue if you think it is relevant.

(edited for typos)

rdeltour (approved):
The Re-organization looks good to me.
I added an inline comment to describe what I briefly mentioned during our last call, but this can be processed in a separate issue once the PR is merged.

epub34/authoring/index.html
iherman: Your first alternative may be slightly more complicated. It can also be an SVG document (which is easy because that is also a possible content document) and it can also be an image provided there is a fallback for it. But an image may be, formally, part of a reflowable document (although I am not sure it is clear how it is rendered); on the other hand, an image does have an intrinsic dimension, ie, window size. Ie, that, by itself, that may not define a fixed layout document. 

Also, that paragraph is for pre-paginated documents, not fixed layout in general. How would roll document fit into the picture? It strikes me that the only difference between roll and pre-paginated might be the package document declaration, not the content documents.

All this to say that it makes it much simpler to follow your second alternative, imho.

As for, the RFC usage issue: yes, as you say, it is a pragmatic goal to make all those documents testable. But that can be done with our usual testing system: we create a document that is 'expected' to behave in a certain manner in an RS, and that is, at the end of the day, tested by humans. What is correct, is that epubcheck would not be able to check that properly, though.

mattgarrish:
> If no spread placement information is specified, the resource referenced from the first [itemref element](https://www.w3.org/TR/epub-34/#dfn-itemref) [[epub-34](http://localhost:120/epub34/rs/#bib-epub-34)] in the spine SHOULD be presented by itself (i.e., with one slot empty).

Jumping back a discussion, I don't like this on second look. It would be clearer to say no placement info is on the first item:

"If the first itemref element</a> [epub-34] in the spine does not specify a spread placement position, it SHOULD be presented by itself (i.e., with one empty slot)."

(Edit: further simplified the start of the sentence)

epub34/authoring/index.html
mattgarrish: > * the `pre-paginated` property declares the intent to consider the document a fixed-layout, and we cannot test intentions.

Right, fixed-layout documents don't exist without a fixed layout. They're entwined. If you don't support pre-paginated or roll layouts, then by default all the documents in the spine become reflowable as the reading system will try to render them as such.

You can't inspect an html or svg document and decide that it's fixed layout simply because you find a viewport meta declaration or viewBox attribute. Those are perfectly valid in reflowable documents, too. (But, again, what is a reflowable svg? One that isn't scaled or have its text properties frozen?)

So, no, a "must" isn't strictly necessary here and we don't have a similar must for declaring roll layouts. It could be phrased as:

```suggestion
					<p>To indicate that an EPUB publication uses a pre-paginated fixed layout, the <a href="#layout"
								><code>rendition:layout</code> property</a> is declared in a <code>meta</code>
						element that specifies the value <dfn id="def-layout-pre-paginated">pre-paginated</dfn>.</p>
```

But then the reflowable MAY requirement should similarly be dropped to a "can":

> If an explicit declaration is preferred, however, the [rendition:layout property](http://localhost:120/epub34/authoring/#layout) can be set in a meta element that specifies the value reflowable.

And then the same question arises about the spine overrides. Are we out of "must" territory, for example, with this statement:

> To do so, the [rendition:layout-pre-paginated spine override](http://localhost:120/epub34/authoring/#layout-pre-paginated) MUST be set in the [properties attribute](http://localhost:120/epub34/authoring/#attrdef-properties) on each spine itemref that is pre-paginated.

The overrides are no more testable than the global intent, at least from a machine checking perspective. The normative keywords only emphasize the need to use the values to someone reading the spec.

mattgarrish: > yes, as you say, it is a pragmatic goal to make all those documents testable. But that can be done with our usual testing system: we create a document that is 'expected' to behave in a certain manner in an RS, and that is, at the end of the day, tested by humans.

Ya, I'm kind of torn on how far we should try to go with this. A test doesn't prove that the assertion in the package document is right or wrong, nothing can verify that, only that a reading system supports the authored intent. I think that's what @rdeltour is getting at here. Only the author of the document will know if they got their layout and overrides set properly. The only testable assertion is that a meta tag with rendition:layout specifies one of reflowable, pre-paginated, or roll (or that a spine override is a known value), but that's already required from the property definitions in the appendix.

iherman: > Only the author of the document will know if they got their layout and overrides set properly. 

Many (most) of our tests rely on that assumption. The author of the tests is expected to describe, as part of the test book, what is _expected_ to be seen, and it is then the job of the tester to make a honest statement whether this is indeed the case (or not). Let us remember that our goal is not to pass a judgement over the RS, but to prove that a specific feature _can_ be properly implemented (with different implementations). This is a very different viewpoint than the epubcheck's.

mattgarrish: > This is a very different viewpoint than the epubcheck's.

Right, that's why I don't want to go too far down this rabbit hole. We had a long discussion about testable assertions with the security folks when we changed our non-normative practices to normative ones. Not everything is testable.

I just had to answer a question again in the KB, for example, about why html strictly disallows cite inside of blockquote for the citation of the quote, even though you can have cite inside of blockquote if the quote cites something. Only the author will know if they got that right. Sometimes you want the emphasis to direct people to what they should be doing.

But in this case I'm okay with not adding normative requirements on setting the layout and overrides both because it's not really critical to understanding their use and also because we have normative requirements on the values in the definitions. If you try to put `<meta property="rendition:layout">gobbledygook</meta>` in the metadata you're going to get an error. If you put `<meta property="rendition:layout">roll</meta>` and you meant to put pre-paginated there's no authoring test to verify this. It's not even a reading system test, as you can't test if an author misapplied a hint.

HadrienGardeur:
> (i.e., with one empty slot)

Well, maybe we need to be even more clear on that one.

The first item could be:

- displayed on its own and centered (the equivalent of `page-spread-center`)
- or displayed on its own with an empty slot next to it (I guess that would be `page-spread-right` if the direction is LTR or `page-spread-left` if the direction is RTL)

It's easier to just use the behavior for `page-spread-center` IMO.

mattgarrish:
> It's easier to just use the behavior for `page-spread-center` IMO.

Okay, I obviously didn't get that from the resolution...

So here's what I've just committed for the authoring spec:

> In this case, the reading system will render the first document on its own (equivalent to setting the rendition:page-spread-center property). It will then create and populate the spread by rendering the next two [EPUB content documents](http://localhost:120/epub34/authoring/#dfn-epub-content-document) (which slot the documents are rendered in is determined by the given [page progression direction](http://localhost:120/epub34/authoring/#attrdef-spine-page-progression-direction) or by local declarations within the [EPUB content documents](http://localhost:120/epub34/authoring/#dfn-epub-content-document)). Both documents in the spread are replaced each time the user progresses forward or backward through the publication.

And the reading systems section now says:

> If the first [itemref element](https://www.w3.org/TR/epub-34/#dfn-itemref) [[epub-34](http://localhost:120/epub34/rs/#bib-epub-34)] in the spine does not specify a spread placement position, it SHOULD be presented by itself (i.e., as though the rendition:page-spread-center property has been set).

Is that closer to what was wanted?

bduga:
> In this case, the reading system will render the first document on its own (equivalent to setting the rendition:page-spread-center property).

Hang on - I don't agree with that, and I don't think it is what we decided on. I consider the exact placement to be a stylistic one, and there is no practical reason why it must be centered. So while this is bad: [1,2],[2,3]... there is no practical difference between [1], [2,3]... and [-,1],[2,3]... In fact, Play Books Web uses the former while PB Android uses the latter. I don't think we should dictate a specific styling, we just need to make sure there spreads are aligned with the correct pages.

mattgarrish:
One more try, then... 😄 

>In this case, the reading system will render the first document on its own — either alone in one of the two available slots (which slot is determined by the given [page progression direction](http://localhost:120/epub34/authoring/#attrdef-spine-page-progression-direction) or by local declarations within the [EPUB content documents](http://localhost:120/epub34/authoring/#dfn-epub-content-document)) or centered on its own (equivalent to setting the rendition:page-spread-center property). After the first document, it will populate the spread by rendering the next two successive [EPUB content documents](http://localhost:120/epub34/authoring/#dfn-epub-content-document). Both documents in the spread are then replaced each time the user progresses forward or backward through the publication.

And for the rs spec:

> If the first [itemref element](https://www.w3.org/TR/epub-34/#dfn-itemref) [[epub-34](http://localhost:120/epub34/rs/#bib-epub-34)] in the spine does not specify a spread placement position, it SHOULD either be presented by itself (i.e., as though the rendition:page-spread-center property has been set) or placed alone in one of the two available spread slots (i.e., with the other slot empty). Which slot to use SHOULD be obtained from the [page-progression-direction attribute](https://www.w3.org/TR/epub-34/#attrdef-spine-page-progression-direction) [[epub-34](http://localhost:120/epub34/rs/#bib-epub-34)] value. For example, in an EPUB publication with a left-to-right page progression direction, the left-hand slot will be unpopulated.

bduga:
> One more try, then... 😄

Sorry! This looks better to me.

shiestyle:
@mattgarrish 
> @toshiakikoike @shiestyle if you can confirm what it applies to, I'll do a last bit of editing to fix either the authoring or rs specs.

I replied that 'rendition:align-x-center' can be used for reflowable contents at the issue thread.
Anyway, it also can be moved to obsolete status because it is rarely used at least in Japan (it was requested from Japan originally).

mattgarrish:
> Anyway, it also can be moved to obsolete status because it is rarely used at least in Japan (it was requested from Japan originally).

Thanks! That actually makes my life a lot easier if we can move it to the outdated (formerly obsolete but conforming) section. It'll continue to be valid to use.

I'll make that change now so we can close the issue with this pull request.

iherman:
Absolutely super minor change before merge: in the authoring change section the 10 October entry refers to the classification of obsolete but non conforming; we now have to note the change to "outdated". I think a separate entry should also be added referring to the reorganization, in particular with the layout classification "reflowable-fixed layout" and fxl classification of "prepaginated-roll". Maybe the fact that spread related properties are now very stricly bound to pre-paginated could also be noted.

(Note to myself: it will be time to refresh the overview document!)

iherman:
Another minor point (it could be added as a separate issue, or reopening https://github.com/w3c/epub-specs/issues/2792): while we agreed on TPAC about that no normative changes will be done regarding images in the spine, I believe we also said that we would add a note somewhere in the document about the fact that it is possible to create an FXL document (pre-paginated or roll) with images in the spine, provided there is a proper fallback for each image.

@reidmore-online @HadrienGardeur

mattgarrish:
> add a note somewhere in the document about the fact that it is possible to create an FXL document (pre-paginated or roll) with images in the spine, provided there is a proper fallback for each image

I added a paragraph to the fxl document intro that mentions that images in spine are often used and are a type of fxl document but require fallbacks. I also added a section to the ICB section that non-normatively says that reading systems will have to get the width and height info from any format they support (both in the authoring and rs specs). This was so that we could avoid having a separate roll document definition and/or having to talk separately about fxl epub content docs and fxl image docs separately.

I mentioned this change in an earlier [comment](https://github.com/w3c/epub-specs/pull/2844#issue-3711075346) and also when we got into questions about the PR on the last call and no one has responded negatively to the change.

mattgarrish:
> I think a separate entry should also be added referring to the reorganization

Thanks, ya, I've added some additional entries and tried to link everything up to issues or discussions. If you think I've missed anything, though, let me know. I guess we'll wait until tomorrow to merge this to get a final okay on the align-x-center issue.

> (Note to myself: it will be time to refresh the overview document!)

And hopefully I didn't make a mess of the tests. I was having issues with all the embedded test tagging as I was restructuring all the properties back into the appendix, but then we deleted them all, so hopefully I didn't break too much in the end.

HadrienGardeur:
> […] while we agreed on TPAC about that no normative changes will be done regarding images in the spine, I believe we also said that we would add a note somewhere in the document about the fact that it is possible to create an FXL document (pre-paginated or roll) with images in the spine, provided there is a proper fallback for each image.

That's a good reminder for several things:

- I need to update example 61 for roll layout to better align with what I had in my PR (images in spine with HTML fallbacks)
- we need to open a separate issue (or discussion ?) to discuss a future group note for comics
- and we're still missing a note that mentions what some Japanese publishers are currently using for roll layout (`pre-paginated` + `scroll-continuous`)

iherman:
> I added a paragraph to the fxl document intro that mentions that images in spine are often used and are a type of fxl document but require fallbacks.

Hm. It does not appear in the preview... sigh. 

But the important point is that you did it! Thx.

iherman:
> > (Note to myself: it will be time to refresh the overview document!)
> 
> And hopefully I didn't make a mess of the tests. I was having issues with all the embedded test tagging as I was restructuring all the properties back into the appendix, but then we deleted them all, so hopefully I didn't break too much in the end.

We will have to look through all the tests anyway, because we have, for example, rename them, some will become obsolete, etc.  I will look at that when the time (and the new year) comes!

mattgarrish:
> * and we're still missing a note that mentions what some Japanese publishers are currently using for roll layout

I've added the following in the last commit:

> Prior to the introduction of roll layouts, some Japanese publishers combined a [pre-paginated layout](http://localhost:120/epub34/rs/#sec-pre-paginated-layouts) with a [rendition:flow property](https://www.w3.org/TR/epub-33/#flow) [[epub-33](http://localhost:120/epub34/rs/#bib-epub-33)] set to the value "scrolled-continuous" to create rolls. If reading systems developers intend to support this content, they will need to treat the combination as synonymous with a roll layout being declared.

Feel free to adjust it, though.

epub34/authoring/index.html
rdeltour: I personally like Matt's proposed rephrasing better than the must.

To me this is about what is a requirement vs what is a definition. In fact, the presence of the layout properties in the package document is what **defines** a document as fixed-layout or not. 

With a MUST statement, my mindset goes "when I have a fixed-layout document, then I MUST declared the property. But wait, why is this a fixed layout document? Oh, because I declared the property." And then it is looping. So I better understand this as: "a content document is fixed-layout when it is declared with this property"; it is not a requirement, it is its definition.

That's said, it is mostly editorial and spec purity, so I definitely won't die on that hill 😊

rdeltour: > And then the same question arises about the spine overrides. 

Yes, my comment applies to these other places too, wherever we a document is defined as fixed-layout due to layout properties (`pre-paginated` or `roll`, and publication-wide or spine overrides).

rdeltour:
Just a nit to pick but I'm not a big fan of the introduction of the "outdated" terminology.
- Before we had : "obsolete" = "obsolete but conforming" + "deprecated"
- now we have "obsolete" = "outdated" + "deprecated"

I'm not convinced it is much clearer without looking into the details of how these terms are defined. Or maybe it is obvious to English native speakers?

I would find it more intelligible as:
- "obsolete" = "obsolete but conforming" + "obsolete and not recommended".
(or the equivalent with "outdated" or "deprecated" instead of "obsolete", I don't have a strong preference here).

i.e. only have one keyword, with the two variants refined by qualifiers.

mattgarrish:
> * "obsolete" = "obsolete but conforming" + "obsolete and not recommended".

Ya, I think that's similar to what I originally proposed (html has obsolete but conforming and non-conforming features as its categories) but we opted not to change deprecated yet again.

The issue raised last week is that "obsolete but conforming" is a mouthful. I can live with outdated because it doesn't, at least in English, have the negative connotations of deprecation. Deprecation means you aren't supposed to use whatever it refers to anymore. Outdated means technology or best practice has moved beyond the feature but if you're still old fashioned you can use it.

iherman:
This was discussed during the [pmwg meeting on 18 December 2025](https://w3c.github.io/pm-wg/minutes/2025-12-18.html#0489).

<h3>Reorganized layout section - <a href="https://github.com/w3c/epub-specs/pull/2844" rel="noopener noreferrer">w3c/<wbr>epub-specs#2844</a></h3>
<p><cite>ivan:</cite> there are over 100 comments</p>
<p><cite>mgarrish:</cite> We were looking at the initial start page, what should happen when there is no spread placement on the first spine item<br>
<span>… we decided we should display it centered in the viewscreen or slotting the page on its own however that best works</span><br>
<span>… we have a note to make an equivalent in rolled layouts</span></p>
<p><cite>Hadrien:</cite> There was a discussion also about the first resource when there isn’t spread information<br>
<span>… there was a discussion with Roman about other things</span></p>
<p><cite>mgarrish:</cite> The issue with Roman was whether this should be a MUST<br>
<span>… this removes a normative requirement that was untestable</span><br>
<span>… I think this works in this case because this is an authoring suggestion</span><br>
<span>… it would be good to move forward since we have a lot of issues in the PR</span></p>
<p><cite>ivan:</cite> it should be merged as soon as possible</p>
<p><cite>&lt;gman&gt;</cite> Hadrien my nickname is the first letter of my first name and 3 first letters of my last name, which makes a half-life reference, how blessed can one be :)</p>
<p><cite>ivan:</cite> we said we would add a not whereby it is possible to make this in an FXL</p>
<p><cite>mgarrish:</cite> it is in the discussion about fixed-layout documents<br>
<span>… and in the icb section it says if you are using these, then the reading system will pick it up</span></p>
<p><cite>ivan:</cite> this should be merged as soon as possible followed by a revamping of the test cases in the new year</p>
<p><cite>duga:</cite> we should submit this asap, there are 100 comments on the PR<br>
<span>… many of them have nothing to do with reorganization</span><br>
<span>… I don’t think this is a good workflow, we are doing real, unrelated editing in this PR</span><br>
<span>… if you see something in a PR that isn’t related to the PR, they should open a separate issue</span><br>
<span>… and I suggest mgarrish pushback on this</span><br>
<span>… obviously somethings would be easier to do at the same time,</span><br>
<span>… but we’ve made so many normative changes in this PR I don’t think I’ve responded to them all</span><br>
<span>… but mgarrish should ask others to open the PR, not be responsible for that himself</span></p>
<p><cite>mgarrish:</cite> I agree this one got out of control</p>
<p><cite>wendyreid:</cite> noted for the future, don’t make mgarrish do so many things in one PR<br>
<span>… I think we are all agreed to merge this as soon as possilbe</span></p>
<p><cite>&lt;wendyreid&gt;</cite> PROPOSED: Merge PR <a href="https://github.com/w3c/epub-specs/issues/2844" rel="noopener noreferrer">#2844</a></p>
<p><cite>&lt;CharlesL1&gt;</cite> +1</p>
<p><cite>&lt;gautierchomel&gt;</cite> +1</p>
<p><cite>shiestyle:</cite> thank you mgarrish for including a note for Japanese rolled comics</p>
<p><strong>RESOLUTION:</strong> Merge PR <a href="https://github.com/w3c/epub-specs/issues/2844" rel="noopener noreferrer">#2844</a></p>
<p><cite>wendyreid:</cite> the big task now will be the tests that need to be edited and renamed</p>
<p><cite>ivan:</cite> I have created a test file with images in the spine</p>
<p><cite>Hadrien:</cite> thank you ivan for the example of images in the spine, I will create a few more<br>
<span>… these will be used by engineers in Thorium to test presentation</span><br>
<span>… we expect implementation on IOS and Readium Mobile</span></p>

<hr /></details>

Issue #2835, opened 2025-11-27
`creator` of an annotation can be a software agent


gautierchomel:
### Section

https://w3c.github.io/epub-specs/epub34/annotations/#1-1-annotation-object

### Describe the problem

Currently, we describe `creator` as follows : 
> The creator of the annotation. This may be either a human or an organization.

The W3C Web Annotation Data Model defines creator as 
> The agent responsible for creating the resource. This may be either a human, an organization or a software agent.

Software agent can provide analysis, point most probable errors or even making wording shift proposals. 

### Describe the fix or new feature you propose

Use W3C Web annotation Data model description.

llemeurfr:
I can understand that a software agent may automatically annotate a publication. 
In such a case, I'd like a use case to be created for expressing the business case. 

If we go there, we must make a clear distinction between such a machine (usually an AI system) and the software on which annotations are imported, processed, and exported from (e.g. Thorium Reader). The latter is already defined as the generator of the AnnotationSet.  

If such a case the Creator's type will become "Type of creator. It MUST be "Person", "Organization" or "Software"."

"Software" is the value defined by the W3C Annotation Model.

gautierchomel:
I am not sure whether this is a new use case or belongs to [Annotations used in the publishing workflow](https://w3c.github.io/epub-specs/wg-notes/annotations-ucr/#annotations-used-in-the-publishing-workflow).

gautierchomel:
Both commits are identical; I had a surprise while publishing my branch and ended up with two similar commits.

iherman:
This was discussed during the [pmwg meeting on 18 December 2025](https://w3c.github.io/pm-wg/minutes/2025-12-18-ann.html#e751).

<details><summary><i>View the transcript</i></summary>

<h3>issue <a href="https://github.com/w3c/epub-specs/issues/2835" rel="noopener noreferrer">#2835</a>: <a href="https://github.com/w3c/epub-specs/issues/2835" rel="noopener noreferrer">w3c/<wbr>epub-specs#2835</a></h3>
<p><cite>LaurentLM:</cite> This is a request to add a software agent as creator of an annotation</p>
<p><cite>gautierchomel:</cite> We already discussed with DAISY, there use cases where we might want these displayed as a set<br>
<span>… and you may have tools that generate annotations, e.g a tool from ACE</span><br>
<span>… So it would be good show this to the user</span></p>
[...]

PR #2856, opened 2025-12-17
Use case regarding software as creator of annotations

gautierchomel:
Following up discussions in #2835


iherman:
I do not know whether such tools exist, but it makes sense to me.

gautierchomel:
This should be merged as the corresponding has been added to https://w3c.github.io/epub-specs/epub34/annotations/#1-2-creator

PR #2861, opened 2025-12-18
Updated some erroneous illustrations for EPUB 3.3

### LINKED ISSUE #2845: Incorrect examples in synthetic spreads section
### Section

https://www.w3.org/TR/epub-33/#spread

### Describe the problem

Examples 58, 59, and 60 all show 3 pages of content turning into 4 pages when spreads are enabled. It appears the intent is to duplicate the right hand page of a spread as the left hand page of the next spread.

Rework the sample to show spreads as [1,2], [3,-] instead of [1,2], [2,3].

iherman:
This is to fix #2845 : the three illustrations, and their alt text descriptions, are updated.

Note that this change _is for the EPUB 3.3 only_. In the EPUB 3.4 case some of these illustrations might become moot (because the relevant property may be declared as outdated). Furthermore, completely new illustrations may have to be used, because xkcd example has only three panels, which is not the best option. However, we have the obligation to handle errors in EPUB 3.3, hence this PR.


iherman (commented):
Once merged, we will have to go through a Recommendation update. This is a purely editorial change, so it should not create process issues.

bduga (approved):
lgtm. It turns out this was also a bad choice because pages 1 and 3 are nearly identical and hard to tell the difference at small screen size (e.g. laptop). But we can worry about that later.

sueneu (commented):
I can’t comment directly on this line here, perhaps there is a better way?

>7463  <p>The first tablet views includes a single, column-like strip (i.e., a rectangle

<p>The first schematic tablet displays a single, column-like strip (i.e., a rectangle

mattgarrish:
The descriptions are really hard to read, even the one that still exists. There aren't multiple tablets, for example. There's one tablet showing a progression of views. etc. etc. If you can update the images in the current spec, if any need updating still, then perhaps we can handle the descriptions separately with the a11y task force.

sueneu:
@mattgarrish Lordy! Don’t break anything for a little edit here and there. The a11y TF sounds like the perfect people to work with if it makes sense to go there right now.

Should I delete my comment?

mattgarrish:
> Note that this change _is for the EPUB 3.3 only_.

Blah, I should have read the first comment instead of looking at the files.

But if we're revising the 3.3 rec anyway, it would still be good to review the descriptions again.

iherman:
> > Note that this change _is for the EPUB 3.3 only_.
> 
> Blah, I should have read the first comment instead of looking at the files.
> 

:-)

> But if we're revising the 3.3 rec anyway, it would still be good to review the descriptions again.

I am not sure. The goal of this PR (and subsequent republication) is to settle an illustration that is visibly and clearly wrong; actually harmful insofar as it leads the reader to a wrong conclusion. Other than that we should not do anything and concentrate on the 3.4 spec for any real improvements...

mattgarrish:
Ya, I'm not too keen on spending time on 3.3. I'm guessing no one's really read these descriptions anyway. They're very literal in their approach - leaning more into what is represented rather than what is being conveyed.

iherman:
@mattgarrish is it o.k. to merge this? I do not want to keep such PR-s for too long, it takes away our attention from the main development.

@shiestyle @reidmore-online @sueneu, once we are in 2026 we should think of re-publishing an editorial update of EPUB 3.3 with this error handled.

mattgarrish:
> is it o.k. to merge this?

Fine with me, but I'm not sure how it works now to republish an old spec. Are you going to manually export and push a new REC?

mattgarrish (approved):

iherman:
> > is it o.k. to merge this?
> 
> Fine with me, but I'm not sure how it works now to republish an old spec. Are you going to manually export and push a new REC?

That is the idea. Will check that with Denis, but this change is a typical case of minor editorial changes of an error, so it should not be complicated. I will pick this up in January.

shiestyle (approved):

PR #2865, opened 2025-12-20
Renew spread illustrations

iherman:
With a number of spread related properties outdated, most of the old illustrations became unused, and are removed (from that part of) the repository. 

The two remaining illustrations have been updated (also handling #2845) using images.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2865.html" title="Last updated on Dec 20, 2025, 12:20 PM UTC (b58535a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2865/6ad59e3...b58535a.html" title="Last updated on Dec 20, 2025, 12:20 PM UTC (b58535a)">Diff</a>


HadrienGardeur:
@iherman did you intentionally moved from a CC-licensed to a public domain publication for any particular reason?

I think that ideally we could use examples that illustrate really well the need for these properties.

This means:

- using a freely licensed continuously scrolled comic for roll
- and a publication where spreads are important for spread placement

iherman:
> @iherman did you intentionally moved from a CC-licensed to a public domain publication for any particular reason?
> 

I did not do it due to the licensing thing. The xkcd that I used previously had the downside that it consists of barely three "plates", which meant that correctly illustrating things was not really possible (see also #2845). This collection is larger and it also has the advantage of making very clear what the correct order/sequence is (ie, letters).

Does it make any difference for us if it is CC or simply public domain? I do not see any.

> I think that ideally we could use examples that illustrate really well the need for these properties.
> 
> This means:
> 
> * using a freely licensed continuously scrolled comic for roll
> * and a publication where spreads are important for spread placement

I am all open for alternatives, but I have the impression that, at least for these two illustrations, this example works. Even if we had a "real" comics, we would end up making screen dumps and illustration like these in the spec itself.

(Although the original of these images is a PDF file on the LoC site, I have the impression that it would make a perfect children's book in FXL, though not necessarily roll. But @sueneu is the person who has experience in this area...)

reidmore-online (approved):

dalerrogers:
I have a poetry comic that I'm willing to relicense for this purpose. It is hybrid. It has reflowable elements, fixed layout elements, and spread content. Let me know.

iherman:
> I have a poetry comic that I'm willing to relicense for this purpose. It is hybrid. It has reflowable elements, fixed layout elements, and spread content. Let me know.

Thanks. We will have to update the test suite for all this section; it may come handy!

HadrienGardeur:
> Does it make any difference for us if it is CC or simply public domain? I do not see any.

I was just wondering if there was a particular reason, which doesn't seem to be the case.

I'll investigate whether I can find any CC-licensed continuously scrolled comic, but in the meantime this looks good to me.

HadrienGardeur (approved):

iherman:
@mattgarrish @sueneu @shiestyle : any problems merging this? This is a purely editorial set of changes, nothing formal about it...

sueneu:
I am OK with merging this

shiestyle (approved):
I'm OK, too.

PR #2873, opened 2026-01-06
Add aural rendering and tts sections

### LINKED ISSUE #2816: The future of TTS and Media Overlays
### Section

https://www.w3.org/TR/epub-rs-34/#sec-media-overlays

### Describe the problem

With these things in mind:

- The timeline for 3.4 implementation and adoption
- The increasing advance of AI voicing for TTS
- The reluctance of specific markets to support media overlays (thinking non-trade, academic as an example)

Is the 3.4 release an opportunity to introduce a broader topic for what is currently Media Overlays, and suggest an approach in the spec to TTS as a whole that could be AI voicing with highlighting OR media overlays?

_No response_

mattgarrish:
Opening this as a draft as this is still somewhat exploratory in terms of what we want to say.

In a nutshell, I've added a new "Aural rendering" section and put TTS and media overlays underneath. I can't think of a lot to say for authors in terms of actually improving TTS, since most of the dedicated technologies don't work. The section ended up informative with a brief recap of what we were originally hoping to do for TTS and how the technologies may yet get revived in a simpler or different form.

For the reading systems spec, I've added a section with support labelled as optional. I added some recommendations for feature support which coupled with optional support would hopefully mean we don't need to show implementations at this time. I also noted some other features that dedicated AT provide that could be useful, but we likely want to go too far down the road of trying to define read aloud as a full-fledged AT.

Feedback welcome on all of this. Consider it exploratory options, especially on the reading system side.

Fixes #2816 

***

Reading Systems: [Preview](https://raw.githack.com/w3c/epub-specs/reorg/add-audio-rendering/epub34/rs/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/rs/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/reorg/add-audio-rendering/epub34/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2873.html" title="Last updated on Jan 22, 2026, 8:24 PM UTC (b92ddea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2873/6f5cdab...b92ddea.html" title="Last updated on Jan 22, 2026, 8:24 PM UTC (b92ddea)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -7772,8 +7772,8 @@ No Entry</pre>
 			<section id="sec-aural-intro">
 				<h3>Introduction</h3>
 
-				<p>Although EPUB 3 is primarily considered a visual reading format, EPUB publications are designed to
-					enable their aural rendering. There are two primary ways this is done:</p>
+				<p>Although EPUB 3 is primarily considered a visual reading format, [=EPUB publications=] are designed
+					to enable their aural rendering. There are two primary ways this is done:</p>
 
 				<ul>
 					<li>through the <a href="#sec-tts">text-to-speech rendering</a> of an EPUB publication's text
@@ -7783,10 +7783,18 @@ No Entry</pre>
 				</ul>
 
 				<p>The experience of reading an EPUB publication via text-to-speech rendering is controlled by a user's
-					reading system using built-in synthetic speech voices. This feature is commonly referred to as "read
-					aloud" capability, although the exact naming can differ from reading system to reading system. There
-					is currently not a lot that can be done to enhance the quality of the text-to-speech experience from
-					the authoring side, except to ensure that EPUB publications are produced accessibly.</p>
+					[=reading system=] using built-in synthetic speech voices. This feature is commonly referred to as
+					"read aloud" capability, although the exact naming can differ from reading system to reading
+					system.</p>
+
+				<p>Although the read aloud functionality provided by EPUB 3 reading systems improves the accessibility
+					for users with disabilities that require voice assistance, it is not the same as the text-to-speech
+					rendering provided by dedicated assistive technologies (what are generally referred to as screen
+					readers). Read aloud functionality is designed for use by all users as it exists within the reading
+					system, not as a separate application. For example, someone travelling might switch from visual to
+					auditory reading to avoid motion sickness. Although both types of text-to-speech renders could exist
+					on a user's system, and could both be used to read EPUB publications, this specification is only
+					focused on the built-in functionality provided by reading systems.</p>
 
 				<p>Media overlays, on the other hand, rely on the producer of the EPUB publication to provide the audio
 					and synchronize it with the text. Media overlays allow an EPUB publication to double as a more
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -7766,622 +7766,709 @@ No Entry</pre>
 				</aside>
 			</section>
 		</section>
-		<section id="sec-media-overlays">
-			<h2>Media overlays</h2>
+		<section id="sec-aural-rendering">
+			<h2>Aural rendering</h2>
 
-			<section id="sec-overlays-introduction" class="informative">
-				<h4>Introduction</h4>
+			<section id="sec-aural-intro">
+				<h3>Introduction</h3>
 
-				<p>Mainstream ebooks, educational tools and ebooks formatted for persons with print disabilities are
-					some examples of works that contain synchronized audio narration. In EPUB 3, these types of books
-					can be created using media overlay documents to describe the timing for the pre-recorded audio
-					narration and how it relates to the [=EPUB content document=] markup. The specification defines the
-					file format for media overlays as a subset of [[smil3]], a W3C Recommendation for representing
-					synchronized multimedia information in XML.</p>
+				<p>Although EPUB 3 is primarily considered a visual reading format, EPUB publications are designed to
+					enable their aural rendering. There are two primary ways this is done:</p>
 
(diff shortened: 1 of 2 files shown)


iherman:
Purely editorial aesthetics: we have now two sections with identical headings in the RS: both §9.1 and §9.2.3.2 are titled "text-to-speech". It is probably easier to rename the latter ('text' element without 'audio' ?). But looking at the content of that section it may require some more surgery and a link back to the former (I am not familiar enough with these to offer explicit solution).

mattgarrish:
> identical headings in the RS: both §9.1 and §9.2.3.2 are titled "text-to-speech"

Ah, right. I was going to change the MO section in the authoring spec to "Invoking text-to-speech" to make it clearer it was an authoring action but opted to leave it as "Text-to-speech rendering". I just assumed the rs spec had the same heading.

@marisademeglio do you have any preference for the title?

marisademeglio:
"Text-to-speech rendering" sounds fine to me.

While I was reading the draft linked above I wondered why section 9.2.3.2 is under "Interacting with the EPUB content document" instead of "Basic playback"?

mattgarrish:
> While I was reading the draft linked above I wondered why section 9.2.3.2 is under "Interacting with the EPUB content document" instead of "Basic playback"?

I don't have an answer to that other than that's where it's always been. I'd guess it's because we had hoped that there'd be better tts support via the markup, at least going by this:

> Reading systems SHOULD use the speech-related information provided in the target EPUB content document to play the audio stream as part of the media overlay rendering.

There's also the problem that it's not required behaviour, so I guess it doesn't really fit as "basic" playback?

marisademeglio:
Oh maybe it made more sense in the context of coming right after this deprecated section: 
https://www.w3.org/TR/epub-rs-33/#sec-embedded-media

Could it move to "basic playback" if it were instead just called "playback"?

mattgarrish:
> Could it move to "basic playback" if it were instead just called "playback"?

Ya, that makes more sense to me. But then do we need the "Interacting with the EPUB content document" section if all it does is wrap "Navigation"?

Could we drop "Interacting with the EPUB content document" and maybe change "Navigation" to be a little more self-explanatory, like "Navigating within EPUB content documents"? (Does that, in turn, fit under "Playback"?)

marisademeglio:
Yes I could see either way being just fine. I have a slight preference for it being its own heading "Navigating within EPUB content documents" at the same hierarchy level as "Playback". 

One day I'd like to stop recommending that we put skip/escape semantics in SMIL and instead use them directly from the content document. So then we might have a section about interacting with the content document with subsections "navigation" and "skip/escape". But that might not be for a 3.4 revision.

mattgarrish:
Removing the draft status on this as the discussion has gone quiet, so either everyone agrees or no one is reading it... 😄 

@rickj any concerns with this, in particular with any of the reading systems recommendations?

@avneeshsingh it would be good to have the a11y task force review the recommendations, too.

rickj:
> Removing the draft status on this as the discussion has gone quiet, so either everyone agrees or no one is reading it... 😄
> 
> @rickj any concerns with this, in particular with any of the reading systems recommendations?
> 
> @avneeshsingh it would be good to have the a11y task force review the recommendations, too.

My silence is agreement with the approach.  I think it resolves my initial concerns.

iherman (approved):
A minor discrepancy

- in RS, §9.2.2.3 says: "… reading systems capable of text-to-speech (TTS) playback _SHOULD_ render the referenced text using TTS"
- In the authoring spec, the corresponding §9.3.2.2.7 says "…In the absence of a sibling [audio](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2873.html#dfn-audio) element, textual content referred to by this element can be rendered via [text-to-speech](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2873.html#sec-mo-tts)."

I.e., _SHOULD_ in one case and the neutral "can" on the other. Shouldn't these two be in sync, i.e., say a _SHOULD_ in the authoring spec as well? (or the other way round?)

Something else: is it worth pushing for new tests for TTS? I would not be the one doing them, not having the right experience.

mattgarrish:
> I.e., _SHOULD_ in one case and the neutral "can" on the other. Shouldn't these two be in sync, i.e., say a _SHOULD_ in the authoring spec as well? (or the other way round?)

No, reading system rendering requirements don't belong in the authoring spec. But if you want it to sound a little closer to the recommendation it could probably be changed to:

> In the absence of a sibling [audio](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2873.html#dfn-audio) element, a reading system is expected to render the textual content referred to by this element via [text-to-speech](https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2873.html#sec-mo-tts).

mattgarrish:
I've made some changes in the last updates to reflect the discussions on the TF call today. I've added more context that read aloud is not a screen reader and trimmed the reading system feature list to match what media overlays offers. The rest I deleted assuming detailed playback features will go in the future note.

iherman:
Any reason you keep this open, @mattgarrish ? I thought this could be merged...

mattgarrish:
> Any reason you keep this open, @mattgarrish ?

Just because of the changes asked for on the last TF call. I'm giving it until this Thursday for any additional feedback and then I'll merge.

PR #2892, opened 2026-01-18
Annotation selectors, first draft

llemeurfr:
(This PR is now outdated, overtaken by #2912.)

I added a section related to Selectors to the Draft. 

---
See:

* For EPUB 3 Annotations:
    * [Preview](https://raw.githack.com/w3c/epub-specs/ann-selectors-1/epub34/annotations/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/annotations/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/ann-selectors-1/epub34/annotations/index.html)

### DIFF
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -479,9 +479,9 @@
 				</pre>
 			</aside>
 
-			<p class="issue"> Should we define another Progression Selector as a value relative to the whole publication,
-				 rather then the current resource? This would help synchronize annotations with publications in a format
-				 different from reflowable EPUB, such as PDF.</p> 
+			<p class="issue"> Should we also define a global progression selector, relative to the whole publication? 
+				 This would help synchronize annotations with publications distributed in a format
+				 different from reflowable EPUB, such as PDF. It would make sense to define it as Meta information (still to be defined)</p> 
 				
 			<!--
 			<h4 id="1-3-3-meta"> Meta </h4>
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -312,13 +312,14 @@
 			<p> A Target with no Selector indicates that the annotation applies to the entire
 				target resource. </p>
             <section>
+
 			<h4 id="1-3-1-source"> Source </h4>
 			<p> The target resource MUST be identified by the URL of an existing resource in the
 				EPUB package. It MUST be one of the `item/@href` values of the [^manifest^] element as
                 defined in [[epub-34]].</p>
 
-            <aside class="example" title=" the source of the annotation is the relative URL identifying an HTML document in an EPUB">
-			<pre>
+      <aside class="example" title=" the source of the annotation is the relative URL identifying an HTML document in an EPUB">
+				<pre>
                 {
                     "@context": [
                         "http://www.w3.org/ns/anno.jsonld",
@@ -335,54 +336,65 @@
                         }
                     }
                 }
-			</pre>
-            </aside>
+				</pre>
+      </aside>
             </section>
+
             <section>
 			<h4 id="1-3-2-selector"> Selector </h4>
 			<p> An annotation refers to a segment of a resource, which is identified by one or more
-                <a data-cite="annotation-model#selectors">Selectors</a>. The nature of the Selectors and methods to describe segments depend on
+        <a data-cite="annotation-model#selectors">Selectors</a>. 
+				The nature of the Selectors and methods to describe segments depend on
 				the resource type. Providing more than one Selector allows an annotation software to
 				choose the most accurate selector from those it can handle and helps accommodate
 				evolutions on the annotated resource. </p>
-			<p> Annotation selectors are specified in <a data-cite="annotation-model#selectors">Web
-					Annotation Data Model</a>. This specification retains
-				selectors deemed useful for annotating publications and details on how to use these
-				selectors. </p>
+			<p> Several annotation selectors are specified in <a data-cite="annotation-model#selectors">Web
+					Annotation Data Model</a>. 
+				This specification retains selectors deemed useful for annotating EPUB publications,
+				 and details on how to use these selectors. It also defines new selectors as needed.</p>
 			<p class="note">New selectors will undoubtedly be defined in the coming months after
 				discussion with members of the W3C Publishing Maintenance Working Group.</p>
 
-			<!--
-			<h5 id="1-3-2-1-text-quote-selector"> Text Quote Selector </h5>
-			<p> This Selector describes a range of text by copying it and including some of the
-				text immediately before (a prefix) and after (a suffix) it to distinguish between
-				multiple copies of the same sequence of characters. </p>
-			<p> Whitespaces present in the source document are preserved in the “prefix”, “exact”,
-				and “suffix” segments and are represented as per the JSON Grammar (e.g. <code> \n
-				</code> , <code> \t </code> etc.). </p>
-			<p> Note: There is no restriction on the preceding and following text that can be
-				included in the selector, but this amount should be left as low as possible. </p>
-			<p> Important: This selector contains complete annotated text segments and, therefore,
-				must not be used when the publication is protected by a DRM, which limits the number
-				of characters that may be copied.</p>
-
-			<p> Sample 3: A text segment represented as a TextQuoteSelector. </p>
+			<h5 id="1-3-2-1-text-fragment-selector"> Text Fragment Selector </h5>
+			<p> A Text Fragment Selector, as defined in the 
+				W3C WICG URL Fragment Text Directives Draft Community Group Report,
+				 contains a shortened version of the annotated text segment,
+				 and optional indication of the preceding and following fragments.</p>
+			<p> In this specification, a Text Fragment Selector is represented with a simplified format: </p>
 			<pre>
-				<code class="lang-json">
-					{
-					  "selector": [
-					    {
-					    "type": "TextQuoteSelector",
-					    "prefix": "trouver quelqu’un    \n  \t\t    comme vous. ",
-					    "exact": "Combien de fois \n\n    ne m’avait-il",
-					    "suffix": " pas \n\n      reproché de travailler ma"
-					    }
-					  ]
-					}
-				</code>
+				[prefix-,]start[,end][,-suffix]
 			</pre>
-			<h5 id="1-3-2-2-cssselector-textpositionselector"> CssSelector +
-				TextPositionSelector </h5>
+
+      <aside class="example" title=" a text segment represented as a TextFragmentSelector">
+				<pre>
+					<code class="lang-json">
+						{
+							"selector": [
+								{
+								"type": "TextFragmentSelector",
+								"value": "an example,text fragment"								}
+							]
+						}
+					</code>
+				</pre>
+			</aside>
+
+			<p class="note"> Because text fragments are not used as fragment URLs,  
+				there is no need for percent-encoding in our context. </p>
+			<p class="note"> This selector contains annotated text segments and, therefore,
+				must be used with great care when the number of characters that can be copied 
+				from the publication is constrained by the publisher.</p>
+			<p class="note"> This selector is a modernized version of the 
+				<a data-cite="annotation-model#text-quote-selector"><code>Text Quote Selector</code></a> 
+				originally defined by the W3C Web Annotation Data Model.</p>
+			<p class="issue">The integration of this selector in reading systems is very challenging: 
+			 the absence of a standard API in web browsers makes mapping a text fragment to a DOM Range difficult.
+			 Rebuilding a DOM range from a textual range using tree walking is cumbersome and suboptimal.
+			 This selector, while appealing, will not be retained unless a public API is rapidly defined and 
+			 implemented in major web browsers.</p>
+
+			<h5 id="1-3-2-2-cssselector-textpositionselector"> Css Selector refined by
+				Text Position Selector </h5>
 			<p>References in the Web Annotation Data Model: <a
 					href="https://www.w3.org/TR/annotation-model/#css-selector"> CSS Selector </a> ,
 					<a href="https://www.w3.org/TR/annotation-model/#text-position-selector"> Text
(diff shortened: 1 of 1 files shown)


epub34/annotations/index.html
```diff
@@ -335,97 +336,100 @@ <h4 id="1-3-1-source"> Source </h4>
...
-				included in the selector, but this amount should be left as low as possible. </p>
-			<p> Important: This selector contains complete annotated text segments and, therefore,
-				must not be used when the publication is protected by a DRM, which limits the number
-				of characters that may be copied.</p>
-
-			<p> Sample 3: A text segment represented as a TextQuoteSelector. </p>
+			<h5 id="1-3-2-1-text-fragment-selector"> Text Fragment Selector </h5>
+			<p> A Text Fragment Selector, as defined in the 
+				W3C WICG URL Fragment Text Directives Draft Community Group Report,
+				 contains a shortened version of the annotated text segment,
```
iherman, line 356:
```suggestion
			<p> A Text Fragment, as defined in the 
				W3C WICG URL Fragment Text Directives Draft Community Group Report,
				 contains a shortened version of the annotated text segment,
```

(The WICG does not define the selector, at least not in our terminology, but the fragment itself.)

iherman:
Comments on the Text Fragment Selector

I wonder whether it is worth defining a separate Selector class for this. After all, a text fragment is a URL fragment, meaning that it can be subsumed by the (existing) [Fragment Selector](https://www.w3.org/TR/annotation-model/#fragment-selector) adding a reference to the text fragments specification to the table.

As a "bonus", we can also get add other fragment selectors into our specs which, I believe, should be added in the first place. From the [table](https://www.w3.org/TR/annotation-model/#fragment-selector) in the Web Annotation spec, we may want to retain the HTML fragment, Media fragments (or subparts of it, see separate discussion with @HadrienGardeur ), and the SVG selector which is a way to add annotation to an SVG file (apart from a CSS selector).

(I realize that the value must be percent encoded then to be stored, but at some point it must be done anyway...)

iherman:
Comment on Css Selector refined by Text Position Selector

I am not sure why this arbitrary combination? I could imagine a Css Selector refined by a text fragment (the same text can be repeated in the same resource, and this may be a way to make it unique).

Wouldn't it be cleaner if we say

1. We accept the following simple selectors: Css, Fragment, Position, etc.
2. We accept refinement of selectors by the combination of these

It may be that some combination should be explicitly excluded if it raises some major implementation problems. I would prefer to do that, rather than fix a single combination.

---

B.t.w., I just realized that the Text Position Selector is somewhat undefined. Does it refer to the position of a byte in a stream (it says something [like that](https://www.w3.org/TR/annotation-model/#:~:text=T-,his%20Selector%20describes%20a%20range%20of%20text%20by%20recording%20the%20start%20and%20end%20positions%20of%20the%20selection%20in%20the%20stream,-.%20Position)). Is that o.k. with us? Shouldn't it refer to Unicode Code Points in our case within the resource?

iherman:
Comments Progression selector

Sounds interesting, but needs a somewhat more precise definition (if at all possible). Percentage of what exactly? Percentage of the characters in the HTML code? Percentage of the visible text content (that means white spaces must be removed if applicable)? Percentage within the file storing the content in terms of bytes?

I would expect there is a way this is done (albeit, as you say, it is not precise) in Reading Systems; is it possible to turn that into a specification text, to make it clear?

epub34/annotations/index.html
```diff
@@ -438,29 +442,46 @@ <h5 id="1-3-2-2-cssselector-textpositionselector"> CssSelector +
...
+								"type": "ProgressionSelector",
+								"value": 0.534234255
+								}
+							]
+						}
+					</code>
+				</pre>
+			</aside>
 
+			<p class="note"> This selector dos not exist in the [[annotation-model]]. </p>
```
iherman, line 475:
```suggestion
			<p class="note"> This selector does not exist in the [[annotation-model]]. </p>
```

iherman:
This was discussed during the [pmwg meeting on 22 January 2026](https://w3c.github.io/pm-wg/minutes/2026-01-22.html#7e88).

<h3>Selectors for Annotations - <a href="https://github.com/w3c/epub-specs/pull/2892" rel="noopener noreferrer">w3c/<wbr>epub-specs#2892</a></h3>
<p><cite>LaurentLM:</cite> [sharing screen]<br>
<span>… in an annotation there is a target</span><br>
<span>… the src references the epub</span><br>
<span>… the file in the epub</span><br>
<span>… the selector just deals with the exact points in the book that are part of the annotation</span><br>
<span>… [walking through example]</span><br>
<span>… in an annotation there can be multiple selectors, but they must point at the same place</span><br>
<span>… our first selector is text fragment</span><br>
<span>… this is from a community group, not a proper standard</span><br>
<span>… we will need to make TextFragmentSelector, as it isn't in webannotations</span><br>
<span>… the format is easy for that selector, it can be made a little trickier</span><br>
<span>… some notes in the annotation doc</span><br>
<span>… 1. we don't need to percent encode since it isn't in a url</span><br>
<span>… if there is a constrained on copied characters, you should use it with care</span><br>
<span>… don't put too much text!</span><br>
<span>… in the original model there was a text quote selector</span><br>
<span>… we can consider this new one as a modernized version of that selector</span><br>
<span>… one negative is there is no API currently in scripting, and no good polyfills</span><br>
<span>… so you have to manually walk the DOM</span><br>
<span>… also 1 question - should we add url fragment syntax? I didn't as I wanted to keep it short, and the static text doesn't seem necessary</span><br>
<span>… so should we add the static text</span></p>
<p><cite>&lt;sueneu&gt;</cite> ack: ivan</p>
<p><cite>ivan:</cite> It is a CG document, but it is also the topic of active work in HTML<br>
<span>… I understand the argument with encoding</span><br>
<span>… but isn't this hiding the underlying URL fragment?</span><br>
<span>… since that is how it will be done in HTML</span><br>
<span>… that avoids parallel specing</span></p>
<p><cite>&lt;Hadrien&gt;</cite> +1 for what ivan just said, we should use FragmentSelector + conformsTo instead of creating TextFragmentSelector</p>
<p><cite>ivan:</cite> and the last thing is, if I select a text, and some is in an html element, how is that handled</p>
<p><cite>LaurentLM:</cite> You just keep the text and ignore the elemenrs</p>
<p><cite>&lt;sueneu&gt;</cite> ack: duga</p>
<p><cite>&lt;sueneu&gt;</cite> duga: How well is the processing model for this group developed?</p>
<p><cite>&lt;sueneu&gt;</cite> …is the detail about how to do this specified somewhere?</p>
<p><cite>&lt;sueneu&gt;</cite> <code>@ivan</code>: it is being considered in the HTML group</p>
<p><cite>&lt;Hadrien&gt;</cite> Old document about this: <a href="https://wicg.github.io/scroll-to-text-fragment/" rel="noopener noreferrer">https://<wbr>wicg.github.io/<wbr>scroll-to-text-fragment/</a></p>
<p><cite>&lt;sueneu&gt;</cite> duga: then it is scary to me to put this in our spec</p>

<hr /></details>

iherman:
As a follow-up on the discussion at the last meeting: as @bduga said, if we define our own Text Fragment selector, then it would be incumbent on us to define the full processing of that selector in an HTML content. Not only would that be a large amount of work, but it would also duplicate efforts happening elsewhere, namely in the HTML Working Group. See

- https://github.com/whatwg/html/pull/11895
- https://github.com/whatwg/html/issues/11938
- https://github.com/whatwg/html/issues/11968

and probably more that I miss. All this reinforces my opinion that we should follow what I described in https://github.com/w3c/epub-specs/pull/2892#issuecomment-3767565866.

We will have to monitor what is happening in the WhatWG on that matter and, hopefully, that work will be finalized by the time we go to Rec.

HadrienGardeur:
I don't see any good reason for defining our own `TextFragmentSelector` either.

We should use `FragmentSelector` together with `conformsTo` to point to the future section of the HTML specification that will contain text fragments.

In the meantime we can use the URI of the old draft for this feature.

```json
{
  "selector": {
    "type": "FragmentSelector",
    "conformsTo": "https://wicg.github.io/scroll-to-text-fragment/",
    "value": "#:~:text=that%20a%20single%20man%20in%20possession%20of%20a%20good%20fortune%20must%20be%20in%20want%20of%20a%20wife"
  }
}
```

HadrienGardeur:
> Comments Progression selector
>
> Sounds interesting, but needs a somewhat more precise definition (if at all possible). Percentage of what exactly? Percentage of the characters in the HTML code? Percentage of the visible text content (that means white spaces must be removed if applicable)? Percentage within the file storing the content in terms of bytes?
>
> I would expect there is a way this is done (albeit, as you say, it is not precise) in Reading Systems; is it possible to turn that into a specification text, to make it clear?

I'm currently working on progression sync in OPDS where this will be the only required value. As you've pointed out, this needs further explanations on how it should be calculated.

Some thoughts on this:

- For pre-paginated publications, it's easy enough: there's a total index of pages and your current page, you can use both information to calculate a progression.
- For roll publications, we ideally need to know the total height of each resource in pixel and then calculate the current scroll value (not as easy to do, ideally we would need images in spine + the height of each of them).
- For reflowable publications it's even harder and you need to remember that a RS might be in a context where the resource is streamed. 

Common scenarios for reflowable publications include: 
- Rendering the entire publication would be the most accurate way of calculating progression, but it's extremely ineffective in terms of performance. 
- Calculating the number of characters would require going through the entire publications, which doesn't sound feasible in a context where you're streaming.
- You could rely on print pagination, but it's not always available and might be tricky to calculate.

In Readium toolkits, we've taken a page from the old RMSDK toolkit from Adobe (pun intended) and we use something that's not perfect but can work in many scenarios:
- we extract the number of bytes of each resource in the spine
- for every 1024 bytes, we consider that the resource contains one "position"
- this gives us a "position list" where we know the total number of positions for the entire publication and we can already calculate the progression for the beginning and end of each resource
- while reading, we can calculate the progression in a given resource by looking at the scroll offset over the entire scroll of that resource
- with a progression in a resource and the position list, we can calculate a total progression in a publication (or the other way around)

This isn't perfect (books with a lot of images tend to have less positions than they should), but it gets the job done and we've successfully implemented this across several plaforms where it yields consistent results.

HadrienGardeur:
> Wouldn't it be cleaner if we say
>
> 1. We accept the following simple selectors: Css, Fragment, Position, etc.
> 2. We accept refinement of selectors by the combination of these

I think that's the right approach as long as we have clear requirements.

IMO, both fragment selector and progression should be required, while CSS Selectors should be optional but useful for refinement.

iherman:
> I don't see any good reason for defining our own `TextFragmentSelector` either.
> 
> We should use `FragmentSelector` together with `conformsTo` to point to the future section of the HTML specification that will contain text fragments.
> 

Actually... we might reduce the formal dependency on the WhatWG work by simply adopting the official [Text Quote selector](https://www.w3.org/TR/annotation-model/#text-quote-selector) as is, and consider the text fragments as an implementation tool (we might add a note on that effect in the spec). After all, the JSON file that we create is primarily meant as an exchange format, the fact that, internally, a URL fragment is used is an implementation issue.

iherman:
Comments on the progression selector, per https://github.com/w3c/epub-specs/pull/2892#issuecomment-3794250014:

I begin to get a bit worried about this thing. The note of @HadrienGardeur shows that it may become impossible (or require an inordinate amount of work) to properly specify this selector. We may get away if we specify it at a high level via  a declarative description of what it is. Accepting that using the selector might not ensure proper interoperability among _different_ RS-s (ie, not among several instances of the same RS), because the implementation strategies might differ a lot.

Would that be acceptable in the spec? I presume the answer is how we define it, more exactly what the usage of it is. I presume the main use case is to define a bookmark that does not depend on a specific reading systems instance: then we might be able to specify this more precisely, and declare that the selector is for a very restricted use only, namely for bookmarks.

---

As an aside: is the RMSDK in public domain? Is it under a patent protection of Adobe? Because if it is the latter, and we want to reuse that algorithm, we have to get the authorization of Adobe...

HadrienGardeur:
> As an aside: is the RMSDK in public domain? Is it under a patent protection of Adobe? Because if it is the latter, and we want to reuse that algorithm, we have to get the authorization of Adobe...

It's not public domain and the ownership of this codebase has changed twice over the last 12 months (from Adobe to Harman and now from Harman to Wipro).

To be clear, I don't have a clue what RMSDK does exactly and the Readium implementation never tried to reverse engineer that specific implementation. 
It's been public knowledge for a long time that they tie positions to a specific number of bytes, that's the only details where the Readium implementation is aligned with the Adobe one.

iherman:
This was discussed during the [pmwg meeting on 29 January 2026](https://w3c.github.io/pm-wg/minutes/2026-01-29.html#5400).

<h3>Selectors in Annotations - <a href="https://github.com/w3c/epub-specs/pull/2892" rel="noopener noreferrer">w3c/<wbr>epub-specs#2892</a></h3>
<p><cite>Laurent:</cite> Providing overview of selectors<br>
<span>… page being explained is <a href="https://www.w3.org/TR/annotation-model/" rel="noopener noreferrer">https://<wbr>www.w3.org/<wbr>TR/<wbr>annotation-model/</a></span></p>
<p><cite>duga:</cite> model syntax is referring to a spine item not a manifest item. A manifest item is not necessarily in the spine.<br>
<span>… referring to an item in the Spine is easier than referring to an item in the manifest.</span></p>
<p><cite>Laurent:</cite> Showing examples of selectors.</p>
<p><cite>ivan:</cite> The text selector was designed for a URL not HTML<br>
<span>… before this will go to the recommendation it will be looked at by other groups.</span><br>
<span>… I would prefer to use the work done by the web development group, rather than define the definition of the selector in the Working Group</span></p>
<p><cite>duga:</cite> What is our fallback position if we propose solution that has not been defined by another group?</p>
<p><cite>Laurent:</cite> There is a different fallback solution.</p>
<p><cite>ivan:</cite> The Text Fragment selector is the preferred solution. but if it is not specified, then there is another solution already proposed.</p>
<p><cite>Laurent:</cite> The Spatial selector will solve problems.</p>
<p><cite>Hadrien:</cite> I would rather reference solutions from other groups, than have to define something so technical in this group.</p>
<p><cite>&lt;sueneu&gt;</cite> +1 Hadrien</p>
<p><cite>ivan:</cite> We will be in contact with that group before this proposal goes for recommendation. So we will be able to adjust the proposal before hand.</p>
<p><cite>Laurent:</cite> The CSS selector is an alternate solution to the Text Fragment selector</p>
<p><cite>sueneu:</cite> Do I need to identify an element in HTML in order to find it a CSS selector?</p>
<p><cite>Laurent:</cite> No, CSS selectors will find content even if it is not specifically identified with an attribute.</p>
<p><cite>Hadrien:</cite> Do we want this to be general enough that it will work in different applications than EPUB? What is the scope?</p>
<p><cite>wendyreid:</cite> EPUBs are rendered on different types of devices. Our solutions need to be aware of the limitation is processing capability of older reading systems.<br>
<span>… it depends on the inner structure on how data is loaded in an older reading system. The backwards compatibility is an issue.</span></p>
<p><cite>Hadrien:</cite> Reading systems will optimize their approach.</p>
<p><cite>duga:</cite> Do we focus on resiliency over performance? +1 for CSS.</p>
<p><cite>ivan:</cite> We have run into this technical hurdle before. CFI?<br>
<span>… let's use the technology CSS provides rather than trying to define something ourselves.</span></p>
<p><cite>&lt;ivan&gt;</cite> +1 wendy</p>
<p><cite>wendyreid:</cite> Having run into problems with bookmarks in the past, we need a robust solution.</p>
<p><cite>duga:</cite> We need to keep the number of selectors small: Text, spatial, audio. We need a solution that will be actually implemented. Don't provide too many options for solving the same problem.</p>
<p><cite>sueneu:</cite> Would the solutions work with older EPUBS?</p>
<p><cite>Hadrien:</cite> It depends.</p>
<p><cite>ivan:</cite> Based on today's discussion. Laurent will come up with different
    wording.</p>

<hr /></details>

iherman:
#2912 being merged, this PR becomes moot. Closing without merge.

Issue #2896, opened 2026-01-21
Adding support for JXL (jpeg-xl)


DieselTech:
### Section

https://w3c.github.io/epub-specs/epub34/authoring/#sec-core-media-types

### Describe the problem

I'd like to see support for JXL spec in the next release of the epub 3 specification. Adoption in other software is already rapidly occurring.

### Describe the fix or new feature you propose

With the addition of [jxl-rs being added to chrome](https://chromium-review.googlesource.com/c/chromium/src/+/7184969?tab=checks) the biggest obstacle to adoption of the format on the web is rapidly coming to an end. To contrast that, there is already plenty of non-browser applications that natively support JXL these days: https://github.com/libjxl/libjxl/blob/main/doc/software_support.md ,including a handful of other browsers that have already adopted the standard but just don't have the market share. A notable example here is Apple adding native decoding directly to the OS since iOS 17. Also throwing in their support for the format is the PDF association who [announced in late 2025 that they will add jxl](https://pdfa.org/wp-content/uploads/2025/10/[PDFDays2025-BreakingBad-Wyatt.pdf](https://pdfa.org/wp-content/uploads/2025/10/PDFDays2025-BreakingBad-Wyatt.pdf#page=11)#page=11) as the preferred solution for embedding High Dynamic Range (HDR) content in the PDF specifications.

The major notable feature for JXL is the space savings of existing images that are in JPEG format. The conversion process can save ~10-20% more than webp conversions while still maintaining the ability to recreate a bit perfect JPEG again when the process is reversed. 

> JPEG XL codec can losslessly transcode a widely supported subset of JPEG files, by directly copying JPEG's DCT block coefficients to 8×8 VarDCT blocks, making smaller file sizes possible due to JPEG XL's superior entropy coding. This process is reversible and it allows for the original JPEG file to be reconstructed bit-for-bit, although constraints limit support for some files

iherman:
I would probably keep this on hold for a while. It will take a while before JXL lands in various EPUB reading systems (which mostly depend on external web rendering engines, so that will happen, eventually). Being picked up by publishers is another matter because it also requires to be present in various authoring tools.

Putting the format into the standard right now would require us to add some tests and prove that it is already implemented by _some_ reading systems. This should happen within, say, a year. At this moment I do not think that is the case, but I am happy if I am proven wrong.

@HadrienGardeur  @bduga @reidmore-online @davemanhall @shiestyle

HadrienGardeur:
[...]

PR #2897, opened 2026-01-21
Fixes for manifest fallback processing

### LINKED ISSUE #2893: Foreign content document is not mentioned for fallback processing in the RS spec
### Section

https://w3c.github.io/epub-specs/epub34/rs/#sec-epub-rs-conf-foreign-res

### Describe the problem

The section describes the fact that a RS _MUST_ process fallbacks for foreign resources if they are rendered. However, that statement does not apply to foreign content documents, e.g., to images (with fallbacks) appearing in the spine. In fact, there is no reference to such documents in the RS. 

We could (a) duplicate the section to describe the behavior for foreign content documents, or (b) try to reformulate the section to handle both cases.

mattgarrish:
Makes the following changes discussed in #2893 to the reading systems spec:

- redirects the manifest fallback processing requirement under the foreign resources section to where fallback processing is defined in the reading system spec rather than to the authoring requirements
- clarifies that the usage context affects support for media types

Also clarifies that core media type resources are foreign content documents in the spine plane section of the authoring spec.

Fixes #2893 

***

Reading Systems: [Preview](https://raw.githack.com/w3c/epub-specs/fix/fallback-support/epub34/rs/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://www.w3.org/publications/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/rs/index.html&doc2=https://www.w3.org/publications/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fix/fallback-support/epub34/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2897.html" title="Last updated on Jan 23, 2026, 1:46 PM UTC (4233aa0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2897/d40315d...4233aa0.html" title="Last updated on Jan 23, 2026, 1:46 PM UTC (4233aa0)">Diff</a>

### DIFF
--- a/epub34/rs/index.html
+++ b/epub34/rs/index.html
@@ -328,9 +328,48 @@
 
 				<p id="confreq-rs-foreign"
 					data-tests="#pub-foreign_image,#pub-foreign_xml-spine,#pub-foreign_xml-suffix-spine,#pub-foreign_bad-fallback,#pub-foreign_json-spine"
-					>Reading systems MAY support an arbitrary set of [=foreign resource=] types, and if a foreign
-					resource is not supported, MUST process fallbacks as defined in <a href="#sec-pkg-doc-manifest"
-					></a>.</p>
+					>Reading systems MAY support an arbitrary set of [=foreign resource=] types.</p>
+			</section>
+
+			<section id="sec-epub-rs-resource-fallbacks">
+				<h4>Resource fallbacks</h4>
+
+				<p>For [=foreign resources=], if a reading system does not support a resource's MIME media type
+					[[rfc2046]]:</p>
+
+				<ul>
+					<li>it MUST give precedence to any fallbacks provided via <a
+							data-cite="epub-34#sec-intrinsic-fallbacks">intrinsic fallback methods</a> [[epub-34]] of
+						the element the resource is referenced from;</li>
+					<li>otherwise, it SHOULD traverse the resource's <a data-cite="epub-34#sec-manifest-fallbacks"
+							>manifest fallback chain</a> [[epub-34]], when one is specified, until it identifies a
+						supported publication resource to use in place of the unsupported resource.</li>
+				</ul>
+
+				<p>When manifest fallbacks are provided for [=core media type resources=], a reading system MAY traverse
+					the fallback chain to determine if a more optimal format is available for rendering.</p>
+
+				<p>For [=foreign content documents=], if a reading system does not support the resource's MIME media
+					type, it MUST traverse the resource's <a data-cite="epub-34#sec-manifest-fallbacks">manifest
+						fallback chain</a> [[epub-34]] until it identifies a supported publication resource to use in
+					place of the unsupported resource.</p>
+
+				<p>When <a data-cite="epub-34#sec-manifest-fallbacks">manifest fallbacks</a> [[epub-34]] are provided
+					for [=epub content documents=], reading systems MAY choose from the available options to find the
+					optimal version to render in a given context (e.g., by inspecting the properties attribute for
+					each).</p>
+
+				<p id="confreq-rs-pkg-manifest-fallback-order">If the reading system supports multiple publication
+					resources in the fallback chain, it MAY select the resource to use based on the resource's <a
+						data-cite="epub-34#attrdef-item-properties"><code>properties</code> attribute</a> [[epub-34]]
+					values, otherwise it SHOULD honor the <a data-cite="epub-34#preferred-fallback-chain">preferred
+						fallback order</a> [[epub-34]].</p>
+
+				<p id="confreq-rs-pkg-manifest-fallback-fail">If a reading system does not support any resource in the
+					fallback chain, it MUST alert the user that it could not display the content. </p>
+
+				<p id="confreq-rs-pkg-manifest-fallback-cycle">A reading system MUST terminate the fallback chain at the
+					first reference to a manifest item it has already encountered.</p>
 			</section>
 
 			<section id="sec-epub-rs-conf-remote-res">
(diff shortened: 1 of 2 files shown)


epub34/rs/index.html
```diff
@@ -907,10 +907,13 @@ <h3>Manifest</h3>
 					security risks from remotely-hosted sources, lack of fallbacks, etc.).</p>
 
 				<p>
-					<span id="confreq-rs-pkg-manifest-fallback" data-tests="#lay-roll-images-mixed,#lay-roll-images-in-spine,#pp-roll-images-mixed,#pp-roll-images-in-spine">A reading system that does not support the MIME media
-						type [[rfc2046]] of a given [=publication resource=] MUST traverse the <a
-							data-cite="epub-34#sec-manifest-fallbacks">manifest fallback chain</a> [[epub-34]] until it
-						identifies a supported publication resource to use in place of the unsupported resource. </span>
+					<span id="confreq-rs-pkg-manifest-fallback"
```
bduga: I am not sure this is the right place to make the change. We just need to clarify how these documents are handled in the spine. Instead we are opening a hole can of processing worms. For instance, if I use a PNG as a stylesheet referenced from HTML, do I really need to find a fallback? Are we overriding what HTML says about processing stylesheets? Or CSS?

Instead, I think we should add a line in 5.5 to just say that the RS should follow the fallback chain for unsupported foreign content documents until they find one that works.

bduga: Which I see you did while I was typing this comment :) That said, I still think this sentence opens a can of worms we should probably leave closed.

mattgarrish: Ya, I didn't see your latest comment until after I made the first batch of changes. The problem of editing in real time!

I put a change into 5.5 instead. Are you okay with that?

bduga: Yes, thumbs up for the 5.5 change! Talk about confusion :)

epub34/rs/index.html
```diff
@@ -945,6 +945,10 @@ <h3>Spine</h3>
 					<li>rendering successive primary items in the order given in the <code>spine</code>.</li>
 				</ul>
 
+				<p>If a reading system does not support the resource type referenced from an <code>itemref</code>
```
```suggestion
				<p id="confreq-rendition-rs-spine-fallback" data-tests="#lay-roll-images-mixed,#lay-roll-images-in-spine,#pp-roll-images-mixed,#pp-roll-images-in-spine">If a reading system does not support the resource type referenced from an <code>itemref</code>
```

I have a branch on my machine for the test suite to be committed if this is accepted.

Note that the same references appear some lines above this. Not sure it is necessary there, but it does not really harm either...

iherman (approved):

mattgarrish:
The last commit is a new attempt at organizing the fallbacks so we aren't repeating requirements and also to cover cases in the authoring spec that aren't mentioned anywhere in the rs spec.

I noticed that we ended up putting intrinsic and manifest fallbacks together under the resource fallbacks section in the authoring spec, but still had manifest fallbacks under the manifest section in the rs. In the interests of symmetry, I moved the processing into a new similarly named section. I took out the repetitive "must" under the spine section and instead added a note to refer to this new section. That avoids the spine section saying fallback chains must be processed for spine items and the manifest processing also saying that manifest fallbacks must be processed.

In the new resource fallbacks section, I've broken out foreign resource and foreign content document processing and tried to cover the expectations based on the details the authoring spec goes into. The requirement to give precedence to intrinsic fallbacks is new, but this has always been expected. And I've put it as a "should" that fallbacks be processed if a chain is specified. This was a must, but support has always been spotty and despite being available for anything we acknowledge in the authoring spec it's only ever been relevant for img tags. It might be worth repeating that in the rs spec, but I assume anyone implementing this will figure it out from the authoring. (It feels like we should make the whole idea of manifest fallbacks for foreign resources obsolete but conforming and only expect them for foreign content documents.)

I also added a paragraph for fallbacks for cmt resources, as we mention that this could be done to offer alternative core media types in the authoring spec. I guess you could put jxl in an img tag and use a manifest fallback to a png, or something, but begs the question why?

Otherwise, the rest of the requirements are from before.

epub34/rs/index.html
```diff
@@ -328,9 +328,48 @@ <h4>Foreign resources</h4>
...
+						supported publication resource to use in place of the unsupported resource.</li>
+				</ul>
+
+				<p>When manifest fallbacks are provided for [=core media type resources=], a reading system MAY traverse
+					the fallback chain to determine if a more optimal format is available for rendering.</p>
+
+				<p>For [=foreign content documents=], if a reading system does not support the resource's MIME media
+					type, it MUST traverse the resource's <a data-cite="epub-34#sec-manifest-fallbacks">manifest
+						fallback chain</a> [[epub-34]] until it identifies a supported publication resource to use in
+					place of the unsupported resource.</p>
```
```suggestion
				<p  id="confreq-rendition-rs-spine-fallback" data-tests="#lay-roll-images-mixed,#lay-roll-images-in-spine,#pp-roll-images-mixed,#pp-roll-images-in-spine">For [=foreign content documents=], if a reading system does not support the resource's MIME media
					type, it MUST traverse the resource's <a data-cite="epub-34#sec-manifest-fallbacks">manifest
						fallback chain</a> [[epub-34]] until it identifies a supported publication resource to use in
					place of the unsupported resource.</p>
```

This is just "transferring" my earlier comment, so that the test references would not get lost. That being said, if and when this section is final, we will have to check whether tests are available for all cases and, if this is not the case, creating them...

iherman (approved):
Re-approving after the latest changes...

epub34/rs/index.html
iherman: Mark it as resolved; in fact it is outdated, but has been transferred to the latest commit.

mattgarrish:
> (It feels like we should make the whole idea of manifest fallbacks for foreign resources obsolete but conforming and only expect them for foreign content documents.)

I woke up this morning convinced we should move this to the outdated section. The fact we have a note telling people we've moved on from it is the essence of a feature being outdated.

Any concerns?

epub34/rs/index.html
sueneu: since we are referring to foreign content documents, using “resource” in this context
>the resource's MIME media type

is potentially misleading, and could be seen as referring to `foreign resources` instead. 

line 352

And also in line 353, 354, 355?

<p  id="confreq-rendition-rs-spine-fallback" data-tests="#lay-roll-images-mixed,#lay-roll-images-in-spine,#pp-roll-images-mixed,#pp-roll-images-in-spine">For [=foreign content documents=], if a reading system does not support the content document's MIME media
					type, it MUST traverse the content document's <a data-cite="epub-34#sec-manifest-fallbacks">manifest
						fallback chain</a> [[epub-34]] until it identifies a supported media type to use in
					place of the unsupported content.</p>

iherman:
> > (It feels like we should make the whole idea of manifest fallbacks for foreign resources obsolete but conforming and only expect them for foreign content documents.)
> 
> I woke up this morning convinced we should move this to the outdated section. The fact we have a note telling people we've moved on from it is the essence of a feature being outdated.
> 
> Any concerns?

I would prefer to close this PR and then raise this issue on the WG. That should not be our decision.

sueneu:
>>(It feels like we should make the whole idea of manifest fallbacks for foreign resources obsolete but conforming and only expect them for foreign content documents.)

>I woke up this morning convinced we should move this to the outdated section. The fact we have a note telling people we've moved on from it is the essence of a feature being outdated.

Moving this to the obsolete section makes sense if we are truly making fallback resources obsolete. We did float [suggesting fallbacks for jxl](https://github.com/w3c/epub-specs/issues/2896#issuecomment-3785217278) in the author spec.

mattgarrish:
> I would prefer to close this PR and then raise this issue on the WG.

But how much effort do want to put into making sure we have foreign resource manifest fallbacks perfect only to turn around and chuck all the processing? That's where it get impossible to separate some of these overlapping issues that arise. We're pursuing clarifications for a feature the WG already took the position that it strongly discourages people from using:

> The original reason for defining a content fallback mechanism was to handle foreign resource images in the [[html](http://localhost:120/epub34/authoring/#bib-html)] [img](https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element) element. As HTML now has intrinsic fallback mechanisms for images, such as the [srcset](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-srcset) attribute and [source](https://html.spec.whatwg.org/multipage/embedded-content.html#the-source-element) element, the use of content fallbacks is strongly discouraged.

> Moving this to the obsolete section makes sense if we are truly making fallback resources obsolete. We did float [suggesting fallbacks for jxl](https://github.com/w3c/epub-specs/issues/2896#issuecomment-3785217278) in the author spec.

There are two types of fallbacks, though. What we strongly discourage is using manifest fallbacks for html because it requires the reading system to pre-process and rewrite the html (literally or by script) to put in the replacement. We got enough support to squeak this through the 3.3 revision, but it's not widespread and also runs counter to our goal of not interfering with web processing.

HTML elements have intrinsic fallback capabilities that you should be using instead. Even `img` now has the ability to provide fallback formats, as called out in the note. Those are the fallbacks you should be using for jxl or any foreign resource you want to embed in an xhtml content document.

mattgarrish:
> But how much effort do want to put into making sure we have foreign resource manifest fallbacks perfect only to turn around and chuck all the processing?

To answer my own question, let's then merge this as it is then and in a new pull request I'll remove manifest fallbacks for content from both documents and if for some reason we opt to keep them we can use that PR to finalize the processing?

PR #2899, opened 2026-01-23
Adding JPEG XL format

### LINKED ISSUE #2896: Adding support for JXL (jpeg-xl)

iherman:
The PR is a _draft_ in trying to resolve #2896, but also considering the issues raised on the [WG discussion](https://w3c.github.io/pm-wg/minutes/2026-01-22.html#526b). It contain the following:

- introducing JXL into the cmt table
- adding an extra column to the table, indicating the EPUB3 version when that particular type has been introduced (mostly 3.0, a few 3.3 and 3.4)
- there was already a note right after the table; it has been separated into a note and an extra caution box. To make the discussion on the draft easier, here is what the caution box says in the PR:

>  The Working Group typically only includes formats as core media type resources when they have
> broad support in web browser cores — the rendering engines that EPUB 3 reading systems build
> upon — as at that stage they can be relied on for rendering in up-to-date reading systems.
>  Nevertheless, authors must be aware that newer formats may not be supported on reading systems
>  that, for example, rely on their own rendering engines, or have not been updated recently.

The PR, eventually, aims at a fix of #2896.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2899.html" title="Last updated on Jan 28, 2026, 11:06 AM UTC (bfea447)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2899/50dc472...bfea447.html" title="Last updated on Jan 28, 2026, 11:06 AM UTC (bfea447)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -958,7 +958,7 @@
 							<th id="tbl-cmt-string">Media Type</th>
 							<th id="tbl-cmt-def">Content Type Definition</th>
 							<th id="tbl-cmt-appl">Applies to</th>
-                            <th id="tbl-cmt-epub-version">EPUB3</th>
+                            <th id="tbl-cmt-epub-version">EPUB</th>
 						</tr>
 					</thead>
 					<tbody>
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -1137,7 +1137,7 @@
 							</td>
 							<td> [[woff2]] </td>
 							<td>WOFF2 fonts</td>
-                            <td>3.3</td>
+                            <td>3.2</td>
 						</tr>
 						<tr>
 							<th colspan="4" id="cmt-grp-other" class="tbl-group">Other</th>
@@ -1195,7 +1195,7 @@
                         The Working Group typically only includes formats as core media type resources when they have
 						broad support in web browser cores — the rendering engines that EPUB 3 reading systems build
 						upon — as at that stage they can be relied on for rendering in up-to-date reading systems.
-                        Nevertheless, authors must be aware that newer formats may not be supported on reading systems
+                        Nevertheless, authors must be aware that newer formats &#8212; introduced after EPUB 3.0 &#8212; may not be supported on reading systems
                         that, for example, rely on their own rendering engines, or have not been updated recently.
                     </p>
 				</div>
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -998,6 +998,13 @@
 							<td> [[jpeg]] </td>
 							<td>JPEG images</td>
 						</tr>
+                        <tr>
+                            <td id="cmt-jxl" data-tests="#pub-cmt-jxl">
+                                <code>image/jxl</code>
+                            </td>
+                            <td> [[iso18181-1]] </td>
+                            <td>JPEG XL images</td>
+                        </tr>
 						<tr>
 							<td id="cmt-png" data-tests="#pub-cmt-png">
 								<code>image/png</code>
@@ -11110,6 +11117,8 @@ EPUB/images/cover.png</pre>
 			<details id="changes-epub-33" open="open">
 				<summary>Substantive changes since <a href="https://www.w3.org/TR/epub-33/">EPUB 3.3</a></summary>
 				<ul>
+                    <li>23-January-2026: Added JPEG XL as a core media type for images. See <a
+                            href="https://github.com/w3c/epub-specs/issues/2896">issue 2896</a>.</li>
 					<li>18-Dec-2025: Renamed "obsolete but conforming features" to "outdated features". As discussed in
 						the <a href="https://w3c.github.io/pm-wg/minutes/2025-12-11.html#59bf">2025-12-11 WG
 						meeting</a>.</li>
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -958,7 +958,7 @@
 							<th id="tbl-cmt-string">Media Type</th>
 							<th id="tbl-cmt-def">Content Type Definition</th>
 							<th id="tbl-cmt-appl">Applies to</th>
-                            <th id="tbl-cmt-epub-version">Introduced in<br/> EPUB3 version</th>
+                            <th id="tbl-cmt-epub-version">EPUB3</th>
 						</tr>
 					</thead>
 					<tbody>
@@ -974,7 +974,7 @@
 							</td>
 							<td>HTML documents that use the <a data-cite="html#the-xhtml-syntax">XML syntax</a>
 								[[html]]</td>
-                            <td>3.0</td>
+                            <td>pre&#8209;3.0</td>
 						</tr>
 						<tr>
 							<th colspan="4" id="cmt-grp-image" class="tbl-group">Images</th>
@@ -985,7 +985,7 @@
 							</td>
 							<td> [[gif]] </td>
 							<td>GIF images</td>
-                            <td>3.0</td>
+                            <td>pre&#8209;3.0</td>
 						</tr>
                         <tr>
                             <td id="cmt-jpeg" data-tests="#pub-cmt-jpeg">
@@ -993,7 +993,7 @@
                             </td>
                             <td> [[jpeg]] </td>
                             <td>JPEG images</td>
-                            <td>3.0</td>
+                            <td>pre&#8209;3.0</td>
                         </tr>
  						<tr>
 							<td id="cmt-png" data-tests="#pub-cmt-png">
@@ -1001,7 +1001,7 @@
 							</td>
 							<td> [[png]] </td>
 							<td>PNG images</td>
-                            <td>3.0</td>
+                            <td>pre&#8209;3.0</td>
 						</tr>
 						<tr>
 							<td id="cmt-svg" data-tests="#pub-cmt-svg,#cnt-svg-support">
@@ -1011,7 +1011,7 @@
 								<a href="#sec-svg">SVG content documents</a>
 							</td>
 							<td>SVG documents</td>
-                            <td>3.0</td>
+                            <td>pre&#8209;3.0</td>
 						</tr>
 						<tr>
 							<td id="cmt-webp" data-tests="#pub-cmt-webp">
@@ -1075,7 +1075,7 @@
 								<a href="#sec-css">CSS Style Sheets</a>
 							</td>
 							<td>CSS Style Sheets</td>
-                            <td>3.0</td>
+                            <td>pre&#8209;3.0</td>
 						</tr>
 						<tr>
 							<th colspan="4" id="cmt-grp-font" class="tbl-group">Fonts</th>
@@ -1166,7 +1166,7 @@
 							</td>
 							<td> [[opf-201]] </td>
 							<td>The [=outdated=] NCX</td>
-                            <td>3.0</td>
+                            <td>pre&#8209;3.0</td>
 						</tr>
 						<tr>
 							<td id="cmt-smil">
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -950,6 +950,7 @@
 						resource has to conform.</li>
 					<li><strong>Applies to</strong>—The publication resource type(s) that the Media Type and Content
 						Type Definition applies to.</li>
+                    <li><strong>In EPUB 3 since</strong>—The version of EPUB 3 which has adopted that Media Type. Note that some formats have a longer history (going back to EPUB 2), not detailed in the table.</li>
 				</ul>
 
 				<table id="tbl-core-media-types">
@@ -958,7 +959,7 @@
 							<th id="tbl-cmt-string">Media Type</th>
 							<th id="tbl-cmt-def">Content Type Definition</th>
 							<th id="tbl-cmt-appl">Applies to</th>
-                            <th id="tbl-cmt-epub-version">EPUB</th>
+                            <th id="tbl-cmt-epub-version">In EPUB 3 since</th>
 						</tr>
 					</thead>
 					<tbody>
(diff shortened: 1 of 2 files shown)


epub34/authoring/index.html
```diff
@@ -1107,16 +1129,18 @@ <h3>Core media types</h3>
...
 							<td>WOFF fonts</td>
+                            <td>3.0</td>
 						</tr>
 						<tr>
 							<td id="cmt-woff2">
 								<code>font/woff2</code>
 							</td>
 							<td> [[woff2]] </td>
 							<td>WOFF2 fonts</td>
+                            <td>3.3</td>
```
mattgarrish:
```suggestion
                            <td>3.2</td>
```

epub34/authoring/index.html
```diff
@@ -1076,6 +1096,7 @@ <h3>Core media types</h3>
 							</td>
 							<td>[[truetype]] </td>
 							<td>TrueType fonts</td>
+                            <td>3.0</td>
```
mattgarrish, line 1102: I think I'm okay with this. We didn't introduce TrueType as a separate entry in the CMT table until 3.2 (adding `font/ttf` at the same time), but you could include TrueType fonts using the `application/font-sfnt` media type we listed for OpenType prior to that. Probably no need to concern people about this technicality by changing it to a 3.2 addition.

epub34/authoring/index.html
```diff
@@ -1162,10 +1189,15 @@ <h3>Core media types</h3>
...
-
-					<p>The Working Group typically only includes formats as core media type resources when they have
+                </div>
+                <div class="caution">
+					<p>
+                        The Working Group typically only includes formats as core media type resources when they have
 						broad support in web browser cores — the rendering engines that EPUB 3 reading systems build
-						upon — as at that stage they can be relied on for rendering in reading systems.</p>
+						upon — as at that stage they can be relied on for rendering in up-to-date reading systems.
+                        Nevertheless, authors must be aware that newer formats may not be supported on reading systems
```
mattgarrish: Might help to qualify that "newer" means post 3.0.

```suggestion
                        Nevertheless, authors must be aware that newer formats &#8212; introduced after EPUB 3.0 &#8212; may not be supported on reading systems
```

reidmore-online:
The added text looks good to me, but I think we may have some formatting issues, the table styling we have in 3.3 seems to have disappeared (or isn't working in the preview), and the caution section is not formatted either.

reidmore-online (approved):

HadrienGardeur (approved):

mattgarrish (approved):

iherman:
> The added text looks good to me, but I think we may have some formatting issues, the table styling we have in 3.3 seems to have disappeared (or isn't working in the preview), and the caution section is not formatted either.

There were some hiccups on github (I could not get to the PR for a while). I have just downloaded the latest version of the branch on my machine, and it works fine.

sueneu (approved):

epub34/authoring/index.html
```diff
@@ -958,11 +958,12 @@ <h3>Core media types</h3>
 							<th id="tbl-cmt-string">Media Type</th>
 							<th id="tbl-cmt-def">Content Type Definition</th>
 							<th id="tbl-cmt-appl">Applies to</th>
+                            <th id="tbl-cmt-epub-version">Introduced in<br/> EPUB3 version</th>
```
bduga: It seems odd to me to have the version in the header of the field that lists versions. I guess the point is to make it clear that we are only talking about when these were added to the EPUB3 spec itself, but may have predated EPUB3? In any case, I think it would be more clear to say just EPUB (which will make life easier if there is ever an EPUB 4.0), and somehow indicate in the table that some media types pre-date 3.0 (e.g. s/3.0/pre-3.0/).

iherman: I agree with the principle. On the details: is s/3.0/pre-3.0/ appropriate? I was looking at https://idpf.org/epub/20/spec/OPS_2.0.1_draft.htm#Section1.4.1 and https://idpf.org/epub/30/spec/epub30-changes-20111011.html#sec-new-changed-audio-video and the way I see it audio formats were not mentioned in pre-3.0. Also, was SMIL used pre-3.0? What about fonts?

iherman: I note that https://www.rfc-editor.org/rfc/rfc7845 is from 2016, so I suspect ogg is definitely not pre-3.0...

mattgarrish: The ops core media types are in [s. 1.3.7](https://idpf.org/epub/20/spec/OPS_2.0.1_draft.htm#Section1.3.7)

Is this just as a historical curiosity, though? Does it really matter if a media type predates 3.0 as far creating conforming (with possible fallbacks) epub 3 content is concerned?

iherman: > The ops core media types are in [s. 1.3.7](https://idpf.org/epub/20/spec/OPS_2.0.1_draft.htm#Section1.3.7)
> 

Yep I saw that. So, essentially, audio, smil, etc, are all 3.0

> Is this just as a historical curiosity, though? Does it really matter if a media type predates 3.0 as far creating conforming (with possible fallbacks) epub 3 content is concerned?

It is the matter of being consistent. If the table head says EPUB3 then it is irrelevant, only 3.0 counts. If the header says EPUB, then we just try to be precise...

I will use that table for epub2 and finalize this...

mattgarrish: I'm just not sure what practical use there is to saying that some of these were core media types in epub 2. RMSDK was the engine for most epub 2 implementations and it didn't survive the transition, so knowing something was a core media type in epub 2 doesn't make it more reliable than media types introduced in 3.0.

But if we're just making this a historical record, I guess that's fine.

bduga: > Is this just as a historical curiosity, though? Does it really matter if a media type predates 3.0 as far creating conforming (with possible fallbacks) epub 3 content is concerned?

The whole point of that table is to say how long things have been around, to give an idea of how likely they are to be implemented and available for use, right? So we decided to pick 3.0 as the epoch for that, but I don't think we should be inaccurate just because something is "old enough". I think saying pre-3.0 is fine, as we are essentially saying anything before that point is so old that it doesn't matter. If we are having trouble deciding what was around that long ago, we could also switch the epoch to, say, 3.2. That is, anything added before 3.2 just becomes pre-3.2, and everything else becomes the specific version.

mattgarrish: > If the table head says EPUB3 then it is irrelevant, only 3.0 counts.

But now we have a column heading of "EPUB3" (requires a space, btw, even if old idpf trademark stuff) but that for values says "pre 3.0". Who other than people in this discussion is going to follow what's going on?

If we have to make this a historical record, a heading of "Introduced in" and values like "EPUB 3.X", or "predates EPUB 3.0" would be clearer, if still confusing.

> The whole point of that table is to say how long things have been around, to give an idea of how likely they are to be implemented and available for use, right?

I just find it strange to bridge the divide between major formats in this particular way. It sounds like you should trust stuff from EPUB 2 even though all support for things EPUB 3 started anew in 2011.

iherman: > But now we have a column heading of "EPUB3" (requires a space, btw, even if old idpf trademark stuff) but that for values says "pre 3.0". Who other than people in this discussion is going to follow what's going on?

That is a bug, not a feature :-) My intention to have "EPUB" in the header, not "EPUB3". I have just changed it.

bduga: I mainly brought this up because I thought it was confusing to say the the version of EPUB3 that introduced XHTML was 3.0. First, it implies 3.0 is a version of EPUB3, which is weird. Does that mean there are 1.0 and 2.0 versions of EPUB3? I thought the 3 was the version number. So you can talk about EPUB (any version) or EPUB 3 (any version with a major revision of 3) or EPUB 3.0 (the specific major/minor of "3" and "0"). This seems to be backed up by the use in the spec - EPUB is a registered trademark, EPUB3, EPUB 3, and EPUB 3.3 are not (or at least do not have an ® symbol in the spec). The spec itself is "EPUB 3.3", not EPUB 3. And it is really confusing to me to say we added XHTML to EPUB in version 3.0!

But looking at it more, I am not sure what the table heading means. It just says EPUB3 now, but what does that mean? It has versions, but no discussion of what those versions mean, or why they are there. There might be a better heading here, but shouldn't the text describing the columns also be updated?

mattgarrish: > version of EPUB3 that introduced XHTML was 3.0

The first _revision_ of EPUB 3 that introduced XHTML was 3.0. If we're calling these versions then, yes, that makes it worse!

It's been 15 years and four, going on five, revisions. We're trying to give people a sense of when stuff was added to EPUB 3 over that time. Whether EPUB 2 did something similar still strikes me as irrelevant. EPUB 2 supported xhtml 1.1, so now I know I can use xhtml 5?? I can't get past this support chasm between the formats.

But I'm probably more concerned about the column making sense on its own.Dropping the 3 still doesn't say why it's there. Yes, if you follow on to the note after you might have an aha moment where you can make sense of it, but that's bad spec design.

What's worse, is it now makes it sound like if something is pre EPUB 3.0, does that mean you can't use it anymore in EPUB 3? What are we saying?

> EPUB is a registered trademark, EPUB3, EPUB 3, and EPUB 3.3 are not (or at least do not have an ® symbol in the spec)

Yes, but if you attach the 3 then it's not the trademark. That was from idpf's legal counsel that specifications must never have the number attached to the name. I think it weakens the trademark. Before W3C we also had to put the trademark on the first use of EPUB not part of a specification name. We dropped the trademarking requirement but just for internal consistency I don't want to see a bunch of different variations!

bduga: > What are we saying?

I don't know! :) I guess we may be in agreement. What information are we trying to convey with this column, who is the target of that information, and what do we expect them to do with it?

iherman: What about making the header more verbose and explicit: 

- "In EPUB since" and then we add the pre-3.0 things; or
- "In EPUB 3 since" and then we roll back and all pre-3.0 becomes 3.0

I am mildly in favour of rolling back.

mattgarrish: One thing we haven't taken advantage of in this pull request is that all the table headers have lengthier definitions right above the table. Maybe that's an option for clarifying that some formats from 3.0 have a longer support history in EPUB without confusing anyone reading the table about why we're getting into that for EPUB 3 support?

Another thought would be to asterisk the media types that weren't in 3.0 and add a footnote to the table that says these were added after 3.0 so older reading systems may not support them as well. That might be the simplest approach.

HadrienGardeur:
FYI, we could already support JPEG XL in EPUB in the Readium Swift Toolkits and we've extended this to CBZ files: https://github.com/readium/swift-toolkit/pull/703

This means that we can now support JPEG XL files in  `<spine>` without fallbacks as well.

epub34/authoring/index.html
iherman: @mattgarrish @bduga I have (a) added some words in the introductory list item on the last column; (b) rolled back the EPUB 2 references (c) also added a back reference to the column in the caution note.

shiestyle (approved):

epub34/authoring/index.html
mattgarrish: Sorry, fell off the radar. Works for me.

iherman:
This was discussed during the [pmwg meeting on 29 January 2026](https://w3c.github.io/pm-wg/minutes/2026-01-29.html#b122).

<h3>JXL in EPUB</h3>
<p><cite>&lt;wendyreid&gt;</cite> <a href="https://github.com/w3c/epub-specs/pull/2899" rel="noopener noreferrer">w3c/<wbr>epub-specs#2899</a>, <a href="https://github.com/w3c/epub-specs/issues/2896" rel="noopener noreferrer">w3c/<wbr>epub-spec#2896</a></p>
<p><cite>duga:</cite> We need to fix the table that references media types regarding FXL</p>
<p><cite>sueneu:</cite> I recommend we merge the PR and make a note to fix the table.</p>
<p><cite>&lt;wendyreid&gt;</cite> Proposed: Merge PR 2899, adding JXL support and information on EPUB version for core media types.</p>
<p><cite>&lt;Dale&gt;</cite> +1</p>
<p><cite>&lt;duga&gt;</cite> 0</p>
<p><cite>&lt;gautierchomel_&gt;</cite> +1</p>
<p><cite>&lt;sueneu&gt;</cite> +1</p>
<p><cite>&lt;gman&gt;</cite> +1</p>
<p><strong>RESOLUTION:</strong> Merge PR 2899, adding JXL support and information on EPUB version for core media types.</p>

<hr /></details>

iherman:
Merging, per latest WG resolution: https://w3c.github.io/pm-wg/minutes/2026-01-29.html#38dc

PR #2902, opened 2026-01-29
Move content fallbacks to outdated features

### LINKED ISSUE #2898: We incorrectly require use of manifest fallbacks
### Section

https://w3c.github.io/epub-specs/epub34/rs/#sec-pkg-doc-manifest

### Describe the problem

The spec currently says

> A reading system that does not support the MIME media type [rfc2046] of a given publication resource MUST traverse the manifest fallback chain [epub-34] until it identifies a supported publication resource to use in place of the unsupported resource.

However, that isn't really true, and in fact manifest resources are *discouraged* for non-spine level fallbacks (see the note at [Manifest fallbacks](https://www.w3.org/TR/epub-34/#sec-manifest-fallbacks)). 

Reading systems should be required to process the fallback mechanism used by the content to specify a supported resources. What happens when there are two or more such mechanisms is unclear.

### LINKED ISSUE #2900: Manifest fallbacks for content should be an outdated feature
### Section

https://www.w3.org/TR/epub-34/#sec-manifest-fallbacks

### Describe the problem

Prior to having an actual outdated feature list, we effectively outdated using manifest fallbacks for content by adding this note to the authoring section that defines it:

> The original reason for defining a content fallback mechanism was to handle foreign resource images in the [[html](https://www.w3.org/TR/epub-34/#bib-html)] [img](https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element) element. As HTML now has intrinsic fallback mechanisms for images, such as the [srcset](https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-srcset) attribute and [source](https://html.spec.whatwg.org/multipage/embedded-content.html#the-source-element) element, the use of content fallbacks is strongly discouraged.

Remove the authoring and processing requirements from the authoring and rs specs and add manifest fallbacks for content to the outdated features section.

Manifest fallbacks should only be used for the spine so we're not interfering in html processing.

mattgarrish:
Title pretty much says it all. We've been advising that manifest fallbacks outside the spine shouldn't be used anymore since 3.3; this just got missed while we were doing the earlier roundup of outdated stuff.

I've kept the intrinsic fallback sections as they're still relevant.

Fixes #2898
Fixes #2900 

***

Reading Systems: [Preview](https://raw.githack.com/w3c/epub-specs/fix/remove-content-fallbacks/epub34/rs/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://www.w3.org/publications/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/rs/index.html&doc2=https://www.w3.org/publications/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/fix/remove-content-fallbacks/epub34/rs/index.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2902.html" title="Last updated on Feb 2, 2026, 2:50 PM UTC (aca95c8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2902/99ef96d...aca95c8.html" title="Last updated on Feb 2, 2026, 2:50 PM UTC (aca95c8)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -1402,26 +1402,23 @@
 						data-tests="https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L280,https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L285,https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L290,https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L296,https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L302,https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L308,https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L313">
 						<h5>HTML <code>img</code> fallbacks</h5>
 
-						<p id="confreq-resources-cd-fallback-img">Due to the variety of sources that can be referenced
-							from the [[html]] [^img^] element, the following fallback conditions apply to its use:</p>
-						<ul>
-							<li>
-								<p>If it is the child of a [^picture^] element:</p>
-								<ul>
-									<li>it MUST reference core media type resources from its <code>src</code> and
-											<code>srcset</code> attributes when those attributes are set; and</li>
-									<li>each sibling [^source^] element MUST reference a [=core media type resource=]
-										from its <code>[^source/src^]</code> and <code>[^source/srcset^]</code>
-										attributes unless it specifies the MIME media type [[rfc2046]] of a [=foreign
-										resource=] in its <code>[^source/type^]</code> attribute.</li>
-								</ul>
-							</li>
+						<p id="confreq-resources-cd-fallback-img">To provide alternative format fallbacks, the [[html]]
+							[^img^] element can be used as the child of a [^picture^] element provided the following are
+							true:</p>
 
-							<li>
-								<p>Otherwise, its <code>[^img/srcset^]</code> attribute MUST NOT be used to provide
-									fallbacks to a core media type.</p>
-							</li>
+						<ul>
+							<li>the <code>img</code> element MUST only reference core media type resources from its
+									<code>src</code> and <code>srcset</code> attributes when those attributes are set;
+								and</li>
+							<li>each [^source^] element MUST only reference [=core media type resources=] from its
+									<code>[^source/src^]</code> and <code>[^source/srcset^]</code> attributes unless it
+								specifies the MIME media type [[rfc2046]] of a [=foreign resource=] in its
+									<code>[^source/type^]</code> attribute.</li>
 						</ul>
+
+						<p>When an <code>img</code> is not the child of <code>picture</code>, its <code>src</code> and
+								<code>srcset</code> attributes MUST NOT be used to provide fallbacks for foreign
+							resources.</p>
 					</section>
 
 					<section>
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -1402,27 +1402,23 @@
 						data-tests="https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L280,https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L285,https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L290,https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L296,https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L302,https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L308,https://w3c.github.io/epub-structural-tests/#03-resources_resources.feature_L313">
 						<h5>HTML <code>img</code> fallbacks</h5>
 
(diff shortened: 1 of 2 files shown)


epub34/authoring/index.html
```diff
@@ -885,17 +885,14 @@ <h4>The content plane</h4>
 					<p>The opposite of core media type resources are foreign resources. These are resources that reading
 						systems are not guaranteed to support the rendering of. As a result, similar to how using
 						foreign content documents in the spine requires fallbacks to ensure their rendering, using
-						foreign resources in content documents also requires fallbacks. These fallbacks are provided in
-						one of two ways: using the capabilities of the host format or via manifest fallbacks.</p>
+						foreign resources in content documents also requires fallbacks. These fallbacks are provided
+						using the capabilities of the host format.</p>
```
```
						foreign resources in content documents also requires fallbacks. These fallbacks are provided
						using the capabilities of the host format.</p>
```

The first sentence of the following paragraph is essentially the same now. Maybe it is worth removing it and merge the remaining to this paragraph? (GH does not let me doing that here.)

epub34/authoring/index.html
```diff
@@ -1362,54 +1355,30 @@ <h5>Manifest fallbacks</h5>
...
+							<code>item</code> elements in the chain.</p>
 				</section>
 
 				<section id="sec-intrinsic-fallbacks">
 					<h4>Intrinsic fallbacks</h4>
 
-					<p>The following sections provide additional clarifications about the intrinsic fallback
-						requirements of specific elements.</p>
+					<p>The use of [=manifest fallback chains=] for use in EPUB content documents and foreign content
+						documents is now [=outdated=]. The following sections provide clarifications about the intrinsic
```
```
					<p>The use of [=manifest fallback chains=] for use in EPUB content documents and foreign content
						documents is now [=outdated=]. 
```

Is this sentence really necessary here? The section is on intrinsic fallbacks, and we do have such statements elsewhere...

epub34/authoring/index.html
```diff
@@ -1452,9 +1421,15 @@ <h5>HTML <code>img</code> fallbacks</h5>
...
-									<code>[^img/srcset^]</code> attributes provided there is also a <a
-									href="#sec-manifest-fallbacks">manifest fallback</a> to a core media type.</li>
+							<li>
+								<p>Otherwise, it MAY reference foreign resources in its <code>[^img/src^]</code> and
+										<code>[^img/srcset^]</code> attributes provided there is also a <a
+										href="#content-fallbacks">content fallback</a> to a core media type.</p>
+								<div class="note">
+									<p>Content fallbacks are now marked as [=outdated=] and their use is strongly
+										discouraged.</p>
+								</div>
```
```
								<p>Otherwise, it MAY reference foreign resources in its <code>[^img/src^]</code> and
										<code>[^img/srcset^]</code> attributes provided there is also a <a
										href="#content-fallbacks">content fallback</a> to a core media type.</p>
								<div class="note">
									<p>Content fallbacks are now marked as [=outdated=] and their use is strongly
										discouraged.</p>
								</div>
```

Do we want to keep the last bullet item (and the note)? I would think that the core specification should not really refer to a discouraged feature...

iherman (approved):

epub34/authoring/index.html
mattgarrish: No, it's not strictly necessary. I just wasn't sure if it would still be helpful to have some context for why we added this section in the first place.

epub34/authoring/index.html
mattgarrish: Ya, I keep struggling with this (as evidenced by my multiple commits) because it doesn't feel right to say nothing about the case simply because it's now outdated.

I tried taking the bullet out and expanding the note to advise only use CMTs when not a child of picture, but that still requires acknowledging the outdated content fallbacks.

My last commit takes a different tack and changes the bullet to say that outside of picture the srcset attribute must not be used to provide CMT fallbacks. That doesn't invalidate manifest fallbacks but also makes it a bit clearer that there isn't an intrinsic fallback mechanism.

Does that work better for you?

iherman: It looks better but… why disallow the `scrset` attribute? Doesn't it provide a way for reasonable fallback intrinsically?

mattgarrish: No, it provides an alternative based on dimensions or pixel density and you can't have two images with the same descriptor. That's why the bullet said you could only reference foreign resources using a manifest fallback.

For selecting by media type you have to use picture/source with a fallback on the img.

mattgarrish: I did a bit more cleanup on this in the last commit. I don't think we need the list of options when the last option is just saying that you can't use sr/srcset for fallbacks outside of picture. I also noticed that the bullets about using in picture said that you must reference a core media type unless a foreign media type is specified on a source element, but it should be more specific that we mean  "must only" reference CMTs.

iherman: Just for the good order, I added labels; I believe it is correct to say this is an editorial issue.

mattgarrish: Ya, I think it warrants a note in the change log but given the caution we already had saying not to use content fallbacks anymore, it's not the same level of change as the other things we moved.

mattgarrish:
Should we move ahead with merging this? We didn't get to it on the call today but it's also only a shuffle from noting not to use the feature to explicitly making it outdated.

PR #2906, opened 2026-02-02
Clarify exemptions can apply to some conforming content

### LINKED ISSUE #2849: Clarify that not all exempt content fails conformance
### Section

https://www.w3.org/TR/epub-a11y-exemption/#sec-evaluator-name

### Describe the problem

Looking at the exemption note today I spotted this:

> When a jurisdiction provides specific legal exemptions for [EPUB publications](https://www.w3.org/TR/epub-33/#dfn-epub-publication), [EPUB creators](https://www.w3.org/TR/epub-33/#dfn-epub-creator) whose content fails to meet the requirements of the EPUB Accessibility standard [[epub-a11y](https://www.w3.org/TR/epub-a11y-exemption/#bib-epub-a11y)] (or any other standard), can use [the exemption property](https://www.w3.org/TR/epub-a11y-exemption/#exemption-property) to indicate when their publications fall under one of these provisions.

This isn't fully accurate, since a publication could claim level A conformance yet still need an exemption from the EAA which requires AA conformance.

Clarify that any publication that conforms below the necessary minimum can also claim exemptions.

(And write out all the "epub creators" while in there. Also fix the "sec-evaluator-name" id on the section as it looks like a clear case of a copy/paste.)

mattgarrish:
The primary fix in this pull request is to reword the first paragraph and note in section 2 so that we aren't inferring that all content that needs an exemption fails accessibility standards. A Level A-conformant publication would need an EAA exemption, for example.

I also got rid of "epub creator" from the document. There weren't many instances, but the one notable change I made was to put a caution before the EAA value definitions to seek legal advice when unsure if they apply. I don't think we need to have a paragraph in every description saying it's the epub creator's responsibility to check.

Fixes #2849 

***

[Preview](https://raw.githack.com/w3c/epub-specs/editorial/issue-2849/wg-notes/a11y-exemption/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://www.w3.org/publications/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/wg-notes/a11y-exemption/index.html&doc2=https://www.w3.org/publications/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/editorial/issue-2849/wg-notes/a11y-exemption/index.html)

### DIFF
--- a/wg-notes/a11y-exemption/index.html
+++ b/wg-notes/a11y-exemption/index.html
@@ -5,6 +5,7 @@
 		<meta name="color-scheme" content="light dark" />
 		<title>The EPUB Accessibility exemption property</title>
 		<script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"></script>
+		<script src="../../common/js/add-caution-hd.js" class="remove"></script>
 		<script src="../../common/js/css-inline.js" class="remove"></script>
 		<script class="remove">
 			var respecConfig = {
@@ -34,6 +35,7 @@
 					branch: "main"
 				},
 				preProcess:[inlineCustomCSS],
+                postProcess:[addCautionHeaders],
 				xref: {
 					profile: "web-platform",
 					specs: ["epub-33"]
--- a/wg-notes/a11y-exemption/index.html
+++ b/wg-notes/a11y-exemption/index.html
@@ -62,9 +62,8 @@
 	<body>
 		<h1 id="title">The EPUB Accessibility <code>exemption</code> property</h1>
 		<section id="abstract">
-			<p>The <code>exemption</code> property allows EPUB creators to indicate that an EPUB publication that does
-				not meet accessibility conformance requirements has an exemption under the applicable jurisdiction's
-				laws.</p>
+			<p>The <code>exemption</code> property indicates that an EPUB publication that does not meet a
+				jurisdiction's mimimum accessibility conformance requirements has an exemption under its laws.</p>
 		</section>
 		<section id="sotd"></section>
 		<section id="sec-intro">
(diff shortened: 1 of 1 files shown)


clapierre (approved):

wg-notes/a11y-exemption/index.html
```diff
@@ -119,10 +117,14 @@ <h3>Accessibility conformance exemptions</h3>
 			</aside>
 
 			<div class="note">
-				<p>EPUB creators do not have to include an accessibility conformance claim of "<code>none</code>" when
-					listing that a publication has an exemption from accessibility requirements, but it is best practice
-					to also provide this information. Reading systems that do not recognize the exemption might
-					otherwise list the accessibility status as unknown.</p>
+				<p>An accessibility conformance claim of "<code>none</code>" is not required when listing that an EPUB
```
sueneu: >				<p>An accessibility conformance claim of "<code>none</code>" is not for an EPUB

opportunity to simplify further

wg-notes/a11y-exemption/index.html
```diff
@@ -119,10 +117,14 @@ <h3>Accessibility conformance exemptions</h3>
...
+					publication has an exemption from accessibility requirements, but it is best practice to also
```
sueneu: >publication has an exemption from accessibility requirements, but it is best practice to do so.

wg-notes/a11y-exemption/index.html
```diff
@@ -119,10 +117,14 @@ <h3>Accessibility conformance exemptions</h3>
...
+					provide this information if no standard has been met. Reading systems that do not recognize the
```
sueneu: Reading systems that do not recognize

sueneu (commented):
I added suggestions for further simplifying lines 120, 121, 122

mattgarrish:
I made some different edits to the note to slim it down, and also put the fact that conforming content can claim exemptions first since it's more important than the stuff about "none" (which I'm not sure we should be saying is best practice here when we don't call that out in the accessibility spec).

If you want to make suggestions to a full paragraph, the best way is to either copy the text and edit it in your comment or, better, use github's suggestion feature. You just have to drag the + sign down to cover all the lines you want to amend and then click the suggestion button. Github will make a diff of whatever you change the suggested text to, plus it can be committed directly. I'll copy a screenshot of what I mean below.

(It's sometimes hard to be sure what people want changed when addressing single lines at a time. The first edit, for example didn't seem to follow into the next line, and the last suggestion appeared to be the same text that was there.)

suggestion

sueneu:
Matt, thank you! 

It would be much easier to comment on the whole paragraph. Sometimes I have that option and sometimes I don’t. I’ll try working in another browser to see if I get more consistent options.

gregoriopellegrino (approved):

PR #2912, opened 2026-02-03
Annotation selectors, second draft

iherman:
In agreement with @llemeurfr (who is on vacations) I created a version of the annotation spec with selectors, based on the previous draft (see #2892) plus the discussions on that PR as well as the [discussion on the WG](https://w3c.github.io/pm-wg/minutes/2026-01-22.html#7e88). I hope I have not forgotten or misunderstood anything.

Some comments:

- I have added, as ***informative*** sections, the definitions of selectors that we just "take over" from the Open Annotation specification (e.g., CssSelector). I did this to make the document more self-contained and to make it easier for implementers. I hope that is the right choice.
- Text fragments appear now at two places: referred to from Fragment Selectors, and noted as a possible implementors' vehicle for text quote selectors. I was not sure which of the two we should retain and, actually, there is room for both. The second option has the advantage that we are not normatively dependent on the text fragment specification. To be discussed.

---

See:

* For EPUB 3 Annotations:
    * [Preview](https://raw.githack.com/w3c/epub-specs/ann-selectors-2/epub34/annotations/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https://www.w3.org/publications/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/epub-specs/epub34/annotations/index.html&doc2=https://www.w3.org/publications/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/epub-specs/ann-selectors-2/epub34/annotations/index.html)

### DIFF
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -380,7 +380,7 @@
                         An annotation refers to a segment of a resource, which is identified by one or more
                         <a data-cite="annotation-model#selectors">Selectors</a>.
                         The nature of the Selectors and methods to describe segments depend on
-                        the resource type. Providing more than one Selectors allows an annotation software to
+                        the resource type. Providing more than one Selector allows an annotation software to
                         choose the most accurate selector from those it can handle and helps to accommodate
                         evolutions on the annotated resource.
                     </p>
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -40,6 +40,7 @@
                     branch: "main"
                 },
                 preProcess: [inlineCustomCSS],
+                postProcess: [addCautionHeaders],
                 testSuiteURL: "https://w3c.github.io/epub-tests/index.html",
                 xref: {
                     profile: "web-platform",
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -4,7 +4,7 @@
     <head>
         <meta charset="utf-8" />
         <meta name="color-scheme" content="light dark" />
-        <title> EPUB Annotations 3.4 </title>
+        <title>EPUB Annotations 1.0</title>
         <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove">
         </script>
         <script src="../../common/js/css-inline.js" class="remove"></script>
@@ -15,7 +15,7 @@
                 group: "pm",
                 wgPublicList: "public-pm-wg",
                 specStatus: "ED",
-                shortName: "epub-ann",
+                shortName: "epub-anno",
                 noRecTrack: false,
                 edDraftURI: "https://w3c.github.io/epub-specs/epub34/annotations/",
                 copyrightStart: "2026",
@@ -48,7 +48,7 @@
                 },
                 lint: { "no-unused-dfns": false },
                 localBiblio: {
-                    "epub-ann-ucr": {
+                    "epub-anno-ucr": {
                         "authors": [
                             "Laurent Le Meur"
                         ],
@@ -74,7 +74,7 @@
         <section id="abstract">
             <p> This document defines a profile of the [[[annotation-model]]] [[annotation-model]]
                 by specifying a subset of the properties allowed in this model, and adding
-                properties deemed useful to satisfy the [[[epub-ann-ucr]]] [[epub-ann-ucr]].
+                properties deemed useful to satisfy the [[[epub-anno-ucr]]] [[epub-anno-ucr]].
             </p>
         </section>
         <section id="sotd"></section>
@@ -460,7 +460,8 @@
                             [^conformsTo^] property. Other URLs MUST NOT be used.
                         </p>
 
-                        <table id="#tbl-core-media-types">
+                        <!-- cheating: I reuse a common css entry done for media types, hence the choice of the ID -->
+                        <table id="tbl-core-media-types">
                             <thead>
                                 <th>Name</th>
                                 <th>Fragment Specification</th>
@@ -492,7 +493,6 @@
                             </tbody>
                         </table>
 
-                        <p>The integration of text fragments into the specification is AT RISK.</p>
 
                         <div class="caution">
                             <p>
@@ -501,10 +501,10 @@
                             </p>
                         </div>
 
-                        <div class="issue">
-                            <p>
-                                The reason the integration of text fragments into the specification is at risk is
-                                two-fold.
+                        <div class="note">
+
+                            <p>The integration of text fragments into the specification is AT RISK.
+                                The reason is two-fold:
                             </p>
 
                             <ol>
@@ -567,7 +567,7 @@
                         </table>
                     </section>
 
-                    <section class="informative">
+                    <!-- <section class="informative">
                         <h5>Text Quote Selector</h5>
                         <p>
                             This Selector describes a range of text by copying it, and including some of the text
@@ -649,7 +649,7 @@
                             See the caveats on using text fragments in the section on <a
                                 href="#fragment-selector">Fragment Selectors</a>.
                         </p>
-                    </section>
+                    </section> -->
 
                     <section class="informative">
                         <h5>Text Position Selector</h5>
@@ -664,21 +664,10 @@
                             as defined in the [[[annotation-model]]] specification, but lists it here for an easier
                             readability.
                         </p>
-                        <p>
-                            For HTML content, the stream of characters correspond to the output
-                            of the <a
-                                href="https://html.spec.whatwg.org/multipage/dom.html#the-innertext-idl-attribute"><code>innerText</code>
-                                attribute</a> on the <code>body</code> element. The selection of the text must be in
-                            terms of unicode code points (the "character number").
-                            Selections should not start or end in the middle of a grapheme cluster.
-                            The selection must be based on the logical order of the text, rather than the visual
-                            order, especially for bidirectional text.
-                        </p>
 
-                        <p class="issue">
-                            The last paragraph is more of a placeholder rather than anything else. The original specification does not
-                            say what "stream" is, and the exact specification is also dependent on whether the selector serves as a refinement or not.
-                            In any case, if indeed this must be specified here, the section should become normative.
+                        <p>
+                            The text must be selected and normalized in the same way as for the <a href="annotation-model#text-quote-selector">Text Quote Selector</a> before counting the number of
+                            characters to determine the start and end positions.
                         </p>
                         <table class="zebra">
                             <thead>
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -22,6 +22,13 @@
                     company: "EDRLab",
                     companyURL: "https://www.edrlab.org",
                     "w3cid": 91888
+                },{
+                    name: "Ivan Herman",
+                    url: "https://www.w3.org/People/Ivan/",
+                    company: "W3C",
+                    w3cid: 7382,
+                    orcid: "0000-0003-0782-2704",
+                    companyURL: "https://www.w3.org",
                 }],
                 includePermalinks: true,
                 permalinkEdge: true,
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -657,7 +657,7 @@
                             selection in the stream.
                             Position 0 would be immediately before the first character, position 1 would be immediately
                             before the second character, and so on.
-                           
+
                             This specification reuses the <a
                                 data-cite="annotation-model#text-position-selector"><code>TextPositionSelector</code></a>,
                             as defined in the [[[annotation-model]]] specification, but lists it here for an easier
@@ -672,6 +672,12 @@
                             Selections should not start or end in the middle of a grapheme cluster.
                             The selection must be based on the logical order of the text, rather than the visual
                             order, especially for bidirectional text.
+                        </p>
+
+                        <p class="issue">
+                            The last paragraph is more of a placeholder rather than anything else. The original specification does not
+                            say what "stream" is, and the exact specification is also dependent on whether the selector serves as a refinement or not.
+                            In any case, if indeed this must be specified here, the section should become normative.
                         </p>
                         <table class="zebra">
                             <thead>
@@ -1338,8 +1344,8 @@
                 <section>
                     <h2 id="4-2-using-multiple-selectors"> Using multiple selectors </h2>
                     <p> It is recommended that Reading Systems export multiple selectors, including at least
-                        one precise selector (e.g. CssSelector + TextPositionSelector).
-                        <!-- and one selector resistant to content modifications (e.g. ProgressionSelector). -->
+                        one precise selector (e.g. CssSelector + TextPositionSelector),
+                        and one selector resistant to content modifications (e.g., based on text fragments).
                     </p>
                     <p> When displaying an annotation, a Reading System is free to use the most precise
                         Selector available. It will select an alternative Selector as a fallback in case the
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -787,7 +787,7 @@
                     </pre>
                         </aside>
                         <p>This selects "q" from "quick" as start position and
-                            "x" from the first instance of "fox" as end position in the following HTML snippet: </p>
+                            "x" from "fox" as end position in the following HTML snippet: </p>
                         <pre>
 &lt;div id="intro">
     &lt;p>Some text.&lt;/p>
@@ -1452,4 +1452,5 @@
 		</section> -->
         <section id="index"></section>
     </body>
+
 </html>
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -1,15 +1,16 @@
 <!DOCTYPE html>
 <html lang="en-US">
-	<head>
-		<meta charset="utf-8" />
-		<meta name="color-scheme" content="light dark" />
-		<title> EPUB Annotations 3.4 </title>
-		<script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove">
-		</script>
-		<script src="../../common/js/css-inline.js" class="remove"></script>
+
+    <head>
+        <meta charset="utf-8" />
+        <meta name="color-scheme" content="light dark" />
+        <title> EPUB Annotations 3.4 </title>
+        <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove">
+        </script>
+        <script src="../../common/js/css-inline.js" class="remove"></script>
         <script src="../../common/js/add-caution-hd.js" class="remove"></script>
-		</script>
-		<script class="remove">
+        </script>
+        <script class="remove">
             var respecConfig = {
                 group: "pm",
                 wgPublicList: "public-pm-wg",
@@ -38,11 +39,11 @@
                     repoURL: "https://github.com/w3c/epub-specs",
                     branch: "main"
                 },
-                preProcess:[inlineCustomCSS],
+                preProcess: [inlineCustomCSS],
                 testSuiteURL: "https://w3c.github.io/epub-tests/index.html",
                 xref: {
                     profile: "web-platform",
-                    specs:[ "epub-34"]
+                    specs: ["epub-34"]
                 },
                 lint: { "no-unused-dfns": false },
                 localBiblio: {
@@ -55,44 +56,47 @@
                         "date": "26 November 2025",
                         "publisher": "W3C"
                     },
-                    "scroll-to-text-fragment" : {
+                    "scroll-to-text-fragment": {
                         "authors": [
                             "Nick Burris",
                             "David Bokan"
                         ],
-                        "title" : "URL Fragment Text Directives",
-                        "publisher" : "W3C",
-                        "href" : "https://wicg.github.io/scroll-to-text-fragment/"
+                        "title": "URL Fragment Text Directives",
+                        "publisher": "W3C",
+                        "href": "https://wicg.github.io/scroll-to-text-fragment/"
                     }
                 }
             };</script>
-	</head>
-	<body>
-		<section id="abstract">
-			<p> This document defines a profile of the [[[annotation-model]]] [[annotation-model]]
+    </head>
+
+    <body>
+        <section id="abstract">
+            <p> This document defines a profile of the [[[annotation-model]]] [[annotation-model]]
                 by specifying a subset of the properties allowed in this model, and adding
-				properties deemed useful to satisfy the [[[epub-ann-ucr]]] [[epub-ann-ucr]].
-			</p>
-		</section>
-		<section id="sotd"></section>
+                properties deemed useful to satisfy the [[[epub-ann-ucr]]] [[epub-ann-ucr]].
+            </p>
+        </section>
+        <section id="sotd"></section>
         <section id="conformance">
             <!-- <p>All algorithm explanations are <em>non-normative</em>.</p> -->
         </section>
 
-		<section id="1-profile-of-the-w3c-annotation-data-model" class="informative">
-			<h1> Subset of the Web Annotation Data Model </h1>
-			<p> This section defines a profile of the [[[annotation-model]]] [[annotation-model]], as used for EPUB
-				Annotations.</p>
+        <section id="1-profile-of-the-w3c-annotation-data-model" class="informative">
+            <h1> Subset of the Web Annotation Data Model </h1>
+            <p> This section defines a profile of the [[[annotation-model]]] [[annotation-model]], as used for EPUB
+                Annotations.</p>
             <p> In the Web Annotation model, the core structure is the
                 <a data-cite="annotation-model#annotations">Annotation object</a>,
                 which contains properties defining the annotation's
                 <a data-cite="annotation-model#bodies-and-targets">Body and Target</a>.
-                EPUB Annotations reuse the <a href="#1-1-annotation-object">same model</a> with some restrictions listed below.
+                EPUB Annotations reuse the <a href="#1-1-annotation-object">same model</a> with some restrictions listed
+                below.
                 Subsequent sections provide more formal definitions for the terms used by this
                 specification.
             </p>
-           <ul>
-                 <li> Web Annotations may have 0 or more Bodies, whereas this document requires to have a single <a href="#1-4-body">Body</a>
+            <ul>
+                <li> Web Annotations may have 0 or more Bodies, whereas this document requires to have a single <a
+                        href="#1-4-body">Body</a>
                     per annotation. Furthermore, in the [[annotation-model]], such Body can be remote or embedded,
                     whereas only embedded Bodies are used in EPUB Annotation
                     for textual comments .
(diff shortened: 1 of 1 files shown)


sueneu:
Adding this here, there is a slight styling issue when viewing this in night mode, perhaps related to the "zebra" class of the table?
Screenshot 2026-02-03 at 10 02 12 AM

iherman:
@sueneu this is strange. It looks fine on my screen:

Screenshot 2026-02-03 at 16 17 50

I tested it on Vivaldi, Firefox, and Safari, they look identical...

sueneu:
I’m in Chrome, perhaps that’s the issue. It’s easy enough to disable the backgrounds in dev tools. 

Chrome Version 144.0.7559.110 (Official Build) (arm64)
MacBook Pro
Sequoia 15.7.3

I’ll try Firefox. Thank you!

epub34/annotations/index.html
```diff
@@ -330,133 +369,465 @@ <h4 id="1-3-1-source"> Source </h4>
...
+                                        }
+                                    }
+                                ]
+                            }
+                        </code>
+                    </pre>
+                        </aside>
+                        <p>This selects "q" from "quick" as start position and
+                            "x" from "fox" as end position in the following HTML snippet: </p>
+                        <pre>
```
sueneu, line 787:
```suggestion
                        </aside>
                        <p>This selects "q" from "quick" as start position and
                            "x" from the first instance of "fox" as end position in the following HTML snippet: </p>
                        <pre>
```

sueneu (commented):
This looks good. One comment/suggestion below. Since there are two instances of "fox" in the CssSelector refined example, let's clarify that the selector is pointing to the first instance of "fox." We could even be clear that using text position as a refinement prevents overreach by the selector and limits the snippet to the first "fox" as intended.

iherman:
> I’m in Chrome, perhaps that’s the issue. It’s easy enough to disable the backgrounds in dev tools.
> 
> Chrome Version 144.0.7559.110 (Official Build) (arm64) MacBook Pro Sequoia 15.7.3
> 
> I’ll try Firefox. Thank you!

Vivaldi uses almost the same engine as Chrome: 144.0.7559.114, and I am also on a MacBook Pro, although on Tahoe and not Sequoia. But I would be surprised if that was the reason...

There are mysteries in this world...

epub34/annotations/index.html
```diff
@@ -30,183 +39,210 @@
...
+                            <th> Required? </th>
+                        </tr>
+                    </thead>
+                    <tbody>
+                        <tr>
+                            <td>
+                                <code> id </code>
+                            </td>
+                            <td>The identity of the annotation. A uuid formatted as a URN is RECOMMENDED.</td>
+                            <td>URL</td>
```
bduga, line 157: I have no idea if this change is correct, but I would kind of like to understand if it is. Is this supposed to be a URL and not a URI like it was before? But if it is a URL, how can it be a URN, as we recommend? And is the change from IRI (from web annotations) intentional? If you really want this to be a URL, you will also have to change the text about recommending a URN, but also I think such a change should be done in a new PR, with an associated bug. Same comment applies for other changes, below.

epub34/annotations/index.html
```diff
@@ -330,133 +369,465 @@ <h4 id="1-3-1-source"> Source </h4>
...
-				</code>
-			</pre>
+                <section>
+                    <h4 id="1-3-2-selector">Selector</h4>
+
+                    <p>
+                        An annotation refers to a segment of a resource, which is identified by one or more
+                        <a data-cite="annotation-model#selectors">Selectors</a>.
+                        The nature of the Selectors and methods to describe segments depend on
+                        the resource type. Providing more than one Selectors allows an annotation software to
```
bduga: I am not sure why this was pluralized (made plural? Pluralled? I don't understand how this language works), but it seems wrong.
```suggestion
                        the resource type. Providing more than one Selector allows an annotation software to
```

epub34/annotations/index.html
```diff
@@ -330,133 +369,465 @@ <h4 id="1-3-1-source"> Source </h4>
...
+                    </section>
+
+                    <section class="informative">
+                        <h5>Text Position Selector</h5>
+                        <p>
+                            This Selector describes a range of text by recording the start and end positions of the
+                            selection in the stream.
+                            Position 0 would be immediately before the first character, position 1 would be immediately
+                            before the second character, and so on.
+                            The start character is thus included in the list, but the end character is not.
```
bduga: I don't understand this sentence. I think the text is clearer without it. And yes I realize this is from the Web Annotations spec, where I also don't understand it.
```suggestion
                           
```

epub34/annotations/index.html
```diff
@@ -330,133 +369,465 @@ <h4 id="1-3-1-source"> Source </h4>
...
+                            This specification reuses the <a
+                                data-cite="annotation-model#text-position-selector"><code>TextPositionSelector</code></a>,
+                            as defined in the [[[annotation-model]]] specification, but lists it here for an easier
+                            readability.
+                        </p>
+                        <p>
+                            For HTML content, the stream of characters correspond to the output
```
bduga: This sentence seems *very* normative. I don't think Web Annotations defines this, but it is stated here as fact even though the section is non-normative. We might want to define this, but then we should do it normatively, or just recommend a method. Also I think this is wrong - do I really need to get the innerText of the body element when using this to refine a position from a CSS selector? Can't I just get the innerText of the node referenced from the CSS?

epub34/annotations/index.html
```diff
@@ -930,111 +1307,121 @@ <h3 id="2-3-serialization"> Serialization </h3>
...
-    				if the publication has been modified after the annotation has been created. </p>
-    			<p> Not all selectors are equally easy to implement. Reading Systems MAY choose to
-    				support only a subset of the selectors defined in this specification. </p>
-    			<p> The W3C Publishing Maintenance Working Group is expected to define one or more selectors
-    				reading systems are required to implement, as a <i>lingua franca</i>. </p>
+                <section>
+                    <h2 id="4-2-using-multiple-selectors"> Using multiple selectors </h2>
+                    <p> It is recommended that Reading Systems export multiple selectors, including at least
+                        one precise selector (e.g. CssSelector + TextPositionSelector).
+                        <!-- and one selector resistant to content modifications (e.g. ProgressionSelector). -->
```
bduga: Why this removal? I agree that the example isn't great, but the recommendation seems sound.

epub34/annotations/index.html
```diff
@@ -330,133 +369,465 @@ <h4 id="1-3-1-source"> Source </h4>
...
+                                        The selector must have exactly 1 <code>value</code> property.</td>
+                                    <td> string </td>
+                                    <td> Yes </td>
+                                </tr>
+                            </tbody>
+                        </table>
+                    </section>
+
+                    <section class="informative">
+                        <h5>Text Quote Selector</h5>
```
bduga, line 571: Did we really decide to go with text quote? I thought we were going straight for text fragment.

bduga (changes requested):
The signal to noise ratio in this PR is quite low. Is there some way in the future to break out all the formatting changes? I had a lot of trouble finding actual changes.

epub34/annotations/index.html
mattgarrish: A URN can be a URL if you're citing the URL standard's definition, but a reference would help.

bduga: Maybe? According to IANA a URN can be a URI, but just because something is a URI doesn't mean it is a URL. Of course, my point is this seems to be an change that is not related to the task at hand, and whose discussion is being lost in the comments of a PR review. I would like to understand if this change makes sense, and if it does we should make it, but I would prefer to have a clear reason for the change stated in a bug where we can discuss and resolve it, with a dedicated PR to make the change so we can easily see what is happening.

mattgarrish: > According to IANA a URN can be a URI, but just because something is a URI doesn't mean it is a URL.

Sure, but the URL standard is trying to break from these distinctions. See their [goals section](https://url.spec.whatwg.org/#goals).

> my point is this seems to be an change that is not related to the task at hand

No comment from me on that part. Fair point given it needs more explanation. I just saw your question about the URL change and assumed the idea is to match up with the change from URIs we made in the core specs.

iherman: > I just saw your question about the URL change and assumed the idea is to match up with the change from URIs we made in the core specs.

That is exactly what happened. The previous version used URI and URL, it was therefore inconsistent. I looked at the authoring spec and took URL to be consistent with it.

epub34/annotations/index.html
iherman: Yeah, good point. I tried to locate this in the Annotation spec and I seem to remember that Laurent said it does say it is the inner text for HTML (I asked this question on the call).

Let us put the normativeness issue aside (if necessary, we can make this section normative) my feeling is that, without it, this selector is underspecified. The original text just say "stream" without saying what it is. We can either drop this selector or properly specify it.

@llemeurfr do you remember what you based your statement on at the call.

epub34/annotations/index.html
iherman: Because Laurent and I propose to remove the progression selector (which is now in comment) as grossly underspecified. Sorry, I should have made it clear in the PR description.

I am happy to put back the core part of the sentence, but I was not sure which selector is resistant to modification; I do not believe any of them really are (although the level of vulnerability is different).

epub34/annotations/index.html
iherman: I do not believe we made a clear statement on any. I think there was an agreement that we should not define a new, dedicated selector for text quote, but then the two approaches in the draft are the two alternatives. As I said in the introduction we may decide to remove one of them.

epub34/annotations/index.html
iherman: I presume the right technical answer is that

- if the selector is used to refine a previous selection, then it is the .innerText of that previous selection;
- otherwise it is the body.innerText

But that is definitely not detailed on the original spec. Which raises the issue for me whether that selector was only defined on a real stream only... in which case we may have to drop it.

epub34/annotations/index.html
iherman: I wonder where to put a reference to URL. Defining a list of terms is an option (selector, target, annotation, that sort of things, which may include the URL), but that may seem a bit too much. Other possibility is to refer, somehow, to the EPUB 3.4 in general, and to https://www.w3.org/TR/epub-34/#sec-overview-relations-url in particular (which would make sense, EPUB 3.4 is the first among equals in the EPUB 3.4 family)

iherman:
@mattgarrish do you have any idea why the "Caution" header does not appear in the caution box in https://raw.githack.com/w3c/epub-specs/ann-selectors-2/epub34/annotations/index.html#1-3-2-selector ? I do include the relevant js file...

epub34/annotations/index.html
bduga: I have no idea what is correct here, and honestly WHATWG isn't really helping the matter. So there used to be URLs and URNs which used to be subclasses of URIs, but now URIs are URLs so URLs are a subclass of themselves? And URNs are subclasses of URLs and also URLs? Or they no longer exist? But why would anyone want a URL here? I guess the entire concept of identity and location being different is out the window, so maybe it is time to ditch UR[I|L] as the foundation of ids. Anyway, this feels outside the scope of this change.

epub34/annotations/index.html
bduga: My issue isn't necessarily that it is based on innerText (though I don't think the web annotations spec says that it is), but rather that the stream is based on the text of the body element. I would think that the stream of text would be the element under consideration, not body. I am fine making this all normative, but that means we really have to think it through,

epub34/annotations/index.html
bduga: I would leave the core without the example. I guess which selector to recommend will depend on what ends up in the spec, but text fragments seem more resilient (immune to unrelated text insertions, deletions, and moves).

epub34/annotations/index.html
bduga: Sounds good.

epub34/annotations/index.html
bduga: > Which raises the issue for me whether that selector was only defined on a real stream only... in which case we may have to drop it.

Given that it probably doesn't make much sense worrying too much about it now. How about leaving it as-is with a note that it is wrong and will either need to be fixed or the selector will need to be dropped?

epub34/annotations/index.html
iherman: > Anyway, this feels outside the scope of this change.

This is an issue for the full EPUB 3.4 family; the change in this document has just made it consistent.

At this moment, the usual approach, taken in all the various specs, is to reference the whatwg spec for URL, and use that term. In theory, the whatwg spec's URL incorporates URNs and URIs and, if there is a problem, it should be handled over there...

epub34/annotations/index.html
iherman: +1. Done in https://github.com/w3c/epub-specs/pull/2912/commits/86673dfe775a44216dfbea7974b52f57bfbd0f71

epub34/annotations/index.html
iherman: Done in https://github.com/w3c/epub-specs/pull/2912/commits/86673dfe775a44216dfbea7974b52f57bfbd0f71

epub34/annotations/index.html
```diff
@@ -30,183 +39,210 @@
                     repoURL: "https://github.com/w3c/epub-specs",
                     branch: "main"
                 },
-                preProcess:[inlineCustomCSS],
-                testSuiteURI: "https://w3c.github.io/epub-tests/index.html",
+                preProcess: [inlineCustomCSS],
```
mattgarrish, line 42: This will fix the missing caution header:
```suggestion
                preProcess: [inlineCustomCSS],
                postProcess: [addCautionHeaders],
```

iherman: Ouch. I am stupid...

thanks!

llemeurfr (approved):
This is a good step forward. Because it mixes content and spacing modifications, we can merge it and work on my comments later. 

- We should add one or more examples of using fragment selectors to avoid syntax errors by implementers. 
- The use of the wicg URL for the text fragment selector is a subject for discussion. If we create implementations now and have to change this URL later (when a spec is approved), it will be backward-compatibility hell. But I don't see a better way. 
- I'm annoyed by the use of percent encoding on the text fragment selector: it makes no sense IMO, as far as this string is not used as a URL fragment. 
- You copy the definition of other selectors with a "non-normative" label. As it is, it can lead to confusion for implementers. In fact, it IS normative, but copied from the parent spec. 
- About the Text Position Selector, the inital spec states "The text MUST be selected and normalized in the same way as for the Text Quote Selector before counting the number of characters to determine the start and end positions." And the Text Quote Slector section states "The text MUST be normalized before recording in the Annotation. Thus HTML/XML tags SHOULD be removed, and character entities SHOULD be replaced with the character that they encode. ". Therefore, the note about processing HTML content can refer to the initial spec, or rephrase it more clearly.

iherman:
> This is a good step forward. Because it mixes content and spacing modifications, we can merge it and work on my comments later.
> 
> * We should add one or more examples of using fragment selectors to avoid syntax errors by implementers.

Absolutely yes.

> * The use of the wicg URL for the text fragment selector is a subject for discussion. If we create implementations now and have to change this URL later (when a spec is approved), it will be backward-compatibility hell. But I don't see a better way.

At this moment we have the choice of using a fragment selector or using the Text Quote Selector only, and rely on implementers to use the wcig stuff. We may want to remove one of the two.

> * I'm annoyed by the use of percent encoding on the text fragment selector: it makes no sense IMO, as far as this string is not used as a URL fragment.

Yeah, that is one of the advantages of using the text quote selector.

> * You copy the definition of other selectors with a "non-normative" label. As it is, it can lead to confusion for implementers. In fact, it IS normative, but copied from the parent spec.

Per W3C rules we cannot do that. We have to make it clear that this is quoted, and we do not do any definitions...

> * About the Text Position Selector, the inital spec states "The text MUST be selected and normalized in the same way as for the Text Quote Selector before counting the number of characters to determine the start and end positions." And the Text Quote Slector section states "The text MUST be normalized before recording in the Annotation. Thus HTML/XML tags SHOULD be removed, and character entities SHOULD be replaced with the character that they encode. ". Therefore, the note about processing HTML content can refer to the initial spec, or rephrase it more clearly.

But that text does not appear in the specification of the text position selector: https://www.w3.org/TR/annotation-model/#text-position-selector (which might be a bug in that spec!) Hence the problem...

iherman:
This was discussed during the [pmwg meeting on 12 February 2026](https://w3c.github.io/pm-wg/minutes/2026-02-12.html#9fc1).

<h3>Annotations</h3>
<p><cite>&lt;wendyreid&gt;</cite> <a href="https://github.com/w3c/epub-specs/pull/2912" rel="noopener noreferrer">w3c/<wbr>epub-specs#2912</a></p>
<p><cite>wendyreid:</cite> and if everyone is good with that, then we can move on to talking about the first public working draft</p>
<p><cite>ivan:</cite> I'm picking up this PR for Laurent who is on vacation. This PR makes the old one obsolete<br>
<span>… I recorded the list of selectors we agreed on two weeks ago</span><br>
<span>… there are two issues for discussion</span><br>
<span>… we discussed one last time, how to handle the text fragment</span><br>
<span>… we didn't agree on how to do that</span><br>
<span>… so I added two approaches, it would be best to remove one of them</span><br>
<span>… I removed some of the selectors from our table, and added text fragment selector</span><br>
<span>… it has two downsides, it is not yet a standard, and we don't know when it will be</span><br>
<span>… we don't know if the syntax will stay on or if it will change</span><br>
<span>… causing issues</span><br>
<span>… if it is a URL it has to be encoded, which would make the text fragment pretty much unreadable</span><br>
<span>… we have a quote selector in the original text</span><br>
<span>… I added that with a note that it is up to the implementers to decide if they want to use it and how</span><br>
<span>… there are two advantages to leaving this to the implementers</span><br>
<span>… the person quoting can set the quote</span><br>
<span>… and we don't have to worry about the syntax change</span><br>
<span>… but we are pushing this into the reading system's court</span><br>
<span>… the second problem: we discussed the text position selector</span><br>
<span>… which counts characters and selects the text between two integers.</span><br>
<span>… how do we do this in HTML</span><br>
<span>… the text position selector was intended for pure text files</span><br>
<span>… there is normative text that would work and is applied to a different selector and doesn't apply here</span><br>
<span>… all other things are editorial and don't need discussion</span></p>
<p><cite>LaurentLM:</cite> The text must be selected before using the text position selector<br>
<span>… when I read the two sections, quote selector, and text position selector</span><br>
<span>… there is a conditional approach, if you have to work with copyrighted protected contents, use the text position selector, otherwise use the quote selector</span></p>
<p><cite>ivan:</cite> I understand now, my comments are withdrawn</p>
<p><cite>duga:</cite> the normalization is ugly, but its OK, because that's what the original text says<br>
<span>… but its not clear where the text stream starts</span><br>
<span>… where do I get the position for the overall text</span></p>
<p><cite>LaurentLM:</cite> I agree that we should specify the text origin</p>
<p><cite>duga:</cite> perhaps it is the inner text of the closest element that we are talikng about, if it there isn't one then it is body.</p>
<p><cite>Hadrien:</cite> I don't want to do the second option, I don't want to have to process all of the text before I start<br>
<span>… the risk of the syntax changing for text fragment isn't very big</span><br>
<span>… it is widely used, and the people working on this know that</span></p>
<p><cite>wendreid:</cite> where does this leave us?</p>
<p><cite>ivan:</cite> we have a three way answer for this<br>
<span>… 1. we keep it as it is in the PR</span></p>
<p><cite>&lt;ivan&gt;</cite> 1. We keep text fragments in both places</p>
<p><cite>&lt;ivan&gt;</cite> 2. remove it from the fragment selector</p>
<p><cite>Laurent:</cite> In both cases you have the string you want to use</p>
<p><cite>&lt;ivan&gt;</cite> 3. we remove it from the text quote selector</p>
<p><cite>LaurentLM:</cite> with the text fragment selector you can keep just the begining and the end<br>
<span>… the choice we've got is not between different models, it is</span><br>
<span>… between a syntax that is within the specification but not widely known</span><br>
<span>… and with one that is widely known but not part of the specifcation</span></p>
<p><cite>duga:</cite> I think we should choose one, and keep in mind we might need to switch</p>
<p><cite>ivan:</cite> I almost agree with Brady, it is safer to keep the text quote selector<br>
<span>… we push the specification out of the problem</span></p>
<p><cite>duga:</cite> as long as we pick one I'm happy</p>
<p><cite>ivan:</cite> the annotation standard we develop is for exchanging annotations between systems<br>
<span>… it is not about the styling, and the text quote selector will be much more readable</span></p>
<p><cite>duga:</cite> the text fragment selector is probably better defined<br>
<span>… having a well defined algorithm is important, but that is a guess</span></p>
<p><cite>LaurentLM:</cite> I agree with Brady, the text fragment selector seems well written<br>
<span>… but the quote selector is not</span><br>
<span>… in the text fragment selector you can leave the middle unsaid</span></p>
<p><cite>duga:</cite> that is a huge difference when we talk about copyright.<br>
<span>… some systems use precentage of the book for copyright</span><br>
<span>… with the fragment selector, we wouldn't be using up so much of the text</span></p>
<p><cite>ivan:</cite> readability is not an option in the HTML spec</p>
<p><cite>ivan:</cite> I change my vote, and propose we remove the quote selector from our spec</p>
<p><cite>LaurentLM:</cite> at the time we want to close the recommendation if there is not standard, should we plan now for what we will do?</p>
<p><cite>GeorgeK:</cite> publishers in education use both EPUB reading systems and HTML reading systems. Using quote selector would bridge those systems</p>
<p><cite>Hadrien:</cite> I think we shouldn't publish this if the syntax isn't final. I think it is fine to wait a bit more, there is no reason to rush</p>
<p><cite>ivan:</cite> I'm not sure I understand what Hardrien said</p>
<p><cite>Hadrien:</cite> I am responding to LaurentLM's question</p>
<p><cite>ivan:</cite> we can make a note in the spec now about the possible fallback<br>
<span>… today we should try to publish a spec that is as close as we can to what we want for its final form</span></p>
<p><cite>LaurentLM:</cite> if the text fragment syntax isn't finalized, I think we should still publish at the end of the year</p>
<p><cite>Hadrien:</cite> but it doesn't need to be normative</p>
<p><cite>wendyreid:</cite> we could leave the annotation spec in the same state as the text fragment spec, as a working group note</p>
<p><cite>duga:</cite> when should we worry about this? we should wait until it was an issue<br>
<span>… rather than solving all possible outcomes</span><br>
<span>… I think it is too early to worry about it</span></p>
<p><cite>ivan:</cite> we should move on publishing to the first working draft<br>
<span>… we get the horizontal reviews</span><br>
<span>… at some point we can say that we suspend the work because we are waiting on a dependent spec</span><br>
<span>… for the time being we should move ahead believing that the syntax will be resolved</span></p>
<p><cite>wendyreid:</cite> our next step is to decide we can announce this as a first draft<br>
<span>… we can still make changes</span><br>
<span>… this is a good time to publish the first draft and invite feedback</span><br>
<span>… is there any opposition to publishing this?</span></p>
<p><cite>ivan:</cite> I don't oppose this. There are two more documents that we may want to publish at some point<br>
<span>… 1. the use case document</span><br>
<span>… we might want to publish this at the same time</span></p>
<p><cite>&lt;ivan&gt;</cite> <a href="https://github.com/w3c/epub-specs/pull/2922" rel="noopener noreferrer">vocab</a></p>
<p><cite>ivan:</cite> the other one is<br>
<span>… I have been developing the vocabulary in parallel</span><br>
<span>… it might be good to publish this together because the annotation has references to it</span><br>
<span>… and to make the links live properly, we would need to have the vocabulary document live</span></p>
<p><cite>wendyreid:</cite> is the vocabulary a note?<br>
<span>… we have to publish this as a working draft, and develop a short name</span></p>
<p><cite>LaurentLM:</cite> is the a problem with "ann"</p>
<p><cite>duga:</cite> it looks alot like "announcements"</p>
<p><cite>LaurentLM:</cite> "annot"? is less confusing?</p>
<p><cite>ivan:</cite> i would go with epub-anno</p>
<p><cite>wendyreid:</cite> that would be consistent with our pattern</p>
<p><cite>ivan:</cite> there is one more question, do we want to add a version number?<br>
<span>… if so, 1.0 or 3.4?</span></p>
<p><cite>LaurentLM:</cite> I would go with 3.4 since it is related to EPUB 3.4</p>
<p><cite>ivan:</cite> the version number would go in the title</p>
<p><cite>shinya:</cite> since the accessibility is 1.2, 1.0 is good</p>
<p><cite>magarrish:</cite> if this is revision bound with 3.4 then I have no problem with it</p>
<p><cite>&lt;AvneeshSingh&gt;</cite> EPUB accessibility is independent of EPUB 3 version.</p>
<p><cite>LaurentLM:</cite> if we don't modify epub, why would we push annotations changes</p>
<p><cite>&lt;shiestyle&gt;</cite> +1 to duga</p>
<p><cite>duga:</cite> it seems like 1.0 makes more sense, then we can update annotations without updating EPUB specs</p>
<p><cite>&lt;wendyreid&gt;</cite> Proposed: Publish EPUB Annotations 1.0 as a FPWD, with the shortname epub-anno</p>
<p><cite>&lt;DaleRogers&gt;</cite> +1</p>
<p><cite>&lt;LaurentLM&gt;</cite> +1</p>
<p><cite>&lt;sueneu&gt;</cite> +1</p>
<p><cite>&lt;gman&gt;</cite> +1</p>
<p><cite>&lt;AvneeshSingh&gt;</cite> +1</p>
<p><strong>RESOLUTION:</strong> Publish EPUB Annotations 1.0 as a FPWD, with the shortname epub-anno</p>
<p><cite>&lt;GeorgeK&gt;</cite> +1</p>
<p><cite>&lt;wendyreid&gt;</cite> Proposed: Publish the EPUB Annotations Vocabulary as a draft note, with the shortname epub-anno-vocab</p>
<p><cite>&lt;GeorgeK&gt;</cite> +1</p>
<p><cite>&lt;sueneu&gt;</cite> +1</p>
<p><cite>&lt;DaleRogers&gt;</cite> +1</p>
<p><cite>&lt;gman&gt;</cite> +1</p>
<p><cite>&lt;AvneeshSingh&gt;</cite> +1</p>
<p><cite>&lt;LaurentLM&gt;</cite> +1</p>
<p><strong>RESOLUTION:</strong> Publish the EPUB Annotations Vocabulary as a draft note, with the shortname epub-anno-vocab</p>
<p><cite>ivan:</cite> is the use UCR ready to publish?</p>
<p><cite>LaurentLM:</cite> yes it is stable</p>
<p><cite>&lt;wendyreid&gt;</cite> Proposed: Publish the EPUB Annotations Use Cases document as a working group note, with the shortname epub-anno-ucr</p>
<p><cite>&lt;LaurentLM&gt;</cite> +1</p>
<p><cite>&lt;gman&gt;</cite> +1</p>
<p><cite>&lt;sueneu&gt;</cite> +1</p>
<p><cite>&lt;DaleRogers&gt;</cite> +1</p>
<p><cite>&lt;GeorgeK&gt;</cite> +1</p>
<p><cite>&lt;AvneeshSingh&gt;</cite> +1</p>
<p><strong>RESOLUTION:</strong> Publish the EPUB Annotations Use Cases document as a working group note, with the shortname epub-anno-ucr</p>

<hr /></details>

iherman:
Per agreement on the meeting, merging.

PR #2939, opened 2026-02-25
Update issue templates for annotations

iherman:
Also added a template for annotations vocabulary

### DIFF
--- a/.github/ISSUE_TEMPLATE/31-epub-annotations.yml
+++ b/.github/ISSUE_TEMPLATE/31-epub-annotations.yml
@@ -1,20 +1,20 @@
 name: EPUB Annotations
-description: Issue related to the EPUB Annotations Recommendation 
+description: Issue related to the EPUB Annotations Recommendation
 labels: Spec-Annotations
-body: 
+body:
   - type: markdown
     attributes:
      value: "## Issue on the EPUB Annotations Recommendation"
   - type: markdown
     attributes:
-     value: __Before opening an issue__, please check that it has not already been resolved in the [latest version](https://www.w3.org/TR/epub-ann-34/)
+     value: __Before opening an issue__, please check that it has not already been resolved in the [latest version](https://www.w3.org/TR/epub-anno-10/)
   - type: input
     id: section
     attributes:
       label: Section
       description: Please provide a link to the relevant section of the specification
     validations:
-      required: false 
+      required: false
   - type: textarea
     id: problem
     attributes:
--- a/.github/ISSUE_TEMPLATE/32-epub-anno-ucr.yml
+++ b/.github/ISSUE_TEMPLATE/32-epub-anno-ucr.yml
@@ -1,10 +1,10 @@
 name: EPUB Annotations UCR
-description: Issue related to the EPUB Annotations Use Cases and Requirement Note
+description: Issue related to the EPUB Annotations Use Cases and Requirement WG Note
 labels: Spec-AnnotationsUCR
 body:
   - type: markdown
     attributes:
-     value: "## Issue on the EPUB Annotations Use Cases and Requirement Note"
+     value: "## Issue on the [EPUB Annotations Use Cases and Requirement](https://www.w3.org/TR/epub-anno-ucr/) WG Note"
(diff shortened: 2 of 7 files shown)


.github/ISSUE_TEMPLATE/33-epub-annotations-vocabulary.yml
```diff
@@ -0,0 +1,30 @@
+name: EPUB Annotations
+description: Issue related to the EPUB Annotations Vocabulary Recommendation
```
mattgarrish: This one's only a note, right?
```suggestion
description: Issue related to the EPUB Annotations Vocabulary Note
```

.github/ISSUE_TEMPLATE/31-epub-annotations.yml
```diff
@@ -1,20 +1,20 @@
 name: EPUB Annotations
-description: Issue related to the EPUB Annotations Recommendation 
+description: Issue related to the EPUB Annotations Recommendation
```
mattgarrish: I don't care too much about this one, but should we call it a recommendation before it's actually a recommendation? I leave it to you...

.github/ISSUE_TEMPLATE/32-epub-anno-ucr.yml
```diff
@@ -1,20 +1,20 @@
 name: EPUB Annotations UCR
```
mattgarrish, line 1: Another really minor thing, but all the other file names have the short name hyphenation. The OCD in me is screaming out that it should be "-ucr.yml" but if you don't want to bother with a rename I'll survive... 😄

iherman: +1. This actually holds for all three annotation-related templates; I have renamed all three.

.github/ISSUE_TEMPLATE/31-epub-annotations.yml
iherman: Yeah... I think leaving the term recommendation around is better, so that to give the required "weight" to the discussion. But I have added a reference to the draft: it now says "Issue related to the Draft of the EPUB Annotations Recommendation".

There was another discrepancy, actually: the EPUB template just refers to EPUB 3 in the description and, more importantly, it refers to the versionless short name: https://www.w3.org/TR/epub/. But then the template refers to the versioned, 3.4., short name for the latest version. Which means that if someone really intends to comments on the 3.4 version then the title URL may be misleading. I guess this is all right because the template is really good for outsiders of the group.

I did not touch this, but follow the same pattern for the new ones (although, in that case, they all refer to the same document).

iherman:
Some more changes, while I was at it:

- The A11y references still used -111 in the URL-s. I changed them to -12 and also added a URL for the description field
- I have also added the anno-* references to the general script

The second item made me think: if someone active in the WG wants to raise an issue, these templates may be a drag, in fact (at least it was true for me). I wonder whether we could do a simplified template for WG members: is this at all possible?

.github/ISSUE_TEMPLATE/32-epub-anno-ucr.yml
iherman: Handled in https://github.com/w3c/epub-specs/pull/2939/commits/f7fce18b772f365d73c5e96f0052def793c99466

.github/ISSUE_TEMPLATE/31-epub-annotations.yml
iherman: Handled in https://github.com/w3c/epub-specs/pull/2939/commits/f7fce18b772f365d73c5e96f0052def793c99466

.github/ISSUE_TEMPLATE/33-epub-annotations-vocabulary.yml
iherman: Handled in https://github.com/w3c/epub-specs/pull/2939/commits/f7fce18b772f365d73c5e96f0052def793c99466

iherman:
Merging; can be updated for stylistic reasons at any time...

PR #2949, opened 2026-03-06
Update of the generated context file for the EA vocabulary

iherman:
The update of yml2vocab tool in the way external context files are handled is applied.


iherman:
@mattgarrish just for the good order, any issues merging this? I have tested the generated context file and found some JSON-LD problems with the generating tool. This PR is just the result of that update...

Issue #2942, opened 2026-02-27
Add markdown to the possible annotation bodies (edited; was HTML instead of markdown)


iherman:
### Section

https://www.w3.org/TR/epub-anno-10/#body

### Describe the problem

At the moment, the `format` property allows for plain text and markdown only. However, in an international setting, we may need HTML as well (e.g., for the inclusion of ruby markup).

### Describe the fix or new feature you propose

Add `text/html` to the possible values of `format`

llemeurfr:
Hi, this is not something use cases and requirements have put forward.  

This merits a deep conversation with RS developers, as accepting HTML for annotations would be VERY impactful for them, as it would mean:
- displaying HTML annotations in a web view, which is heavy for native applications (the annotations are not always displayed as part of the EPUB reading webview).
- adding a Tiny-HTML editor to the body field during edition
- deal with nerdy CSS requests (and their impact if the annotations are displayed as part of the EPUB reading webview). 
- sanitizing HTML annotations both during editing and before displaying annotations. 

HTML is not an easy peasy feature

iherman:
I agree/understand it is not easy to add. But I am a bit worried that i18n (and maybe a11y) considerations may make this unavoidable. Think also annotating with mathematics.

We may try considering a restriction of HTML using some of the [tag categories](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Content_categories) defined by the HTML specs. Or even restricting one of those.

Alternatively, we may explicitly refer to CommonMark as the accepted markdown dialect: common mark includes [html tags](https://spec.commonmark.org/0.31.2/#html-blocks). Would that be a possible middle road? Is that easier to handle for implementers?

iherman:
Hm. The CommonMark does not include ruby markup either. Whether it would work nevertheless may depend on markdown implementations...

iherman:
Adding the i18n issue, inviting the i18n experts to chime in...

murata2makoto:
Ruby annotations are not possible without using HTML markup.  For better support of annotations in Mainland China, Taiwan, and Japan, I support the addition of HTML as bodies.

murata2makoto:
> But I am a bit worried that i18n (and maybe a11y) considerations may make this unavoidable. 

[...]

PR #2950, opened 2026-03-12
Use json-ld keywords for language and direction

### LINKED ISSUE #2854: Use the JSON-LD keywords for international texts
### Section

https://w3c.github.io/epub-specs/epub34/annotations/#body

### Describe the problem

The `Body` object uses the `language` and `textDirection` terms for international texts. This is inherited from the Web Annotation spec, which was created when text directionality was not properly handled in JSON-LD. However, JSON-LD now has the `@language` and `@direction` keywords that properly covers this, see [String Internationalization](http://www.w3.org/ns/anno.jsonld).

I would propose to rely on the JSON-LD spec, possibly aliasing the `language` and `direction` terms to the keywords.

iherman:
This is to fix #2854, implementing the approach in  https://github.com/w3c/epub-specs/issues/2854#issuecomment-3935262418. 

Except that: I did not describe the mechanism for setting the language/direction for an individual string. As I say https://github.com/w3c/epub-specs/issues/2854#issuecomment-3935262418 there are some awkwardnesses; besides, for people not familiar with JSON-LD, it probably feel confusing. On the other hand, I think that the scoped context approach covers our need; I did not see any example where that mechanism would be absolutely necessary. (There is a relation to the discussion in #2942: for more complex settings nothing would beat the facilities offered by HTML…) (Actually, the draft is in the text now, but commented out.)

Note that the whole section is informative. We do ***not*** formally specify anything: we just reuse a mechanism specified elsewhere.

I added the string internationalization as a subsection on Internationalization considerations; we may have other topics coming up. I that is not the case we can make it the only i18n issue later.

---

See:

* For EPUB Annotations 1.0:
    * [Preview](https://raw.githack.com/w3c/epub-specs/anno-use-json-ld-for-language-and-direction/epub34/annotations/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fannotations%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Fanno-use-json-ld-for-language-and-direction%2Fepub34%2Fannotations%2Findex.html)
* For EPUB Annotations Vocabulary 1.0:
    * [Preview](https://raw.githack.com/w3c/epub-specs/anno-use-json-ld-for-language-and-direction/epub34/annotations-vocab/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fannotations-vocab%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Fanno-use-json-ld-for-language-and-direction%2Fepub34%2Fannotations-vocab%2Findex.html)

### DIFF
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -305,7 +305,7 @@
                                 <code> name </code>
                             </td>
                             <td> The name of the creator. </td>
-                            <td> string </td>
+                            <td> natural language string </td>
                             <td> No </td>
                         </tr>
                     </tbody>
@@ -840,7 +840,7 @@
                                 <code> value </code>
                             </td>
                             <td> The textual content of the annotation. </td>
-                            <td> string </td>
+                            <td> natural language string </td>
                             <td> Yes </td>
                         </tr>
                         <tr>
@@ -870,22 +870,6 @@
                         </tr>
                         <tr>
                             <td>
-                                <code> language </code>
-                            </td>
-                            <td> The language of the annotation. </td>
-                            <td> [[BCP47]] </td>
-                            <td> No </td>
-                        </tr>
-                        <tr>
-                            <td>
-                                <code> textDirection </code>
-                            </td>
-                            <td> The direction of the text; left-to-right by default. </td>
-                            <td> "ltr" | "rtl" </td>
-                            <td> No </td>
-                        </tr>
-                        <tr>
-                            <td>
                                 <code> <dfn>tags</dfn> </code>
                             </td>
                             <td> Free text categorizing the annotation. </td>
@@ -896,8 +880,6 @@
                 </table>
 
                 <p class="note">Read “Best practices for Reading Systems” about using tags in an annotation. </p>
-                <p class="issue" data-number="2854"></p>
-
                 <aside class="example" title="An annotation Body">
                     <pre>
    {
@@ -990,7 +972,7 @@
                             <code><dfn>title</dfn></code>
                         </td>
                         <td> A title to help identifying the set. </td>
-                        <td> string </td>
+                        <td> natural language string </td>
                         <td> No </td>
                     </tr>
                     <tr>
@@ -1042,7 +1024,7 @@
                                 <code> name </code>
                             </td>
                             <td> The name of the generator software. </td>
-                            <td> string </td>
+                            <td> natural language string </td>
                             <td> Yes </td>
                         </tr>
                         <tr>
@@ -1083,7 +1065,7 @@
                                 <code> dc:title </code>
                             </td>
                             <td> The title of the publication. </td>
-                            <td> string </td>
+                            <td> natural language string </td>
                             <td> No </td>
                         </tr>
                         <tr>
@@ -1099,7 +1081,7 @@
                                 <code> dc:publisher </code>
                             </td>
                             <td> The name of the publisher. </td>
-                            <td> string </td>
+                            <td> natural language string </td>
                             <td> No </td>
                         </tr>
                         <tr>
@@ -1107,7 +1089,7 @@
                                 <code> dc:creator </code>
                             </td>
                             <td> The author(s) of the publication. </td>
-                            <td> array of strings </td>
+                            <td> array of natural language strings </td>
                             <td> No </td>
                         </tr>
                         <tr>
@@ -1298,6 +1280,140 @@
             <h1> JSON Schema </h1>
 
             <p>T.B.D.</p>
+        </section>
+
+        <section>
+            <h1>Internationalization Considerations</h1>
+
+            <section class="informative">
+                <h2>Natural language strings</h2>
+
+                <p>
+                    The EPUB Annotation model inherits the <a data-cite="json-ld11#string-internationalization">string internationalization</a> features
+                    defined by [[[json-ld11]]]. The <a data-cite="json-ld11#dfn-default-language">default language</a> and
+                    <a data-cite="json-ld11#dfn-base-direction">base direction</a>
+                    of a natural language string can be set with the `@language` and `@direction` keywords, respectively.
+                    The value for `@language` is a string representing a [[BCP47]] language tag.
+                    The value for `@direction` is a string whose value must be `"ltr"`, `"rtl"`.
+                    Using these keywords, the language and the direction of natural language strings can be set either globally,
+                    e.g., for all annotations in an [=AnnotationSet=], or for a specific like, for example a [=Body object=].
+                </p>
+
+                <p>
+                    For a global setting, the language and direction may be set by adding an extra, dedicated context object to
+                    `@context` (using the JSON array notation) setting the necessary values.
+                </p>
+
+                <aside class="example" title="Setting language and direction globally">
+                    <pre>
+                    {
+                        "@context": [
+                            "https://www.w3.org/ns/epub-anno.jsonld",
+                            {
+                                "@language": "en",
+                                "@direction": "ltr"
+                            }
+                        ],
+                        "type": "AnnotationSet",
+                        "id": "urn:uuid:123-456-789",
+                        "items" : [
+                            {
+                                "type": "Annotation",
+                                "id": "urn:uuid:123-456-789-a1",
+                                "body": {
+                                    "type": "TextualBody",
+                                    "value": "This is a comment in English"
+                                }
+                            },
+                            …
+                        ]
+                    }
+                    </pre>
+                </aside>
+
+                <p>
+                    For setting the values locally, the same `@context` notation can be used within any
+                    object. This is referred to as the <a data-cite="json-ld11#scoped-contexts">"scoped" context</a>,
+                    i.e., a context that is only valid within the enclosing object and its descendent objects.
+                    Possible global values are overwritten.
+                </p>
+
+                <aside class="example" title="Setting language and direction locally">
+                    <pre>
+                        {
+                            "@context": [
+                                "https://www.w3.org/ns/epub-anno.jsonld",
+                                {
+                                    "@language": "en",
+                                    "@direction": "ltr"
+                                }
+                            ],
+                            "type": "AnnotationSet",
+                            "id": "urn:uuid:123-456-789",
+                            "items" : [
+                                …
+                                {
+                                    "id": "urn:uuid:123-456-789-a2",
+                                    "type": "Annotation",
+                                    "body": {
+                                        "@context": {
+                                            "@language": "he",
+                                            "@direction": "rtl"
+                                        },
+                                        "type": "TextualBody",
+                                        "value": "התגובה הזו בעברית"
+                                    }
+                                },
+                                …
+                            ]
+                        }
+                        </pre>
+                 </aside>
+                <!-- <section>
+                    <h3>Setting for a specific string</h3>
+                    <p>
+                        This is the finest granularity setting; instead of relying on the global or the scoped context, the
+                        language and direction are set on a specific string. This relies on an an additional, property called `stringValue`,
+                        and is used as follows:
+                    </p>
+
+                    <aside class="example" title="Setting language and direction locally">
+                        <pre>
+                            …
+                                "items": [
+                                    {
+                                        …
+                                    },
+                                    {
+                                        "id": "urn:uuid:123-456-789-a3",
+                                        "type": "Annotation",
+                                        "body": {
+                                            "type": "TextualBody",
+                                            "value": {
+                                                "@language"   : "he",
+                                                "@direction"  : "rtl",
+                                                "stringValue" : "התגובה הזו בעברית"
+                                            }
+                                        }
+                                    },
+                                    …
+                                ]
+                            …
+                        </pre>
+                    </aside>
+
+                    <p>
+                        `stringValue` is actually an alias set to the JSON-LD `@value` keyword by the core EPUB Annotation `@context`
+                        file. It has been introduced to avoid a possible clash with the `value` property, inherited from the [[annotation-model]] specification.
+                    </p>
+                    <div class="issue">
+                        <p>
+                            It is not clear whether the string level language setting is of any use for annotations. It
+                            may be simpler ignore this possibility and require the usage of scoped contexts everywhere.
+                        </p>
+                    </div>
+                </section> -->
+            </section>
         </section>
 
         <section>
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -1295,7 +1295,7 @@
                     <a data-cite="json-ld11#dfn-base-direction">base direction</a>
                     of a natural language string can be set with the `@language` and `@direction` keywords, respectively.
                     The value for `@language` is a string representing a [[BCP47]] language tag.
-                    The value for `@direction` is a string whose value must be `"ltr"`, `"rtl"`.
+                    The value for `@direction` is a string whose value must be either `"ltr"` or `"rtl"`.
                     Using these keywords, the language and the direction of natural language strings can be set either globally,
                     e.g., for all annotations in an [=AnnotationSet=], or for a specific object like a [=Body object=].
                 </p>
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -305,7 +305,7 @@
                                 <code> name </code>
                             </td>
                             <td> The name of the creator. </td>
-                            <td> natural language string </td>
+                            <td> Localizable text </td>
                             <td> No </td>
                         </tr>
                     </tbody>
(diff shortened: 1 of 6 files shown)


epub34/annotations/index.html
```diff
@@ -1300,6 +1282,154 @@ <h1> JSON Schema </h1>
...
+                <h2>Natural language strings</h2>
+
+                <p>
+                    The EPUB Annotation model inherits the <a data-cite="json-ld11#string-internationalization">string
+                    internationalization</a> features defined by [[[json-ld11]]].
+                    The <a data-cite="json-ld11#dfn-default-language">default language</a> and
+                    <a data-cite="json-ld11#dfn-base-direction">base direction</a>
+                    of a natural language string can be set with the `@language` and `@direction` keywords, respectively.
+                    The value for `@language` is a string representing a [[BCP47]] language tag.
+                    The value for `@direction` is a string whose value must be `"ltr"`, `"rtl"`.
```
bduga:
```suggestion
                    The value for `@direction` is a string whose value must be either `"ltr"` or `"rtl"`.
```

At least, I assume that is true and it can't be both.

bduga (commented):
I am not enough of an expert to understand all these json-ld changes, some of which look unrelated to an untrained eye. Let me know if it is important enough for me to invest time in understanding the changes and their implications. I just have the one comment on the actual text.

epub34/annotations/index.html
iherman: Thanks. You're right.

iherman:
> I am not enough of an expert to understand all these json-ld changes, some of which look unrelated to an untrained eye. Let me know if it is important enough for me to invest time in understanding the changes and their implications. I just have the one comment on the actual text.

Most of our readers will ***not*** be json-ld experts, and they should not become one to understand the spec. Ie, I believe, your comments will be welcome!

iherman:
One issue I did not address in the text: do we expect or require that the AnnotationSet defines a default language (and direction) when exporting the data and, if yes, what is it? Should it reuse the `xml:lang` attribute in the package document? Or do we expect to set the language values for each annotation explicitly?

iherman:
cc @pchampin ; not as a formal reviewer but as a json-ld expert...

bduga (approved):
This seems fine to me, though I continue to have roughly zero knowledge of json-ld.

iherman:
This was discussed during the [pmwg meeting on 19 March 2026](https://w3c.github.io/pm-wg/minutes/2026-03-19.html#89bd).

<h3>Use JSON-LD keywords for language and direction - <a href="https://github.com/w3c/epub-specs/pull/2950" rel="noopener noreferrer">w3c/<wbr>epub-specs#2950</a></h3>
<p><cite>&lt;ivan&gt;</cite> <a href="https://github.com/w3c/epub-specs/issues/2854" rel="noopener noreferrer">w3c/<wbr>epub-specs#2854</a></p>
<p><cite>wendyreid:</cite> ivan, can you explain this topic?</p>
<p><cite>ivan:</cite> we have to talk about language and language direction in strings in annotations</p>
<p><cite>&lt;LaurentLM&gt;</cite> link to the body language and textDirection: <a href="https://w3c.github.io/epub-specs/epub34/annotations/#body" rel="noopener noreferrer">https://<wbr>w3c.github.io/<wbr>epub-specs/<wbr>epub34/<wbr>annotations/#body</a></p>
<p><cite>ivan:</cite> the current draft has a field for text direction that has two problems.<br>
<span>… it is only for the body and doesn't solve other text, like titles and descriptions</span><br>
<span>… json LD now has a built-in feature to set language and direction</span><br>
<span>… my original issue is that we should use what we get in jsonLD</span><br>
<span>… and not duplicate things. It would solve the missing setting for other natural language text fields</span><br>
<span>… I've defined the solution and written a PR that we use JSON LD</span><br>
<span>… I described part of the JSON LD that are usable in my view in annotaitons</span><br>
<span>… i dont want to go into detail in the syntax</span><br>
<span>… if we set another block we can use JSON LD to set the language and direction</span><br>
<span>… it would then apply to everything in the tree underneath</span><br>
<span>… you can still have local overrides</span><br>
<span>… two open questions that need a decision</span><br>
<span>… there is a way in the JSON LD syntax to set the language and direction in a single text field</span><br>
<span>… which uses a different syntax and can be confusing</span><br>
<span>… we may not have to specify values for single strings</span><br>
<span>… I don't think we need to add this to the spec</span><br>
<span>… I would like some feedback on this</span><br>
<span>… is it a must to add language and direction?</span><br>
<span>… or is it a should? We should do the same for the annotations spec</span><br>
<span>… this has a slight effect on the vocabulary so we should make a decision</span></p>
<p><cite>mgarrish:</cite> we don't have a requirement in the package document, but it is required for accessibility<br>
<span>… setting the language of the publication is required, but not setting the language of the HTML</span></p>
<p><cite>ivan:</cite> in the pacakge doc, if we say it is a must to specify the DC language, the we should do the same for the annotations spec</p>
<p><cite>LaurentLM:</cite> web annotations are based on JSON LD<br>
<span>… programmers work with JSON not JSON LD</span><br>
<span>… to go too far with JSON LD is dangerous for developers</span><br>
<span>… using language direction is a small improvement for annotaitons</span><br>
<span>… I would like to see that use case for adding language and language direction before we implement this</span><br>
<span>… we had this same discussion with web publications. The outcome was to</span><br>
<span>… web publication spec can give us some continuity</span></p>
<p><cite>&lt;LaurentLM&gt;</cite> Readium Web Publications: <a href="https://readium.org/webpub-manifest/contexts/default/#title" rel="noopener noreferrer">https://<wbr>readium.org/<wbr>webpub-manifest/<wbr>contexts/<wbr>default/#title</a></p>
<p><cite>LaurentLM:</cite> readium web publications have chosen another path<br>
<span>… if you want multiple languages you use the language code, language is mapping to JSON LD</span></p>
<p><cite>&lt;LaurentLM&gt;</cite> W3C WEb Publications: <a href="https://www.w3.org/TR/pub-manifest/#manifest-lang-dir-local" rel="noopener noreferrer">https://<wbr>www.w3.org/<wbr>TR/<wbr>pub-manifest/#manifest-lang-dir-local</a></p>
<p><cite>LaurentLM:</cite> this seems easier to master than the example in the web publication W3C spec</p>
<p><cite>ivan:</cite> I realized while doing this, that what we have in web publications is buggy<br>
<span>… you cannot use the alias within the context</span></p>
<p><cite>LaurentLM:</cite> having the abilty specify language locally is important<br>
<span>… to have it at the top, and setting the direction, at the top makes sense</span></p>
<p><cite>ivan:</cite> it is clear that we cannot use language and direction in the body that the way it is now<br>
<span>… it is not now the JSON LD specification, you cannot specify the language and have it be recognized</span><br>
<span>… the mapping you have in Readium is valid JSON LD and can be used</span><br>
<span>… but since it doesn't handle direction we might have problems with internationalization</span><br>
<span>… we rarely have a case where we want to express the same thing with 3 different languages</span></p>
<p><cite>sueneu:</cite> Use cases are rare where you want to say things in different languages<br>
<span>… but I do this all the time</span></p>
<p><cite>ivan:</cite> In general that is true, but the question is does this happen in annotations</p>
<p><cite>sueneu:</cite> Absolutely. This could be very helpful<br>
<span>… you could have English annotations</span></p>
<p><cite>ivan:</cite> that is not really what we are talking about. We need to specify the language you made the annotation in</p>
<p><code>@laurentLM</code> it looks like this needs more discussion</p>
<p><cite><code>@gman</code>:</cite> I think there is a use case here, which could be annotation of a translation and discussion that could take place in two languages</p>
<p><cite>wendyreid:</cite> this could be more about the accessiblity issue where if the language isn't marked up properly screen readers can lead to some interesting publications<br>
<span>… in theory, I could add a word in a second language within an annotation</span></p>
<p><cite>ivan:</cite> which leads us to the discussion of HTML in annotations<br>
<span>… if we can use markdown in the body, like the markdown in github</span><br>
<span>… that does allow some HTML, so you could manually set a span with the correct language tag, it would cover that issue</span><br>
<span>… if we really what to mix the language inside the body, we could use multiple bodies</span></p>
<p><cite>duga:</cite> Github technically doesn't support HTML but doesn't strip them out<br>
<span>… so it displays correctly in a browser</span><br>
<span>… I don't know if there is a version of Markdown that allows Ruby</span><br>
<span>… I'm not even sure markdown will make it through rec</span></p>
<p><cite><code>@ivan</code>:</cite> for those issues, we will have face an international review,<br>
<span>… if we cannot set the language and direction in the right place, we will not get the sign off</span></p>
<p><cite>wendyreid:</cite> Let's leave it at that, we have more to do on the language question<br>
<span>… otherwise, thank you everyone for a great meeting. See you all next week</span></p>

<hr /></details>

iherman:
Just for reference:

This issue came up during the last call: one cannot use an alias to a keyword within a context (hence the necessity to use `@language` as opposed to `language`). Here is a [reference to the json-ld spec](https://www.w3.org/TR/json-ld11/#aliasing-keywords:~:text=Aliased%20keywords%20may%20not%20be%20used%20within%20a%20context%2C%20itself).

(Yes, the whole section is informative, the normative part is deep in the JSON-LD API spec.)

pchampin (approved):

mattgarrish:
> one cannot use an alias to a keyword within a context (hence the necessity to use `@language` as opposed to `language`)

That doesn't mean you can't have an alias for `@language` in the context, does it? Isn't it only requiring that you not subsequently use the alias "`language`" for any other definitions you want to define a language for?

Otherwise, aren't you saying that the [publication manifest](https://www.w3.org/ns/pub-context) context is wrong? It uses "`@language`" everywhere in the context but then aliases it to "`language`" at the end. Our examples in the specification then use "`language`" everywhere.

I followed the link you gave and ended up at the `@language` keyword definition, which says:

> The `@language` keyword MAY be aliased and MAY be used as a key in a [value object](https://www.w3.org/TR/json-ld11/#dfn-value-object).

(Or am I not understanding the issue here?)

iherman:
> > one cannot use an alias to a keyword within a context (hence the necessity to use `@language` as opposed to `language`)
> 
> That doesn't mean you can't have an alias for `@language` in the context, does it? Isn't it only requiring that you not subsequently use the alias "`language`" for any other definitions you want to define a language for?

Calling in @bigbluehat and @pchampin to make sure I do not misunderstand the rules. I believe that the aliases for keywords can only be used _outside_ of _any_ context. Only in the real data, so to say. 

When I set the language for an annotation within an annotation set, or a body within a annotation, I use a "scoped context", ie, a context local to that object. That is where `@language` must be used.

I have made some tests using the [json-ld playground](https://json-ld.org/playground/next/) or the [RDF playground](https://champin.net/2023/sowasm/) and they both behave the way I understand this. (Actually, it is by making some tests there that I realized the issue.)

I agree that this rule is very much between the lines, let alone the fact that following the links to the JSON-LD API is the only way to find a normative reference to this. (@bigbluehat, @pchampin, I really believe this fact should be made much more explicit and precise in the spec.)

> 
> Otherwise, aren't you saying that the [publication manifest](https://www.w3.org/ns/pub-context) context is wrong? It uses "`@language`" everywhere in the context but then aliases it to "`language`" at the end. Our examples in the specification then use "`language`" everywhere.

I think it is all right, but it chose a different path, namely that all natural language value is tagged explicitly for a different language or direction than the default values set in the overall context, see [example 13](https://www.w3.org/TR/pub-manifest/#example-13-a-publication-in-arabic-with-the-title-also-given-in-english). The alias is used in a value object, which is all right.

This is the "third" option that I put forward in https://github.com/w3c/epub-specs/issues/2854#issuecomment-3935262418, and we both agreed it is awkward. The approach in this PR proposes to use scoped contexts, ie, a context within an annotation or within an annotation body; it in those scoped contexts that the usage of `language` is not possible.

Yes, we can move completely to the Publ. manifest way, but it becomes more awkward: if all natural language values within a single annotation must be, say, in Japanese within a set that is predominantly English, then all the individual values must be written down as a separate object with language setting. This is what is simplified via a scoped context.

I believe that using _both_ scoped contexts _and_ individual language settings would be even more confusing... Ie, it should be either or, not both. Ie, if we go towards the Publ. manifest way, then language setting may be set once in the top context (in the annotation set) using `@language`, and then any variation of the language must use a value object. 

I am willing to create a separate PR to explore that, and we can then choose among them.

(Scoped contexts did not really exist in the publ. manifest days...)

> 
> I followed the link you gave and ended up at the `@language` keyword definition, which says:
> 
> > The `@language` keyword MAY be aliased and MAY be used as a key in a [value object](https://www.w3.org/TR/json-ld11/#dfn-value-object).

Yes. Value objects can use `language` and, actually, they are the only ones that can use them.

mattgarrish:
> I believe that the aliases for keywords can only be used _outside_ of _any_ context. Only in the real data, so to say.

Right, I thought you were saying that you couldn't use "language" in the data because you couldn't alias "`@language`" in the context, which seemed to contradict what we did previously. But it sounds like we actually agree. 😄

iherman:
> > I believe that the aliases for keywords can only be used _outside_ of _any_ context. Only in the real data, so to say.
> 
> Right, I thought you were saying that you couldn't use "language" in the data because you couldn't alias "`@language`" in the context, which seemed to contradict what we did previously. But it sounds like we actually agree. 😄

My apologies if I was not clear. The point is: if we want to set the language for a full annotation or the full annotation set, we cannot avoid using `@language`. Ie, the source of confusion remains.

epub34/annotations/index.html
```diff
@@ -868,22 +868,6 @@ <h2>Body</h2>
                             <td> "solid" | "underline" | "strikethrough" | "outline" </td>
                             <td> No </td>
                         </tr>
-                        <tr>
```
sueneu, line 871: Should we keep the language row here if we continue to use the language sub property in our examples? (see line 893).

epub34/annotations/index.html
```diff
@@ -868,22 +868,6 @@ <h2>Body</h2>
...
-                            <td>
-                                <code> textDirection </code>
-                            </td>
-                            <td> The direction of the text; left-to-right by default. </td>
-                            <td> "ltr" | "rtl" </td>
-                            <td> No </td>
-                        </tr>
                         <tr>
                             <td>
                                 <code> <dfn>tags</dfn> </code>
```
sueneu, line 873: Some of the data in the names column are tagged <dfn> and some are not. I'm just confirming that this is intentional?

epub34/annotations/index.html
```diff
@@ -1300,6 +1282,154 @@ <h1> JSON Schema </h1>
...
+                    <p>
+                        This is the finest granularity setting; instead of relying on the global or the scoped context, the
+                        language and direction are set on a specific string. This relies on an an additional, property called `stringValue`,
+                        and is used as follows:
+                    </p>
+
+                    <aside class="example" title="Setting language and direction locally">
+                        <pre>
+                            …
+                                "items": [
```
sueneu, line 1395: > "items": [
what is "items" here? Is it the whole hebrew text string? Or an id for the hebrew text? Is there another term more specific than "items" Like "text-string-id" that might help people new to JSON granularity markup?

epub34/annotations/index.html
iherman: Oops. Just about the opposite: the example should be modified, ie, those entries removed.

epub34/annotations/index.html
iherman: Yes. `<defn>` are for terms that we define in this specification. The others (like `value`) are inherited from the Web Annotation specification, and are just listed for completeness.

(Maybe at some point we will have to add some terminology section at the beginning of the spec to make these nuances clear, @llemeurfr ?)

epub34/annotations/index.html
iherman: `items` is defined for Annotation sets: see [definition](https://raw.githack.com/w3c/epub-specs/anno-use-json-ld-for-language-and-direction/epub34/annotations/index.html#dfn-items). It is a list of annotations; syntactically, that is what the `[` means in JSON.

iherman:
Closing without merge, in favor of #2967

PR #2953, opened 2026-03-18
[epub34-rs] fixed referenced spec ID

himorin:
link target of property is CSS2 spec, but specref points csssnapshot which is non rec-track and cannot appear in normative part.


mattgarrish:
It's not a good way to reference specific properties, but I wouldn't go quite so far as to say it can't appear in normative sections. If that were the case, it would blow up how we cite css support for reading systems (it's technically a bit of indirection that we're really normatively citing specs that the snapshot identifies).

But as a property reference it's kind of a lazy handwave that the property exists somewhere in some CSS spec, especially when the linking on the property itself is specific to a document. There are a few other cases of the snapshot being used as a reference for css properties that should be fixed, too, before merging.

mattgarrish:
I just realized this is a forked branch, so I'll merge this and fix the other instances in a separate PR.

PR #2954, opened 2026-03-18
Simplify the use of the refines attribute with accessibility metadata

### LINKED ISSUE #2846: Simplify evaluator metadata
### Section

https://www.w3.org/TR/epub-a11y-12/#sec-conf-reporting-eval

### Describe the problem

Considering the difficulty in extracting evaluator info for the metadata guide given the refines chains, does it still make sense to have so many refines chains going on?

For example, does it really make more sense that the evaluation date is attached to the evaluator name instead of directly to the conformance statement? Yes, the evaluator performed it on the specified date, but it's just as simple to say conformance was determined on the date. The same holds for the more detailed report.

And what do you do if you don't want to specify the evaluator, only the date. Then it has to bump up a level.

The only property that has to be attached to the evaluator is their credential.

Drop the extra refines chain involving the evaluator and have these properties linked to the conformance statement.

mattgarrish:
This changes the guidance on refines to emphasize it is only needed if there are multiple conformance claims.

Only the certifier credential would still prefer refines since it's specific to the evaluator. I don't see much gain and lots of room for misinterpretation if we drop that case.

I haven't added a change log entry since this guidance is non-normative and actually re-aligns us with the 1.0 spec.

Fixes #2846 

***
[Preview](https://raw.githack.com/w3c/epub-specs/fix/a11y-refines/epub34/a11y/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fa11y%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Ffix%2Fa11y-refines%2Fepub34%2Fa11y%2Findex.html)

### DIFF
--- a/epub34/a11y/index.html
+++ b/epub34/a11y/index.html
@@ -1230,6 +1230,107 @@
 				<section id="sec-conf-reporting-eval">
 					<h4>Evaluator information</h4>
 
+					<section id="sec-conf-num-express" class="informative">
+						<h5>Differences in expression based on the number of claims</h5>
+
+						<p>How to add the evaluator information to an EPUB publication depends on the number of
+							conformance claims that are made.</p>
+
+						<p>Most publications will only contain a single <code>dcterms:conformsTo</code> [[dcterms]]
+							property with the <a href="#sec-conf-reporting-pub">accessibility claim</a> for the content.
+							In this case, it is not necessary to use EPUB 3's <a data-cite="epub/#subexpression"
+								>metadata association mechanism</a> [[epub-3]] to connect the various metadata
+							statements.</p>
+
+						<p>Consider the following example metadata:</p>
+
+						<pre>&lt;metadata …>
+   …
+   &lt;meta
+       property="dcterms:conformsTo"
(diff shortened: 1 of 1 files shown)


clapierre (approved):

mattgarrish:
I've made some more changes to this pull request based on the discussions this morning. I've added a new informative section to explain how the number of conformance claims affects the need for `refines` and show a couple of complete examples. I also added an non-refines example of a credential.

It's all still informative changes, but I've added a change log entry just because it's still a significant clarification that both methods are accepted.

clapierre (approved):

gregoriopellegrino (approved):

PR #2957, opened 2026-03-18
fixed CSS reference targets

himorin:
both properties links to CSS2 spec, but spec reference points to csssnapshot but not css2.


mattgarrish:
Thanks, but closing without merging as this duplicates some of the changes already in #2956.

PR #2967, opened 2026-03-23
alternate version for use json-ld keywords for language and direction

iherman:
Based on the discussion in #2950, this is an alternate version for the same issue. It emphasizes the setting of individual strings, just as it is done in the publication manifest. However a global setting is also possible for annotation sets.

As in #2950, the section is informative. We do ***not*** formally specify anything; we reuse a mechanism specified elsewhere, and use the keyword aliasing possibilities of JSON-LD.

See also https://github.com/w3c/epub-specs/pull/2950#issuecomment-4102225210

---

See:

* For EPUB Annotations 1.0:
    * [Preview](https://raw.githack.com/w3c/epub-specs/anno-alternate-use-json-ld-for-language-and-direction/epub34/annotations/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fannotations%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Fanno-alternate-use-json-ld-for-language-and-direction%2Fepub34%2Fannotations%2Findex.html)
* For EPUB Annotations Vocabulary 1.0:
    * [Preview](https://raw.githack.com/w3c/epub-specs/anno-alternate-use-json-ld-for-language-and-direction/epub34/annotations-vocab/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fannotations-vocab%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Fanno-alternate-use-json-ld-for-language-and-direction%2Fepub34%2Fannotations-vocab%2Findex.html)

### DIFF
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -305,7 +305,7 @@
                                 <code> name </code>
                             </td>
                             <td> The name of the creator. </td>
-                            <td> string </td>
+                            <td> natural language string </td>
                             <td> No </td>
                         </tr>
                     </tbody>
@@ -840,7 +840,7 @@
                                 <code> value </code>
                             </td>
                             <td> The textual content of the annotation. </td>
-                            <td> string </td>
+                            <td> natural language string </td>
                             <td> Yes </td>
                         </tr>
                         <tr>
@@ -870,22 +870,6 @@
                         </tr>
                         <tr>
                             <td>
-                                <code> language </code>
-                            </td>
-                            <td> The language of the annotation. </td>
-                            <td> [[BCP47]] </td>
-                            <td> No </td>
-                        </tr>
-                        <tr>
-                            <td>
-                                <code> textDirection </code>
-                            </td>
-                            <td> The direction of the text; left-to-right by default. </td>
-                            <td> "ltr" | "rtl" </td>
-                            <td> No </td>
-                        </tr>
-                        <tr>
-                            <td>
                                 <code> <dfn>tags</dfn> </code>
                             </td>
                             <td> Free text categorizing the annotation. </td>
@@ -896,8 +880,6 @@
                 </table>
 
                 <p class="note">Read “Best practices for Reading Systems” about using tags in an annotation. </p>
-                <p class="issue" data-number="2854"></p>
-
                 <aside class="example" title="An annotation Body">
                     <pre>
    {
@@ -990,7 +972,7 @@
                             <code><dfn>title</dfn></code>
                         </td>
                         <td> A title to help identifying the set. </td>
-                        <td> string </td>
+                        <td> natural language string </td>
                         <td> No </td>
                     </tr>
                     <tr>
@@ -1042,7 +1024,7 @@
                                 <code> name </code>
                             </td>
                             <td> The name of the generator software. </td>
-                            <td> string </td>
+                            <td> natural language string </td>
                             <td> Yes </td>
                         </tr>
                         <tr>
@@ -1083,7 +1065,7 @@
                                 <code> dc:title </code>
                             </td>
                             <td> The title of the publication. </td>
-                            <td> string </td>
+                            <td> natural language string </td>
                             <td> No </td>
                         </tr>
                         <tr>
@@ -1099,7 +1081,7 @@
                                 <code> dc:publisher </code>
                             </td>
                             <td> The name of the publisher. </td>
-                            <td> string </td>
+                            <td> natural language string </td>
                             <td> No </td>
                         </tr>
                         <tr>
@@ -1107,7 +1089,7 @@
                                 <code> dc:creator </code>
                             </td>
                             <td> The author(s) of the publication. </td>
-                            <td> array of strings </td>
+                            <td> array of natural language strings </td>
                             <td> No </td>
                         </tr>
                         <tr>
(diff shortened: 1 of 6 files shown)


llemeurfr:
I don't think that an AnnotationSet should get such information. An AnnotationSet is an envelope; a transient bag of annotations, used for the purpose of moving annotations, possibly from different authors, from one system to another. 
There is no hierarchy between the AnnotationSet level and the Annotation level. No semantic can pass from an "upper" level to an "inner" level.

epub34/annotations/index.html
```diff
@@ -1300,6 +1280,126 @@ <h1> JSON Schema </h1>
...
+                            "language": "en"
+                        },
+                        "items" : [
+                            {
+                                "id": "urn:uuid:123-456-789-a1",
+                                "type": "Annotation",
+                                "value": {
+                                    "language": "en",
+                                    "stringValue": "HTML is a markup language.",
+                                    "language": "en"
```
llemeurfr: language is already defined in the object

epub34/annotations/index.html
```diff
@@ -1300,6 +1280,126 @@ <h1> JSON Schema </h1>
...
+                        "id": "urn:uuid:123-456-789",
+                        "title" : {
+                            "stringValue": "This is an example for an annotation text",
+                            "language": "en"
+                        },
+                        "items" : [
+                            {
+                                "id": "urn:uuid:123-456-789-a1",
+                                "type": "Annotation",
+                                "value": {
```
llemeurfr: The `body` object is missing. It contains `type` and `value`.

epub34/annotations/index.html
```diff
@@ -1300,6 +1280,126 @@ <h1> JSON Schema </h1>
...
+            <section class="informative">
+                <h2>Localizable texts</h2>
+
+                <p>
+                    The EPUB Annotation model inherits the <a data-cite="json-ld11#string-internationalization">string
+                    internationalization</a> features defined by [[[json-ld11]]].
+                    These may be used for values that are defined as
+                    <a data-cite="i18n-glossary#dfn-localizable-text">localizable texts</a>.
+                    To set the right language metadata, the value is a separate
+                    object with the string value set explicitly through the `stringValue` property.
```
llemeurfr, line 1295: I would vote for `text` instead. Just bikeshedding.

llemeurfr (changes requested):

epub34/annotations/index.html
iherman: I took the first term that came to my mind. If we prefer `text`, that is also fine with me.

epub34/annotations/index.html
iherman: Oops, sorry. Added

epub34/annotations/index.html
iherman: ~~No it is not. The definition of language in the title is only valid in its containing object, used by the `title` property.~~ oops sorry, I misread the comment. Yes, taking care of it.

iherman:
> I don't think that an AnnotationSet should get such information. An AnnotationSet is an envelope; a transient bag of annotations, used for the purpose of moving annotations, possibly from different authors, from one system to another. There is no hierarchy between the AnnotationSet level and the Annotation level. No semantic can pass from an "upper" level to an "inner" level.

Well, first of all, it is more than "just" a bag. It has its own metadata on the enclosing publication and the generator, which is a common characteristics for each annotation. The default language is not much different.

Technically, It is a bona fide JSON-LD "node". From that point of view, semantics passes "down" to the contained objects per definition. For example, if the annotation set has the proper `@context` setting, it is unnecessary to repeat the same `@context` setting to the enclosed annotation objects because it just "inherits" that context. Repeating it just unnecessarily complicates implementers' life. 

The same is valid for language setting. Of course, technically it is possible to repeat the language setting either explicitly as in the first example, or via an scoped context, ie, adding the context in the annotation. Then repeating the same pattern for each and every annotation. But this is really not a DRY approach.

llemeurfr:
After some intra-EDRLab/Readium discussions, let's go for this version, which is more "JSON-friendly". 
We prefer "text" rather than "stringValue". 

```json
"body": {
          "type": "TextualBody",
           "value": {
               "language": "he",
                "direction": "rtl",
                "text": "היא שפת סימון."
             }
         },
```

The difference between this local notation and the global, JSON-LD-friendly notation in the AnnotationSet is not considered a real issue. 

Note also that allowing the localization of the title of an AnnotationSet, and the title, authors, and publisher of the associated book, is clean in theory, but will be unused in practice, because real EPUBs do not express `xml:lang` and `dir` on `dc:title`, `dc:creator`and `dc:publisher` (I'm ok to be proved wrong).

We also think the generator property should remain a string. 

see also: https://github.com/edrlab/thorium-reader/issues/2926#issuecomment-4182933191

sueneu:
Minor point— if there is a period to indicate end of sentence it likely should be on the left of the text string. 
>"היא שפת סימון."

AbeJellinek:
The actual character order in `"text": "היא שפת סימון."` is correct; the bidi algorithm causes the period to show on the wrong end.

epub34/annotations/index.html
bduga: `text` works for me. `stringValue` seems redundant, but I could also live with just `string`.

bduga:
Although, this text is missing part of the string - it should start with
HTML, which is at least part of why direction is needed and can't be
computed.

sueneu:
Can we assume if BiDi is correctly implemented in a reading system that the text will display correctly? 

In other words, is this a problem we would need to solve in the annotation's JSON? Or something that a reading system will know how to implement from the language declaration?

bduga:
That is the point of the language and direction attributes. Looking at the
string, it's impossible to tell if it is an English string with Hebrew
embedded in it, or a Hebrew string with English embedded in it. Your email
reader assumes the former, as does GitHub. A reading system can use those
attributes to know that the specific string is meant to be the latter. So
no change needed for the JSON, and yes, reading systems should know what to
do (or at least they can if they choose to)..

iherman:
> After some intra-EDRLab/Readium discussions, let's go for this version, which is more "JSON-friendly". We prefer "text" rather than "stringValue".
> 
> ```json
> "body": {
>           "type": "TextualBody",
>            "value": {
>                "language": "he",
>                 "direction": "rtl",
>                 "text": "היא שפת סימון."
>              }
>          },
> ```
>

Changed the alias to `text` in the latest json ld context file, and changed the examples.
 
> The difference between this local notation and the global, JSON-LD-friendly notation in the AnnotationSet is not considered a real issue.
> 
> Note also that allowing the localization of the title of an AnnotationSet, and the title, authors, and publisher of the associated book, is clean in theory, but will be unused in practice, because real EPUBs do not express `xml:lang` and `dir` on `dc:title`, `dc:creator`and `dc:publisher` (I'm ok to be proved wrong).
> 

That may be correct, but that is because they rely on the default [`xml:lang` setting](https://www.w3.org/TR/epub-34/#attrdef-xml-lang) on the package element. The default setting on the `AnnotationSet` level plays exactly the same functional role (syntax issue put aside).

> We also think the generator property should remain a string.
> 

Agreed. Made the change in the latest iteration.

PR #2969, opened 2026-03-24
Add note about linked its documents

### LINKED ISSUE #2964: Internationalization Tag Set (ITS) : Embedded or detached
### Section

https://www.w3.org/TR/epub-34/#sec-xhtml-its

### Describe the problem

> The [[its20](https://www.w3.org/TR/epub-34/#bib-its20)] specification defines a set of attributes that MAY be used in [XHTML content documents](https://www.w3.org/TR/epub-34/#dfn-xhtml-content-document) to add support for internationalization, translation, and localization.

I find that the use of "Internationalization Tag Set" (ITS) is now allowed in EPUB 3.4. This is a small addition but it might have significant impacts. I can imagine that AI will honor ITS instructions when it translates EPUB publications.

ITS has two usage scenarios: 1) embedded ITS within HTML, and 2) detached ITS that references HTML by XPath. EPUB 3.4 now allows 1) only, but I think that 2) is better since XHTML content documents can be kept clean.  Just like  CSS can be embedded within or detached from XHTML content documents, it should be possible to detach ITS from XHTML content documents.

In the context of EPUB 3.4, the second scenario (detached ITS) could naturally be realized as a combination of three mechanisms:

- ITS documents are listed in the EPUB package document;
- The relationship between ITS documents and XHTML content documents is established using the EPUB refines mechanism; and
- ITS selectors reference XHTML elements using XPath, as defined in ITS 2.0.

The intent is to define a clear packaging and processing model at the EPUB level, without altering the semantics of ITS itself.

Package-level ITS would enable consistent application of translation and internationalization rules across multiple content documents within a publication. It also aligns with publishing workflows in which localization metadata is managed separately from source content.

This design preserves compatibility with existing ITS processing models while providing a scalable structure for EPUB publications. It also opens the possibility of structured integration with AI-assisted translation pipelines, without requiring reading systems to interpret ITS.

mattgarrish:
Adds the text from https://github.com/w3c/epub-specs/issues/2964#issuecomment-4110329744

Fixes #2964

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2969.html" title="Last updated on Mar 24, 2026, 1:50 PM UTC (de49b6f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2969/e6d9db2...de49b6f.html" title="Last updated on Mar 24, 2026, 1:50 PM UTC (de49b6f)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -2502,7 +2502,7 @@
 									container, not parts of files.</p>
 
 								<p id="encryption-restrictions">The following files MUST NOT be encrypted:</p>
-								
+
 								<ul class="nomark">
 									<li>
 										<code>mimetype</code>
@@ -2527,7 +2527,7 @@
 									</li>
 									<li>[=package document=]</li>
 								</ul>
-								
+
 								<p>Encrypted data replaces unencrypted data in an OCF abstract container. For example,
 									if an image named <code>photo.jpeg</code> is encrypted, the contents of the
 										<code>photo.jpeg</code> resource is replaced with its encrypted contents.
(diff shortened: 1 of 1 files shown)


iherman (approved):

mattgarrish:
Merging this as there hasn't been any further feedback.

Issue #1660, opened 2021-05-02
Use of cautions and warnings


mattgarrish:
We used to have a caution class in IDPF for some messages about lack of support and future invalidity with respect to html and svg. See, for example, http://idpf.org/epub/31/spec/epub-contentdocs.html#sec-svg

It appears a couple of the cautions got turned into warning classes when we translated the documents, which, despite being undocumented, respec turns into red warning boxes: https://w3c.github.io/epub-specs/epub33/core/#sec-reserved-prefixes

We should harmonize all these instances, but the questions this raises for me are:

1) Is a "warning" an appropriate replacement for a "caution" or does it sound too severe? (Should we add a custom script to change the titles and backgrounds to cautions?)
2) Should we be relying on an undocumented feature? (Would be good to ask if it's legit but forgotten in the documentation, or likely to be dropped in the future.)
3) Should we use notes and avoid the issue entirely, or does that minimize the message too much?

mattgarrish:
I also discovered there's an "advisement" class.

Have asked about these classes in https://github.com/w3c/respec/issues/3503

iherman:
I must admit I did not know about this built-in behaviour either...

I believe that the current 'warning' is "over the top": it should be reserved to situations where a specific features may have, say, security implications. This is not our case. As, (seeing https://github.com/w3c/respec/issues/3503#issuecomment-831011421) this seems to be a feature built-in the core TR style sheets, I do not think we should change the outlook pragmatically (besides, we may, either now or in the future, hit a situation where the current warning style _is_ the appropriate one!).

I would be in favour of either use a note, or come up with our own style and class name for 'caution'.

mattgarrish:
> I do not think we should change the outlook pragmatically

Right, I wasn't even aware there were organization-specific styles. For all I knew, we just "use respec".

Sounds like we'd need to request a new caution style, too, so maybe we should just go with the existing "advisement" class? They're just notes we want to draw a little extra attention to, after all.

iherman:
> Sounds like we'd need to request a new caution style, too,

There is absolutely nothing wrong if we define our own class in our own css file. Whether respec will take over later, all the better, but we should not have a dependency on this...

mattgarrish:
> There is absolutely nothing wrong if we define our own class in our own css file.

Right, but is it okay to define something like a new caution box on our own?

[...]

PR #2970, opened 2026-03-24
Fix reference to charmod-norm

### LINKED ISSUE #2965: [epub3.4] referencing charmod-norm in normative part
### Section

4.2.3 File paths and file names

### Describe the problem

> All file names within the same directory MUST be unique following Unicode canonical normalization [[uax15](https://www.w3.org/TR/epub-34/#bib-uax15)] and then full case folding [[unicode](https://www.w3.org/TR/epub-34/#bib-unicode)]. (Refer to [Unicode Canonical Case Fold Normalization Step](https://www.w3.org/TR/charmod-norm/#CanonicalFoldNormalizationStep) [[charmod-norm](https://www.w3.org/TR/epub-34/#bib-charmod-norm)] for more information.)

Even sentence with charmod-norm is in parentheses and could be considered as some sort of additional note (or annotation?), this is normative part of spec where Note is not suitable to be refrerred. This sentence should be turned into Note or something.

(some other similar references exist, like datetime in [5.6.3.2.4 The dc:date element](https://www.w3.org/TR/epub-34/#sec-opf-dcdate))

Referencing "Unicode Canonical Case Fold Normalization Step" from charmod-norm is additional description/information (e.g. see also or for further reading contents) to MUST clause to apply canonical normalization and full case folding, I propose just moving sentence in parentheses into new Note block.

mattgarrish:
Changes the link to informative so that charmod-norm isn't listed as a normative reference.

Also clarifies that all statements in parentheses are informative.

Fixes #2965

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2970.html" title="Last updated on Mar 27, 2026, 7:17 PM UTC (d3a4311)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2970/e6d9db2...d3a4311.html" title="Last updated on Mar 27, 2026, 7:17 PM UTC (d3a4311)">Diff</a>

### DIFF
--- a/epub34/a11y/index.html
+++ b/epub34/a11y/index.html
@@ -151,7 +151,7 @@
 				<p>This specification, EPUB Accessibility, addresses two key needs in the EPUB ecosystem:</p>
 
 				<ul>
-					<li>discoverability of the accessible qualities of <a data-cite="epub/#dfn-epub-publication">EPUB
+					<li>discoverability of the accessible qualities of <a data-cite="epub-3#dfn-epub-publication">EPUB
 							publications</a>; and</li>
 					<li>evaluation and certification of accessible EPUB publications.</li>
 				</ul>
@@ -188,7 +188,7 @@
 				<h3>Success techniques</h3>
 
 				<p>This specification takes an abstract approach to the accessibility requirements for <a
-						data-cite="epub/#dfn-epub-publication">EPUB publications</a>, similar to how WCAG [[wcag2]]
+						data-cite="epub-3#dfn-epub-publication">EPUB publications</a>, similar to how WCAG [[wcag2]]
 					separates its accessibility guidelines from the techniques to achieve them. This approach allows the
 					guidelines to remain stable even as the format evolves.</p>
 
@@ -222,7 +222,7 @@
 			<section id="sec-application" class="informative">
 				<h3>Application to older versions of EPUB</h3>
 
-				<p>This specification is applicable to any <a data-cite="epub/#dfn-epub-publication">EPUB
+				<p>This specification is applicable to any <a data-cite="epub-3#dfn-epub-publication">EPUB
 						publication</a>, regardless of what version of EPUB it conforms to.</p>
 
 				<p>For example, even though EPUB 2 [[opf-201]] was produced before this specification, and so makes no
@@ -231,7 +231,7 @@
 					version of EPUB 3 to get access to the most advanced accessibility features and techniques.</p>
 
 				<p>Note that EPUB 2 does not support all metadata expressions defined in this specification as it does
-					not have an equivalent to the <a data-cite="epub/#attrdef-refines"><code>refines</code>
+					not have an equivalent to the <a data-cite="epub-3#attrdef-refines"><code>refines</code>
 						attribute</a> [[epub-3]]. If metadata expressions that require a <code>refines</code> attribute
 					cannot be avoided, there will be a certain amount of ambiguity in the statements (i.e.,
 					relationships between expression might only be apparent by their placement in the package document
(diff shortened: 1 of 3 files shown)


iherman (approved):

mattgarrish:
FYI, I changed "statements" to "explanatory statements" in the last commit just to try and add a bit of additional clarity on what this applies to - to avoid things like our status labels being considered statements.

mattgarrish:
I did a more thorough run through of our normative references and found a few oddities.

Because the caution boxes are our own creation, references inside them aren't treated as informative like with note boxes. This primarily affected the sha-1 references which I've made informative and added caution boxes to the list of things that are informative.

There were also a couple of references used in "such as" examples that were not marked informative.

I also found a few cases of introductions that hadn't been marked as informative.

And in the accessibility spec, we had two different ways of referring to the latest epub spec going - one using the "epub" reference to the latest version and another using the undated "epub-3" reference in the local bibliography (other than generating two references, it also split all the referenced terms in the index). I've standardized them all to the undated reference.

iherman:
I am just wondering... respec has (now? it is the first time I see it) a separate `.warning` class, which is handled as a note but with a different header and color. I presume it takes care of the normative/non-normative stuff as well. Do you think it would be worth changing our home made caution boxes to warning boxes? A warning has, roughly, the same semantic meaning, but it would save us some trouble by having one less extra js file and css to care about...

I duplicated a caution box in our spec and made screen dumps in light and dark modes; here is what they look like:

Screenshot 2026-03-28 at 09 06 17


Screenshot 2026-03-28 at 09 06 53

mattgarrish:
We had this discussion a number of years ago: https://github.com/w3c/epub-specs/issues/1660

I don't really care about the naming too much, although you did convince me back then that warning was a bit of overkill.

Caution and advisement seem to be equivalent creations but are incomplete which is why we have to script in the heading. <strike>(I thought we styled them, too, but appears not.)</strike> (Edit: I forgot to look at common.css. Looks like we copied caution from advisement.)

But the normative/informative problem seems to exist for all of the classes because they're just W3C styling options. Respec isn't aware that the boxes are informative.

iherman:
> But the normative/informative problem seems to exist for all of the classes because they're just W3C styling options. Respec isn't aware that the boxes are informative.

Really? I thought respec "knows" about these. In any case, they are part of the respec documentation: https://github.com/speced/respec/wiki/warning

But if you are right, it is not worth the trouble changing anything, we would have the problem of normative references.

(I have completely forgotten about this discussion in 2021. You have an amazing memory...)

mattgarrish:
> In any case, they are part of the respec documentation: https://github.com/speced/respec/wiki/warning

That page appears to have only been added 10hrs ago. It's not in the main doc yet: https://respec.org/docs/

I finally found .warning listed as a variant of the note class in the big doc, though. Odd that it's treated as normative text if it's a note variant, but maybe that's the intention since the conformance boilerplate doesn't list warnings as informative?

In any case, if we use that class it definitely treats references inside as normative so they still have to be overridden.

(From a visual perspective, I much prefer our cautions to the super-orangey warnings!)

iherman:
> 
> I finally found .warning listed as a variant of the note class in the big doc, though. Odd that it's treated as normative text if it's a note variant, but maybe that's the intention since the conformance boilerplate doesn't list warnings as informative?

Odd indeed.

> 
> In any case, if we use that class it definitely treats references inside as normative so they still have to be overridden.
> 

The it beats the purpose of changing.

> (From a visual perspective, I much prefer our cautions to the super-orangey warnings!)

Then (2nd time:-) let us keep to what we have...

mattgarrish:
Ya, there's also the concern that if we change the labelling people might think the importance level has changed. Better to leave things as they are.

I'm going to merge this before I find other changes to try and sneak into it... 😄

PR #2973, opened 2026-04-06
Add support for audiovisual notes

### LINKED ISSUE #2925: Annotations: the model may need a mechanism to include audiovisual comments.
At the moment, only textual annotations are described. Audovisual comments may come to the fore in the future.

llemeurfr:
Evolution in the specification (I didn't look if the json schema has to be modified) to handle audiovisual annotation bodies. 
The serialization section is heavily modified; the exported file is a zip which contains the serialized AnnotationSet and, if present, a set of audiovisual resources.   

[Preview](https://raw.githack.com/w3c/epub-specs/ann-audiovisual-notes/epub34/annotations/index.html)
[Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fannotations%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Fann-audiovisual-notes%2Fepub34%2Fannotations%2Findex.html)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/2973.html" title="Last updated on Apr 22, 2026, 12:12 PM UTC (5547e6d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/2973/094c6cd...5547e6d.html" title="Last updated on Apr 22, 2026, 12:12 PM UTC (5547e6d)">Diff</a>

### DIFF
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -11,6 +11,7 @@
         <script src="../../common/js/add-caution-hd.js" class="remove"></script>
         <script class="remove">
             var respecConfig = {
+                subtitle: "An interchange format for user notes, highlights, and bookmarks",
                 group: "pm",
                 wgPublicList: "public-pm-wg",
                 specStatus: "ED",
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -186,7 +186,7 @@
                                 <code> <dfn>motivation</dfn> </code>
                             </td>
                             <td>The motivation for the annotation's creation.</td>
-                            <td>"bookmarking" | "highlighting"</td>
+                            <td>"bookmarking" | "commenting" | "highlighting"</td>
                             <td>No</td>
                         </tr>
                         <tr>
@@ -241,10 +241,10 @@
 
                 <p class="note">The type of annotation should be considered when determining the value of the
                     [=motivation=] property.
-                    The motivation is "highlighting" if the Selector 
-                    defines a range of characters, a space in an image or a time period.
-                    It is "bookmarking" if the Selector references a single character in the resource.
-                    A textual or audiovisual body may be present in both cases.
+                    An annotation with a Body structure corresponds to a "comment". 
+                    An annotation without Body structure corresponds to a "highlight" if its Selector defines a range of characters, 
+                    a region in an image, or a time period, 
+                    and a "bookmark" if it does not.
                 </p>
 
                 <p class="ednote">We should specify whether a property may appear at most once (body, target) because
@@ -840,16 +840,8 @@
                             </td>
                             <td> The body type. </td>
                             <td> "TextualBody" | "Image" | "Audio" | "Video" </td>
-                            <td> The body type. It MUST be “TextualBody”. </td>
+                            <td> The body type. </td>
                             <td> string </td>
-                            <td> Yes </td>
-                        </tr>
-                        <tr>
-                            <td>
-                                <code> value </code>
-                            </td>
-                            <td> The textual content of the annotation. </td>
-                            <td> Localizable text </td>
                             <td> Yes </td>
                         </tr>
                         <tr>
@@ -857,7 +849,8 @@
                                 <code> format </code>
                             </td>
                             <td> The media-type of the note. Only plain text is allowed for textual notes. 
-                                Audiovisual notes can be of any media type supported by EPUB, see EPUB Media Types.
+                                Audiovisual notes can be of any media type supported by EPUB, 
+                                see <a data-cite="epub34#sec-core-media-types">EPUB Media Types</a>.
                             </td>
                             </td>
                             <td> string </td>
@@ -878,6 +871,22 @@
                             <td> The relative URL of the audiovisual note. </td>
                             <td> URL </td>
                             <td> Yes, if type is "Image", "Audio", or "Video". Not applicable for textual notes. </td>
+                        </tr>
+                        <tr>
+                            <td>
+                                <code> <dfn>color</dfn> </code>
+                            </td>
+                            <td> The color of the annotation; yellow by default. </td>
+                            <td> "pink" | "orange" | "yellow" | "green" | "blue" | "purple" </td>
+                            <td> No </td>
+                        </tr>
+                        <tr>
+                            <td>
+                                <code> <dfn>highlight</dfn> </code>
+                            </td>
+                            <td> The style of the annotation; solid background by default. </td>
+                            <td> "solid" | "underline" | "strikethrough" | "outline" </td>
+                            <td> No </td>
                         </tr>
                         <tr>
                             <td>
@@ -1194,16 +1203,6 @@
                 <code>application/ld+json;profile="http://www.w3.org/ns/anno.jsonld"</code>
             </p>
 
-            <p>
-                Following the requirements of JSON-LD, each annotation file (whether a single annotation or an
-                annotation set) MUST start with a context declaration referring to the following, EPUB Annotation
-                specific context: <code>https://www.w3.org/ns/epub-anno.jsonld</code>.
-                This context file imports the "core" context file of the [[[annotation-model]]] [[annotation-model]], i.e.,
-                <code>https://www.w3.org/ns/anno.jsonld</code>.
-                The EPUB Annotation specific context file contains terms extending, and sometimes overriding the terms
-                in the core annotation context file, as defined in this specification.
-            </p>
-
             <aside class="example" title="Required context declaration">
                 <pre>
                     {
@@ -1228,6 +1227,18 @@
             </p>
             
             <section>
+
+                <p> An AnnotationSet can be shared as a detached file, or embedded in an EPUB package. 
+                    The advantage of detached annotations is that they can be shared independently of the publication, 
+                    and that they can be associated with a publication without modifying it. 
+                    The advantage of embedded annotations is that they are always available to users of the publication, 
+                    without any need to import them.
+
+                <p class="note">
+                    Reading Systems are not required to store annotations with such a format: this format is for export and import purposes.
+                </p>
+            
+                </p>
 
                 <h2>Detached annotations</h2>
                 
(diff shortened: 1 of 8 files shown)


llemeurfr:
Fixes #2925

llemeurfr:
As this PR deals with the serialization of annotation sets and individual annotations are never exchanged outside such sets, I'm willing to remove the `@context` object from all samples that relate to individual annotations. 
I'm hesitating to do this in this PR for the sake of clarity. Please let me know if I should do it.

iherman:
(practicality) You should probably merge the current 'main' branch into this one to avoid merge conflicts...

iherman:
> As this PR deals with the serialization of annotation sets and individual annotations are never exchanged outside such sets, I'm willing to remove the `@context` object from all samples that relate to individual annotations. I'm hesitating to do this in this PR for the sake of clarity. Please let me know if I should do it.

I think they should be removed.

iherman:
> Evolution in the specification (I didn't look if the json schema has to be modified) to handle audiovisual annotation bodies.

You mean the vocabulary? Yes, it has to be modified. My proposal is that we try to finalize this PR and I will then do a separate PR to update the vocabulary. No need to mix the two.

epub34/annotations/index.html
```diff
@@ -1213,98 +1234,142 @@ <h1>Serialization</h1>
...
+                    as specified in ISO/IEC 21320-1:2015 ([[ISO21320]] and [[zip]]). </p>
+                
+                <p> This specification introduces a dedicated file extension for detached annotations: 
+                    <code>.annotations</code>.
+                </p>
+
+                <p> The media type of this file is
+                    <code>application/zip;profile="https://www.w3.org/TR/epub-anno-10/"</code>. </p>
+            
+                <p> The serialized AnnotationSet is stored in the Root Directory as <code>annotations.json</code>.
```
iherman: I presume "Root Directory" is a term defined in one of the zip specifications. We should have a (normative) reference to the definition.

epub34/annotations/index.html
```diff
@@ -831,58 +839,52 @@ <h2>Body</h2>
...
                         </tr>
                         <tr>
                             <td>
                                 <code> format </code>
                             </td>
-                            <td> The media-type of the annotation value. It
-                                SHOULD be "text/plain".
+                            <td> The media-type of the note. Only plain text is allowed for textual notes. 
+                                Audiovisual notes can be of any media type supported by EPUB, see EPUB Media Types.
+                            </td>
```
iherman, line 852:
```suggestion
                            <td> The media-type of the note. Only "text/plain" is allowed for textual notes. 
                                Audiovisual notes can be of any media type supported by EPUB, see <a data-cite="epub34#sec-core-media-types>EPUB Media Types</a>.
                            </td>
```

The downside of this reference is that the EPUB spec does not mention any specific video media type, only audio and images. We may decide to gloss over this, or say a few words in a note; I am not sure.

epub34/annotations/index.html
```diff
@@ -1213,98 +1234,142 @@ <h1>Serialization</h1>
...
+                <p> To simplify associating annotations with a publication, a Reading System MUST
+                    offer a way to select a publication before selecting an annotation set. An automatic association of an annotation set with a publication (e.g. via drag and
+                    drop) MAY also be proposed, butidentifying the proper publication from 
+                    the metadata in the annotation set is error prone. </p>
+                <p> When importing an annotation set, a Reading System SHOULD display a message with the
+                    title of the annotation set and the number of annotations in the set. The Reading
+                    System MUST offer the user the choice to abort the import. </p>
+                <p> Each annotation is uniquely identified. If during the import of an annotation set,
+                    one or more annotations are re-imported, the Reading System MUST offer to the user
+                    the choice to override existing annotations or abort the import of the annotation
```
iherman: Each annotation is uniquely identified _within_ an Annotation set. But it seems that "import" means to take each imported annotation and add it to the local collection. What happens if, when doing that, there is an ID clash? Also: what happens with the metadata (Generator, About) of the incoming AnnotationSet? Is it lost?

But maybe I misunderstand, and the local and imported annotations co-exist side-by-side within the EPUB. But then this would clash with the assumption of having a fixed name for the annotations.

epub34/annotations/index.html
```diff
@@ -1213,98 +1234,142 @@ <h1>Serialization</h1>
                 Such implementations may safely ignore the context
                 declarations and are not required to dereference the respective URLs.
             </p>
+            
+            <section>
+
+                <h2>Detached annotations</h2>
```
iherman, line 1223: I believe an informal introduction paragraph should be added, explaining why detached and embedded annotations are introduced in the first place. Ie, that these are the two approaches to export annotations for interchange. (I know §6 includes subsections on some of this, but that is the "best practices" section; that is not the same I think.)

epub34/annotations/index.html
```diff
@@ -1213,98 +1234,142 @@ <h1>Serialization</h1>
...
+
+                <h2>Annotations embedded in EPUB</h2>
+
+                <p> The AnnotationSet is stored in the META-INF directory as <code>annotations.json</code>. 
+                    Audiovisual notes, if present, MAY be in any location descendant from the META-INF directory, or in the META-INF directory itself. Annotation bodies within the AnnotationSet MUST reference these resources via relative-URL strings [[url]].</p>
+
+                <div class="note">
+                    <p> Storing audiovisual notes in the META-INF directory has implications for the size of the EPUB package, and is therefore not recommended.</p>
+                </div>
+
```
iherman, line 1259: An extra note making it clear that this setup is only necessary when the goal is to _export_. RS-s are not _required_ to hold the annotations this way, it is up to their internal organization.

epub34/annotations/index.html
```diff
@@ -241,10 +240,10 @@ <h2>Annotation Object</h2>
 
                 <p class="note">The type of annotation should be considered when determining the value of the
                     [=motivation=] property.
-                    An annotation with a Body structure corresponds to a "comment".
-                    An annotation without Body structure corresponds to a "highlight" if its Selector defines a range of
-                    characters, a space in an image
-                    or a time period, and a "bookmark" if it does not.
+                    The motivation is "highlighting" if the Selector 
+                    defines a range of characters, a space in an image or a time period.
+                    It is "bookmarking" if the Selector references a single character in the resource.
```
bduga, line 245: That seems wrong - you might have a bookmark in a fixed layout volume that has no characters (is entirely images). Maybe:
```
Otherwise, it is "bookmarking".
```
Which is similar to the previous text.

epub34/annotations/index.html
llemeurfr: The new wording should clarify this.

epub34/annotations/index.html
llemeurfr: mmh I found it in the LPF spec. But it is, in fact, defined in this spec as https://www.w3.org/TR/lpf/#sec-terminology. A zip has file names, each with an optional relative path.

I think "the root of the ZIP file" is something everybody will understand. Let's try it.

llemeurfr:
Spec updated with today's W3C discussions.

epub34/annotations/index.html
iherman: Ok, let us leave it for now, but maybe we should follow the LPF spec in adding a terminology section, eventually.

epub34/annotations/index.html
iherman: Ok for now. As we agreed on the call, once this PR is merged, we will have to specify the import mechanism normatively...

epub34/annotations/index.html
```diff
@@ -234,10 +240,10 @@ <h2>Annotation Object</h2>
...
```
sueneu, line 245:
```suggestion
                    It is "bookmarking" if the Selector references a single item in the resource.
```

Users can annotate a single image, like a chart or graph, or a video/audio, too? Or is selecting an instance of an image or video "highlighting"? See above.

Can a space or `<hr>` or blank line `<p></p>` be referenced?

epub34/annotations/index.html
```diff
@@ -234,10 +240,10 @@ <h2>Annotation Object</h2>
 
                 <p class="note">The type of annotation should be considered when determining the value of the
                     [=motivation=] property.
-                    An annotation with a Body structure corresponds to a "comment".
-                    An annotation without Body structure corresponds to a "highlight" if its Selector defines a range of
-                    characters, a space in an image
-                    or a time period, and a "bookmark" if it does not.
+                    The motivation is "highlighting" if the Selector 
+                    defines a range of characters, a space in an image or a time period.
```
sueneu, line 244: > defines a range of characters, a space in an image or a time period.

I'm not sure how one would select the just the space in an image's content. Isn't it all pixels generally speaking? How would an RS differentiate between foreground and space?

I'm guessing I have misunderstood this. Do we mean that if an image is selected, the motivation is "highlighting"?

bduga: > select the just the space in an image's content.

I think this is meant to be `space` as in `area` or `region`. Maybe we should use one of those words instead.
```suggestion
                    defines a range of characters, a region of an image or a time period.
```

epub34/annotations/index.html
bduga: This change suggestion has revealed to me how broken this concept is. Take this snippet:
```
<p id="p1">Now is the time ... <p>
```
If I use the CSS selector `#p1`, is this a highlight or a bookmark? Either is possible. Maybe I want to highlight the entire paragraph and color it red. Or maybe this is just the first paragraph on a page and I want to add a bookmark. And if it is possible to determine the motivation of an annotation by the thing that is referenced, why do we even need a motivation field in the first place?

Shouldn't the motivation be based on what the user did in the UI for the reading system? The reading system probably knows what the user was trying to do - so if they clicked in the upper right corner of the page in an area where a bookmark toggle is displayed in this particular RS, then the motivation is `bookmarking`. If they dragged a finger over the entire text of the paragraph, and chose a color to highlight with, then it is `highlighting`. This shouldn't reflect the data we already have, it should be informed by the actions of the user that are not otherwise visible in this data.

iherman: I believe your description is correct, @bduga. From the data model point of view both bookmarking and highlighting are identical in the sense that they have no annotation body. The only difference is what the value of "motivation" is, and that is up to the RS to determine.

However... shouldn't this be a separate issue? It does not seem to have a relevance to the PR's focus.

bduga: Am I misreading this change? It looks like the pr removes `commenting` as a motivation, and modified the definition of `bookmarking`, which is where my comments are directed. If changing these values is outside the scope of the pr, I am fine with moving to a new issue/pr.

iherman: Hm, you are right, my apologies. It indeed removes `commenting`. @llemeurfr why? How does that relate to the original goals of the PR?

llemeurfr: OK, I tested this change when thinking about the mapping of audiovisual notes to motivcations, but it is merits its own discussion. I revert the modification and open https://github.com/w3c/epub-specs/issues/2987.

PR #2975, opened 2026-04-07
Removing a reference to the json-ld import mechanism

iherman:
For technical reasons this approach is no longer in use in yml2vocab, which generates the context file for the annotation vocabulary. I forgot to remove the reference to it in the spec.


iherman:
Merging this; it is was a mistake not to remove that at the last iteration of the vocabulary.

PR #2982, opened 2026-04-12
Added a subtitle

iherman:
To move forward a discussion (see [slack discussion](https://w3ccommunity.slack.com/archives/C0AP01YQ34N/p1775847228401349)) on adding more "clues" to the essence of the annotation spec.

---

See:

* For EPUB Annotations 1.0:
    * [Preview](https://raw.githack.com/w3c/epub-specs/annotations/added-subtitle/epub34/annotations/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fannotations%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Fannotations%2Fadded-subtitle%2Fepub34%2Fannotations%2Findex.html)

### DIFF
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -11,7 +11,7 @@
         <script src="../../common/js/add-caution-hd.js" class="remove"></script>
         <script class="remove">
             var respecConfig = {
-                subtitle: "An interchange format for annotations, highlights, and bookmarks",
+                subtitle: "An interchange format for user notes, highlights, and bookmarks",
                 group: "pm",
                 wgPublicList: "public-pm-wg",
                 specStatus: "ED",
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -11,6 +11,7 @@
         <script src="../../common/js/add-caution-hd.js" class="remove"></script>
         <script class="remove">
             var respecConfig = {
+                subtitle: "An interchange format for annotations, highlights, and bookmarks",
                 group: "pm",
                 wgPublicList: "public-pm-wg",
                 specStatus: "ED",


HadrienGardeur:
"An interchange format for annotations, bookmarks and highlights" would imply that annotations are separate from bookmarks and highlights. 

I still think that we have a vocabulary issue:
- by annotation, do we mean the textual, auditory or visual note that can be optionally added to a highlight/bookmark? 
- or do we consider that annotations encompass both highlights and bookmarks?

iherman:
My "mental model" is the former. I have the impression that this is the case for most.

I realize that, strictly speaking, our data model is technically the second option. But the subtitle should not necessarily have to be 100% precise; it is an eye-catcher one liner for readers.

HadrienGardeur:
I think that a lot of people interchangeably use "Annotations" to refer to "Highlights", somehow implying that they're always associated to a note (which is incorrect).

I don't think that there's a well established term across RS to refer to this textual, auditory and/or visual content, which is why I'm using "note" since it's less entangled with the concept of highlighting.

mattgarrish:
Reusing "annotation" in the subtitle makes you wonder where highlights and bookmarks suddenly came from, even if I get where you're going with this. Is there a way to avoid it and also make clear that this is not publisher content, like:

"An interchange format for user notes, highlights and bookmarks"

iherman:
> Reusing "annotation" in the subtitle makes you wonder where highlights and bookmarks suddenly came from, even if I get where you're going with this. Is there a way to avoid it and also make clear that this is not publisher content, like:
> 
> "An interchange format for user notes, highlights and bookmarks"

I like that. Before I change it, any other opinion?

dalerrogers:
Since we are using "words", I go back to the dictionary. Annotation: https://www.dictionary.com/browse/annotate. I have always used the term to refer to notes or comments about a piece of content. A bookmark is a location. A highlight doesn't add commentary. So, I wonder how far we want to generalize the word annotate? My $0.02.

bduga:
+1 to @mattgarrish's suggestion. As for what we mean by annotation, we need to consider the audience of this spec, which seems to be reading system implementors. In that space, the term `annotation` seems to have become a term of art that refers to an entire category of things users add to books, like notes (text and audio), highlights, underlines, bookmarks, doodles, and even reading progress. This set of things with a shared set of properties doesn't have a great name in English, at least that I am aware of, so annotation has come to encompass that. At least, at my previous jobs, that is what we called them. If we want to reconsider use of that term, I think it should come from the reading systems implementors.

dalerrogers:
@bduga  You bring up a good point... audience. Is the spec limited to reading system implementers? Or is a publisher, or author, also considered to be the audience? That is one of the most challenging things about writing: knowing the audience. If the primary audience is the implementers of the reading systems, then I agree, we should go with the vocabulary they are already using. For the "spec"  (or its translation) designed for authors, it would be good to define the term as well.

bduga:
> Is the spec limited to reading system implementers? Or is a publisher, or author, also considered to be the audience?

Questioning the target audience is reasonable, and there are certainly parts of our specs where there is a very legitimate mixed audience. But for this particular spec, it seems like the main audience (perhaps the only one) is reading system implementors. This is very squarely targeted at exporting annotations from reading systems with the intent of importing them into other reading systems. Are there other valid users? Probably. But the vast majority will be exchange of annotations across reading systems. And I believe our list of interested implementors is entirely comprised of that use case.

But, as @iherman also likes to ask, are we ok with the change @mattgarrish suggested? If so, let's accept it. I, for one, approve of the text.

dalerrogers:
Given the discussion and audience, I +1 @mattgarrish  language.

iherman:
Adopted the Matt's text.

iherman:
Discussion went silent; change is cosmetic that can be reopened if necessary; merging

Issue #2884, opened 2026-01-13
Target is not necessarily a specific segment, but we say it is


bduga:
### Describe the problem

In the section titled "Subset of the Web Annotation Data Model" we say:

> The Target of an EPUB Annotation is the content being annotated, which is a specific segment in a document within the EPUB package defined by its relative [Source](https://htmlpreview.github.io/?https://github.com/w3c/epub-specs/blob/2e222e197b9c71681ba13b30562b2d5a08beb180/epub34/annotations/index.html#1-3-1-source) URL and a [Selector](https://htmlpreview.github.io/?https://github.com/w3c/epub-specs/blob/2e222e197b9c71681ba13b30562b2d5a08beb180/epub34/annotations/index.html#1-3-2-selector).

However, we do not require the target to be a specific segment in a document, unless we are including the entire document with no range specified as a "segment in" that document. I am actually not sure why we need to define target here at all? There are a lot of nuances (for instance, I am not sure from this definition if selector is required), and this is just a list of changes from web annotations, which is really only interesting to people that know what is in web annotations, and hence are likely aware of how target is used.

### Describe the fix or new feature you propose

 I propose deleting the second sentence of this bullet point.

llemeurfr:
We'll discuss this at the next call, but I agree with you: this section lists the constraints applied in this specification, relative to the W3C Annotations Data Model: we don't need to add definitions there.

llemeurfr:
The term "segment" comes from the W3C Annotation Model sections [ 3.2.3 Segments of External Resources](https://www.w3.org/TR/annotation-model/#segments-of-external-resources) and [4.2 Selectors](https://www.w3.org/TR/annotation-model/#selectors). 

In our case, a target is a combination of a source and a selector. The source identifies an EPUB resource, and the selector references a "Segment (of Interest)" (extract from the first paragraph). 

Therefore, I agree that we can get rid of the sentence describing what a Target is in the non-normative Introduction (which explains the main contraints applied to the Wev Annotation Model in this profile). But I think that using the term "segment" in [3.3 Target](https://w3c.github.io/epub-specs/epub34/annotations/#target) is adequate. 

To make it clear I propose to start 3.3 with "Target object</dfn> of an annotation associates the annotation with a specific Segment (of Interest) of a resource in the current publication.". I don't see the need for a hyperlink to the W3C Annotation Model here; the capital 'S' attracts attention, the "of Interest" gives some context, and a curious user may have a look into the Annotation Model.

[...]

PR #2990, opened 2026-04-22
Add some context to the use of the term Segment in the Target section.

llemeurfr:
Associated with issue #2884.

---

See:

* For EPUB Annotations 1.0:
    * [Preview](https://raw.githack.com/w3c/epub-specs/anno-segment/epub34/annotations/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fannotations%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Fanno-segment%2Fepub34%2Fannotations%2Findex.html)

### DIFF
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -113,9 +113,6 @@
 
                 <li> Web Annotations have 1 or more Targets, while only a single <a href="#target">Target</a> is
                     allowed for an EPUB Annotation.
-                    The Target of an EPUB Annotation is the content being annotated, which is
-                    a specific segment in a document within the EPUB package defined by its relative
-                    <a href="#source">Source</a> URL and a <a href="#selector">Selector</a>.
                 </li>
 
                 <li> Web Annotations may have multiple motivations, while only a single motivation is allowed for EPUB
@@ -320,7 +317,7 @@
 
             <section>
                 <h2>Target</h2>
-                <p>The <dfn>Target object</dfn> of an annotation associates the annotation with a specific segment of a
+                <p>The <dfn>Target object</dfn> of an annotation associates the annotation with a specific Segment (of Interest) of a
                     resource in the current publication.</p>
                 <p> This document defines three target sub-properties: </p>
                 <table class="zebra">


iherman (approved):

epub34/annotations/index.html
```diff
@@ -320,7 +317,7 @@ <h2>Creator</h2>
 
             <section>
                 <h2>Target</h2>
-                <p>The <dfn>Target object</dfn> of an annotation associates the annotation with a specific segment of a
+                <p>The <dfn>Target object</dfn> of an annotation associates the annotation with a specific Segment (of Interest) of a
```
bduga, line 320: I still have an issue with this - in our spec, we allow a target to specify an entire document, with no selector. So, is an entire document a Segment of Interest of itself? I don't see a proper definition for the term, but from the use it seems to imply a selector that pulls out some portion of the document. Perhaps:

```suggestion
                <p>The <dfn>Target object</dfn> of an annotation associates the annotation with a specific resource or Segment (of Interest) of a
```

iherman:
This was discussed during the [pmwg meeting on 21 May 2026](https://w3c.github.io/pm-wg/minutes/2026-05-21.html#adf3).

<h3>Add some context to the use of the term Segment in the Target section <a href="https://github.com/w3c/epub-specs/pull/2990" rel="noopener noreferrer">w3c/<wbr>epub-specs#2990</a></h3>
<p><cite>LaurentM:</cite> we know we can create bookmarks and annotations. A bookmark is a placeholder in a text or image but not so precise for a user<br>
<span>… an annotation is a highlight of text or other media, and has some range in the content</span><br>
<span>… the selectors we define can isolate a range even for a bookmark, since there is no specific marker for a book mark</span><br>
<span>… will we accept that the selector of a bookmark can be a range or must be a single point?</span></p>
<p><cite>Hadrien:</cite> reading systems usually have a specific affordance for bookmarks<br>
<span>… like an icon on the corner of the page, we look at what is currently displayed to the users</span><br>
<span>… and check if there is a bookmark there</span><br>
<span>… if there is a superlong range, the text could be displayed across many screens and cause problematic behavior</span><br>
<span>… i suspect rs will want to define a bookmark that won't cross the boundary of multiple screens</span></p>
<p><cite>ivan:</cite> this is related to the previous discussion because th choice of selectors is done behind the scenes by the rs<br>
<span>… so this is a matter of the affordance that LaurentM was talking about. So the situation you describe won't happen</span><br>
<span>… so reading systems should do this properly. We should say if a readers intent is to set a bookmark, we should use a single selector</span></p>
<p><cite>LaurentM:</cite> a text position selector has a beginning and end, and is a range<br>
<span>… do you agree that a bookmark should be a single location we should add that to the spec</span></p>
<p><cite>LaurentLM:</cite> if reading system A generates a bookmark, it will still work on reading system A, but export that annotation to reading system B and B handles bookmarks differently, we could have problems<br>
<span>… even if we include a specification, reading system B will have a sanitisation process</span><br>
<span>… that will then decide if the bookmark would be at the beginning or end of a range</span><br>
<span>… we can try to force whatever and people will do what they want</span><br>
<span>… I don't think we will be able to control what people will do</span><br>
<span>… it might be good to have some recommendations but reading systems shouldn't expect it</span></p>
<p><cite>DaleRogers:</cite> a reading system can do what it wants with our declarations, and we can make those declarations</p>

<hr /></details>

reidmore-online:
Will be overwritten by an upcoming PR to address more recent discussion on #2884

PR #3001, opened 2026-05-13
Proposing a text on merging.

### LINKED ISSUE #2997: Merging Annotation Sets
### Section

_No response_

### Describe the problem

We should discuss the case of merging annotation sets, particularly around resolving locators. 

Two main cases come to mind, though there may be more:
1. I am reading and annotating a book on my device, and a friend shares a file containing their annotations, I want to review it alongside mine by importing it to my reading system. 
2. I download multiple annotation sets from different sources for the same book, and I want to view them all in the same reading experience. 

The biggest potential issue that comes to mind is what to do in cases where there are multiple annotations on the same locator, or overlapping annotations (this may be purely a UI issue though). 

_No response_

### LINKED ISSUE #2998: How can I be sure this annotation file is for my book?
### Section

_No response_

### Describe the problem

In the case where I am reading a book on my device and want to open an annotation file I acquired separately from the book, how can I be sure that the annotation file I downloaded is compatible with my book/based on the content of my book? 

EPUB has a tricky ID problem, where it’s entirely possible for IDs to change, be reused (ISBNs), or drift between distribution systems or versions. While dc:identifier is a required metadata field, what goes in it can change. 

I think this is a problem that can possibly be tackled both from a reading system direction and user-driven direction. 

From the spec side, we can state something like the “identifier” of a book is a combination of its dc:identifier, dc:title, and author/creator. Reading systems can check this on ingestion of an annotations file. On the user side, in the case of a mismatch or incomplete check (2/3 match for example), the user can confirm/reject uploading the annotation file. 

Another consideration for the reading systems might be an addition check of the locators in the annotation set with the OPF, if for example the annotation locators are “chapter_one.xhtml#(location)”, but the book’s chapters are in the format “ch1.xhtml” there is a strong possibility this is the wrong file.

iherman:
This is an attempt to add a small paragraph providing a fix for #2997 and #2998.

---
See:

* For EPUB Annotations 1.0:
    * [Preview](https://raw.githack.com/w3c/epub-specs/annotations/notes-on-merging/epub34/annotations/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fannotations%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Fannotations%2Fnotes-on-merging%2Fepub34%2Fannotations%2Findex.html)

### DIFF
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -1267,12 +1267,12 @@
                     <h3>Importing detached annotations</h3>
 
                     <p>
-                        When a reading system imports detached annotations in the form of, say, a ZIP file, it SHOULD check
-                        whether those annotations are indeed for the publication being rendered. 
+                        When a reading system imports detached annotations, it SHOULD check
+                        whether those annotations are indeed for the publication being rendered.
                         Because there are no universally accepted identifiers for EPUB publications, this process involves some
                         heuristics that depend on the reading system, such as comparing the content of the [=package document=]
-                        with the metadata assigned to the imported annotations, or checking whether the target resources 
-                        of the annotation selectors are valid. 
+                        with the metadata assigned to the imported annotations, or checking whether the target resources
+                        of the annotation selectors are valid.
                         This process MAY require user interaction via the reading system's user interface.
                         In any case, information about the imported annotations SHOULD be provided to the user.
                     </p>
--- a/epub34/annotations/index.html
+++ b/epub34/annotations/index.html
@@ -1263,6 +1263,20 @@
                         allow a broad interoperability among operating systems.</p>
                 </div>
 
+                <section>
+                    <h3>Importing detached annotations</h3>
+
+                    <p>
+                        When a reading system imports detached annotations in the form of, say, a ZIP file, it SHOULD check
+                        whether those annotations are indeed for the publication being rendered. 
+                        Because there are no universally accepted identifiers for EPUB publications, this process involves some
+                        heuristics that depend on the reading system, such as comparing the content of the [=package document=]
+                        with the metadata assigned to the imported annotations, or checking whether the target resources 
+                        of the annotation selectors are valid. 
+                        This process MAY require user interaction via the reading system's user interface.
+                        In any case, information about the imported annotations SHOULD be provided to the user.
+                    </p>
+                </section>
             </section>
 
             <section>


sueneu:
Looks good!

llemeurfr:
I'm worried about having a section defining RS behaviour outside the "Reading System" Section 6. 
So far, most of the specifications are about the annotation format, and processing rules are concentrated in Section 6. 
More specifically, Section 6.5 is about "Importing Annotation", which is what this text is about. 

Therefore, I propose to complement Section 6.5 with this text, which is ok in itself. The first part of the section would be about the recommended UX, and the second part would present these processing rules.

iherman:
My problem is that, while placing it into §6.5 is, editorially, a logical thing to do, the whole section §6 may be glossed over by an everyday user who just want to understand what can be done and how. The title, ie, "Best practices for reading systems" reads like "Read this if you implement a reading system, otherwise just skip it".

Maybe we should look at §6 more critically, and see items which are to be expected by a user from a reading system, and separate from those which are really for implementers only. And find an alternative title for the section.

epub34/annotations/index.html
```diff
@@ -1263,6 +1263,20 @@ <h2>Detached annotations</h2>
                         allow a broad interoperability among operating systems.</p>
                 </div>
 
+                <section>
+                    <h3>Importing detached annotations</h3>
+
+                    <p>
+                        When a reading system imports detached annotations in the form of, say, a ZIP file, it SHOULD check
```
sueneu: > in the form of, say, a ZIP file,

We do not need this phrase if processing all imported annotations require this check. It implies that different file formats will require different handling

iherman: You are right. It is a leftover reaction, we only use ZIP these days.

iherman: Handled in https://github.com/w3c/epub-specs/pull/3001/commits/0d01a963b880f6590ff21a82f0acb84043ab71ee, resolving this comment.

iherman:
This was discussed during the [pmwg meeting on 21 May 2026](https://w3c.github.io/pm-wg/minutes/2026-05-21.html#6b56).

<h3>Proposing a text on merging [annotations]  <a href="https://github.com/w3c/epub-specs/pull/3001" rel="noopener noreferrer">w3c/<wbr>epub-specs#3001</a></h3>
<p><cite>ivan:</cite> I have a question about when annotaions are merged. What can we describe normatively<br>
<span>… it turns out there is not much consistency in book metadata so we have to rely on the heuristics in the rs</span><br>
<span>… so we need a paragraph to make clear to readers what they can expect on import</span><br>
<span>… if I am a user and not an implementor, I may never see it</span><br>
<span>… I don't want to hide the text from reading systems either</span><br>
<span>… if we are able to separate the information into implementer and user sections, then this text should go into the user part</span></p>
<p><cite>LaurentM:</cite> most of section six is about what a user can expect<br>
<span>… there are sentences that are aimed at developers</span><br>
<span>… we can change the language to make it clear that implementers should pay attention to</span><br>
<span>… rather than duplicating information</span></p>
<p><cite>DaleRogers:</cite> I notice section 6 is called "best practices for reading systems" we could in each section say what audience a section is aimed at</p>
<p><cite>LaurentM:</cite> the risk is duplication and being unclear<br>
<span>… developers know how to deal with a use case, if we reorient this section more like use cases we can avoid splitting and duplicating</span></p>
<p><cite>ivan:</cite> perhaps we could rewrite the section as use cases, and then have other sections or notes for implementors<br>
<span>… most of the sections would be use case oriented</span><br>
<span>… editorially, the simplest thing is to merge the PR and then we will move it during rewriting</span></p>
<p><cite>LaurentM:</cite> I can take a run on rewriting it, I'll show you the branch before the PR</p>
<p><cite>sueneu:</cite> Do users look at the reading system spec? I expect not, so rewriting it with a use case framing makes sense</p>
<p><cite>LaurentM:</cite> I think users won't read the spec but there is an interim party, people who make articles and usage notes who would need this information<br>
<span>… and they will pick up on this vocabulary</span></p>

<hr /></details>

iherman:
Per discussion on the call, merging this, and moving the text in a later PR.

PR #3003, opened 2026-05-19
Addiing "Software" as an alias to as:Application

iherman:
(This was forgotten): we use the ter "Software" for Generators but the OA refers to an activity stream term only. An alias is put into the vocabulary definition


iherman:
My mistake. It turns out that the web annotation json-ld context file already includes this aliasing, it is unnecessary to repeat it.

PR #3007, opened 2026-05-25
Fix glossterm HTML usage context.

emilis:
Since `dl[epub:type="glossary"] > dd` is specified as implied to be _glossdef_, I assume _glossterm_ should be implied from `... > dt`.

### DIFF
--- a/wg-notes/ssv/index.html
+++ b/wg-notes/ssv/index.html
@@ -1158,7 +1158,7 @@
 					<dd>
 						<p>
 							<span class="subproplabel">HTML usage context: </span>The glossterm property is implied on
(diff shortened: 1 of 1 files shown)


mattgarrish:
As a copy/paste typo, I agree, but as the actual implied term, no, I think the SSV is wrong on that count. The `dt` element contains the term but there's supposed to be a `dfn` tag around the term itself. A `dt` can contain more text than just the term being defined.

But I expect this is one of the more unused corners of epub as the indexing spec never got traction, so it probably doesn't matter too much if implying the term from `dt` doesn't work properly all the time.

PR #3012, opened 2026-05-27
Move 1.2 strings to top of conformance list

mattgarrish:
I noticed that the list of conformance strings starts with the 1.1 strings first. Since this is the 1.2 standard, it seems more appropriate that the list begin with 1.2 conformance strings.

Also, since we recommend always conforming to the latest version of WCAG, I reordered the list so that the 2.2 identifiers are first, then 2.1, and finally 2.0.

Finally, to deemphasize old versions and to avoid the list of strings occupying greater and greater space, I added two details elements to group them. The first is open by default and lists the strings for the current version. The second is collapsed and will be where we can hide all the old identifiers that are still valid but not relevant to the current document anymore.

Thoughts on these changes welcome.

***

[Preview](https://raw.githack.com/w3c/epub-specs/editorial/conf-id-order/epub34/a11y/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fa11y%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Feditorial%2Fconf-id-order%2Fepub34%2Fa11y%2Findex.html)

### DIFF
--- a/epub34/a11y/index.html
+++ b/epub34/a11y/index.html
@@ -1078,27 +1078,28 @@
 &lt;/package></pre>
 					</aside>
 
-					<p>The following conformance strings are valid as of publication of this specification:</p>
+					<p id="conf-strings">The following conformance strings are valid as of publication of this
+						specification:</p>
 
 					<ul>
-						<li>EPUB Accessibility 1.1 - WCAG 2.0 Level A</li>
-						<li>EPUB Accessibility 1.1 - WCAG 2.0 Level AA</li>
-						<li>EPUB Accessibility 1.1 - WCAG 2.0 Level AAA</li>
-						<li>EPUB Accessibility 1.1 - WCAG 2.1 Level A</li>
-						<li>EPUB Accessibility 1.1 - WCAG 2.1 Level AA</li>
-						<li>EPUB Accessibility 1.1 - WCAG 2.1 Level AAA</li>
-						<li>EPUB Accessibility 1.1 - WCAG 2.2 Level A</li>
-						<li>EPUB Accessibility 1.1 - WCAG 2.2 Level AA</li>
-						<li>EPUB Accessibility 1.1 - WCAG 2.2 Level AAA</li>
-						<li>EPUB Accessibility 1.2 - WCAG 2.0 Level A</li>
-						<li>EPUB Accessibility 1.2 - WCAG 2.0 Level AA</li>
-						<li>EPUB Accessibility 1.2 - WCAG 2.0 Level AAA</li>
-						<li>EPUB Accessibility 1.2 - WCAG 2.1 Level A</li>
-						<li>EPUB Accessibility 1.2 - WCAG 2.1 Level AA</li>
-						<li>EPUB Accessibility 1.2 - WCAG 2.1 Level AAA</li>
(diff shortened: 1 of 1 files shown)


mattgarrish:
To jump straight to where the changed list is, try this preview link:
https://raw.githack.com/w3c/epub-specs/editorial/conf-id-order/epub34/a11y/index.html#conf-strings

iherman:
I presume it has its logic but, as an outsider, it looks odd that an older version (say EPUB A11y 1.1) is bound to the same WCAG 2.2 as the EPUB A11y 1.2 version). Isn't the list of older version some sort of a history, that should be bound to whatever it used to be?

clapierre (approved):

mattgarrish:
> Isn't the list of older version some sort of a history, that should be bound to whatever it used to be?

Yes, it's like a full record of every valid conformance string that can be used with dcterms:conformsTo, not just the ones that are valid for this version. That's why I'd like to hide the old stuff.

And for either of 1.1 or 1.2 you can conform to any version and level of WCAG 2 which is why we have all the different permutations of strings.

But it's a valid question whether we should bother to keep a record of every possible conformance string in the current version. I'd be fine dropping all the 1.1 strings if we change the lead-in paragraph to:

> The following strings are valid for identifying conformance to this specification:

That would leave it open-ended that other strings are also valid for older versions.

mattgarrish:
Maybe we can even go further. We don't need to have a variable for the current accessibility standard in this pattern above the strings:

> EPUB Accessibility [A11Y-VER](https://raw.githack.com/w3c/epub-specs/editorial/conf-id-order/epub34/a11y/index.html#acc-ver) - WCAG [WCAG-VER](https://raw.githack.com/w3c/epub-specs/editorial/conf-id-order/epub34/a11y/index.html#wcag-ver) Level [WCAG-LVL](https://raw.githack.com/w3c/epub-specs/editorial/conf-id-order/epub34/a11y/index.html#wcag-lvl)

The lead-in here says: "To indicate conformance to the accessibility requirements of this specification". So the number isn't variable. It should be:

> EPUB Accessibility 1.2 - WCAG [WCAG-VER](https://raw.githack.com/w3c/epub-specs/editorial/conf-id-order/epub34/a11y/index.html#wcag-ver) Level [WCAG-LVL](https://raw.githack.com/w3c/epub-specs/editorial/conf-id-order/epub34/a11y/index.html#wcag-lvl)

We established this pattern in 1.1 and didn't want it applied to 1.0 conformance which has its own URLs, but that the EPUB number changes is just something we know as a group. It doesn't need to be in the pattern.

mattgarrish:
The last commit makes the changes discussed in the previous comments. I also noticed the note after the conformance strings was still suggesting the list may grow to accomodate new wcag 2 versions. I deleted that paragraph since that is no longer expected and just left the text about addressing wcag 3 changes whenever they come to pass.

clapierre (approved):

PR #3015, opened 2026-05-29
Add section for redundant entry

mattgarrish:
See if you agree with my interpretation of this one. Since it hinges on multi-step processes, and those are pretty rare in publications, I can't think of a lot of cases for it. But I'm not completely confident about my take on it.

What I can't be sure of from the definition is if this would apply to a quiz where, for example, you can fix your answers and submit again. Does the process of validation and resubmission turn a basic test into a process in which you have to ensure that the information the user submitted is still set? Sometimes I think yes, other times no, since it's almost never required to redo a quiz or even to finish one (outside of an LMS, perhaps, but that still wouldn't be a content level requirement but a course one).

https://raw.githack.com/w3c/epub-specs/understand/redundant-entry/epub34/a11y-understand/index.html#redundant-entry

***

[Preview](https://raw.githack.com/w3c/epub-specs/understand/redundant-entry/epub34/a11y-understand/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fa11y-understand%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Funderstand%2Fredundant-entry%2Fepub34%2Fa11y-understand%2Findex.html)

### DIFF
--- a/epub34/a11y-understand/index.html
+++ b/epub34/a11y-understand/index.html
@@ -243,8 +243,43 @@
 			<section id="redundant-entry">
 				<h3>Redundant entry</h3>
 
-				<p>The [[wcag2]] <a data-cite="wcag2#redundant-entry">redundant entry success criterion (3.3.7)</a>
-					&#8230;</p>
+				<p>The goal of the [[wcag2]] <a data-cite="wcag2#redundant-entry">redundant entry success criterion
+						(3.3.7)</a> is to prevent users from having to enter the same information again when completing
+					a multi-step process on the web. This reduces memory requirements and stress for users with
+					cognitive and learning disabilities.</p>
+
+				<p>A common example of when this success criterion applies to web content is when checking out at an
+					ecommerce store. Users' billing and shipping addresses are most often the same, so requiring them to
+					re-input the same information in separate steps increases the cognitive difficulty of the checkout
+					process. A checkbox to automatically use the billing address as the shipping address is now a
+					ubiquitous pattern to avoid this issue.</p>
+
+				<p>But this is one of the WCAG success criteria that has greater application on the web than in the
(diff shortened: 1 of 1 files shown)


iherman (approved):

clapierre (approved):
I think this is correct, another idea would be I would think that even a single page if there was error checking / validating say making sure the date or email was formatted correctly to ensure that all the fields wouldn't be blanked out if an error occurred. Now this would probably also fail other WCAG criteria, but it would also force the user to re-enter all fields.

mattgarrish:
Ya, having had the weekend to ponder this one some more I think I'm leaning back to quiz validation/resubmission (or similar) falling under this criterion. Redundant entry becomes a bit ridiculous if it only applies to a form broken up into stages, like an ecommerce site having different steps to enter your billing info, then shipping info, then verifying shipping, etc.

If  a quiz could be wiped clean after an initial verification, then what are you left with but redundantly entering all the same answers again? But there would be an exception for a test that lets you try it again from scratch, in which case it becomes essential to start the user over.

And there is a close example to this listed in the understanding doc:

> A user submits a checkout form with an incorrect credit card number in it. The page updates, showing an error message. Submitted information, such as credit card number, is not cleared from the form.

I'll do some rewrites on this PR...

mattgarrish:
See if the changes in the last commit improve this one.

PR #3017, opened 2026-06-03
Add section on inapplicable success criteria

mattgarrish:
This pull request updates some prose I had started when I was first structuring the document.

https://raw.githack.com/w3c/epub-specs/understand/inapplicable-sc/epub34/a11y-understand/index.html#eval-not-applicable

***

[Preview](https://raw.githack.com/w3c/epub-specs/understand/inapplicable-sc/epub34/a11y-understand/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fa11y-understand%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Funderstand%2Finapplicable-sc%2Fepub34%2Fa11y-understand%2Findex.html)

### DIFF
--- a/epub34/a11y-understand/index.html
+++ b/epub34/a11y-understand/index.html
@@ -772,26 +772,27 @@ aside.sidebar {
 			<section id="eval-not-applicable">
 				<h3>Reporting inapplicable success criteria</h3>
 
-				<p>[[[epub-a11y-12]]] [[epub-a11y-12]] does not define a format for evaluation reports. Evaluators have the flexibility to use a format that best meets personal and/or regional reporting
-					needs. In general, though, it is expected that the report will list all of the WCAG success
-					criteria that were evaluated and the result of testing for each. For any criteria that were not met, include an explanation of what caused the failure.</p>
+				<p>[[[epub-a11y-12]]] [[epub-a11y-12]] does not define a format for evaluation reports. Evaluators have
+					the flexibility to use a format that best meets personal and/or regional reporting needs. In
+					general, though, it is expected that the report will list all the WCAG success criteria that were
+					evaluated and the result of testing for each. For any criteria that were not met, an explanation of
+					what caused the failure is typically also provided.</p>
 
-				<p>While this form of reporting is traditional for any kind of web content accessibility evaluation,
-					evaluators sometimes incorrectly assume that success criteria that do not apply to the content
-					should receive a pass. The idea being to confirm that the content meets all the requirements of WCAG
-					for the specified level.</p>
+				<p>While this form of reporting is traditional for accessibility evaluations, confusion sometimes arises
+					around how to report success criteria that do not apply to the content. The content does not fail
+					these success criteria, but it only passes because it is not relevant to check. The content does not
+					pass because the requirements have been met.</p>
 
-				<p>But if a success criterion does not apply to an EPUB publication, it should never be reported as an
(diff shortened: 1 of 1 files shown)


epub34/a11y-understand/index.html
```diff
@@ -772,16 +772,28 @@ <h3>Content type considerations</h3>
...
-				<p>When reporting success criteria that do not apply to a publication, it is recommended to label their
-					status as "not applicable" (or an equivalent language-specific term). If a "pass" value must be
-					assigned by the reporting template used, then a comment should be included that clearly states that
-					the publication contains no applicable content.</p>
-				-->
+				<p>[[[epub-a11y-12]]] [[epub-a11y-12]] does not define a format for creating evaluation so that
+					evaluators have the flexibility to use templates that best meet personal and/or regional reporting
+					needs. In general, though, it is expected that an evaluation report will list all the WCAG success
+					criteria that were evaluated, the result of testing each, and, for any success criteria that failed,
+					an explanation of what caused the failure.</p>
```
```suggestion
				<p>[[[epub-a11y-12]]] [[epub-a11y-12]] does not define a format for evaluation reports. Evaluators have the flexibility to use a format that best meets personal and/or regional reporting
					needs. In general, though, it is expected that the report will list all of the WCAG success
					criteria that were evaluated and the result of testing for each. For any criteria that were not met, include an explanation of what caused the failure.</p>
```

epub34/a11y-understand/index.html
```diff
@@ -772,16 +772,28 @@ <h3>Content type considerations</h3>
...
+
+				<p>While this form of reporting is traditional for any kind of web content accessibility evaluation,
+					evaluators sometimes incorrectly assume that success criteria that do not apply to the content
+					should receive a pass. The idea being to confirm that the content meets all the requirements of WCAG
+					for the specified level.</p>
```
sueneu: I'm not sure I understand. Does the content not receive a pass if there is criteria that doesn't apply? Like if there is no audio in my book, and therefore no accessible controls for audio in the book, does my book then fail the WCAG evaluation?

clapierre: @sueneu, I think what we don't want to do here is to say Passes WCAG 1.2.1 Audio-only and Video-only (Prerecorded) when there is no audio or video at all in the book, in that case that should not be reported.

mattgarrish: Right, that's what the rest of the section continues to explain. Designating the criteria as "not applicable" or making that clear in a template that requires pass/fail is what we want. A publication doesn't "pass" the success criterion in the sense that the requirements of the SC have been met.

For example, probably the most common method of reporting web site accessibility, at least in the US, is via a VPAT and those require a "not applicable" designation if the content is not present, not a pass.

But I can add another paragraph to further explain that inapplicable SCs count as passes as far as declaring whether a publication meets a conformance level of WCAG.

sueneu: You're right! The next two paragraphs are very clear. Do we need the above paragraph?

clapierre (approved):

iherman (approved):

PR #3020, opened 2026-06-12
Add section on filtering success criteria

mattgarrish:
This section has had me going in circles. It's a challenge to definitively say what SC apply to what types of content. It all depends on what the content is, how it's used, etc. Scripting can bring pretty much every success criteria into play, for example, because scripts can inject any type of content.

But rather than go down the rabbit hole of what can apply, what I've done is try to isolate the ones that definitely have no application unless the content type is present. That can be very small, only images of text definitely doesn't apply if there are no images, or lengthier in the case of all the scripting-specific SCs.

https://raw.githack.com/w3c/epub-specs/understand/content-types/epub34/a11y-understand/index.html#eval-filter

***

[Preview](https://raw.githack.com/w3c/epub-specs/understand/content-types/epub34/a11y-understand/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fa11y-understand%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Funderstand%2Fcontent-types%2Fepub34%2Fa11y-understand%2Findex.html)

### DIFF
--- a/epub34/a11y-understand/index.html
+++ b/epub34/a11y-understand/index.html
@@ -963,10 +963,12 @@ aside.sidebar {
 						<p>When it comes to determining if there are images to check, be careful about using only the
 							package document manifest. While the media types list in the manifest are a reliable
 							indicator for EPUB 2 publications, EPUB 3 allows <a data-cite="epub-3#sec-data-urls">data
-								links</a> [[epub-3]]. Data links allow the raw image data to be embedded directly in
-							EPUB 3 content documents meaning no manifest entry is required. Consequently, a search for
-							URLs that begin with "<code>data:</code>" will also be required to ensure there are no
-							images.</p>
+								links</a> [[epub-3]] and scripting.</p>
+						<p>Data links allow the raw image data to be embedded directly in EPUB 3 content documents
+							meaning no manifest entry is required. Consequently, a search for URLs that begin with
+								"<code>data:</code>" will also be required to ensure there are no images.</p>
+						<p>Scripts can be used both to draw images on the [[html]] [^canvas^] element and to inject data
+							links, neither of which would require a manifest entry.</p>
 					</div>
 				</section>
 
--- a/epub34/a11y-understand/index.html
+++ b/epub34/a11y-understand/index.html
@@ -937,8 +937,8 @@ aside.sidebar {
 						applications beyond images. It is not possible to ignore the success criterion simply because an
 						EPUB publication does not contain any images.</p>
 
-					<p>There are many other success criteria like non-text content that can apply to images but also
-						apply to many other types of content. Success criteria related to the <a
+					<p>There are many other success criteria like non-text content that can apply to images and also
+						apply to other types of content. Success criteria related to the <a
 							data-cite="wcag2#use-of-color">use of color</a> and to <a data-cite="wcag2#contrast-minimum"
 							>contrast</a> [[wcag2]] are just a couple of examples. Images can also cause flashing
 						hazards and be used as links and buttons, bringing all those related success criteria into play,
--- a/epub34/a11y-understand/index.html
+++ b/epub34/a11y-understand/index.html
@@ -908,74 +908,59 @@ aside.sidebar {
 		<section id="eval">
 			<h2>Evaluation considerations</h2>
 
-			<section id="content-type">
-				<h3>Content type application</h3>
+			<section id="eval-filter">
+				<h3>Filtering success criteria</h3>
 
-				<p>Although there are fifty six success criteria that fall at level A or AA in WCAG 2.2 [[wcag2]], it is
+				<p>Although there are fifty-six success criteria that fall at level A or AA in WCAG 2.2 [[wcag2]], it is
 					rarely the case that all of these will apply when evaluating an EPUB publication. WCAG breaks the
-					success critieria up by the <em>POUR</em> categorization &#8212; perceivable, operable,
+					success criteria up by the <em>POUR</em> categorization &#8212; perceivable, operable,
 					understanding, and robust &#8212; but many of the success criteria can also be categorized by the
 					type of content they apply to.</p>
 
 				<p>Between the EPUB package document and accessibility evaluation tools, it is possible to determine
 					what types of content are, and are not, present in a EPUB publication. With this information it is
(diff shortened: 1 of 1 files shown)


epub34/a11y-understand/index.html
```diff
@@ -908,60 +908,244 @@ <h4>Skippability</h4>
 		<section id="eval">
 			<h2>Evaluation considerations</h2>
 
-			<section id="content-type">
-				<h3>Content type considerations</h3>
+			<section id="eval-filter">
+				<h3>Filtering success criteria</h3>
+
+				<p>Although there are fifty-six success criteria that fall at level A or AA in WCAG 2.2 [[wcag2]], it is
```
sueneu, line 914:
```suggestion
				<p>There are more than fifty success criteria that fall at level A or AA in WCAG 2.2 [[wcag2]]. It is
``
If we are less specific about the number of sc, we won't have to change it every time a new sc is created.

File: epub34/a11y-understand/index.html [hunk 2]
```diff
@@ -908,60 +908,244 @@ <h4>Skippability</h4>
...
+					rarely the case that all of these will apply when evaluating an EPUB publication. WCAG breaks the
+					success criteria up by the <em>POUR</em> categorization &#8212; perceivable, operable,
+					understanding, and robust &#8212; but many of the success criteria can also be categorized by the
+					type of content they apply to.</p>
```
sueneu, line 918:
```suggestion
					rarely the case that all of these will apply when evaluating an EPUB publication. Many of the success criteria can be categorized by the type of content they apply to.</p>
```
Is the POUR categorization mentioned elsewhere? If so, since this paragraph is about content specific cr, it isn't necessary here.

epub34/a11y-understand/index.html
```diff
@@ -908,60 +908,244 @@ <h4>Skippability</h4>
...
+						making them accessible, the most recognized success criterion is <a
+							href="wcag2#content-type-non-text">non-text content (1.1.1)</a> [[wcag2]]. This success
+						criterion determines if alternative text and/or extended descriptions are required depending on
+						the information the images convey.</p>
+
+					<p>But sometimes overlooked is that the non-text content success criterion is not exclusive to
+						images. It also requires that audio and video have a label that describes the content of the
+						clip, for example. In fact, most success criteria that can apply to images also have
+						applications beyond images. It is not possible to ignore the success criterion simply because an
+						EPUB publication does not contain any images.</p>
```
sueneu, line 938:
```suggestion
					<p>Most success criteria that can apply to images can apply to other content as well. For example, the non-text content success criterion also applies to audio and video. Each clip should have alternative text and/or an extended description.  It is not possible to ignore the success criterion simply because an EPUB publication does not contain any images.</p>
```
Is it possible that a text-only epub with no images, audio, or video could ignore the success criterion for images?

epub34/a11y-understand/index.html
```diff
@@ -908,60 +908,244 @@ <h4>Skippability</h4>
...
+
+					<p>There are many other success criteria like non-text content that can apply to images but also
+						apply to many other types of content. Success criteria related to the <a
```
```suggestion
					<p>There are many other success criteria like non-text content that can apply to images and also
						apply to other types of content. Success criteria related to the <a
```

epub34/a11y-understand/index.html
mattgarrish: WCAG 2 is not being updated anymore so there will never be more than 56 success criteria at level A and AA. Being non-specific when we know the number doesn't seem helpful.

clapierre (approved):

epub34/a11y-understand/index.html
mattgarrish: The danger is that POUR is the default categorization of success criteria and why content-specific attempts to categorize success criteria don't work well. We need to be careful about how we frame discussions like this because we can mislead people both about what does and does not apply. Content-type categorization just doesn't work without a lot of caveats.

But I might go back and rewrite the intro to better explain why POUR matters and why this section is not helpful when it comes to evaluating content that does contain the content types we get into. It's a bit scattered right now.

epub34/a11y-understand/index.html
sueneu: Makes sense.

epub34/a11y-understand/index.html
sueneu: Sounds good Matt.

epub34/a11y-understand/index.html
mattgarrish: That's the problem with content-type categorization and exclusion. It starts to become a game of mix and match to figure out which success criteria don't apply depending on what you have or don't have in the publication.

You have to have no images, audio, video, or scripting before you default assume that the non-text content success criterion can't apply. As soon as scripting is present you have to see what it's doing before you can skip the SC.

And even when all those aren't present there are some SCs that look like they won't apply until you realize that they also cover links.

(I'm still not convinced this section is a great idea - it feels like it belongs to evaluation tools to try and filter out SC since they can get more info about a publication to make better decisions.)

iherman: I understand the hesitation, but I think we say that books that

> […] have no images, audio, video, or scripting […]

are probably an overwhelming majority of published books, at least in the North American and European market (putting aside the weight of manga on the Japanese/Korean market). For that reason emphasizing this "subclass" of books, for which evaluation is much simpler, makes lots of sense to me.

iherman (approved):

epub34/a11y-understand/index.html
mattgarrish: Maybe, but the whole danger of this section is that it will lead people who shouldn't be doing accessibility evaluations into thinking they've been given a map to what to check. The ability to filter only knowing the formats used (plus scripting), and not how they are used, is limited. If we get into cumulative exclusions then everything comes into play.

But I'm closing this pull request. I've got an idea on a different approach that might work better, but if it fails I'm washing my hands of this section!

PR #3022, opened 2026-06-16
Add section on success criteria filtering

mattgarrish:
Okay, new attempt. Instead of subsections for each content type and having to re-explain the limitations, I've restructured it so there is now an intro to explain once what we're trying to achieve. I then put all the guidance on identifying the content types into a separate section. And there's a final section with a table to cover the content types and what can be excluded.

The non-text content still doesn't fit with the table since it's additive. To put it in the table we'd have to re-list all the success criteria for the other content types. If we assume that CSS animations are almost never used, we can also add a few other success criteria. So, I've listed these additional SC after the table to try and make this easier to follow.

https://raw.githack.com/w3c/epub-specs/understand/sc-filter/epub34/a11y-understand/index.html#eval-filter

Edit: I also removed the script to fix the errata links from this document and the techniques because notes don't get errata.

***

[Preview](https://raw.githack.com/w3c/epub-specs/understand/sc-filter/epub34/a11y-understand/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fa11y-understand%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Funderstand%2Fsc-filter%2Fepub34%2Fa11y-understand%2Findex.html)

### DIFF
--- a/epub34/a11y-understand/index.html
+++ b/epub34/a11y-understand/index.html
@@ -1130,9 +1130,6 @@ aside.sidebar {
 												(1.4.13)</a>
 										</li>
 										<li>
-											<a data-cite="wcag2#keyboard">keyboard (2.1.1)</a>
-										</li>
-										<li>
 											<a data-cite="wcag2#no-keyboard-trap">no keyboard trap (2.1.2)</a>
 										</li>
 										<li>
--- a/epub34/a11y-understand/index.html
+++ b/epub34/a11y-understand/index.html
@@ -12,7 +12,7 @@
 				group: "pm",
 				wgPublicList: "public-pm-wg",
 				specStatus: "ED",
-				shortName: "epub-a11y-tech-12",
+				shortName: "epub-a11y-understand-12",
 				noRecTrack: true,
 				edDraftURI: "https://w3c.github.io/epub-specs/epub34/a11y-understand/",
                 copyrightStart: "2026",
--- a/epub34/a11y-understand/index.html
+++ b/epub34/a11y-understand/index.html
@@ -939,12 +939,12 @@ aside.sidebar {
 						organized by whether they improve the perceivability, operability, understanding, or robustness
 						of the content.</p>
 
-					<p>Why this organization matters is because success criteria across multiple categories can apply to
-						different content types, the most common being images, audio, video, and interactive content.
-						The <a data-cite="wcag2#non-text-content">non-text content success criterion</a>, for example,
-						can apply to all these types of content: static images and dynamic images drawn on an [[html]]
-						[^canvas^] element need alternative text and/or descriptions; audio and video need labels to
-						describe the nature of the content.</p>
+					<p>This organization matters because success criteria across multiple categories can apply to
+						different content types. The <a data-cite="wcag2#non-text-content">non-text content success
+							criterion</a>, for example, can apply to images, audio, video, and interactive content. It
+						requires that static images and dynamic images drawn on an [[html]] [^canvas^] element have
+						alternative text and/or descriptions. And for audio and video, it requires labels to describe
+						the nature of the content.</p>
 
 					<p>At the same time, there are various success criteria that only apply to a single type of content.
 						The <a data-cite="wcag2#time-based-media">time-based media success criteria</a>, for example,
@@ -958,14 +958,14 @@ aside.sidebar {
 						to an EPUB publication because so much depends on the nature of the content. But it is possible
 						to help steer evaluators away from success criteria that do not apply.</p>
 
-					<p>The <a href="#eval-not-applicable">mapping table provided in this section</a> identifies success
-						criteria that cannot apply when images, audio, video, and scripted content are not present.
-						While it is possible to encounter these types of content in EPUB publications, they were chosen
-						for analysis because there are many more publications that omit them than have them.</p>
+					<p>Many publications do not contain images, audio, video, or scripted content. The <a
+							href="#eval-filter-table">filtering table</a> identifies success criteria that can be
+						skipped when they are not present in EPUB publications. </p>
 
-					<p>It is possible to determine what types of content are, and are not, present in a EPUB publication
-						using validation tools and by inspecting the package document. With this information these
-						success criteria can be filtered out of an evaluation when they do not apply.</p>
+					<p>It is possible to determine what types of content are present, or not present, in an EPUB
+						publication by using validation tools and inspecting the package document. With this
+						information, the <a href="#eval-filter-table">filtering table</a> can be used to skip any
+						success criteria that do not apply.</p>
 				</section>
 
 				<section id="eval-filter-identify">
@@ -1031,7 +1031,7 @@ aside.sidebar {
 				</section>
 
 				<section id="eval-filter-table">
-					<h4>Content-specific success criteria</h4>
+					<h4>Filtering table</h4>
 
 					<p>The following table provides a list of success criteria that can be skipped if an EPUB
 						publication does not contain the specified type of content.</p>
--- a/epub34/a11y-understand/index.html
+++ b/epub34/a11y-understand/index.html
@@ -932,8 +932,8 @@ aside.sidebar {
 			<section id="eval-filter">
 				<h3>Filtering success criteria</h3>
 
-				<section id="eval-filter-overview">
-					<h4>Overview</h4>
+				<section id="eval-filter-intro">
+					<h4>Introduction</h4>
 
 					<p>There are fifty-six success criteria that fall at level A or AA in WCAG 2.2 [[wcag2]]. They are
 						organized by whether they improve the perceivability, operability, understanding, or robustness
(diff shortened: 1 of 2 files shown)


epub34/a11y-understand/index.html
```diff
@@ -908,60 +929,285 @@ <h4>Skippability</h4>
...
+										<li>
+											<a data-cite="wcag2#identify-input-purpose">identify input purpose
+												(1.3.5)</a>
+										</li>
+										<li>
+											<a data-cite="wcag2#content-on-hover-or-focus">content on hover or focus
+												(1.4.13)</a>
+										</li>
+										<li>
+											<a data-cite="wcag2#keyboard">keyboard (2.1.1)</a>
```
clapierre: Couldn't 2.1.1 or 2.1.2 be caused with incorrect use of tabindex's, or malformed SVG images?

epub34/a11y-understand/index.html
```diff
@@ -908,60 +929,285 @@ <h4>Skippability</h4>
...
-					of content they apply to.</p>
+				<section id="eval-filter-intro">
+					<h4>Introduction</h4>
 
-				<p>Accessible scripting is the focus of a number of success criteria given the highly dynamic nature of
-					the modern web. The following success criteria are all conditional on a publication even having
-					scripted content:</p>
+					<p>There are fifty-six success criteria that fall at level A or AA in WCAG 2.2 [[wcag2]]. They are
+						organized by whether they improve the perceivability, operability, understanding, or robustness
+						of the content as these categories best describe the focus of the success criteria.</p>
```
```suggestion
						of the content.</p>
```
The rest of the sentence is implied.

epub34/a11y-understand/index.html
```diff
@@ -908,60 +929,285 @@ <h4>Skippability</h4>
...
-					<li>3.3.2 Labels or Instructions</li>
-					<li>3.3.3 Error Suggestion</li>
-					<li>3.3.4 Error Prevention (Legal, Financial, Data)</li>
-					<li>3.3.7 Redundant Entry</li>
-					<li>3.3.8 Accessible Authentication (Minimum)</li>
-					<li>4.1.2 Name, Role, Value</li>
-					<li>4.1.3 Status Messages</li>
-				</ul>
+					<p>Why this organization matters is because success criteria across multiple categories can apply to
+						different content types, the most common being images, audio, video, and interactive content.
```
```suggestion
					<p>Success criteria across multiple categories can apply to
						different content types. The most common types being images, audio, video, and interactive content.
```

epub34/a11y-understand/index.html
```diff
@@ -908,60 +929,285 @@ <h4>Skippability</h4>
...
-					<li>1.4.2 Audio Control</li>
-					<li>1.4.5 Images of Text</li>
-				</ul>
-				-->
+					<p>It is not possible for a document like this to determine exactly which success criteria can apply
+						to an EPUB publication because so much depends on the nature of the content. But it is possible
+						to help steer evaluators away from success criteria that do not apply.</p>
+
+					<p>The <a href="#eval-not-applicable">mapping table provided in this section</a> identifies success
+						criteria that cannot apply when images, audio, video, and scripted content are not present.
```
```suggestion
					<p>The <a href="#eval-filter-table">filtering table provided in this section</a> identifies success
						criteria that cannot apply to EPUB publications that don't contain images, audio, video, or scripted content.  Many publications omit these types of content. 
```

epub34/a11y-understand/index.html
```diff
@@ -908,60 +929,285 @@ <h4>Skippability</h4>
...
+						While it is possible to encounter these types of content in EPUB publications, they were chosen
+						for analysis because there are many more publications that omit them than have them.</p>
```
```suggestion
						</p>
```

epub34/a11y-understand/index.html
```diff
@@ -908,60 +929,285 @@ <h4>Skippability</h4>
...
+
+					<p>It is possible to determine what types of content are, and are not, present in a EPUB publication
+						using validation tools and by inspecting the package document. With this information these
+						success criteria can be filtered out of an evaluation when they do not apply.</p>
```
```suggestion
					<p>It is possible to determine what types of content are present, or not present, in an EPUB publication by using validation tools and inspecting the package document. With this information, use <a href="#eval-filter-table">the filtering table</a> to filter out any success criteria that do not apply.</p>
```

epub34/a11y-understand/index.html
mattgarrish: For 2.1.1, tabindex could be a fail, although it's probably an edge case. Without scripting, you're having to take away focus access from regular hyperlinks or maybe an image map with plain hyperlinks. But since it can happen I agree with striking it out.

But can you cause a keyboard trap with tabindex alone? Usually the culprit is some scripting in the page or in embedded content that takes away focus and then you can't get out of the content. I wouldn't think a malformed svg, if you just mean invalid markup, would trap a user. The dom parsing should correct that.

sueneu:
This is making so much more sense to me now. Thank you!

Can we rename the mapping table? Since it is only for filtering success criteria, maybe if we name it something that indicates that people won't be tempted to misuse it.

Something like 
"Inapplicable Success Criteria"
"When Success Criteria Can Be Skipped"

epub34/a11y-understand/index.html
```diff
@@ -908,60 +929,285 @@ <h4>Skippability</h4>
...
+						determine if scripting is present without the help of a validator. Scripts embedded using
+						[[html]] [^script^] tags and scripting inlined in the markup using event handler attributes like
+						{{HTMLElement/onclick}} and {{HTMLElement/onkeypress}} are a couple of ways that scripting can
+						be present at the content level. Trying to locate these possibilities manually is prone to
+						mistakes. But so long as the EPUB publication is validated prior to carrying out an
+						accessibility evaluation, detecting scripting will be straightfoward.</p>
+				</section>
+
+				<section id="eval-filter-table">
+					<h4>Content-specific success criteria</h4>
```
```suggestion
					<h4>Filtering guide</h4>
```

mattgarrish: Changing it to "Inapplicable success criteria" makes it sound like the section defines SC that never apply to EPUB. Which ones are inapplicable is a mix and match.

At one point I was using the title "Filtering table" for the section (hence the id). I'd lean to something more like that to keep it connected with the main section title. "Filtering guide" might also work.

clapierre (approved):

murata2makoto:
If the sentence "The content does not pass because the requirements have been met" still remaints, it is a bit hard to understand for non-native speakers.

Addition after seeing comments from @sueneu and @mattgarrish:  Sorry about my mistake.  I thought that 4.2 was also added as part of this PR.

sueneu:
@murata2makoto 
I gave this a try. Please see CR
[Update accessibility reporting explanation #3024](https://github.com/w3c/epub-specs/pull/3024)

mattgarrish:
> I believe it would be best to make a new PR for this

Yes, it confused me at first because I didn't find it in the text for this new section so I thought it was something from the closed PR. Let's not start a side discussion here.

epub34/a11y-understand/index.html
```diff
@@ -908,60 +929,282 @@ <h4>Skippability</h4>
...
+
+					<div class="caution">
+						<p>This table does not provide a list of all the success criteria have to be checked if the
+							specified content types are present. The <a
+								href="https://www.w3.org/WAI/WCAG22/quickref/?currentsidebar=%23col_customize">How to
+								Meet WCAG (Quick Reference) guide</a> provides the ability to filter WCAG success
+							criteria to show all that apply to different content types as do evaluation tools.</p>
+					</div>
+
+					<table class="zebra">
```
iherman: I was just wondering whether it is not more appropriate to add an `@id` on the table itself and use that in `<a>` elements above, instead of jumping to the section header...

epub34/a11y-understand/index.html
```diff
@@ -908,60 +929,282 @@ <h4>Skippability</h4>
...
+					<p>The <a href="#eval-not-applicable">mapping table provided in this section</a> identifies success
+						criteria that cannot apply when images, audio, video, and scripted content are not present.
+						While it is possible to encounter these types of content in EPUB publications, they were chosen
+						for analysis because there are many more publications that omit them than have them.</p>
+
+					<p>It is possible to determine what types of content are, and are not, present in a EPUB publication
+						using validation tools and by inspecting the package document. With this information these
+						success criteria can be filtered out of an evaluation when they do not apply.</p>
+
+					<p>But when using this section as a guide, do not try to reverse the guidance when the content types
```
iherman: I am not sure if I understand the "reverse the guidance" here...

iherman (approved):

epub34/a11y-understand/index.html
mattgarrish: I'm just worried that if we jump people straight to the table they miss the caution. But no harm in having an ID for referencing from other documents.

epub34/a11y-understand/index.html
mattgarrish: It was trying to highlight that the table doesn't work both ways. You can't read the table as a way of skipping success criteria when the content isn't included and also the only success criteria that apply when the content is present.

But I added the caution note before the table after writing the intro, so I don't think we need to repeat this warning here. The intro already gets into the total being greater than the exclusive set, so I dropped this paragraph in the last commit.

PR #3028, opened 2026-06-22
Fixes term links

naglis:
Currently these two term links are not rendered:

([section](https://www.w3.org/TR/epub-33/#sec-encryption.xml-encryption))
Screenshot of EPUB 3.3 spec showing a link of `package document` term that is not rendered

([section](https://www.w3.org/TR/epub-33/#publication-resources-example))
Screenshot of EPUB 3.3 spec showing a link of `EPUB spine` term that is not rendered


mattgarrish:
Thanks!

PR #3039, opened 2026-07-06
Wrong revision number in the abstract...

iherman:
(The /TR version for the CR has also been changed.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/3039.html" title="Last updated on Jul 8, 2026, 6:32 AM UTC (625d05b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/3039/c986f2e...625d05b.html" title="Last updated on Jul 8, 2026, 6:32 AM UTC (625d05b)">Diff</a>

### DIFF
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -88,7 +88,7 @@
 				format provides a means of representing, packaging, and encoding structured and semantically enhanced
 				web content — including HTML, CSS, SVG, and other resources — for distribution in a single-file
 				container.</p>
-			<p>This specification defines the authoring requirements for [=EPUB publications=] and represents the third
+			<p>This specification defines the authoring requirements for [=EPUB publications=] and represents the fourth
 				major revision of the standard.</p>
 		</section>
 		<section id="sotd" class="updateable-rec"></section>
--- a/epub34/authoring/index.html
+++ b/epub34/authoring/index.html
@@ -88,8 +88,7 @@
 				format provides a means of representing, packaging, and encoding structured and semantically enhanced
 				web content — including HTML, CSS, SVG, and other resources — for distribution in a single-file
 				container.</p>
-			<p>This specification defines the authoring requirements for [=EPUB publications=] and represents the fourth
-				major revision of the standard.</p>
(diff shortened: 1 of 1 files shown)


mattgarrish:
I'm wondering if it would be better to drop the last part of the sentence. Does it really need saying that 3.4 is the fourth major revision? Seems like this will get missed again if we have to update it every time.

The reading systems spec has the same first paragraph in its abstract but then qualifies what a reading system is in the second. Could we do something similar here by compacting the epub publication definition a bit:

> This specification defines the authoring requirements for [EPUB publications](https://www.w3.org/TR/epub-34/#dfn-epub-publication) — works of intellectual or artistic content that are represented by a set of interrelated resources and packaged for distribution in a specialized ZIP container.

(Edit: tweaked the proposed text.)

iherman:
Do be honest, I did not consider a change: I just wanted to get all ducks in a row for the CR publication and this was glaring bug. 

Thinking about it: I agree it is more consistent with the RS to do it the way you propose. It is an editorial change that can be done at any time, but it would probably be cleaner to do it today or tomorrow, so that I can still change the official version and the queued up administrative emails. Ideally today or tomorrow...

epub34/authoring/index.html
```diff
@@ -88,7 +88,7 @@
 				format provides a means of representing, packaging, and encoding structured and semantically enhanced
 				web content — including HTML, CSS, SVG, and other resources — for distribution in a single-file
 				container.</p>
-			<p>This specification defines the authoring requirements for [=EPUB publications=] and represents the third
+			<p>This specification defines the authoring requirements for [=EPUB publications=] and represents the fourth
 				major revision of the standard.</p>
```
mattgarrish: The change you've made to the CR document is fine for now, but adding the suggestion I made for inclusion in the first CRD we publish after the snapshot is published on the 21st:

```suggestion
			<p>This specification defines the authoring requirements for [=EPUB publications=] — works of intellectual or artistic content that are represented by a set of interrelated resources and packaged for distribution in a specialized ZIP container.</p>
```

iherman:
I have adopted your changes @mattgarrish

PR #3040, opened 2026-07-06
Fix unmarked normative refs

mattgarrish:
I took a look at the [link report](https://labs.w3.org/normative-references/check?url=https%3A%2F%2Fwww.w3.org%2FTR%2F2026%2FCR-epub-34-20260721%2F) mentioned in the transition request and these looked like the only references that could be made normative.

Most of the items it flagged were in the iana registrations where we wouldn't use normative or informative references.

Do you think we should add a new substantive changes section to the change log to mention this @iherman? If we were updating a REC I think we'd have to note this as a class 2 change and log it, but it doesn't seem all that substantive when we're just tidying up a CR doc.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/pull/3040.html" title="Last updated on Jul 8, 2026, 2:14 AM UTC (a4c7094)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/epub-specs/3040/c986f2e...a4c7094.html" title="Last updated on Jul 8, 2026, 2:14 AM UTC (a4c7094)">Diff</a>

### DIFF
--- a/epub34/authoring/biblio.js
+++ b/epub34/authoring/biblio.js
@@ -74,6 +74,11 @@ var biblio = {
 		"href": "https://www.itu.int/ITU-T/recommendations/rec.aspx?rec=13189",
 		"date": "2017-04-13"
 	},
+	"iso14977": {
+		"title": "ISO/IEC 14977:1996 — Information technology — Syntactic metalanguage — Extended BNF",
+		"href": "https://www.iso.org/standard/26153.html",
+		"date": "1996-12"
+	},
 	"iso22424": {
 		"title": "ISO/IEC TS 22424-1:2020 — Digital publishing — EPUB3 preservation",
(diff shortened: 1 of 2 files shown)


iherman:
> Do you think we should add a new substantive changes section to the change log to mention this @iherman? If we were updating a REC I think we'd have to note this as a class 2 change and log it, but it doesn't seem all that substantive when we're just tidying up a CR doc.
> 

If we do this _now_, ie, at the time of publishing the CR, it would be o.k. to do this without further ado. Once we are in CR, I would probably prefer to add a change log item (which we will have to have listing changes since the last CR snapshot).

We can combine this with #3039 insofar as we make an update on /TR; you will have to generate a HTML version for this. I will be out of town on Wednesday...

iherman (approved):

Issue #2958, opened 2026-03-19
Use http or https for URIs


himorin:
### Section

e.g. 9.3.2.2

### Describe the problem

In EPUB 3.4 spec, several examples or namespaces use URI with http or https.

For namespace references, some references, like in 9.3.2.2 of epub-34, SMIL namespace is noted as https://www.w3.org/ns/SMIL , but [accessing this page](https://www.w3.org/ns/SMIL) namespace is defined using http. I understand recent discussion around switching entire [www.w3.org](http://www.w3.org/) into HSTS, all http URIs are redirected into https, but not sure how we should use/refer/define namespace URI now.
Note, I heard that most of tools, except for some specific in data space (RDF or some), tool will not access to the URI specified through xmlns attribute, and mostly no harm to use either. But just not sure what is the best (or better?) way now.

For some examples, like example 2, 40, or 41 of epub-34, uses http for (normal) link target. but 1) in epub-rs-34 3.4 https URI scheme is required as SHOULD, and 2) in recent web ecosystem https is strongly encouraged. So I suppose examples are encouraged to use https URIs instead of http.

iherman:
Both in XML and in RDF the namespace URL-s are compared by characters. Ie, they should keep to the http and not to https (unless, mainly for newer cases, they have been defined with https). I would think that the SMIL namespace should actually say http. Ie, what you found is probably a bug...

mattgarrish:
> I would think that the SMIL namespace should actually say http.

Ya, I wonder if it got copied in from the redirected landing page. In any case, as you say, namespaces are identifiers and you can't modify them without changing them so the "https" is wrong and needs fixing.

sueneu:
Is using http instead of https a security issue? 

> namespaces are identifiers and you can't modify them without changing them

does this mean that changing from http to https would break the identifier and since most http addresses are force forwarded to https anyway, there is little security risk for people following the links?

iherman:
> Is using http instead of https a security issue?
> 
> > namespaces are identifiers and you can't modify them without changing them
> 
> does this mean that changing from http to https would break the identifier and since most http addresses are force forwarded to https anyway, there is little security risk for people following the links?

Exactly.

[...]

PR #3049, opened 2026-07-11
Fix SMIL namespace URL

naglis:
The SMIL 3.0 namespace URL [has `http` scheme](https://www.w3.org/TR/SMIL3/smil-scalabilityFramework.html#smilScalabilityNS-BasicDoc-RulesConformant-Docs).

### DIFF
--- a/epub33/core/index.html
+++ b/epub33/core/index.html
@@ -8043,7 +8043,7 @@ No Entry</pre>
 				<section id="sec-overlays-def">
 					<h3>Media overlay document definition</h3>
 
-					<p>All elements [[xml]] defined in this section are in the <code>https://www.w3.org/ns/SMIL</code>
+					<p>All elements [[xml]] defined in this section are in the <code>http://www.w3.org/ns/SMIL</code>
 						namespace [[xml-names]] unless otherwise specified.</p>
 
 					<section id="sec-smil-smil-elem">


iherman (approved):
That is correct. Note that all the example in the document us the correct `http:`. (SMIL was created prior to the wide-spread usage of `https:`).

However, I believe this change, which is a bug handling, should be listed at the end of the document under the "changes" section. The extra editorial complication is that this change will come into effect once the CR is published. @mattgarrish , what is the best way to this to avid merge conflicts?

mattgarrish:
Okay, I'm not losing me mind. i was sure we already fixed this and indeed, it's corrected in 3.4:
https://www.w3.org/TR/epub-34/#sec-overlays-def

It was reported previously in https://github.com/w3c/epub-specs/issues/2958. This pull request is against the old 3.3 standard.

But it looks like we fixed it as an editorial change. If we close this issue should we go back to the change log and add an entry?

iherman:
***You*** are not losing your mind, I am! I have not paid attention that the PR is on EPUB 3.3😒

No, of course we should not do it on 3.3 now that we have 3.4 on its way.

@naglis we should not make a change on the previous version of the standard. That one is frozen. 3.4 should be published sometimes at the end of this year or early next year, which is the place for handling this type of bugs (among other things). See [epub 3.4 version](https://www.w3.org/TR/epub-34/#sec-overlays-def).

naglis:
Sorry for causing a needless ruckus, please feel free to close.

PR #3055, opened 2026-08-11
(editorial) fixing url typo - removing double slash after domain name

svanteschubert:
see title

### DIFF
--- a/epub33/overview/index.html
+++ b/epub33/overview/index.html
@@ -717,7 +717,7 @@
 				<p>The documents themselves have been restructured. The primary motivation for this restructuring, as
 					well as an extensive editorial revision, was to make the documents more readable. Also, as part of
 					the thorough testing regime developed by the <a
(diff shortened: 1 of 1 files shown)


iherman:
Thank you. However, that version of the overview document is frozen, in favour of the [epub 3.4 version](https://www.w3.org/TR/epub-overview-34/). The new version is o.k. I do not think it is worth changing older versions.

/cc @mattgarrish

mattgarrish:
Agree, it's not worth the effort to temporarily retrofit fixes into 3.3. If it's fixed in 3.4, then let's close and wait on the upcoming obsoleting of 3.3.

svanteschubert:
Sorry, I was obviously not reading the latest document. Thanks for your quick response. Greetings from the ISO/IEC JTC 1/SC 34 meeting in Seoul; still orienting myself, I rejoined after 16 years. :-)

PR #3058, opened 2026-08-19
Create .gitattributes

iherman:
Lately a number of PR-s in other repository had the problems when text with different line ending are put in PR-s. When that happens, all PR reviews become difficult because the whole text is marked as "changed". Setting .gitattributes seems to be a good way of handling this (at least per Claude...)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+* text=auto eol=lf
+*.html text eol=lf
+*.htm  text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,2 @@
-* text=auto eol=lf
-*.html text eol=lf
-*.htm  text eol=lf
+* text=auto
+


mattgarrish (approved):
Seems harmless. My client already does line ending conversions for windows, so it's kind of redundant for me.

.gitattributes
```diff
@@ -0,0 +1,3 @@
+* text=auto eol=lf
```
bduga: I don't think you want to set eol here (or below), as it will control what happens when files are checked out of the db. Basically you are forcing `lf` as the end-of-line marker for everyone, even Windows users. We only care about what happens when files are written to the db, so `text=auto` should be enough. As I understand it, you only want to use `eol` when you want to avoid native line endings (e.g. a `.sh` file on Windows may still wand to use `lf`, not `crlf`).

.gitattributes
```diff
@@ -0,0 +1,3 @@
...
+*.html text eol=lf
```
bduga: I am agnostic about adding these lines, but there seem to be two schools of thought here. The first is to just trust git's auto detection and not override all your file types. The second is to take full control and provide entries for all the file types in your repository, especially binary files (jpg, png, etc). So I would either remove these, or add all the others. If you decide to go with option 2 (full control), these are the file types we currently use:

css
doc
epub
htm
html
js
json
jsonld
md
nvdl
png
rnc
sch
sh
svg
ttl
yml

bduga (commented):
Definitely a good idea!

danielweck:
I am primarily a MacOS user but I occasionally clone git repositories on Windows via the command line (I alternate between the "MSDOS" console, PowerShell and Bash). My git client is the de-facto / mainstream one installed from:

https://gitforwindows.org
https://git-scm.com/install/windows

...both website actually point to the GutHub releases:

https://github.com/git-for-windows/git

My experience is that the client automatically handles line endings (i.e. conversion between reads/writes phases). This default behaviour is configurable during Git installation, it is also configurable post-installation.

I do not understand / have not researched the ramifications of forcing a line ending policy in the .git-attributes.

.gitattributes
iherman: To be honest, I am a bit at loss here. I do not know how exactly the github diff operates on a PR review v.a.v. what you call the db. If the lf vs crlf difference is ignored because we force, say, lf for all lines at retrieval time, then I am not sure 'auto' will be enough.

But, frankly, I do not really know...

.gitattributes
iherman: I would be in favour of a "ain't broken, don't fix it" principle. So far we did not have any problems and, in other places, we only had problems with the review of html files (at least in other repos).

iherman:
> I am primarily a MacOS user but I occasionally clone git repositories on Windows via the command line (I alternate between the "MSDOS" console, PowerShell and Bash). My git client is the de-facto / mainstream one installed from:
> 
> https://gitforwindows.org https://git-scm.com/install/windows
> 
> ...both website actually point to the GutHub releases:
> 
> https://github.com/git-for-windows/git
> 
> My experience is that the client automatically handles line endings (i.e. conversion between reads/writes phases). This default behaviour is configurable during Git installation, it is also configurable post-installation.
> 
> I do not understand / have not researched the ramifications of forcing a line ending policy in the .git-attributes.

I do git in various tools (sometimes on command line on my macos, sometimes from within VSCode, sometimes the SourceTree program) and I frankly do not know which versions of git they all use...

iherman:
Actually, what I also learned is that the magic `?w=1` helps in the PR review, see, for example:

 https://github.com/w3c/vc-data-integrity/pull/359/changes?w=1

maybe it is enough to keep that in mind if we hit the problem...

.gitattributes
bduga: > github diff operates on a PR review v.a.v. what you call the db

I am not an expert in git (actually, I kind of hate git - why can't we get hghub?!), but I am fairly sure you are comparing the files from the db. So when I clone a repo, I get a new db with the same .gitattributes of the source repo, so when I add files to it, if they are text, the line endings are normalized. That is, when I did the commit to my local fork, the line endings were forced to `lf`. So regardless of native line endings the PR should review properly. So I think `auto` should resolve the issue (if/when it comes up).

.gitattributes
bduga: The issue is, if the auto detection of git fails to identify some files correctly, then some binary files may become corrupted. You can leave these lines, but they are unnecessary - if we are counting on the auto detection to not corrupt our binary files then we can safely count on auto detection to realize html is text. These lines are something of a placebo, since the `text=auto` line turned on automatic line end conversions for every file type, not just the ones you list here.

bduga:
> the client automatically handles line endings

Correct, but this leaves proper behavior up to the correct configuration of the client. Using `.gitattributes` moves the config responsibility to the repo. There is just less chance of someone messing up their local configs (though, it seems we haven't really experienced that).

bduga:
I am in favor of taking this change with either all used types added, or the specific html lines removed. I verified on a local fork that all files are currently normalized (`git add --renormalize .` did not modify any files).

iherman:
@bduga just to be very safe in all your comments: you tested the .gitattributes files which is a single line, as of:

```
* text=auto
```

Is this where we are?

iherman:
I tried to understand a little bit what is happening, eg, in https://thelinuxcode.com/gitattribute-file-in-git/. My understanding is that  a plain `* text=auto` means to leave things as they are, but it is not a defense if someone's editor forces a different line ending then what is in the repo right now. The only clean way is to enforce a uniform behaviour by doing the following steps:

> An administrator of the site must choose an equilibrium point, when there are no major outstanding PR-s. On his/her local copy, he/she:
>
> - Adds/modifies the `.gitattributes` file in the repository to include
> ```
> * text=auto eol=lf
> ```
> - Executes the following commands:
> ```sh
> git add --renormalize .
> git commit -m "Normalize line endings per .gitattributes"
> git push
> ```
> - Announces it to the group
> - All participants _MUST_ make a `git pull` _before_ any new PR is submitted, to put their repo in sync
>
> Not sure what the best strategy is with the outstanding PR-s, though. But using the `?w=1` trick may make it possible to get the reviews done, and the PR authors may do a rebase (or local merge of main) into their local branch and then push that into the PR.

So the question we have is whether we want to do this, or trust our luck that this will not be an issue for us?

iherman:
I have opened a separate issue in https://github.com/w3c/vc-wg/issues/16. That WG maintains 20+ repositories and is now plagued with this problem. Worth following the discussion there...

mattgarrish:
Wouldn't it be simpler to tell the person trying to commit files with crlf to fix their client if it ever comes up? The default for git on windows is to check out as windows and commit as unix, so someone has to go out of their way to cause an issue. I assume we'll get a really ugly diff if they have their settings incorrect.

(An aside, but the most common problem we run into is that a lot of people/authoring tools don't format files so line _lengths_ get to be a mess. Oxygen formats lines to 120 characters for me, but we've never tried to standardize that. I've been manually reformatting the files for years now to get consistency.)

bduga:
> Is this where we are?

Correct, and this appears to be the proper way of normalizing. From the git man page:

> If you want to ensure that text files that any contributor introduces to the repository have their line endings normalized, you can set the text attribute to "auto" for all files.
> 
> *	text=auto

This differs from leaving in unspecified:

> Unspecified
> If the text attribute is unspecified, Git uses the core.autocrlf configuration variable to determine if the file should be converted.

And if it is set:

> Set
> Setting the text attribute on a path enables end-of-line conversion on checkin and checkout as described above. Line endings are normalized to LF in the index every time the file is checked in, even if the file was previously added to Git with CRLF line endings.

bduga:
> I tried to understand a little bit what is happening, eg, in https://thelinuxcode.com/gitattribute-file-in-git/. My understanding is that a plain `* text=auto` means to leave things as they are

I looked at that link, and it doesn't seem to be very well written. since it starts with eol, which is not really what you want to use to normalize line endings - in fact, it only changes read behavior for the database, not write.

Then there is this line which just seems wrong:

> The special text=auto setting detects and preserves the original line endings on files:

This is just wrong, Specifically, from the git man page on `text` being set ` When a matching file is added to the index, the file’s line endings are normalized to LF in the index` and `when the file is copied from the index to the working directory, its line endings may be converted from LF to CRLF depending on the eol attribute, the Git config, and the platform`. It looks like this was just sloppy writing, as the very next lines in that document say:

> `* text=auto`
> With just a single line, you can eliminate all line ending problems for a repository!

Way to bury the lede.

bduga:
> Wouldn't it be simpler to tell the person trying to commit files with crlf to fix their client if it ever comes up? The default for git on windows is to check out as windows and commit as unix, so someone has to go out of their way to cause an issue. I assume we'll get a really ugly diff if they have their settings incorrect.

I think it is simpler to just enforce the rule now, before there is a problem. That way we don't to enforce client behavior for anyone who will ever commit to the repo, for any platform or client they use. Normalizing at source of truth seems like a sensible safeguard.
> 
> (An aside, but the most common problem we run into is that a lot of people/authoring tools don't format files so line _lengths_ get to be a mess. 

Yeah. I hate the line length mess. But that is a topic for another day. Or at least another PR.

bduga:
> * Adds/modifies the `.gitattributes` file in the repository to include
> 
> ```
> * text=auto eol=lf
> ```
> 
> * Executes the following commands:
> 
> ```shell
> git add --renormalize .
> git commit -m "Normalize line endings per .gitattributes"
> git push
> ```

I don't know where these snippets come from, but they are almost identical to the man page recommendation:

```
$ echo "* text=auto" >.gitattributes
$ git add --renormalize .
$ git status        # Show files that will be normalized
$ git commit -m "Introduce end-of-line normalization"
```

With the exception of adding `eol=lf`, which we should not do, and will make things hell for Windows users as it will override their clients `core.autocrlf` settings (all files will suddenly have UNIX line endings on Windows). I did run that locally and no files where modified, so there should be no issue with pending PRs - adding the .gitattibutes should not change any files in the repo.

iherman:
The VC WG has decided to put a single-line .gitattributes into all its repos (https://github.com/w3c/vc-wg/issues/16) and advise people not to use the github built in editor, which seems to be the source of the problems. They have decided to use 

```
* text=auto eol=lf
```

I would propose we keep an eye on what is happening there (I am on that WG, too, so if problems do arise, I will see it), and adopt that. Alternatively, we do what @bduga  proposes right away, ie, almost the same except for the `eol=lf`.

The very few people in this WG that really submit PR-s did not seem to have had problems, so it is probably not urgent. In the other WG this problems became a plague, it suspended the review of some important PR-s...

bduga:
Let me try some local testing and verify that my understanding is correct. But from what I can tell, adding the `eol=lf` is just trading one problem for another.

bduga:
I spent some time messing with this today. All tests run on Windows using the github desktop app.

With no .gitattributes in my fork of the epub-specs repo, and setting autocrlf=false I was able to change the line endings of all files in the repo, with diffs showing changes on every line. This is pretty obvious when applying to the entire repo, but for a single new file it is easy to miss (nothing will tell you the line endings are CRLF).

I then added .gitattributes with `* text=auto`, committed it, then added a new text file with autocrlf=false. While this file has CRLF line endings locally in my working directory, it only uses LF in the repo (tested by downloading the raw file from github).

Next I added `eol=lf` to that .gitattributes and confirmed that files only have LF as the line ending on Windows regardless of the autocrlf setting. 

I think "correct" here is for line endings to be normalized in the repo (to LF), but whatever native line ending you use on your platform for working directories. Testing seems to show that `* text=auto` is what we want. Leaving things as-is means we can accidentally introduce mixed line endings, and adding `eol=lf` is unnecessary and may be a burden for Windows users. I support adding a .gitattributes with `* text=auto`, though there are recommendations that we specify the text status for every file type we use.

iherman:
I have made the change, @bduga, it is now a one-liner with auto.

Any objection from anyone to merge this?

PR #3072, opened 2026-09-20
Annotations, updates on the specification and UCR

llemeurfr:
- Refine the introduction, i.e. the comparison between Web Annotations and EPUB Annotations.
- Refine the use of the Motivation property, the expression of the "highlight" vs "bookmark" model, and remove "commenting" (may be re-added later if we specify annotation chains).
- Add some precision on the algorithm to adopt for the Text Position Selector.
- Correct a heading for the Sharing annotationSets section.
- Editorial update of a use case in the UCR document; replace "associates" with "imports" for clarity.  

---
See:

* For EPUB Annotations 1.0:
    * [Preview](https://raw.githack.com/w3c/epub-specs/anno-updates-oct26/epub34/annotations/index.html)
    * [Diff](https://services.w3.org/htmldiff?doc1=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fw3c.github.io%2Fepub-specs%2Fepub34%2Fannotations%2Findex.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fepub-specs%2Fanno-updates-oct26%2Fepub34%2Fannotations%2Findex.html)


iherman (approved):
I cannot directly comment through GitHub, hence doing it here: did we say that the values of the About object properties _SHOULD_ (if not _MUST_) mirror the corresponding property values in the package document?

bduga:
I think you mentioned that (requiring values to match), but I don't think
we decided on it. I am not yet convinced, but might be. As an example, what
if a Reading System gets metadata via an ONIX feed? That metadata is often
considered better, and some vendors will always favor it to the content in
the epub itself. In that case, if the epub says the title is "mob_dic.docx"
and ONIX says it is "Moby Dick; or, The Whale", must it really use
"mob_dic.docx"?

sueneu:
>As an example, what if a Reading System gets metadata via an ONIX feed? 

Is there a practical way to access metadata from sources other than the ebook? Would this be adding complexity? Did we decide to limit the metadata to what is in the EPUB, and that we cannot enforce the reliability of such data?

bduga:
On Mon, Sep 21, 2026 at 12:24 PM Susan Neuhaus ***@***.***>
wrote:

> *sueneu* left a comment (w3c/epub-specs#3072)
> <https://github.com/w3c/epub-specs/pull/3072#issuecomment-5766195355>
>
> As an example, what if a Reading System gets metadata via an ONIX feed?
>
> Is there a practical way to access metadata from sources other than the
> ebook?
>
Yes. ONIX and user supplied are the two that come to mind. Metadata from
external sources is fairly normal in the ebook market.


> Would this be adding complexity?
>
No, it reduces complexity since it removes a requirement. For instance,
Google Play Books ingests EPUBs but will grab metadata from other sources
when supplied. Only the "best" (per some calculation of best) is sent to
the clients, and EPUB metadata is never considered "best". So the client
only knows the title that was sent, not the title that was in the EPUB
metadata. Requiring that it be the exact title that was in the EPUB would
mean additional information sent to all the clients on the off chance they
want to export annotations. And the end result is worse data.

> Did we decide to limit the metadata to what is in the EPUB, and that we
> cannot enforce the reliability of such data?
>
I believe we decided to limit the metadata fields. I don't think we agreed
that the values of the fields must exactly match. We definitely cannot
enforce the reliability of the data.

llemeurfr:
Re. expressing that the About metadata is extracted from the EPUB info: there is a sentence in this PR, at the start fo the About section, which states:

"The following table lists recommended metadata fields, **directly extracted from the EPUB**."

That aligns with what Ivan proposes. 

I agree that, on some occasions, the best book metadata doesn't come from the EPUB, but from another source the reading system can access. Thorium Reader has access to OPDS metadata for EPUBs downloaded from an OPDS feed, and OPDS metadata usually comes from ONIX metadata. We don't replace the EPUB title, publisher, or authors with info from the OPDS feed (we simply extract tags from the OPDS subjects), but we could. Also, users request to manually update an ebook's title and cover in the reading app.

iherman:
This was discussed during the [pmwg meeting on 24 September 2026](https://w3c.github.io/pm-wg/minutes/2026-09-24.html#8a1a).

<h3><a href="https://github.com/w3c/epub-specs/pull/3072" rel="noopener noreferrer">w3c/<wbr>epub-specs#3072</a></h3>
<p><cite>wendyreid:</cite> I'm asking that we have a pointer to it</p>
<p><cite>LaurentLM:</cite> this is from a previous meeting, there are several small modifications that we previously agreed on.<br>
<span>… looking at the modifications, I removed the generator object, changed the note at the beginning</span></p>
<p><cite>Duga:</cite> we raised the issue here because of the comment thread, do we require the values of the metadata field to match those of the metadata in the Ebook?<br>
<span>… one argument is to make it "dumb" and match just what is in the EPUB</span><br>
<span>… another is that there is sometimes better information from external sources. Like if a user types in a more accurate title. We know that internal EPUB data can be inaccurate</span></p>
<p><cite>Ivan:</cite> unless we provide an annotation on the package document, so the author can make a comment or change the metadata, then I don't see why we should let the reader invent new metadata</p>
<p><cite>wendyreid:</cite> I think we probably should have a way for metadata to be edited. I've opened books where I needed to modify or fill in the blanks in the metadata. Especially in an academic setting. If I need to provide the city of publication, which is almost never provided in an Ebook, I have to add it manual<br>
<span>… there is a necessary need for that editing</span></p>
<p><cite>ivan:</cite> that raises another issue. The about object and all the fields there are filled in by the rs, and readers can annotation the about object, then why would the rs use anything other than the metadata in the ebook.<br>
<span>… for now in the specification, it is the rs that does something with metadata</span></p>
<p><cite>Duga:</cite> in Playbooks I think none of the implementers know the metadata title of an ebook. Publishers can edit that at many points. These get overwritten. The user/reader sees the title the reading system provides. They will be completely surprised when they see the title from the metadata, which could be something obscure like "mob_dick_v2"<br>
<span>… we know the data in the epub is the worst data. We are asking reading systems to work harder to display bad data</span></p>
<p><cite>wendyreid:</cite> I think ivan and I are closer in concept. I can totally see that some reading systems will allow readers to edit the metadata, like a Vital Source having these features, so we can leave this in the hands of implementers.<br>
<span>… if implementers hear from their users they can respond. Series data is an example of that, users always want the ability to update that</span></p>
<p><cite>LaurentLM:</cite> it is largely requested for users to change the title, etc. within the publication. We don't implement that now, but people will expect to see their title, not the one from the metadata</p>
<p><cite>Hadrien:</cite> there are 2 types of reading systems, some of them provide ability to edit metadata or even bring in 3rd party data. These types get better data from PDS<br>
<span>… another type is libraries and institutions, where all the metadata are coming from ONIX or similar services.</span><br>
<span>… for category two, it is odd to expect users to change the metadata. It is more common for category one, where users have more time to catalog their own files.</span><br>
<span>… so it is hard to come up with one recommendation for both. It will be hard to convince type 2 to change their ways.</span></p>
<p><cite>ivan:</cite> it is clearly a clash between theory and practice. I have to accept that the practice is much messier. For this issue, I yield to the practice. But none of the problems we have talked about here are addressed in our recommendations<br>
<span>… perhaps we should come back to this in the charter. But here we should include a note about why we don't take into account the package metadata.</span><br>
<span>… because this might seem illogical</span></p>
<p><cite>wendyreid:</cite> this is probably a good EPUB 3.5 conversation. We really do have two very different types of users and use cases. We have focused more on the ecosystem publisher/distributor/end user pipeline. Where gaps are shored up by ONIX<br>
<span>… there is a very large community that consumes EPUB outside of that system, even if the use it sometimes. They be managing their own metadata. That community talks about annotations more. Like the Calibre community. Where people are customizing their library, even writing their own reading software. We should probably document it better</span></p>
<p><cite>Hadrien:</cite> this group is unrepresented here. I'm in touch with them because they all use OPDS. They do sometimes contribute to Github issues. Right now we are not connected to that community and it is huge.<br>
<span>… do we want to serve this community and how do we want to reach out. They want to move fast and test things. They know a lot about what goes wrong in EPUB production in general.</span><br>
<span>… this is a bit off topic but we should come back to this as WG and decide if we want to serve them and how we reach out</span></p>
<p><cite>wendyreid:</cite> we should add this to charter talks?</p>
<p><cite>ivan:</cite> Yes, but should we add something about not allowing metadata in the annotations?</p>
<p><cite>LaurentLM:</cite> I can change the metadata source to something more generic, like "that is used by the reading system" with a caveat that this may vary from one rs to the next.</p>
<p>Thanks, Hadrien</p>
<p><cite>LaurentLM:</cite> I had an ai help me review this for security issues and it found few things to fix</p>
<p><cite>ivan:</cite> the W3C is leaning toward making threat models, and it is a lot of work. The emphasis of the threat model is on processing operations. Our situation might be simpler.<br>
<span>… I propose that you put what you have in the spec, and when we send it for horizontal review, we will see if they want the new proceedure</span></p>
<p><cite>wendyreid:</cite> we did a threat model for 3.3?</p>
<p><cite>ivan:</cite> these threat models are new. We should have a discussion in Dublin.</p>
<p><cite>Duga:</cite> I sent a few comments on the privacy section, and I didn't read the security section because I thought it might be re-written, but I can take some time to read it.</p>
<p>Are we OK with this PR, or do we need others?</p>
<p><cite>LaurentLM:</cite> I will wait for Duga to have a look by Monday</p>

<hr /></details>

llemeurfr:
The About section now reads: 

The following table lists recommended metadata fields, directly extracted **from the metadata of the publication.**

And a note is added: 

 >The metadata of the publication may differ from the EPUB package document metadata. 
 > This happens when the publication metadata is manually updated by the user of the reading system, 
   or when the reading system fetches metadata from an external source.

